PR #84, opened 2013-05-13
Addition of RDS instance available/deleted waiters

tomsmaddox:
See https://forums.aws.amazon.com/thread.jspa?messageID=450707


rquadling:
Could this be extending to work on DBSnapshots?

When I copy a db snapshot (for example) being able to wait until it is available would be useful.

rquadling:
I'm happy to generate a PR, but I think my knowledge of the internals of AWS RDS services mean that I don't know all the available states or transitions to make me an ideal candidate for that. Though willing to learn.

rquadling:
Alternatively, is there a way to build a generic waiter?

kstich:
@rquadling Would you mind opening up a separate issue for this request?

Issue #161, opened 2013-10-04
Duplicate headers are mishandled by PHP SDK


pcolby:
See [this topic](https://forums.aws.amazon.com/thread.jspa?threadID=136223)  (forums.aws.amazon.com) for background.

The SignatureV4.php implementation **incorrectly** sorts header values, in direct contradition to the [V4 documentation](http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) which says:

> Append a comma-separated list of values for that header. If there are duplicate headers, the values are comma-separated. **Do not sort the values in headers that have multiple values.**

I see that the PHP SDK makes use of the official [Signature Version 4 Test Suite](http://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html) - unfortunately, the only two tests in that suite that deal with multiple headers (`get-header-key-duplicate` and `get-header-value-order`) are both wrong, as tested against SQS.

I'm happy to help fix the issue (the code change should be very easy), but I think that my claim that the official test suite is wrong needs to be addressed first, otherwise the fix will result in failing unit tests.

Thanks.

PR #162, opened 2013-10-05
Corrected the multi-header mishandling (issue #161)

pcolby:
Corrected the multi-header mishandling (issue #161), and disabled two related (now failing) unit tests that are believed to be invalid.

**Note:** This pull request should not be accepted until someone confirms that the related V4 test suite files are indeed invalid (see issue #161).  If/when the V4 test suite is corrected, then the skip added to the `SignatureV4Test.php` file can be discarded.

Let me know if there's anything more I can do to help.

Cheers.


ShadoWhore:
IMG_20221206_205744_586
****

PR #344, opened 2014-08-29
Add waiters for StackDeleted, StackUpdated, StackCreated

radford:


samsamm777:
@jeremeamia I cant find any docs on waiters for cloudformation?

jeremeamia:
Well, this request never made it out of the backlog. The implementation that was provided in this PR was not quite right. The SDK does not have waiters for everything, only the ones that have been requested the most or contributed. I don't work at AWS anymore though, so you will have to continue discussion with the current devs by reopening this issue or starting a new one.

radford:
@samsamm777, FWIW I'm still interested in this, and I'd be happy to try and update my patch given some feedback and/or direction.

samsamm777:
@radford So I found out this is all implemented in the SDK version 3. Its **NOT** documented anywhere though.

```php
$this->cloudFormationClient->waitUntil("StackCreateComplete", [
    "StackId" => $myStackId
]);
```

radford:
@samsamm777, thanks for pointing that out.  They weren't in v3 last time I looked.  That's enough motivation to try upgrading again.

jeskew:
It's possible they were added in a different repository. The JSON waiter definitions are shared among the AWS SDKs. v2 uses an artifact built from the shared files, whereas v3 uses them more or less directly, so there's a lot of functionality that v3 gets for free.

Issue #419, opened 2014-12-04
v3: Call to undefined method Command::search() from AwsClient::getIterator()


radford:
when calling:
```php
     $ec2client->getIterator('DescribeInstances');
```

I notice that this failing code is only reached because ```DescribeInstances``` does not have ```input_token```  and ```output_token``` defined in the ec2 paginators, though it looks like it could.

mtdowling:
Closed by https://github.com/aws/aws-sdk-php/commit/fbaa02d06ccea42fc177df5bda9e2ee27f321240

PR #421, opened 2014-12-04
Support pagination (of Instances) in DescribeInstances

radford:
This gives the behavior I would expect, and is similar to what how ```getIterator()``` used to work for DescribeInstances in the sense that it flatten out the reservations. It's hard to tell if ```getIterator()``` is still supposed to work this way since it currently doesn't work at all (see #419).

### DIFF
--- a/src/Common/Resources/api/ec2-2014-09-01.paginators.php
+++ b/src/Common/Resources/api/ec2-2014-09-01.paginators.php
@@ -34,7 +34,10 @@
       'result_key' => 'InstanceStatuses',
     ],
     'DescribeInstances' => [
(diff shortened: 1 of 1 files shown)


mtdowling:
This information is sourced from an internal build tool. Because this is an automated process, we'll push the relevant changes through our internal build process and then deploy any code changes as a result of that build step. 

We won't be able to change the `result_key`  because it loses the association of an instance and its reservation. Other SDKs also use the current `result_key` setting, so changing this upstream would be a breaking change and would make the PHP SDK inconsistent with the other AWS SDKs.

PR #437, opened 2014-12-21
Always add SignatureListener, even on NullCredential requests

rbayliss:
Creating a new S3Client using NullCredentials, then setting legitimate credentials later on causes Authorization errors.  This is because the SignatureListener is only added when the client is created with NullCredentials, meaning the listener cannot respond to the client.credentials_changed event.  This pull request changes things so the listener is always added, and moves the check for NullCredentials into the SignatureListener plugin.


jeremeamia:
I've submitted this for review. Thanks for sending the CLA.

jeremeamia:
Everything is good on our end. Merging.

PR #438, opened 2014-12-21
Elastic Container Service client

nickschuch:
I've started to add the ECS client but looking for some guidance around:

* Are you ok with the naming etc?
* Handle the result responses (I will continue to look at other services as well).


mtdowling:
Thanks for taking a stab at this! I've now created a generated client using our internal build system. This has been added in https://github.com/aws/aws-sdk-php/commit/c62d4ac30c1bfbcbb1ed3d086ec8021830b5d426.

nickschuch:
All by hand :)

Thanks for adding this! Ill take it for a spin.

PR #447, opened 2015-01-08
Removing some compat layers and cleaning up conf.

mtdowling:
* Removing the Compat class and Aws class.
* Cleaning up the client factory configuration system.
* Added a way to add deprecation functions to the client factory without
  causing a runtime penalty for normal options.
* Removed the ability to set a custom exception_class


src/CloudSearchDomain/CloudSearchDomainFactory.php
```diff
@@ -26,7 +26,7 @@ protected function addDefaultArgs(&$args)
         if (strpos($args['endpoint'], 'http') !== 0) {
             $args['endpoint'] = Url::buildUrl([
                 'host'   => $args['endpoint'],
-                'scheme' => $args['scheme'],
+                'scheme' => 'https',
```
jeremeamia: Why this change?

src/Common/Aws.php
```diff
@@ -1,52 +0,0 @@
...
-
-    /**
-     * @param string     $name
-     * @param bool|array $throwAway
-     *
-     * @return AwsClientInterface
-     */
-    public function get($name, $throwAway = false)
-    {
-        if (!isset($this->clients[$name]) || $throwAway) {
```
jeremeamia: Are we going to provide any kind of client "caching" feature (i.e., `$sdk->getS3() === $sdk->getS3()`)?

src/Common/ClientFactory.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Aws\Common;
 
+use InvalidArgumentException as IAE;
```
jeremeamia: :wink:

jeremeamia:
Cool!

src/CloudSearchDomain/CloudSearchDomainFactory.php
mtdowling: Looks like I was wrong about this change.

This function is now invoked *before* default parameters are merged in. This function relied on the fact that "scheme" was set to "https" by default.

I was thinking about updating this function, but now I'm wondering if we should even do this. "endpoint" is supposed to be set to an absolute URL. This code here attempts to create a URL using a scheme and treat the endpoint as a host. No other client does this, and I'd like to add clarifications to the "endpoint" config option to clarify that it must be absolute. Does that sound OK?

jeremeamia: Hmmm... there is a potential problem, because `CloudSearchClient::describeDomains` does not include a scheme when you retrieve the endpoint. This is because the endpoints are accessible via http and https, and it is left up to the user to choose. I had the code set up the way it was, because it is likely that the endpoint provided won't include a scheme, especially when using the [`CloudSearchClient::getDomainClient` method](https://github.com/aws/aws-sdk-php/blob/v3/src/CloudSearch/CloudSearchClient.php#L21).

mtdowling: What if we just updated the getDomainClient() function to return an absolute URL? That could do the user a favor of assuming a default scheme of https, while not introducing a one-off behavior of a single client (i.e., endpoint accepting a domain and not an actual endpoint).

jeremeamia: Sounds good to me!

src/Common/Aws.php
mtdowling: I don't think we should. Maybe to make this distinction more clear, `get*()` should become `create*()` (e.g., `createS3()`)?

PR #450, opened 2015-01-13
Updating waiters to be async and use the new waiters2 config format.

jeremeamia:


src/Common/AwsClient.php
```diff
@@ -224,19 +222,21 @@ public function getPaginator($name, array $args = [], array $config = [])
             . "of {$this->api['serviceFullName']} cannot be paginated.");
     }
 
-    public function getWaiter($name, array $args = [], array $config = [])
+    public function waitUntil($name, array $args = [], array $config = [])
     {
+        // Prepare waiter config and args.
         $config += $this->api->getWaiterConfig($name);
+        $args += ['@future' => false];
```
mtdowling: Is this used the waiter object somehow? If not, I think this could be removed and the later check for this attribute could be changed to `empty()`.

tests/Common/Api/ServiceTest.php
```diff
@@ -135,55 +135,54 @@ public function dataForWaiterConfigTest()
...
-
-        // Get the resolved config and verify its correctness
-        $actual = $api->getWaiterConfig($name);
-        /** @var array $expected */
-        foreach ($expected as $key => $value) {
-            $this->assertEquals($value, $actual[$key]);
-            $this->assertEquals($name, $actual['waiter_name']);
-        }
-    }
+//    /**
```
mtdowling: ?

GrahamCampbell:
Do we really need to have a "waiters2" format. Could we not just change the existing one given that 3.0 is still a wip.

tests/Common/Api/ServiceTest.php
jeremeamia: Doh! Looks like I forgot about something there...

jeremeamia:
> Do we really need to have a "waiters2" format? Could we not just change the existing one given that 3.0 is still a wip.

For now, yes. The "2" is important to our team's internal processes.

GrahamCampbell:
Ok. :)

mtdowling:
LGTM

jeremeamia:
Cool. I'm going to hold off until #451 is merged, and then reapply my changes with updated namespaces and such.

mtdowling:
I just pushed up all the docs models and waiters2 models that are currently published to the v3 branch. This PR will need to be reapplied to the v3 branch. I used the models published to the Ruby SDK as the basis for what is currently published. Closing until then.

PR #451, opened 2015-01-14
Shorter namespaces, PHP 5.5, and fixing stream wrapper.

mtdowling:
This pull request shortens the namespace structure of the SDK.

- Clients have been moved from `Aws\Service\ServiceClient` to `Aws\ServiceClient`. For example, `Aws\S3Client`.
- Exceptions have been moved from `Aws\Service\Exception\ServiceException` to `Aws\Exception\ServiceException` (e.g., `Aws\Exception\S3Exception`.
- Moved things out of Common and up a namespace level. This gives us shallower namespaces, but makes the root namespace larger.

In order to make the transition from v2 namespaces to v3 namespaces easier, I've included a legacy autoloader that will alias old class names to new class names when they are loaded. I've also added checks for common v2 classes that are no longer utilized or have changed significantly to the point that aliasing would not work. When one of these classes is attempted to be loaded, an exception will be thrown with information on how a user can migrate to an alternative class.

This PR also updates the SDK to PHP 5.5 so that we can use generators. Generators have helped significantly reduce the amount of code needed to to implement lazy filtering and mapping, and has made it easier to work around PHP iterator quirks that caused the PHP stream wrapper to behave badly (i.e., skipping the first element of dir_readdir() due to RecursiveDirectoryIterator being broken in the constructor). PHP 5.5 is a viable choice for the SDK because nearly half of our users already use PHP 5.5, PHP 5.5 is available on Amazon Linux, PHP 5.5 can be used on RHEL systems using software collections, PHP 5.5 is the default version of PHP on the latest LTS Ubuntu release, and CentOS users can use the remi repo for PHP 5.5 support.

With the addition of PHP 5.5, I've added a dependency on transducers.php so that we do not need to reimplement stuff like filter, map, mapcat, etc.


jeremeamia:
:open_mouth: Wow... I'll look through this some more in the morning.

mtdowling:
I got the build passing after fixing the `@covers` annotations and making a `goto` related recursion change to guzzlehttp/command: https://github.com/guzzle/command/commit/3303240d33b53c4ced7d94695dd2faa1712fdc58

src/Multipart/AbstractUploader.php
```diff
@@ -159,12 +159,9 @@ public function upload($concurrency = 1, callable $before = null)
         }
 
         // Create iterator that will yield UploadPart commands for each part.
-        $commands = new MapIterator($this->parts, function (array $partData) {
+        $commands = t\to_iter($this->parts, t\map(function (array $partData) {
             return $this->createCommand(static::UPLOAD, $partData);
-        });
-
-        // This is needed to get the MapIterator ready for the valid() method.
-        $commands->rewind();
```
mtdowling: Note that this change removed our reliance on PHP's FilterIterator which had the requirement of rewinding it first before the iterator was valid.

src/ResultPaginator.php
```diff
@@ -66,31 +67,15 @@ public function __construct(
...
     }
 
     /**
-     * Returns either a single config value or entire paginator's config.
-     *
-     * @param string $key
-     *
-     * @return string|array|null
-     */
-    public function getConfig($key = null)
```
mtdowling: I removed this function because I didn't see a need for it (at least yet). It seems like an implementation detail of the iterator leaking through.

src/ResultPaginator.php
```diff
@@ -101,25 +86,6 @@ public function getNextToken()
     }
 
     /**
-     * Fetch the next result for the command managed by the paginator. You may
-     * pass additional arguments, the most appropriate being a limiter.
-     *
-     * @param array $args
-     * @return Result|null
-     */
-    public function getNext(array $args = [])
```
mtdowling: I removed this as well. Modifying the paginator options after you begin paginating seems potentially problematic. Maybe we could add a setLimit() function if necessary as that seems to be the only option someone would/should change after they begin iterating.

src/S3/ApplyMd5Subscriber.php
```diff
@@ -11,7 +11,7 @@
  *
  * @internal
  */
-class ApplyMd5 implements SubscriberInterface
+class ApplyMd5Subscriber implements SubscriberInterface
```
mtdowling: Note that I added a Subscriber suffix and shallowed up the namespaces.

src/S3/StreamWrapper.php
```diff
@@ -350,23 +354,21 @@ public function dir_opendir($path, $options)
...
-                            && (!isset($key['Key'])
-                                || substr($key['Key'], -1, 1) !== '/');
+                            && (!isset($key['Key']) || substr($key['Key'], -1, 1) !== '/');
                     }
                 );
-            }
+            })
         );
 
-        $this->objectIterator->rewind();
```
mtdowling: No need to rewind now that we're using a generator.

src/S3/StreamWrapper.php
```diff
@@ -325,7 +329,7 @@ public function rmdir($path, $options)
      */
     public function dir_opendir($path, $options)
     {
-        $this->clearStatInfo();
+        $this->openedPath = $path;
```
mtdowling: By tracking the opened path, we can reopen the path when dir_rewinddir is called (used in things like the Symfony Finder).

src/S3/StreamWrapper.php
```diff
@@ -428,7 +430,7 @@ public function dir_readdir()
 
         // Cache the object data for quick url_stat lookups used with
         // RecursiveDirectoryIterator.
-        self::$nextStat = [$key => $stat];
+        self::$nextStat[$key] = $stat;
```
mtdowling: I changed nextStat to a hash so that it caches more than just a single element, but rather caches each element as it is iterated. This significantly reduces the number of requests made in the stream wrapper, and these cached items are cleared out when a dir resource is closed.

src/S3/StreamWrapper.php
```diff
@@ -436,12 +438,7 @@ public function dir_readdir()
 
     private function formatKey($key)
     {
-        $k = "s3://{$this->openedBucket}/";
-        if ($this->openedBucketPrefix) {
-            $k .= rtrim($this->openedBucketPrefix, '/') . '/';
-        }
-
-        return $k . $key;
+        return "s3://{$this->openedBucket}/{$key}";
```
mtdowling: This appending of the openedBucket prefix was actually causing cache misses and wasn't necessary.

src/S3/Transfer.php
```diff
@@ -109,19 +109,32 @@ public function __construct(
...
+     * Returns a recursive directory iterator that yields filenames only and
+     * is not broken like PHP's built-in DirectoryIterator (which will read
+     * the first file from a stream wrapper, then rewind, then read it again).
      *
      * @param string $path Path on disk to traverse
      *
      * @return \Iterator Returns an iterator that yields absolute filenames.
      */
-    public static function fileIterator($path)
+    public static function recursiveDirIterator($path)
```
mtdowling: Here I've created my own implementation of RecursiveDirectoryIterator that doesn't have weird rewind semantics like PHP's built-in RecursiveDirectoryIterator. The built-in iterator will immediately read the first file using dir_readdir, then it will be rewound and call dir_rewinddir when you begin iterating. This caused the first element of our generators to be skipped.

Here I've just reimplemented RecursiveDirectoryIterator using a generator and a queue of values to yield. This works on any type of stream wrapper and does not call unnecessary rewinds.

src/S3/Transfer.php
```diff
@@ -129,7 +142,7 @@ public static function fileIterator($path)
      */
     public function transfer()
     {
-        if (!$this->source->valid() && !($this->source instanceof \Generator)) {
+        if (!$this->source->valid()) {
```
mtdowling: The generator check was not necessary here.

jeremeamia:
We have a lot of things in the SDK that require iterator-like behavior. Bumping to 5.5 to use generators will be amazing. We've run into a lot of "gotchas" with the SPL iterator classes that aren't worth dealing with. It's also a pain to have to write our own iterator classes for things, when you can do the same thing with a single generator function. I'm totally in favor of this version requirement change. If you think you're transducers library is the right solution for the SDK, I'm not opposed to that. It seems pretty useful. I still have a few more things to check out on this PR, before you merge.

README.md
```diff
@@ -21,17 +21,32 @@ downloading a single zip or phar file from our [latest release][latest-release].
...
+The recommended way to install the AWS SDK for PHP is through Composer.
+
+1. Install Composer
+
+        curl -sS https://getcomposer.org/installer | php
+
+2. Next, run the Composer command to install the latest stable version of
+   Guzzle:
+
+        composer require guzzlehttp/guzzle
```
jeremeamia: You might want to review/update this section a little. :wink:

GrahamCampbell: lol

mtdowling: Whoops! :)

mtdowling: Fixed

jeremeamia:
:ship:

GrahamCampbell:
AWSome. :)

PR #453, opened 2015-01-15
Moving client factories to a single directory

mtdowling:
1. Allows us to see which clients have custom factories.
2. More consistent with how exceptions are now namespaced.

### DIFF
--- a/src/ClientFactory/ClientFactory.php
+++ b/src/ClientFactory/ClientFactory.php
@@ -1,10 +1,13 @@
 <?php
-namespace Aws;
+namespace Aws\ClientFactory;
 
 use InvalidArgumentException as IAE;
+use Aws\Sdk;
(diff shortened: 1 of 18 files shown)


jeremeamia:
LGTM. I did notice that the Sqs client factory is not marked as internal though.

mtdowling:
Cool, I'll fix that.

GrahamCampbell:
I wonder if all the client classes should be moved to their own namespace too?

mtdowling:
I like them at the root namespace because the namespace is shorter to type, there's no namespace collision possibility thanks to the "Client" suffix, and every client that we support will have a Client class. Not every client we support will have a custom factory, and end users will never need to interact with factories, so it makes sense to put them in a specific namespace.

GrahamCampbell:
k

PR #454, opened 2015-01-15
Adding hasWaiter() and hasPaginator() to Service

mtdowling:


src/Api/Service.php
```diff
@@ -194,8 +194,7 @@ public function getSigningName()
      */
     public function getProtocol()
     {
-        return $this->getMetadata('protocol')
-            ?: $this->getMetadata('type');
+        return $this->getMetadata('protocol');
```
mtdowling: "type" is no longer used btw.

jeremeamia:
Cool, that should do the trick. :ship:

PR #458, opened 2015-01-20
Partial revert of #451

### LINKED ISSUE #456: S3Client name already in use
Getting this error recently:

```
PHP Fatal error:  Cannot use Aws\S3Client as S3Client because the name is already in use in /base/data/home/apps/s~crack-talent-783/1.381614094212936608/vendor/aws/aws-sdk-php/src/S3/BucketStyleSubscriber.php on line 4
```

mtdowling:
This commit partially reverts #451, moving clients and exceptions back
to their former v2 locations. The autoloader that was put in place that
allowed us to move the classes to new locations without breaking v2 code
had an edge case in clients that have custom code in a client specific
namespace (e.g., S3 classes attempting to import an S3Client that was
already aliased into its namespace). The old Common namespace is still
removed.

Closes #456.


jeremeamia:
LGTM!

PR #461, opened 2015-01-21
Adding a system for more helpful factory errors.

mtdowling:
This commit adds a way for us to create custom error messages for
missing required client configuration options. In this commit, I've
added helpful error messages for a missing region (i.e., a link to
rande.html), and a helpful error message for a missing version (i.e., a
list of available versions on disk and a link to the API docs).

### DIFF
--- a/src/ClientFactory.php
+++ b/src/ClientFactory.php
@@ -32,17 +32,9 @@ class ClientFactory
         'key' => ['type' => 'deprecated'],
         'ssl.certificate_authority' => ['type' => 'deprecated'],
         'curl.options' => ['type' => 'deprecated'],
-        'scheme' => [
-            'type'     => 'value',
-            'default'  => 'https',
-            'required' => true
-        ],
-        'region' => [
-            'type'     => 'value',
-            'required' => true,
-            'default'  => true
-        ],
         'service' => ['type' => 'value', 'required' => true],
+        'scheme' => ['type' => 'value', 'default' => 'https', 'required' => true],
+        'region' => ['type' => 'value', 'required' => true],
         'endpoint' => ['type' => 'value'],
         'version' => ['type' => 'value', 'required' => true],
         'defaults' => ['type' => 'value'],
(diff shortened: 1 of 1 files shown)


mtdowling:
This has the following error message for a missing version:

----

A 'version' configuration value is required when creating an API client. For example, when using Amazon S3, you can lock your API version to '2006-03-01' to ensure that your code will be unaffected by a change in the web service.

Your build of the SDK has the following versions of 's3': 
- 2006-03-01

You may provide 'latest' to the 'version' configuration value to utilize the most recent available API version that your client's API provider can find (the default API provider will scan the /Users/dowling/projects/sdk3/src/data directory for *.api.php and *.api.json files). Using 'latest' in a production application is not recommended.

A list of available API versions can be found on each client's API documentation page: http://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html. If you are unable to load a specific API version, then you may need to  update your copy of the SDK.

----

And the following for a missing region:

----

A "region" configuration value is required when connecting to the dynamodb service (e.g., us-west-2). A list of available public regions and endpoints can be found at http://docs.aws.amazon.com/general/latest/gr/rande.html

----

src/ClientFactory.php
```diff
@@ -80,7 +72,11 @@ public function create(array $args = [])
                     // Merge defaults in when not present.
                     $args[$key] = $a['default'];
                 } elseif (!empty($a['required'])) {
-                    throw new IAE("{$key} is a required client setting");
+                    // Allows custom error messages for missing values.
+                    $message = method_exists($this, "missing_{$key}")
+                        ? call_user_func([$this, "missing_{$key}"], $args)
```
jeremeamia: Could also do `$this->{"missing_{$key}"}($args)`. I usually avoid `call_user_func*` when I can. It doesn't really matter much though, so whatever you prefer is fine.

src/ClientFactory.php
```diff
@@ -484,4 +480,42 @@ private function deprecated_curl_options($value, array &$args)
...
+        $error = 'A "region" configuration value is required when connecting to '
+            . 'the %s service (e.g., us-west-2). A list of available public '
+            . 'regions and endpoints can be found at %s';
+
+        return sprintf($error, $args['service'], 'http://docs.aws.amazon.com/general/latest/gr/rande.html');
+    }
+
+    private function missing_version(array $args)
+    {
+        $error = "A 'version' configuration value is required when creating "
```
jeremeamia: I almost feel like maybe a `<<<` would be better here. Meh...

jeremeamia:
I inlined some suggestions. Take them or leave them. Looks like a good addition for the user experience though. :smile:

src/ClientFactory.php
mtdowling: Good idea

PR #462, opened 2015-01-21
Adding a signature provider

mtdowling:
This PR adds a signature provider that can be used to create signature objects by name. This is now utilized in the ClientFactory, and could be utilized by customers to create custom signature objects to, for example, create a presigned URL.

### DIFF
--- a/src/ClientFactory.php
+++ b/src/ClientFactory.php
@@ -1,6 +1,7 @@
 <?php
 namespace Aws;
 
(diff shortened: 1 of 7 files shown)


mtdowling:
Fixed the `@covers` annotation to fix the tests.

jeremeamia:
Seems like a good way to separate out the signature logic.

PR #463, opened 2015-01-22
Adding a Utils class and using less dir buffering

mtdowling:
This commit adds a Utils class where we can place useful and generic
functionality (e.g., a directory iterator that doesn't call readdir()
and rewinddir() when constructed).

This commit also updates the recursive directory iterator to not require
each level of directory traversal be cached in memory.


jeremeamia:
Hooray generators!

PR #464, opened 2015-01-22
Refactored MUPs to use generators

jeremeamia:
Refactored the Multipart Uploaders for S3 and Glacier to use generators and have a more succinct codebase.


src/Multipart/UploadState.php
```diff
@@ -35,7 +35,8 @@ public function __construct(array $uploadId)
     }
 
     /**
-     * Returns
+     * Get the upload's ID, which is a triad of parameters that can uniquely
+     * identity the upload.
```
mtdowling: *identify*

src/Multipart/UploadState.php
```diff
@@ -44,21 +45,54 @@ public function getUploadId()
...
     }
 
+    /**
+     * Returns whether a part has been uploaded.
+     *
+     * @param int $partNumber
+     *
+     * @return bool
+     */
+    public function hasPartBeenUploaded($partNumber)
```
mtdowling: Should this be named `isUploaded` to be consistent with the other `is*` methods?

src/Multipart/UploadState.php
```diff
@@ -44,21 +45,54 @@ public function getUploadId()
...
     public function setPartSize($partSize)
     {
         $this->partSize = $partSize;
     }
 
+    /**
+     * Marks a part as being uploaded.
+     *
+     * @param int   $partNumber
+     * @param array $partData
```
mtdowling: What is "partData" here? Can that be documented?

src/Multipart/AbstractUploadBuilder.php
```diff
@@ -9,7 +9,7 @@
 abstract class AbstractUploadBuilder
```
mtdowling: I suggest either marking this class as internal or refactoring it to use private properties with protected accessor methods.

mtdowling:
Great changes! I had a couple points of feedback, but otherwise :shipit:

PR #465, opened 2015-01-24
Added more functionality to client factory options

mtdowling:
Added a new static method for client factories that allows you to get a
hash of valid arguments. This result can be used to generate
documentation, and I've added a build script for generating the Sdk
class docblock that describes client factory arguments.
You can now provide a "doc" string for each argument.
You can now provide a "valid" key for an argument to describe the valid
types or class names. These will then be automatically validated when
handling client factory options.
The class_name argument now must be a full class name.
Added the exception_class argument.
Removed the no longer needed per-argument validation.

### DIFF
--- a/build/option-docs.php
+++ b/build/option-docs.php
@@ -1,0 +1,49 @@
+<?php
+require __DIR__  . '/../vendor/autoload.php';
+
+use GuzzleHttp\Ring\Core;
+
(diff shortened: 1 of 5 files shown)


jeremeamia:
I like it. Seems like a good way to contain all that information. :ship:

PR #466, opened 2015-01-26
Update Marshaler.php

xterr:
Suggestion: Change this method to protected so it can be easily extended. Sometimes in JSON we can have empty strings and an exception will be thrown. In my implementation, it is ok to mark that value as NULL instead. Or, maybe, add a config param to marshalItem/marshalJSON to automatically do this when encountered.


jeremeamia:
Seems like a reasonable change. I'm inclined to just make it `public` though. Same with `unmarshalValue`. Do you want to update this pull request?

jeremeamia:
Well, I'll take of it. Thanks!

PR #470, opened 2015-01-28
Add support for "endpoint" for forward compat w/ v3

mtdowling:
This now shows "endpoint" only in the docs, but "base_url" will continue to be supported.


jeremeamia:
Cool. That looks good. :ship:

PR #471, opened 2015-01-28
Adding forward compat for credentials array setting

mtdowling:
This commit updates the SDK to accept a hash of options for the credentials
option so that it is forward compatible with v3. This commit also
updates the docs to favor the new format. I've removed instances of
hardcoding credentials from the docs where they are not necessary.


jeremeamia:
Cool! :ship:

PR #472, opened 2015-01-29
Added the DynamoDB Marshaler

jeremeamia:
This includes support for binary and set values via special methods/classes.

### DIFF
--- a/src/DynamoDb/BinaryValue.php
+++ b/src/DynamoDb/BinaryValue.php
@@ -1,0 +1,36 @@
+<?php
+namespace Aws\DynamoDb;
+
+use GuzzleHttp\Stream\Stream;
+
+/**
+ * Special object to represent a DynamoDB binary (B) value.
+ */
+class BinaryValue implements \JsonSerializable
+{
+    /** @var string Binary value. */
+    private $value;
+
+    /**
+     * @param mixed $value A binary value compatible with Guzzle streams.
+     *
+     * @see GuzzleHttp\Stream\Stream::factory
+     */
+    public function __construct($value)
+    {
+        if (!is_string($value)) {
+            $value = Stream::factory($value);
+        }
+        $this->value = (string) $value;
+    }
+
+    public function jsonSerialize()
+    {
+        return $this->value;
+    }
+
+    public function __toString()
+    {
+        return $this->value;
+    }
+}
--- a/src/DynamoDb/Marshaler.php
+++ b/src/DynamoDb/Marshaler.php
@@ -1,0 +1,320 @@
+<?php
+namespace Aws\DynamoDb;
+
+use GuzzleHttp\Stream\StreamInterface;
+
+/**
+ * Marshals and unmarshals JSON documents and PHP arrays into DynamoDB items.
+ */
+class Marshaler
+{
+    /** @var array Default options to merge into provided options. */
+    private static $defaultOptions = [
+        'ignore_invalid'  => false,
+        'nullify_invalid' => false,
+        'wrap_numbers'    => false,
+    ];
+
+    /** @var array Marshaler options. */
+    private $options;
+
+    /**
+     * Instantiates a DynamoDB Marshaler.
+     *
+     * The following options are valid.
+     *
+     * - ignore_invalid: (bool) Set to `true` if invalid values should be
+     *   ignored (i.e., not included) during marshaling.
+     * - nullify_invalid: (bool) Set to `true` if invalid values should be set
+     *   to null.
+     * - wrap_numbers: (bool) Set to `true` to wrap numbers with `NumberValue`
+     *   objects during unmarshaling to preserve the precision.
+     *
+     * @param array $options Marshaler options
+     */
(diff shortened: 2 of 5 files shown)


jeremeamia:
@Sazpaimon you might be interested in this.

src/DynamoDb/SetValue.php
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @return array
+     */
+    public function getValues()
+    {
+        if (empty($this->values)) {
+            throw new \RuntimeException('DynamoDB does not allow empty sets.');
+        }
+
+        return array_map('strval', array_keys($this->values));
```
mtdowling: Is the `strval` necessary here? Aren't the keys always strings?

src/DynamoDb/BinaryValue.php
```diff
@@ -0,0 +1,35 @@
+<?php
+namespace Aws\DynamoDb;
+
+use GuzzleHttp\Stream\Stream;
+use GuzzleHttp\Stream\StreamInterface;
```
mtdowling: Unused use here

src/DynamoDb/Marshaler.php
```diff
@@ -0,0 +1,236 @@
...
+ */
+class Marshaler
+{
+    /**  @var callable Called if the marshaler encounters an invalid value. */
+    private $safetyNetFn;
+
+    /**
+     * Instantiates the DynamoDB Marshaler.
+     *
+     * The `$safetyNetFn` allows the user to provide custom logic to handle
```
mtdowling: Could this be called something else like "errorHandler"?

src/DynamoDb/Marshaler.php
```diff
@@ -0,0 +1,236 @@
...
+     * invalid values (e.g., an empty string). The signature of this function
+     * should look like `function (&$type, &$value) : bool`. It should modify
+     * `$type` and `$value` as needed in order to marshal the value to a valid
+     * DynamoDB attribute. The function should return `true` when successful,
+     * or `false` on failure, so that the normal error state can be reached.
```
mtdowling: Does this need to pass by reference and return true/false? Could it just return a new value?

src/DynamoDb/Marshaler.php
jeremeamia: Sure.

src/DynamoDb/Marshaler.php
jeremeamia: Well, you need to be able to change the type and value. Maybe it could return the same type of array that marshalValue returns or null. That might be better.

src/DynamoDb/SetValue.php
jeremeamia: You would think that, but PHP treats numerical keys as numbers no matter how hard you try to make them strings.

mtdowling:
Should there also be a Number wrapper that ensures a value is serialized as a number, but the actual value can be internally stored as a string so as to not lose precision?

jeremeamia:
@mtdowling pushed a new commit with a `NumberValue` wrapper and a few other changes. Let me know what you think.

src/DynamoDb/Marshaler.php
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Instantiates a DynamoDB Marshaler.
+     *
+     * The following options are valid.
+     *
+     * - skip_invalid: (bool) Set to `true` if invalid values should be skipped
+     *   (i.e., not included) during marshaling.
+     * - wrap_numbers: (bool) Set to `true` to wrap numbers with `NumberValue`
+     *   objects during unmarshaling to preserve the precision of numbers.
+     * - invalid_handler: (callable) Allows user to provide custom logic to
```
mtdowling: With the addition of `ignore_invalid` is an `invalid_handler` still necessary?

jeremeamia: Well, they are 2 solutions to the same problem. With ignore_invalid, they are simply dropped from the marshaled data. With invalid_handler, you can massage the data (e.g., convert empty strings to null values). I'm not sure I would want to remove one or the other. What do you think?

src/DynamoDb/SetValue.php
```diff
@@ -0,0 +1,111 @@
+<?php
+namespace Aws\DynamoDb;
+
+use GuzzleHttp\ToArrayInterface;
+
+/**
+ * Special object to represent a DynamoDB set (SS/NS/BS) value.
+ */
+class SetValue implements \JsonSerializable, \ArrayAccess, \Countable, \IteratorAggregate, ToArrayInterface
```
mtdowling: Looking at how Set works now, I wonder if we actually should make it immutable. This would push validation into the constructor of the Set, cast each value to string (which it looks like is a requirement), and we could drop getValues(). Also note that the other wrappers are immutable.

jeremeamia:
@mtdowling Updated SetValue stuff to make it immutable. Take another look. If you are good with this now, I'll rebase and merge.

mtdowling:
:shipit:

PR #473, opened 2015-01-29
Add pre-signed URLs for CopySnapshot

mtdowling:
I decided to pass in a serializer and endpoint provider into the subscriber rather than have the subscriber create a new client with a different endpoint and signer (like we did with v2).

### DIFF
--- a/src/Ec2/CopySnapshotSubscriber.php
+++ b/src/Ec2/CopySnapshotSubscriber.php
@@ -1,0 +1,68 @@
+<?php
+namespace Aws\Ec2;
+
+use Aws\AwsClientInterface;
+use Aws\Signature\SignatureV4;
+use GuzzleHttp\Command\CommandInterface;
+use GuzzleHttp\Command\CommandTransaction;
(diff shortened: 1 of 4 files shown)


mtdowling:
@jeremeamia You happy with this? Good to merge?

src/Ec2/CopySnapshotSubscriber.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private function createPresignedUrl(
+        AwsClientInterface $client,
+        CommandInterface $cmd
+    ) {
+        $newCmd = $client->getCommand('CopySnapshot', $cmd->toArray());
+        // Serialize a request for the CopySnapshot operation.
+        $trans = new CommandTransaction($client, $newCmd);
+        $request = call_user_func($this->requestSerializer, $trans);
```
mtdowling: Note: I used `call_user_func` here because it doesn't cause significant overhead and it circumvents the need to use a temporary variable.

jeremeamia:
This looks good. :ship:

PR #474, opened 2015-01-31
Adding bucket endpoint functionality.

mtdowling:
If you provide "bucket_endpoint" => true when creating an S3Client, then
requests are sent to a bucket-specific endpoint, allowing users to send
requests and create pre-signed URLs for CNAME style buckets, local S3
mocks.


jeremeamia:
That looks good! Cool.

PR #478, opened 2015-02-03
Adding signature providers.

mtdowling:
This commit updates clients to no longer provide a getSignature()
method, but rather utilizes an internal signature provider, that
provided a version, service, and region returns a SignatureInterface
object used to sign requests. This allows customers to inject custom
signature providers, allows customers to provide a signature version
override to use when signing requests (e.g., for S3, use v4 instead),
and allows clients to vary the signature version on a per/operation
basis (e.g., this is already a possibility as some operations are not
signed while others are signed).

This commit also updates the SDK to use a few recent Guzzle Command
changes: client options are now immutable, so some of the setOption
calls have been refactored to use a new "config" client factory
option type, which means that the option will be passed through to
the getConfig() array of a client. For example, "signature_version"
is now always available in the getConfig() array and is the
preferred signature version of the client. This information is used
in the SDK in places like the Multipart upload abstraction to know
which type of hash to add to an upload.

### DIFF
--- a/src/Ec2/CopySnapshotSubscriber.php
+++ b/src/Ec2/CopySnapshotSubscriber.php
@@ -4,7 +4,6 @@ namespace Aws\Ec2;
 use Aws\AwsClientInterface;
 use Aws\Signature\SignatureV4;
 use GuzzleHttp\Command\CommandInterface;
-use GuzzleHttp\Command\CommandTransaction;
 use GuzzleHttp\Command\Event\InitEvent;
 use GuzzleHttp\Event\RequestEvents;
 use GuzzleHttp\Event\SubscriberInterface;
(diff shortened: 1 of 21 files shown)


src/Ec2/CopySnapshotSubscriber.php
```diff
@@ -47,9 +42,9 @@ private function createPresignedUrl(
         CommandInterface $cmd
     ) {
         $newCmd = $client->getCommand('CopySnapshot', $cmd->toArray());
+        $newCmd->getEmitter()->detach($this);
         // Serialize a request for the CopySnapshot operation.
-        $trans = new CommandTransaction($client, $newCmd);
-        $request = call_user_func($this->requestSerializer, $trans);
+        $request = $client->initTransaction($newCmd)->request;
```
mtdowling: I've used `initTransaction` here now that it's public.

jeremeamia:
:ship: Looks good.

PR #480, opened 2015-02-05
Client constructor refactor

mtdowling:


jeremeamia:
Nice! I like where this is going.

GrahamCampbell:
Me too! :)

mtdowling:
Ok, all done. This PR is a huge refactor of the client factory system such that it is now gone. The goal of this PR is to allow you to call `new S3Client($args)` instead of having to use the factory method (note: the factory method is still available for BC).

Client constructor arguments are declaratively specified per/client using a static `getArguments()` method. Custom factories are now gone in favor of simply overriding the client constructor and/or modifying the return value of `getArguments()` (for example, you can make "endpoint" required for CloudSearchDomain and provide a custom default "region" value using a closure).

Constructor error messages are much more verbose and helpful now. For example, if you provide an invalid type to a specific setting, it'll tell you:

    Invalid configuration value provided for "region". Expected string, but got int(-2)'

When required arguments are missing, it will describe the argument so that the customer doesn't need to look anything up, but rather can just make the necessary changes using the error message:

```
Fatal error: Uncaught exception 'InvalidArgumentException' with message ' in /Users/dowling/projects/sdk3/src/ClientResolver.php on line 449

InvalidArgumentException: Missing required client configuration options: 

region: (string)

  A "region" configuration value is required for the "dynamodb" service
  (e.g., "us-west-2"). A list of available public regions and endpoints can be
  found at http://docs.aws.amazon.com/general/latest/gr/rande.html;

version: (string)

  A "version" configuration value is required. Specifying a version constraint
  ensures that your code will not be affected by a breaking change made to the
  service. For example, when using Amazon S3, you can lock your API version to
  "2006-03-01".
  
  Your build of the SDK has the following version(s) of "dynamodb":
  * "2012-08-10"
  
  You may provide "latest" to the "version" configuration value to utilize the
  most recent available API version that your client's API provider can find
  (the default api_provider will scan the /Users/dowling/projects/sdk3/src/data
  directory for *.api.php and *.api.json files). Note: Using 'latest' in a
  production application is not recommended.
  
  A list of available API versions can be found on each client's API documentation
  page: http://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html. If you are
  unable to load a specific API version, then you may need to update your copy of
  the SDK. in /Users/dowling/projects/sdk3/src/ClientResolver.php on line 449
```

I know it's hard to review huge diffs like this... sorry. @jeremeamia  @GrahamCampbell: does this look OK? Note: we'll be updating the docs/ directory soon in a different PR.

mtdowling:
I'm going to go ahead and merge this. We can fix any issues that need to be addressed in the v3 branch before we push out an RC.

jeremeamia:
:+1: Yeah, this looks great.

benmadin:
Hi,

We seem to have lost track of how to instantiate a new instance - service builder no longer works for us since the latest update, with \Aws\Common\Aws no longer available (replaced by \Aws\Sdk)

When we try something like :

```php
use Aws\Ec2\Ec2Client;
$aws = new Ec2Client([
    'region'  => 'ap-southeast-1',
    'version' => 'latest',	
    'key'    => 'AKWI49IFK3FKSDLWO3<-NOT!',
    'secret' => '1t_i5_a_working_secret',
    ]
);
```

We are getting *Aws\Exception\UnresolvedCredentialsException* 'Could not load credentials'

I'm sure we've missed something...! but I'm not even sure where in the api we find out way through this?

cheers

Ben

PR #487, opened 2015-02-13
Adding in the ObjectNotExist waiter dsl as it comes in handy for pages i...

awlodarski:
...n which we delete an object and attempt to use the resource on the same page.

Change-Id: Ib3fe817122eebd19c26f0eb0476455009c9088d2


jeremeamia:
Cool. Can you amend the commit to change your tabs to 4 spaces to match the rest of the file?

PR #488, opened 2015-02-13
Updating the FilesystemApiProvider to use a manifest file

jeremeamia:
Updating the `FilesystemApiProvider` to use a manifest file generated from metadata in the service models, to support specifying past API versions not shipped in the SDK.

For example, in the following code, all 3 EC2 client objects will use the 2014-10-01 API version. The resolution to that version happens via a simple lookup in the manifest instead of having to glob the `src/data/` directory at runtime.

```php
$client1 = new Ec2Client([..., 'version' => 'latest', ...]);
$client2 = new Ec2Client([..., 'version' => '2014-10-01', ...]);
$client3 = new Ec2Client([..., 'version' => '2014-09-01', ...]);
```


src/Api/ApiProvider.php
```diff
@@ -0,0 +1,217 @@
...
+     * @throws UnresolvedApiException
+     */
+    public static function resolve(callable $provider, $type, $service, $version)
+    {
+        $result = $provider($type, $service, $version);
+        if (is_array($result)) {
+            return $result;
+        }
+
+        throw new UnresolvedApiException($service
```
mtdowling: Can you include the $type value in the message as well?

src/Api/ApiProvider.php
```diff
@@ -0,0 +1,217 @@
...
+     *   'ecs' => [...],
+     *   'elasticache' => [...],
+     *   ...
+     * ]
+     *
+     * @param array $manifest The API version manifest data.
+     *
+     * @return callable
+     */
+    public static function manifest(array $manifest, $dir = null)
```
mtdowling: What would you think about making $dir required here, and the `defaultProvider` function passes in the relative directory pointing to the models shipped with the SDK?

src/Api/ApiProvider.php
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * Retrieves a list of valid versions for the specified service.
+     *
+     * @param string $service Service name
+     *
+     * @return array
+     */
+    public static function getServiceVersions($service)
```
mtdowling: Is there anything we can do to show that the versions returned from this function are specific to the manifest file?

jeremeamia: I wasn't sure what to do with this, since there are no instances of providers any more. Any specific ideas?

src/ClientResolver.php
```diff
@@ -425,7 +425,7 @@ public static function _apply_validate($value, array &$args, EmitterInterface $e
 
     public static function _default_api_provider()
     {
-        return new FilesystemApiProvider(__DIR__ . '/data');
+        return ApiProvider::defaultProvider();
```
mtdowling: I think this can be removed in favor of providing the string "Aws\Api\ApiProvider::defaultProvider" to the "fn" attribute of this option.

src/Api/ApiProvider.php
mtdowling: One idea might be to make a ManifestApiProvider class that has the static function getApiVersions().

src/ClientResolver.php
jeremeamia: Ah, nice!

mtdowling:
I like the changes. Is there a reason why there isn't something like a ManifestProvider and a separate FilesystemProvider?

jeremeamia:
They share so much logic, that it didn't seem necessary to separate them. And patterned after the other Provider objects in the SDK, the `ApiProvider` has static methods that return a callable.

mtdowling:
I want us to look into possibly not having the build step of creating .php files for the models and instead lazily create .php files in a cache directory if an API provider is wrapped in a caching function. With the caveat that going that route might require us to refactor some of this (and assuming it's performant enough): LGTM

PR #492, opened 2015-02-16
Removing ringphp_handler option in favor of allowing a callable client option

mtdowling:
This PR allows the `client` option to be a callable that returns a Client instance. This removes the ringphp_handler argument which helps to better decouple the SDK from a Guzzle implementation detail.


jeremeamia:
Nice. I like this change. :ship:

PR #501, opened 2015-03-13
Reimplemented waiters using promise coroutines.

jeremeamia:
Here's the waiter's coroutine implementation. Unit and integ tests (at least for the waiters themselves) are passing with 100% coverage. I wanted to get this up here so you can see it.

#### Blocking Usage

```php
$client->waitUntil('TableExists', ['TableName' => 'Foo']);
```

#### Non-blocking Usage

```php
$promise = $client->waitUntil('TableExists', [
    'TableName' => 'Foo',
    '@future' => true
]);
```

### DIFF
--- a/src/AwsClient.php
+++ b/src/AwsClient.php
@@ -258,16 +258,17 @@ class AwsClient implements AwsClientInterface
         return new ResultPaginator($this, $name, $args, $config);
     }
 
-    public function getWaiter($name, array $args = [], array $config = [])
+    public function waitUntil($name, array $args = [], array $config = [])
     {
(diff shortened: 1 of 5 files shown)


mtdowling:
This looks good.

I liked having a getWaiter() and a waitUntil() method though. It would work well with your changes such that getWaiter() would return the waiter object and not initiate anything.  You'd then call promise() to kick off the async waiting and could `then` off of it. waitUntil() would call `getWaiter(...)->promise()->wait();`.

mtdowling:
I'll go ahead and merge this. We can discuss getWaiter() or waitUntil() later.

PR #503, opened 2015-03-16
Removing `@future` from commands.

mtdowling:
This commit removes the `@future` system from the SDK. Using `@future`
to change how the result of a command behaves was awkward because it
conditionally changed error handling semantics of an operation from
thenning to try/catching. It is also cumbersome to have to constantly
set `@future` to true for each async operation.

This commit updates the SDK to use separate methods for synchronous and
asynchronous calls. I added `executeAsync` to execute a command
asynchronously, and added magic methods that are suffixed with "Async"
to send commands asynchronously (e.g., PutObjectAsync).

This change made it no longer necessary to have a ResultPromise object
as the async methods now just return promises. I've removed executeAll
from AwsClientInterface in favor of adding a CommandPool object (in a
future PR).


jeremeamia:
Cool! LGTM. :smile:

PR #505, opened 2015-03-16
Adding a CommandPool abstraction

mtdowling:


jeremeamia:
The only thing I'm not sure about is "limit", which was kind of confusing. I thought, at first, that "limit" meant the total limit, and had to read through everything to see that it meant the limit of one batch.

mtdowling:
The "limit" attribute stands for the limit of the number of outstanding promises. It's based on async.js: https://github.com/caolan/async#eachLimit. The abstraction used in the SDK is a light abstraction on top of the Guzzle promises each abstraction which also uses "limit". I like limit personally, but if we were to change it, what would you suggest?

jeremeamia:
I just think it's confusing since PHP devs are used to thinking about limit like a SQL LIMIT. It's probably not a big deal. Other terms we've used in the past include `size`, `pool_size`, `batch_size`, and `concurrency`; but I'm not sure if those are better. Let's just leave it as limit.

PR #511, opened 2015-03-18
Adding retries

mtdowling:
- Removed retry_logger option. This can be done using a custom
  middleware if needed.
- Updating the XML error parser to set the error code to "NotFound" when
  there is no body and the status code is 404. Removed custom S3 error
  parsing for bucket vs object not found.
- Updated DynamoDB to uses a response parser that ensures the response
  x-amz-crc32 header is valid.


jeremeamia:
Looks good!

PR #525, opened 2015-03-30
Made multipart uploads async-compatible

jeremeamia:
Integration tests pass on GuzzleV5. Still working on unit test updates.


jeremeamia:
@mtdowling Code and tests are complete now, but some are failing on Travis (looking into it). Ready for CR if you haven't gotten around to it yet.

src/Glacier/MultipartUploader.php
```diff
@@ -0,0 +1,207 @@
...
+            });
+    }
+
+    protected function determinePartSize()
+    {
+        // Make sure the part size is set.
+        $partSize = $this->config['part_size'] ?: $this->info['part']['min_size'];
+
+        // Calculate list of valid part sizes.
+        static $validSizes;
```
mtdowling: If this is static, doesn't it leave a reference to this uploader around forever?

jeremeamia: Probably. Stupid closure bindings. If I make the closure static, it should fix that, right?

mtdowling: You can't because it's using $this in the closure. I'd just not make a static cache of the closure.

jeremeamia: Oh wait, it's not storing the closure. It's calling `array_map` and just storing the result.

mtdowling: ah right... Sorry! How many valid sizes are there? Is there a reason to compute them at runtime?

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7;
+use InvalidArgumentException as IAE;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamableInterface as Stream;
+
+/**
+ * Encapsulates the execution of a multipart upload to S3 or Glacier.
+ */
+abstract class AbstractUploader
```
mtdowling: What about making this implement PromisorInterface so that calling `promise()` will start an async upload, and calling cancel on the promise will abort the upload? That could possibly remove the need for an uploadAsync method.

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+    protected $info = [];
+
+    /** @var PromiseInterface Promise that represents the multipart upload. */
+    protected $promise;
+
+    /** @var Stream Source of the data to be uploaded. */
+    protected $source;
+
+    /** @var UploadState State used to manage the upload. */
+    protected $state;
```
mtdowling: There's some weirdness around $this->state. It's either a promise or an UploadState object. Because this is a protected property, I would suggest making this always an UploadState or always a promise. I would suggest making it always a promise. When synchronously calling `getState()`, just return the wait result of the promise. By making this always a promise, you can remove a lot of duplicated conditional checks and make subclasses have to know less about how this works.

Also note that the functions that yield the upload state can just always yield the value regardless of if it's a promise or not.

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+     *
+     * @return UploadState
+     */
+    public function getState()
+    {
+        if ($this->state instanceof Promise\PromiseInterface) {
+            $this->state = $this->state->wait();
+        }
+
+        return $this->state;
```
mtdowling: This could just become `return $this->state->wait()`

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+    public function uploadAsync()
+    {
+        if ($this->promise) {
+            return $this->promise;
+        }
+
+        return $this->promise = Promise\coroutine(function () {
+            // Determine the state of the upload
+            if ($this->state instanceof Promise\PromiseInterface) {
+                $this->state = (yield $this->state);
```
mtdowling: You can yield regular values and promises. I suggest yielding regardless of it's a promise or not (but I also suggest always making it a promise).

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+     */
+    public function abortAsync()
+    {
+        if ($this->promise) {
+            return $this->promise;
+        }
+
+        return $this->promise = Promise\coroutine(function () {
+            if ($this->state instanceof Promise\PromiseInterface) {
+                $this->state = (yield $this->state);
```
mtdowling: Same as above: could yield always

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,455 @@
...
+        $id = [$uploadIdParam => $this->config['upload_id']];
+        foreach ($required as $key => $param) {
+            if (!$this->config[$key]) {
+                throw new IAE('You must provide a value for "' . $key . '" in '
+                    . 'your config for the MultipartUploader for '
+                    . $this->client->getApi()->getServiceFullName() . '.');
+            }
+            $id[$param] = $this->config[$key];
+        }
+        $this->state = new UploadState($id);
```
mtdowling: Could become `$this->state = Promise\promise_for(new UploadState($id));`

mtdowling:
Lookin good! I really like that there isn't an upload builder any more. I felt like that added a lot of complexity and state that wasn't necessary.

src/Multipart/AbstractUploader.php
jeremeamia: I use `$this->state->{methodOnUploadState}` everywhere though, so it needs to be an `UploadState`. Maybe I should track the promise that resolves to the `UploadState` in a different variable?

src/Multipart/AbstractUploader.php
jeremeamia: Aborting as a cancel action doesn't make sense. It's a decision you make. For example, you might create a Uploader object for the sole purpose of aborting the upload. That is why I made upload/uploadAsync and abort/abortAsync. Trying to combine them is very weird. I at least chose the same convention as our client objects though, with the async suffix.

src/Glacier/MultipartUploader.php
jeremeamia: 13 items. I'll just go head and precompute them.

mtdowling:
:shipit:

src/Glacier/MultipartUploader.php
```diff
@@ -0,0 +1,241 @@
...
+use Aws\PhpHash;
+use Aws\ResultInterface;
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7;
+use Psr\Http\Message\StreamableInterface as Stream;
+
+/**
+ * Encapsulates the execution of a multipart upload to Glacier.
+ */
+class MultipartUploader extends AbstractUploader
```
mtdowling: Can you document how to pass custom parameters to the different operations?

src/Multipart/AbstractUploader.php
```diff
@@ -0,0 +1,388 @@
...
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7;
+use InvalidArgumentException as IAE;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamableInterface as Stream;
+
+/**
+ * Encapsulates the execution of a multipart upload to S3 or Glacier.
+ */
+abstract class AbstractUploader implements Promise\PromisorInterface
```
mtdowling: Is there a way to pass custom parameters to each operation? If so, can this be documented here?

src/S3/MultipartUploader.php
```diff
@@ -0,0 +1,202 @@
...
+use Aws\PhpHash;
+use Aws\ResultInterface;
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7;
+use Psr\Http\Message\StreamableInterface as Stream;
+
+/**
+ * Encapsulates the execution of a multipart upload to S3 or Glacier.
+ */
+class MultipartUploader extends AbstractUploader
```
mtdowling: Same as the Glacier class: can we document how parameters work here?

src/Glacier/MultipartUploader.php
jeremeamia: Added to the constructor.

src/Multipart/AbstractUploader.php
jeremeamia: There is in the `params` option. This is now documented in each concrete MUP.

src/S3/MultipartUploader.php
jeremeamia: Added to the constructor

src/Glacier/MultipartUploader.php
```diff
@@ -0,0 +1,278 @@
...
+
+        return $state;
+    }
+
+    /**
+     * Creates a multipart upload for a Glacier archive.
+     *
+     * The valid configuration options are as follows:
+     *
+     * - account_id: (string, required, default=string('-')) Account ID for the
```
mtdowling: Why does account_id need special treatment here? I thought that it was added when creating a Glacier command: https://github.com/aws/aws-sdk-php/blob/abstract-http/src/Glacier/GlacierClient.php#L42

jeremeamia: It's needed so it can be overridden if the archive is for an account other than the one performing the operations, but I updated the docblock to remove "required" and added a clarification.

jeremeamia:
Gonna merge now. We can make other updates if needed, but this is looking good.

PR #526, opened 2015-04-02
Updating the SDK to use JSON files directly.

mtdowling:
This commit updates the SDK to use the JSON models directly so that a
pre-build step is not required. The models that were added in this
commit are taken from the aws/aws-sdk-ruby repository and renamed to
match the new hierarchical directory format of service/version/*.

JSON files are first checked if they are cached in the opcode cache. If
so, then the file is loaded directly. If not, then the JSON data is
parsed, saved to disk as a PHP file for subsequent cache hits, and the
JSON data is returned. The cached JSON files are stored in the `cache/`
directory of the SDK by default, but this can be customized using the
`AWS_PHP_CACHE_DIR` environment variable.

Because we are now using the JSON files directly various JSON to PHP
build scripts have been removed.

@jeremeamia, @trevorrowe


mtdowling:
The diff times out when viewing in the changes tab. You can see the full diff here: https://github.com/aws/aws-sdk-php/pull/526.diff or browse the source here: https://github.com/aws/aws-sdk-php/tree/json-directly

Note: the models are stored in the "src/data" directory.

mtdowling:
Trevor reviewed this and said the model structure looks good. I'm going to merge this. We can iterate on the feature if needed. It's pretty much unreviewable due to the model updates.

jeremeamia:
:+1:

PR #535, opened 2015-04-17
Fix assuming tests are run from the project root.

deviantintegral:
When running individual tests in PHPStorm, the cwd is set to the path of the file you are testing. This fixes not being able to resolve `test_services.dist.json`.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -54,7 +54,7 @@ if (!isset($_SERVER['PREFIX']) || $_SERVER['PREFIX'] == 'hostname') {
 }
 
 // Instantiate the service builder
-$aws = Aws\Common\Aws::factory(isset($_SERVER['CONFIG']) ? $_SERVER['CONFIG'] : (__DIR__) . '/../test_services.dist.json');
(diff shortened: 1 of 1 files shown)


tests/bootstrap.php
```diff
@@ -54,7 +54,7 @@
 }
 
 // Instantiate the service builder
-$aws = Aws\Common\Aws::factory(isset($_SERVER['CONFIG']) ? $_SERVER['CONFIG'] : 'test_services.dist.json');
+$aws = Aws\Common\Aws::factory(isset($_SERVER['CONFIG']) ? $_SERVER['CONFIG'] : (__DIR__) . '/../test_services.dist.json');
```
GrahamCampbell: you don't need brackets around `__DIR__`

mtdowling:
Looks good. As Graham suggested, we don't need to parens. If it's too much of a hassle, then it doesn't really matter.

deviantintegral:
That was easily the most taxing commit of the day.

mtdowling:
Hahaha :) thanks for the PR!

PR #536, opened 2015-04-17
Support caching url_stat() calls

deviantintegral:
I'm in the process of upgrading the <a href="https://www.drupal.org/project/amazons3">amazons3</a> module to the 2.x SDK. One of the features the module has is to cache stat calls. For example, Drupal will generate image derivatives on demand, which involves a few stat calls to see if the file exists, followed by uploading it. With this patch, and an attached persistent cache, that same process requires no API calls at all once the image derivative is uploaded.

* No stat caching: ~4.5s
* ArrayCache() stat caching: ~3s
* Drupal's default MySQL cache: ~0.05s

Of course, this assumes that no other clients are connecting to S3 directly, so the cache is off by default (and saves a Doctrine\Cache dependency). But, I wonder if there's a benefit in defaulting to using an ArrayCache()?

### DIFF
--- a/tests/Aws/Tests/S3/StreamWrapperTest.php
+++ b/tests/Aws/Tests/S3/StreamWrapperTest.php
@@ -18,6 +18,9 @@ namespace Aws\Tests\S3;
 
 use Aws\S3\S3Client;
 use Aws\S3\StreamWrapper;
+use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Common\Cache\PhpFileCache;
+use Guzzle\Cache\DoctrineCacheAdapter;
 use Guzzle\Http\Message\Response;
 use Guzzle\Http\EntityBody;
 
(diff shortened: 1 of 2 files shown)


tests/Aws/Tests/S3/StreamWrapperTest.php
```diff
@@ -591,6 +594,76 @@ public function testStatsRegularObjects()
...
+        $ts = strtotime('Tuesday, April 9 2013');
+        $this->setMockResponse($this->client, array(new Response(200, array(
+          'Content-Length' => 5,
+          'Last-Modified'  => gmdate('r', $ts)
+        ))));
+
+        clearstatcache('s3://bucket/key');
+        $stat = stat('s3://bucket/key');
+        clearstatcache('s3://bucket/key');
+        sleep(2);
```
deviantintegral: I really don't like this. I'd expected to be able to do at least a 1 second sleep, but on my local it looks like the "1 second" cache is valid up to around 1.5 seconds later (and isn't consistent). We could implement our own cache that extends ArrayCache() and implements subsecond cache lifetimes, but that seems bit overkill. Thoughts?

mtdowling:
> I'm in the process of upgrading the amazons3 module to the 2.x SDK.

Just to give you a heads up: we are actively working on a v3 of the PHP SDK that should be GA in the next month or so: https://github.com/aws/aws-sdk-php/tree/abstract-http. This branch will be merged into the v3 branch shortly, which will be merged into the master branch and tagged as 3.0.0 soon. It might be better to update the module to use our V3 instead as v2 will eventually be deprecated.

> One of the features the module has is to cache stat calls.

This is one of those things that I'd prefer just be added to our V3. We actually already added an LRU cache to our v3 stat calls: https://github.com/aws/aws-sdk-php/blob/abstract-http/src/S3/StreamWrapper.php.

When using our v3, the following code only sends a single request:

```php
<?php

$c = new Aws\S3\S3Client([
    'region'  => 'us-east-1',
    'version' => 'latest',
]);

$c->getHandlerList()->append('sign', Aws\Middleware::tap(function () {
    echo "Sending request\n";
}));

$c->registerStreamWrapper();

var_export(is_file('s3://t1234/test'));
var_export(is_file('s3://t1234/test'));
var_export(is_file('s3://t1234/test'));
```

In summary, we'd prefer to not add major features to v2 now that v3 is about to be released. Given that v3 supports caching stat calls by default, is it acceptable for you to use our v3 in this Drupal module?

deviantintegral:
That's great! I didn't realize the V3 branch was that close to release. And I'm kicking myself for not checking the abstract-http branch.

What would you think about changing the static array to Doctrine's ArrayCache(), so it can be swapped out? The cache methods are private, so I'd still have to patch the class to inject something else.

What's your definition of "soon"? I've got to have a beta out of the Drupal module next week, and need to be production ready in 3-4 weeks.

mtdowling:
> What would you think about changing the static array to Doctrine's ArrayCache(), so it can be swapped out?

It needs to be a static value because you need to be able to persist the cache between created instances of the stream wrapper. As for using Doctrine's array cache: what's the use case for this? I don't really want to expose the caching behavior publicly or add a dependency on a cache until there is a need.

> I've got to have a beta out of the Drupal module next week, and need to be production ready in 3-4 weeks, which puts me before the GA release. Do you know of anyone already using the v3 branch in production?

We are hoping to ship v3 shortly after PSR-7 is accepted. I can't commit to a hard deadline (because I don't know how PSR-7 will play out), but I'd estimate sometime between May 11 through the 15th.

> Do you know of anyone already using the v3 branch in production?

I'm not aware of anyone using it in production yet, but I think by mid-next week we will have the abstract http branch merged into the v3 branch and it will be relatively stable before our GA.

deviantintegral:
> It needs to be a static value because you need to be able to persist the cache

Yeah, that's why this PR defines the cache as a static CacheAdapterInterface.

> As for using Doctrine's array cache: what's the use case for this?

Drupal's image styles are generated on request, and not ahead of time. Many complex sites define dozens of image styles, but only a subset may be actually used for a given image. To regenerate an image, the derivative is deleted. Generating all derivatives in a queue is problematic because you're either blocking a site editor for around 4 seconds times the number of derivatives. Background queuing will cause images to 404 until they are generated.

The image module does a few stat calls to make sure that any directories and the image exist, which adds up to several S3 calls. By caching the stat calls across multiple clients, we can get a serious performance boost. Even in the simplest one-call test to see if an image exists, this cache is dropping an image derivative request from ~275ms to ~50ms.

(if only Lambda had events for 404s!)

So to answer your question; I don't really care about using an ArrayCache, but I just need to be able to pass in a CacheAdapterInterface so we can make the cache shared across requests.

mtdowling:
I'm adding support for an injectable cache to our v3. Will send a PR tomorrow.

mtdowling:
Support for cache stat's has now been added into our WIP v3: https://github.com/aws/aws-sdk-php/pull/548. Closing this in favor of this being a new feature in v3.

PR #538, opened 2015-04-19
Method name fix for creating PresignedUrl

fustundag:
If presigned url created for PutObject using V4 Signature ( e.g. V4 is required for eu-central-1 region ), url not working because presigned url created with 'GET' mehtod name.


mtdowling:
Thanks. I've taken your suggested fix, merged it, added tests, and added a case for changing the method to GET if the request is a application/x-www-form-urlencoded POST request.

PR #542, opened 2015-04-21
Add missing use statement for ResourceIteratorInterface

hacfi:


hacfi:
Found 2 more: ConfigServiceClient & KinesisClient

PR #543, opened 2015-04-21
Fix inconsist @var PhpDoc & use statements

hacfi:
Going through the code I noticed some inconsistencies.

See http://www.phpdoc.org/docs/latest/references/phpdoc/tags/var.html


jeremeamia:
Looks good. PHPStorm must be too forgiving about the ordering on the `@var` declarations.

PR #545, opened 2015-04-21
Fix S3 NoSuchTagSetException

hacfi:
This exception was never thrown because of the wrong name.

### DIFF
--- a/src/Aws/S3/Exception/NoSuchTagSetException.php
+++ b/src/Aws/S3/Exception/NoSuchTagSetException.php
@@ -19,4 +19,4 @@ namespace Aws\S3\Exception;
 /**
  * There is no TagSet associated with the bucket.
  */
-class NoSuchTagSetErrorException extends S3Exception {}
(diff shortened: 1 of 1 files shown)


jeremeamia:
How did you determine that this PR has the correct name?

hacfi:
Debugged it with XDebug and noticed it looked for a class that didn't exist.

jeremeamia:
Just reproduced this problem, and will accept the PR. Thanks.

Note: We do not encourage catching granular exceptions like this. See [this](https://github.com/aws/aws-sdk-php/blob/master/UPGRADING.md#preparations-for-version-3-of-the-sdk) for an explanation.

hacfi:
@jeremeamia Thanks for the hint! I only catch this exception during a `getBucketTagging` call to trigger a `putBucketTagging` request. Look forward to version 3 and guzzle 5!

jeremeamia:
Yep, that makes sense. Thanks for reporting these issues and providing PRs..

PR #548, opened 2015-04-22
Added support for injectable cache to stream wrapper.

### LINKED ISSUE #531: Specify the protocol name
Could Aws\S3\S3Client::registerStreamWrapper accept a protocol name?

To work with two clients is a lite hard. Having different (s3Custom1, s3Custom2) will help us.

mtdowling:
- You can now inject a cache into the stream wrapper in a contect or in
  the default context options.
- Added a CacheInterface for simple get/set/remove caching.
- Added a single cache implementation LruArrayCache that implements an
  LRU cache using PHP's ordered associative arrays.
- Added support for creating a stream wrapper using a custom protocol.

Closes #531. Closes #536.

Example:

```php
<?php

$s3 = new Aws\S3\S3Client([
    'region'  => 'us-east-1',
    'version' => 'latest',
    // use debug to ensure that no requests are sent.
    'debug'   => true
]);

// Create an LRU cache
$cache = new \Aws\LruArrayCache();
// Create a dummy return value.
$cache->set('s3://t1234/test', ['size' => 123, 7 => 123]);
// Register the stream wrapper with the cache and a protocol name.
\Aws\S3\StreamWrapper::register($s3, 's3', $cache);
// Show that we got the dummy size value.
var_dump(filesize('s3://t1234/test'));
```

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -441,18 +441,22 @@ class StreamWrapper
             return false;
         }
 
+        // First we need to create a cache key. This key is the full path to
+        // then object in s3: protocol://bucket/key.
+        // Next we need to create a result value. The result value is the
+        // current value of the iterator without the opened bucket prefix to
+        // emulate how readdir() works on directories.
+        // The cache key and result value will depend on if this is a prefix
+        // or a key.
         $cur = $this->objectIterator->current();
         if (isset($cur['Prefix'])) {
             // Include "directories". Be sure to strip a trailing "/"
             // on prefixes.
-            $prefix = rtrim($cur['Prefix'], '/');
-            $key = $this->formatKey($prefix);
-            $result = str_replace($this->openedBucketPrefix, '', $prefix);
-            $stat = $this->formatUrlStat($prefix);
+            $result = rtrim($cur['Prefix'], '/');
+            $key = $this->formatKey($result);
+            $stat = $this->formatUrlStat($key);
         } else {
-            // Remove the prefix from the result to emulate other
-            // stream wrappers.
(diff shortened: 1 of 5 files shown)


src/S3/StreamWrapper.php
```diff
@@ -609,14 +633,9 @@ private function openReadStream()
         $client = $this->getClient();
         $command = $client->getCommand('GetObject', $this->getOptions());
         $command['@http']['stream'] = true;
-
         $result = $client->execute($command);
         $this->body = $result['Body'];
 
-        if ($result['ContentLength']) {
```
mtdowling: This was dead code that isn't necessary

src/S3/StreamWrapper.php
```diff
@@ -693,8 +712,6 @@ private function formatUrlStat($result = null)
                 // Pluck the content-length if available.
                 if (isset($result['ContentLength'])) {
                     $stat['size'] = $stat[7] = $result['ContentLength'];
-                } elseif (isset($stat['Size'])) {
```
mtdowling: More dead code

tests/S3/StreamWrapperTest.php
```diff
@@ -772,27 +785,4 @@ public function testCanSetDelimiterStreamContext()
         $r = opendir('s3://bucket', $context);
         closedir($r);
     }
-
```
mtdowling: No longer relevant

jeremeamia:
¡Me gusto mucho! :+1:

deviantintegral:
This looks great. Thanks!

@mtdowling I'm backporting this to the 2.x branch just so what we're using isn't vastly different. Should I open a PR for that when I'm done, or would you rather keep this in v3 only?

PR #550, opened 2015-04-22
Adding RefreshableCredentialsInterface

mtdowling:
Adding a way to refresh credentials. This would be useful if you know that credentials have been revoked OR if you get an exception about invalid credentials before the credentials expire and want to force them to refresh.


jeremeamia:
Simple. :ship:

PR #551, opened 2015-04-23
Adding documentation override system

mtdowling:
This PR adds a documentation override system that allows us to modify the generated service docs for API clients by version.

Each client has a static `applyDocFilters` method which accepts an array of service and docs data and returns an array of service and doc data. This function can then change the data it is passed to add in parameters or modify documentation.

In order to easily achieve this, I've refactored the Service class to accept the data as an argument instead of having to invoke the provided api provider to get the service data. This not only simplified the code and mocking service's, it made filtering docs in this way possible.

In order to generate docs, I needed a way to map the endpointPrefix to the service namespace. I did this by calculating the namespace when generating the api manifest. I moved the list of versions into a "versions" key and added a top level "namespace" key. This allows us to get the namespace of a client and call the corresponding `applyDocFilters` method. Because it's no longer just a version manifest, I renamed the file to "manifest.json".


jeremeamia:
:ship:

PR #552, opened 2015-04-23
Adding regions to S3Client and removing PathStyle as a per/command param...

mtdowling:
...eter


jeremeamia:
LGTM. Should we add any notes about eu-central-1?

mtdowling:
I added more information about sigv4 only regions to the error message.

PR #554, opened 2015-04-24
Testing multiple clients and adding PHP 7. Closes #547

### LINKED ISSUE #547: Create TravisCI test matrix for v3 that uses Guzzle 5 and 6
We need to setup travis to somehow run our unit tests using Guzzle 5 and then Guzzle 6 for each supported PHP version.

mtdowling:


jeremeamia:
Cool!

PHP7...

.travis.yml
```diff
@@ -3,18 +3,24 @@ language: php
...
+  - COMPOSER_OPTS=""
+  - COMPOSER_OPTS="--prefer-lowest"
+
 matrix:
   allow_failures:
     - php: hhvm
+    - php: 7
   fast_finish: true
 
-sudo: false
```
GrahamCampbell: ?

mtdowling:
By counting the dots, I think it's segfaulting on test 90, which I think is Crc32ValidatingParserTest::testNothingWhenValidChecksum.

.travis.yml
mtdowling: I saw that in the flysystem .travis.yml file and copied it over. Does it not do anything?

GrahamCampbell: It tells travis to use the new container architecture which is way faster. Old repos don't have it enabled by default on travis, so you must include it for now if you want the container architecture.

PR #561, opened 2015-04-29
Return results for GetCheckerIpRanges

radford:
Use the correct element name when extracting CIDRs from a GetCheckerIpRangesResponse.


jeremeamia:
Thanks for bringing this to our attention. I've verified that this patch makes the operation work correctly. Since we generate service description files before putting them in the SDK, there may be a change we need to make to our upstream tools. We'll get back to you soon.

PR #562, opened 2015-04-29
Return results for GetCheckerIpRanges

radford:
This (untested) patch is analogous to #561, but for `v3`.


jeremeamia:
This is not needed for v3. It works fine as is. The service description formats are little different from v2.

PR #567, opened 2015-05-05
Updating waiters (and docs) and paginators.

mtdowling:
- Adding waiter documentation.
- Waiters are now fulfilled with the command that last succeeded. This
  provides a relatively useful indicator of which waiter succeeded when
  working with multiple promises and pairs well with the addition of the
  before argument.
- Waiters now accept a "before" callback that takes the command about
  the be executed and the number of attempts.
- Removed the "retry" callback from waiters.
- Custom waiter configs are now injected in the `@waiter` key value pair
  of the `$args` parameter for `getWaiter()` and `waitUntil()`.
- Removed the `$config` array from paginators as this does not have a
  valid use case.

### DIFF
--- a/docs/guide/waiters.rst
+++ b/docs/guide/waiters.rst
@@ -2,132 +2,155 @@
 Waiters
 =======
 
+
 Introduction
 ------------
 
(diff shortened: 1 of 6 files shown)


jeremeamia:
Looks good. Nice updates.

PR #568, opened 2015-05-07
Fix bool typecasting of request param values

lifeofguenter:
As via gitter.im - I had a problem with `ElasticBeanstalk::describeEnvironments` when passing through `IncludeDeleted => false` as param.

Currently it would typecast it into a string, which PHP will just make a empty string out of it. With this path it will check if value is a boolean and replace it with actual 'true' or 'false' strings.


mtdowling:
Cool thanks. I think this can be added to the double-dispatch though by creating a new method called `format_boolean` that matches the other double dispatch methods.

lifeofguenter:
Yes makes more sense - was actually searching for an example only to find them in the same file ;)

I changed the commit accordingly.

PR #569, opened 2015-05-08
[v3] Signature::presign and S3Client::getPresignedRequest

jeremeamia:
Updated the SignatureInterface to change createPresignedUrl to presign. Changed S3's createPresignedUrl to createPresignedRequest.

Consumers can get the presigned URL (and other information like headers) from the presigned request object. Formerly, only the URLs were returned, and getting the header values needed to actually make requests to the presigned URLs was not possible with custom middleware.


mtdowling:
This is a great change. Ship it.

PR #570, opened 2015-05-10
Making credential providers async.

mtdowling:
- Moving serialize from client to functions.php.
- Removed refreshable credentials. This is something that will be done
  in the memoized credential provider.
- Updated credential collaborators to use promise->wait()


src/S3/Transfer.php
```diff
@@ -338,6 +341,7 @@ private function addDebugToBefore($debug)
                     $source = "s3://{$command['Bucket']}/{$command['Key']}";
                     $dest = $command['@http']['sink'];
                     break;
+                case 'CompleteMultipartUpload':
```
mtdowling: This was missing from the debug list for some reason

src/S3/Transfer.php
```diff
@@ -277,7 +276,12 @@ private function createUploadPromise()
 
         // Create an EachPromise, that will concurrently handle the upload
         // operations' yielded promises from the iterator.
-        return $this->promise = Promise\each_limit($files, $this->concurrency);
+        return Promise\each_limit(
+            $files,
+            $this->concurrency,
+            null,
+            $this->getRejectFn()
```
mtdowling: This ensures that the first failed transfer fails the aggregate promise

src/S3/Transfer.php
```diff
@@ -254,7 +252,8 @@ private function createDownloadPromise()
             // Create a GetObject command pool and return the promise.
             return (new Aws\CommandPool($this->client, $commands, [
                 'concurrency' => $this->concurrency,
-                'before'      => $this->before
+                'before'      => $this->before,
+                'rejected'    => $this->getRejectFn()
```
mtdowling: This ensures that the first failed transfer fails the aggregate promise.

jeremeamia:
Reviewing this now.

src/S3/Transfer.php
jeremeamia: It's already there, a few lines down.

jeremeamia:
Looks good, except for the one comment I made on the S3 Transfer changes.

src/S3/Transfer.php
mtdowling: Nah, all that's there is:

- PutObject
- UploadPart
- CreateMultipartUpload
- CompleteUploadPart

Should I just replace `CompleteUploadPart` with `CompleteMultipartUpload`?

jeremeamia: Yep. Apparently, I can't read or write.

mtdowling:
I've fixed the CompleteMultipartUpload issue with the switch statement.

PR #572, opened 2015-05-11
Better failover for INI credentials.

jeremeamia:
Updates credential chain so that it fails over to instance profile credentials when ini credentials are invalid.

### DIFF
--- a/src/Aws/Common/Credentials/Credentials.php
+++ b/src/Aws/Common/Credentials/Credentials.php
@@ -143,11 +143,11 @@ class Credentials implements CredentialsInterface, FromConfigInterface
             $profile = self::getEnvVar(self::ENV_PROFILE) ?: 'default';
         }
 
-        if (!file_exists($filename) || !($data = parse_ini_file($filename, true))) {
+        if (!is_readable($filename) || ($data = parse_ini_file($filename, true)) === false) {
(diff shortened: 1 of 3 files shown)


src/Aws/Common/Credentials/Credentials.php
```diff
@@ -147,7 +147,7 @@ public static function fromIni($profile = null, $filename = null)
             throw new \RuntimeException("Invalid AWS credentials file: {$filename}.");
```
mtdowling: I think this block of code needs to be updated to not thrown when the credentials file is empty. `parse_ini_file()` could return `[]` on an empty file, which would cause an exception. Perhaps instead we should check if the result if `=== false`.

jeremeamia: Yeah, that makes sense. That way, the next block of code throws the exception, which is a more accurate message for that case. I'll update the PR.

jeremeamia: Updated.

mtdowling: just for readability, I'd probably move the assignment of parsing the ini file into it's own line and then add another conditional to ensure that it is not false.

mtdowling:
LGTM

PR #573, opened 2015-05-11
Temporary credentials with cross-region CopySnapshot

jeremeamia:
Fixes a bug in the EC2 CopySnapshotListener so it works with temporary credentials.


mtdowling:
:ship:

PR #576, opened 2015-05-13
Making the step name part of the method for HandlerList

mtdowling:
This commit updates the HandlerList to include the step name as part of
adding a middleware to a handler. This adds methods for each step
prefixed with "append" or "prepend". I believe this to be a better API
as it does not require string parsing (e.g., "init:foo") and makes the
steps explicit and auto-completable.


jeremeamia:
:+1:

PR #581, opened 2015-05-18
Removing S3's old Signature; use Signature Version 4 exclusively

jeremeamia:
* Removes `Aws\Signature\S3Signature` and related code and tests.
* Removes `'force_path_style'` S3 client option, because path_style will not be used exclusively.
* Removes `'calculate_md5'` option, because Signature Version 4 already does sha256 calculations.
* Simplifies several middleware and the multipart uploader due to these changes.
* Fixes issues with `LocationConstraint` in `GetBucketLocation`.


src/S3/BucketStyleMiddleware.php
```diff
@@ -15,28 +15,22 @@
 class BucketStyleMiddleware
 {
     private static $exclusions = ['GetBucketLocation' => true];
-    private $bucketEndpoint;
```
mtdowling: Why was this removed? Did it go somewhere else?

mtdowling:
I like it, but can you tell me how `$bucketEndpoint` is implemented?

src/S3/BucketStyleMiddleware.php
jeremeamia: Not needed because this middleware's sole reason to exist now is to handle bucket endpoints.

jeremeamia:
BucketStyle middleware only gets added if bucket_endpoint config is specified: https://github.com/aws/aws-sdk-php/pull/581/files#diff-4ca0826aef3c8f72db542dfae087e933R70

mtdowling:
Ah ok. If that's all it does, what would you think about renaming it to `BuckeEndpointMiddleware`? Other than that, :ship:

mtdowling:
:ship:

PR #584, opened 2015-05-21
Now converting timestamps to DateTimeResult objects

mtdowling:
This change updates the SDK to parse timestamp shapes into custom DateTimeResult objects. These objects extend DateTime and implement `__toString` and `JsonSerializable`. This allows the object to work with JMESPath as well.

This is a good change because it normalizes the timestamp result type across different protocols but still allows users to work with it as if it is an ISO 8601 date string.


jeremeamia:
Looks great! :ship:

Issue #587, opened 2015-05-22
getimagesize() fails when seeking is required


deviantintegral:
<a href="https://php.net/getimagesize">`getimagesize()`</a> is a GD function to return metadata for an image. If it seeks on an S3 stream, it fails for two reasons:

1. The function takes a URI, so we can't pass in any stream context options. The extension code is assuming that seeking is available, but the <a href="https://php.net/manual/en/streamwrapper.stream-seek.php">StreamWrapper docs explicitly state that seeking is not required to be supported</a>.
1. If we force seeking to be enabled in the stream wrapper, `getimagesize()` still fails because it seeks without first reading the stream. <a href="https://guzzle3.readthedocs.org/http-client/entity-bodies.html">CachingEntityBody documents that it can only seek on read bytes</a>, but this is an implementation detail we can't expect calling code to be aware of.

I can get a successful image size call if I hack in a few stream_read() calls inside of stream_seek(). Strangely, $length always seems to be in 8K segments even when I pass in another value, but I haven't dug any deeper.

I haven't isolated exactly where seeking becomes required, if it's based off of the internal JPEG structure or if it's simply due to file size. I could see this breaking for other image formats if they store metadata at the end of a large file.

At this point, this looks like two bugs to me - one in GD for not handling unseekable streams and falling back to loading the whole file into memory, and one in the SDK to handle calls to seek without previous reads.

deviantintegral:
Turns out this is triggered by some subset of EXIF info. When the <a href="https://github.com/php/php-src/blob/php-5.5.9/ext/standard/image.c#L535">M_APP15</a> JPEG marker is found, php_skip_variable() calls _php_stream_seek(). I was just able to trigger this issue with an 85KB image, while a 6MB image with no EXIF data works fine. It's not just "having EXIF" either - images from my Nexus 5 work fine, but images from my standalone camera fail.

mtdowling:
> The function takes a URI, so we can't pass in any stream context options. The extension code is assuming that seeking is available, but the StreamWrapper docs explicitly state that seeking is not required to be supported.

There's nothing we can do here in the SDK, but you could use the default stream options features from PHP to work around this: http://php.net/manual/en/function.stream-context-set-default.php

> If we force seeking to be enabled in the stream wrapper, getimagesize() still fails because it seeks without first reading the stream. CachingEntityBody documents that it can only seek on read bytes, but this is an implementation detail we can't expect calling code to be aware of.

[...]

PR #589, opened 2015-05-25
re #587: Ensure streams are read before seeking

deviantintegral:
See https://github.com/aws/aws-sdk-php/issues/587

If this approach is good, it still needs test coverage.


mtdowling:
Fixed in the caching stream. See https://github.com/guzzle/psr7/pull/23

PR #595, opened 2015-05-28
Using older composer version syntax to support older builds of composer

mtdowling:
This commit updates the SDK to use older composer.json syntax to support older builds of composer. This converts `^` usage to `~` and `||` usage to `|`. I'm not sure if `|` is exactly equivalent to `||`, but it seems to be based on testing it locally.

/cc @GrahamCampbell

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require": {
         "php": ">=5.5",
-        "guzzlehttp/guzzle": ">=5.3|>=6.0.1,<7",
+        "guzzlehttp/guzzle": ">=5.3|~6.0.1|~6.1",
         "guzzlehttp/psr7": "~1.0",
         "guzzlehttp/promises": "~1.0",
         "mtdowling/jmespath.php": "~2.2"
--- a/composer.json
+++ b/composer.json
@@ -17,10 +17,10 @@
     },
     "require": {
         "php": ">=5.5",
-        "guzzlehttp/guzzle": "^5.3||^6.0.1",
-        "guzzlehttp/psr7": "^1.0",
-        "guzzlehttp/promises": "^1.0",
-        "mtdowling/jmespath.php": "^2.2"
+        "guzzlehttp/guzzle": ">=5.3|>=6.0.1,<7",
+        "guzzlehttp/psr7": "~1.0",
+        "guzzlehttp/promises": "~1.0",
+        "mtdowling/jmespath.php": "~2.2"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: composer has supported this for over 6 months now.

mtdowling:
I appreciate that. Unfortunately we need to make this change for now.

I was hoping you could inform me on this: do you know if the syntax of `|` and `||` are equivalent?

GrahamCampbell:
I'd have to check the source to be 100% sure.

jeremeamia:
Could we do `>=5.3,<7`?

mtdowling:
Kind of, but it's really fragile. For example, what if we only wanted 6.1 and up because of a new required feature? Or 6.0.1 and up for a bug fix?

composer.json
```diff
@@ -17,10 +17,10 @@
     },
     "require": {
         "php": ">=5.5",
-        "guzzlehttp/guzzle": "^5.3||^6.0.1",
-        "guzzlehttp/psr7": "^1.0",
-        "guzzlehttp/promises": "^1.0",
-        "mtdowling/jmespath.php": "^2.2"
+        "guzzlehttp/guzzle": ">=5.3|>=6.0.1,<7",
```
hannesvdvreken: I would do [`~5.3||^6.0.1`](http://semver.mwl.be/#?package=guzzlehttp%2Fguzzle&version=~5.3%7C%7C%5E6.0.1) or [`~5.3||~6.0.1||~6.1`](http://semver.mwl.be/#?package=guzzlehttp%2Fguzzle&version=~5.3%7C%7C~6.0.1%7C%7C~6.1) if you want to avoid the caret sign.

GrahamCampbell: `~5.3||^6.0.1` wouldn't avoid it, and would result in a parse error.

GrahamCampbell: `~5.3||~6.0.1||~6.1` would work fine.

hannesvdvreken: what parse error?

GrahamCampbell: The one you get if you use a really old verion of composer...

hannesvdvreken: what you said is what I said then.

GrahamCampbell: Well, i can't remember the exact error off the top of my head. If you're interested try it out I guess.

mtdowling:
As per @GrahamCampbell's suggestion, I've update the version constraint to `">=5.3|~6.0.1|~6.1`

PR #602, opened 2015-06-01
Fixing MachineLearningClient to use PredictEndpoint

jeremeamia:
Fixing the Machine Learning client to use the provided PredictEndpoint as the host for the Predict operation.


mtdowling:
:ship:

Issue #604, opened 2015-06-03
SHA256 client mismatch server when using s3:// resource


stuartcarnie:
```php
$resource = fopen('s3://some/object', 'rb')
```

followed by

```php
$obj = [
    'ACL'    => 'public-read',
    'Body'   => $resource,
    'Key'    => '<key>',
    'Bucket' => '<bucket>',
];
$s3->putObject($obj);
```

Fails due to the client and server calculated SHA256 mismatching. When stepping through, the client does calculate the correct SHA256 when `\GuzzleHttp\Psr7\hash` is called, suggesting the data sent to the server is incorrect. I have not dug in to the stream wrappers, but I suspect an issue with the caching stream wrapper…

Worth noting that if I create a temporary stream via `fopen('php://temp)` and then `stream_copy_to_stream`, from the `$resource` to the temporary stream, the `putObject` call succeeds.

mtdowling:
When executing commands with the SDK (v3), you can use the `debug` option to get a much more detailed view into what's going on. To enable debug mode, pass the `debug` option in to the client's constructor:

```php
$s3 = new Aws\S3\S3Client([
    'region' => 'my-region',
    'version' => '2006-03-01',
    'debug' => true
]);
```

Can you please run your test with debug mode enabled and post the debug output?

stuartcarnie:
I added that, and received an error:

`fprintf(): 51 is not a valid stream resource` in `/var/www/develop/odin/core/vendor/guzzlehttp/guzzle/src/Handler/StreamHandler.php:367`:

Which is [here](https://github.com/guzzle/guzzle/blob/6d2bab840c2e1fb0b6dd6ac9b72009c6b74ba36f/src/Handler/StreamHandler.php#L367)

mtdowling:
I made several enhancements to the SDK based on this issue. Here's the related pull request: https://github.com/aws/aws-sdk-php/pull/605.

To solve your immediate issue, you will need to create a seekable stream with the PHP stream wrapper. This can be done through a stream context that sets `seekable` to true.

Here's a working example of how this would work (I also forced the use of the PHP stream wrapper as well in case you are using the stream wrapper as your client. Note that you could omit the `http_handler` option altogether and this will still work).

```php
<?php
require 'vendor/autoload.php';

$streamHandler = new \GuzzleHttp\Handler\StreamHandler();

$s3 = new \Aws\S3\S3Client([
    'region'  => 'us-east-1',
    'version' => 'latest',
[...]

PR #605, opened 2015-06-03
Various enhancements related to #604

mtdowling:
This commit makes a few enhancements to the SDK based on #604:

1. Context parameters like cache and client are now removed from the
   parameters that are sent when executing commands in the S3 stream
   wrapper. When these parameters were present, they resulted in
   lengthy debug messages with the TraceMiddleware, and even caused
   stream resources to enter an invalid state after dumping them.
2. Updated TraceMiddleware to better provide debug information about
   exceptions. These debug messages previously var_dump'd the exception,
   causing a massive dump to the debug output. This is now handled by
   extracting specific pieces of information from exceptions and
   formatting in a more readable way.
3. Added a try/catch to the signature version 4 signer such that if a
   stream claims it's readable, but creating the hash fails for some
   reason, and CouldNotCreateChecksumException will still be thrown,
   which will provide more context as to why the checksum failed to
   create.

### DIFF
--- a/src/Exception/CouldNotCreateChecksumException.php
+++ b/src/Exception/CouldNotCreateChecksumException.php
@@ -3,7 +3,7 @@ namespace Aws\Exception;
 
 class CouldNotCreateChecksumException extends \RuntimeException
 {
-    public function __construct($algorithm)
+    public function __construct($algorithm, \Exception $previous = null)
     {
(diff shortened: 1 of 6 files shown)


jeremeamia:
Cool. LGTM.

PR #606, opened 2015-06-04
Handle empty maps in JSON serializer

jeremeamia:
Added a check to the json serializers to ensure empty maps get serialized to {} instead of []

### DIFF
--- a/src/Api/Serializer/JsonBody.php
+++ b/src/Api/Serializer/JsonBody.php
@@ -69,11 +69,13 @@ class JsonBody
                 return $value;
 
             case 'map':
+                if (empty($value)) {
+                    return new \stdClass;
+                }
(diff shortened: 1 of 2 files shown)


src/Api/Serializer/JsonBody.php
```diff
@@ -74,6 +74,10 @@ private function format(Shape $shape, $value)
                     $v = $this->format($values, $v);
                 }
 
+                if (empty($value)) {
```
mtdowling: Could this be moved up to the beginning of the case statement and simply return a new `stdclass()`?

jeremeamia: good call

mtdowling:
:shipit:

PR #611, opened 2015-06-04
Always using Unix style paths for Windows compat.

### LINKED ISSUE #610: Wrong path normalization in JsonCompiler.php
Hello all,
I think I've found a problem in the normalize function of JsonCompiler.php.
Whenever the path given to normalize is in the form of '\\xxx\sd' for example in a VirtualBox Windows Server + Apache + network shared folder, the normalize functions spits out a resolved pattern in the form of '\xxx\sd' which is wrong and this raises an InvalidArgumentException since PHP cannot locate the right files (manifest.json / endpoints.json) and so on.

mtdowling:
Closes #610.

This commit updates the JsonCompiler class to normalize paths to always
use "/" instead of using DIRECTORY_SEPARATOR. This works for all SDK
usage because the SDK uses relative paths for everything, and any
absolute paths that would have been mangled through this transformation
(e.g., 'C:/path/to/sdk/foo/bar.json.php') are modified to remove the base
directory of the SDK from the path (e.g., 'foo/bar.json.php' ->
'foo_bar.json.php).

This commit also removes a duplicate call to normalizePath().

Phar testing has been updated to include loading a paginator.

Phars are now built and tests on every Travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,6 @@ install:
   - export AWS_SECRET_ACCESS_KEY=bar
   - travis_retry composer update $COMPOSER_OPTS --no-interaction --prefer-source
 
-script: make test
+script:
+  - make test
+  - [ -z "$COMPOSER_OPTS" ] && make package
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ integ:
 
 # Packages the phar and zip
 package:
-	time php build/packager.php $(SERVICE)
+	php build/packager.php $(SERVICE)
 
 guide:
 	cd docs && make html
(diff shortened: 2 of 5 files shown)


jeremeamia:
```
$ make package
time php build/packager.php 
make: time: Command not found
make: *** [package] Error 127

The command "make package" exited with 2.
```

mtdowling:
I've pushed changes that will eventually make the travis builds succeed, but it seems that Travis is not receiving any of my pushes to trigger new builds. If you review and give this a :shipit:, I'll merge.

jeremeamia:
LGTM me. Thanks. :shipit:

PR #613, opened 2015-06-05
Added the ability to add a second leading slash during path normalization

smagnaschi:
Added the ability to add a second leading slash in case the original path was a network path.


jeskew:
The JSON compiler has been moved to a build step, so you shouldn't be affected by the issue this was causing anymore. Please feel free to reopen if you have any question.

PR #615, opened 2015-06-07
AWS Bundle link for Symfony

seferov:


jeremeamia:
We may be releasing an official Symfony bundle soon, so I don't think we are going to accept this PR. However, your bundle looks nice. I added it to our forum: https://forums.aws.amazon.com/ann.jspa?annID=1976, and also mentioned it in a tweet.

PR #628, opened 2015-06-15
Add a manifest for AWS Config Service

jeskew:


.gitignore
```diff
@@ -4,8 +4,8 @@ build/artifacts
 phpunit.xml
 phpunit.functional.xml
 test_services.json
-Config
-./Makefile
+/Config
```
GrahamCampbell: ?

.gitignore
```diff
@@ -4,8 +4,8 @@ build/artifacts
...
+/Makefile
```
GrahamCampbell: ?

jeremeamia:
Awesome. You're official now. :ship:

.gitignore
jeskew: This line was meant to exclude a file at the project root named `Config` but was also excluding a folder from the AWS manifests for a service named `config`.

GrahamCampbell: Ah right. :)

.travis.yml
```diff
@@ -25,4 +25,4 @@ install:
 
 script:
   - make test
-  - [ -z "$COMPOSER_OPTS" ] && make package
```
mtdowling: Why this change? Was it not working before?

jeskew: Travis was failing to interpret the manifest file and wasn't testing any new builds.

mtdowling: Ah I see.

PR #631, opened 2015-06-16
Pass the same uri object url to the canned policy creation

Gator92:
When using the CloudFront component for url signing, if a url contains non RFC compliant characters, the url passed to the canned policy creation will not match the actual url generated by the GuzzleHttp\Psr7\Uri component, resulting in the generation of an invalid signed url. This pr simply passes the correct url, which may have characters encoded in some use cases, to the createCannedPolicy method.

### DIFF
--- a/src/CloudFront/UrlSigner.php
+++ b/src/CloudFront/UrlSigner.php
@@ -69,16 +69,16 @@ class UrlSigner
 
         // Get the real scheme by removing wildcards from the scheme
         $scheme = str_replace('*', '', $urlSections[0]);
+        $uri = new Uri($scheme . '://' . $urlSections[1]);
 
         if ($policy) {
             $isCustom = true;
         } else {
             $isCustom = false;
-            $policy = $this->createCannedPolicy($scheme, $url, $expires);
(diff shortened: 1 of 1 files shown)


jeskew:
As currently written, this will not handle all characters. Below the change you made, there is a separate call to `parse_str`, which `urldecode`s query keys and converts a few characters (like spaces and periods) to underscores.

Gator92:
@jeskew The goal here is to send the same url to the canned policy that is sent to the signing. So what you're saying is if the original url has a query string, that would have to be taken into consideration as well? The pull I made does not consider this, but is a step above sending the raw url to the canned policy creation. Will adjust and squash the pull so that the same url that is used for sigining is also used for the canned policy. Not sure if there are unit tests here but will append those as necessary.

jeskew:
@Gator92 The test case I tried to add was something along the lines of 'https://example.com/path.ext?query parameter=query value'. The query did not survive the call to `parse_str`, even when properly encoded.

jeskew:
This is waiting on https://github.com/guzzle/guzzle/pull/1261 to get finalized and merged into a release.

PR #633, opened 2015-06-16
Fixing an issue with path encoding in SigV4 that affects Cognito

jeremeamia:
Related to forum post: https://forums.aws.amazon.com/thread.jspa?threadID=182601

CR PLZ: @mtdowling @jeskew


mtdowling:
Do the integration tests all pass?
Can you add an integration for this for S3 and for Cognito?

jeremeamia:
@mtdowling New integ tests added for this. Other tests were passing as well (tried them all before the PR).

jeskew:
The tests look nice, too. :shipit:

PR #636, opened 2015-06-17
Add shared SDK cucumber tests to PHP repo

jeskew:


tests/Integ/SmokeContext.php
```diff
@@ -0,0 +1,363 @@
...
+    Context,
+    SnippetAcceptingContext
+{
+    protected static $configDefaults = [
+        'version' => 'latest',
+        'region' => 'us-east-1',
+        'profile' => 'integ',
+    ];
+
+    protected static $services = [
```
jeremeamia: You could use the [manifest file](https://github.com/aws/aws-sdk-php/blob/master/src/data/manifest.json) for this data.

jeremeamia:
Looks great aside from my one comment. Nice work on this. Thank you.

tests/Integ/SmokeContext.php
jeskew: The tags on the feature files aren't 1-to-1 matches to what's in the manifest file. I removed the exact matches from the latest commit but there's still a sizable config array.

jeremeamia: Gotcha. I'm working on something that will allow you to remove the rest of that array except the EFS region.

jeskew: I'll hold on this PR until #641 is in master.

jeskew: `\Aws\manifest` worked for everything except ElasticLoadBalancing, so I just changed the tag on that feature.

jeremeamia:
What are we waiting on for this PR?

PR #641, opened 2015-06-18
Use manifest file to resolve service names when creating clients

jeremeamia:
This PR:

- Centralizes the relationship between endpoint prefixes and namespaces, without the need to manually maintain the list in the SDK code.
- Makes it easier for users to create a client from the `Aws\Sdk`, because you don't need to know the casing, and specifying either the service's namespace or endpoint prefix will work.
- Reuses the existing manifest file.
- Fixes #635 a different way.

/cc @mtdowling @jeskew

### DIFF
--- a/src/Api/ApiProvider.php
+++ b/src/Api/ApiProvider.php
@@ -91,9 +91,7 @@ class ApiProvider
      */
     public static function defaultProvider()
     {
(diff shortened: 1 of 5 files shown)


jeskew:
:+1: This is great. I really wasn't looking forward to maintaining a separate alias => namespace map in the behat suite.

PR #647, opened 2015-06-24
Fixes #479. Affects serialization for specific ELB and CFN operations.

### LINKED ISSUE #479: Can't unset ELB policies using this SDK
It's doesn't seem possible currently to unset LB policies using this SDK (neither latest 2.7 nor master). We need to remove all policies from the LB.

[SDK spec](http://docs.aws.amazon.com/aws-sdk-php/latest/class-Aws.ElasticLoadBalancing.ElasticLoadBalancingClient.html#_setLoadBalancerPoliciesOfListener) implies ELBClient's "setLoadBalancerPoliciesOfListener" needs to be used for this (and indeed, can successfully be used from command line). The spec says that one needs to pass an empty list as `PolicyNames`.

HOWEVER this throws up an API exception "Policy Name cannot be empty":

	$elbClient->setLoadBalancerPoliciesOfListener(array(
		'LoadBalancerName' => $loadBalancer,
		'LoadBalancerPort' => $port,
		'PolicyNames' => array()
	);

If one tries setting `PolicyNames` to null, empty strings, or variants of thereof, Guzzle spits out validation error instead.

I've debugged the actual Guzzle HTTP request structures, and this SDK produces:

      string(33) "SetLoadBalancerPoliciesOfListener"
      'Version' =>
      string(10) "2012-06-01"
      'LoadBalancerName' =>
      string(21) "platformers-test-dev1"
      'LoadBalancerPort' =>
      int(80)
      'PolicyNames.member' =>
      string(0) ""

If I hack the magic special case for this API call in AwsQueryVisitor::resolveArray to do `$query['PolicyNames'] = ''` instead of what it currently does, then it produces something that seems to work with the API:

      'PolicyNames =>
      string(0) ""

(note ".member" is now gone).

Is it a bug with this SDK? Or am I just plain stupid and can't figure out the proper way to do this? :-)

Many thanks.

jeremeamia:
Fixes how empty lists are serialized in some cases.

/cc @mtdowling

### DIFF
--- a/src/Aws/Common/Command/AwsQueryVisitor.php
+++ b/src/Aws/Common/Command/AwsQueryVisitor.php
@@ -99,6 +99,9 @@ class AwsQueryVisitor extends AbstractRequestVisitor
         // For BC, serialize empty lists for specific operations
         if (!$value) {
             if (isset($serializeEmpty[$this->fqname])) {
+                if (substr($prefix, -7) === '.member') {
+                    $prefix = substr($prefix, 0, -7);
+                }
                 $query[$prefix] = '';
             }
             return;
--- a/tests/Aws/Tests/Common/Command/AwsQueryVisitorTest.php
+++ b/tests/Aws/Tests/Common/Command/AwsQueryVisitorTest.php
@@ -259,7 +259,15 @@ class AwsQueryVisitorTest extends \Guzzle\Tests\GuzzleTestCase
             'type' => 'object',
             'location' => 'aws.query',
             'properties' => array(
-                'test' => array(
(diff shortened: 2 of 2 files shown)


mtdowling:
Looks good, but should this replace the old test or add a new test?

jeremeamia:
I think replacing is fine, since this addition makes the test more complete.

mtdowling:
Ok. Maybe add another field with a sentAs that is consumed by the conditional `if (substr($prefix, -7) === '.member') {`? Just to prevent regressions. Other than that: :ship:

PR #657, opened 2015-06-25
Updating stream wrapper to check response status

### LINKED ISSUE #609: S3 StreamWrapper ignores HTTP error response code
If you don't have permissions to access a particular file, the function `file_get_contents()` returns the wrong content. Instead of the content, the data retrieved is just an error response like `<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>123456789</RequestId><HostId>garblebase64digook=</HostId></Error>`. I have debugged this to `openReadStream()`, which returns `true`, while the HTTP response says "403 Forbidden" (in `$this->body->customData->response_headers`).

I'm using version 2.8.8 of the SDK, btw.

mtdowling:
Closes #609

@jeremeamia @jeskew


jeremeamia:
Looks good.

Issue #650, opened 2015-06-24
aws kinesis put recrord throws exceptin


BardiaAfshin:
here's the php code snippet

```php
...
$result = $client->putRecords(array(
    // Records is required
    'Records' => array(
        array(
            // Data is required
            'Data' => 'string',
            'ExplicitHashKey' => 'string',
            // PartitionKey is required
            'PartitionKey' => 'string',
        ),
        // ... repeated
    ),
    // StreamName is required
    'StreamName' => 'string',
));
...
```

and the exception being returned
```log

Fatal error: Uncaught exception 'Aws\Kinesis\Exception\KinesisException' with message 'Error executing "PutRecords" on "https://kinesis.us-east-1.amazonaws.com"; AWS HTTP error: Client error: 400 SerializationException (client): 'string' can not be converted to a Blob: Base64 encoded length is expected a multiple of 4 bytes but found: 6 - {"__type":"SerializationException","Message":"'string' can not be converted to a Blob: Base64 encoded length is expected a multiple of 4 bytes but found: 6"}'
...

Process finished with exit code 255
```

jeremeamia:
What version of the SDK are you using?

BardiaAfshin:
Tested on 3.0 and 2.8




Same problem



—

jeremeamia:
What patch version of 2.8? We'll investigate this issue.

BardiaAfshin:
here's a copy of my composer.json (originally I was using the aws-sdk-php version 3.0)

```json
{
    "name": "bafshin/php-kinesis",
    "require": {
        "aws/aws-sdk-php": "^2.0"
    },
    "authors": [
        {
            "name": "me",
            "email": "me@me.com"
        }
    ],
    "prefer-stable": true
}
```

composer update (just ran it)

```json
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Removing guzzlehttp/guzzle (6.0.1)
  - Removing guzzlehttp/psr7 (1.1.0)
  - Removing psr/http-message (1.0)
  - Removing guzzlehttp/promises (1.0.1)
  - Removing mtdowling/jmespath.php (2.2.0)
  - Installing symfony/event-dispatcher (v2.7.1)
    Loading from cache

  - Installing guzzle/guzzle (v3.9.3)
    Loading from cache

  - Removing aws/aws-sdk-php (2.8.11)
  - Installing aws/aws-sdk-php (2.8.12)
    Downloading: 100%

[...]

PR #658, opened 2015-06-25
Replace input values with formatted values in JSON list serialization

jeskew:
Addresses #650

/cc @jeremeamia && @mtdowling

### DIFF
--- a/tests/Api/test_cases/protocols/input/json.json
+++ b/tests/Api/test_cases/protocols/input/json.json
@@ -45,6 +45,82 @@
     ]
   },
   {
+    "description": "Lists",
+    "metadata": {
+      "protocol": "json",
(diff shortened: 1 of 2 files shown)


tests/Api/test_cases/protocols/input/json.json
```diff
@@ -45,6 +45,82 @@
...
+        "type": "list",
+        "member": {
+          "shape": "StructureType"
+        }
+      },
+      "StructureType": {
+        "type": "structure",
+        "members": {
+          "a": {
+            "type": "string"
```
mtdowling: These should all be shape references

mtdowling:
Minor changes needed in the compliance test, but otherwise :ship:

PR #659, opened 2015-06-25
Allow custom stream wrapper protocol

### LINKED ISSUE #639: StreamWrapper can not register protocol other than 's3'
I am using version 3.0.4 of the SDK

The `StreamWrapper::register` function accepts a `$protocol` parameter, however, if you pass in anything other than `s3` then `StreamWrapper::getClient()` will throw exception.

Issue is in `getOptions`

```
    private function getOptions($removeContextData = false)
    {
        // Context is not set when doing things like stat
        if ($this->context === null) {
            $options = [];
        } else {
            $options = stream_context_get_options($this->context);
            $options = isset($options['s3']) ? $options['s3'] : [];
        }

        $default = stream_context_get_options(stream_context_get_default());
        $default = isset($default['s3']) ? $default['s3'] : [];
        $result = $this->params + $options + $default;

        if ($removeContextData) {
            unset($result['client'], $result['seekable'], $result['cache']);
        }

        return $result;
    }
```

The initial `register` will set the `stream_context_get_default` with the correct key (the passed protocol), but `getOptions` is hardcoded to always check for `s3`. 

I would include failing unit tests but these are private functions

mtdowling:
This pull request allows the stream wrapper to use a custom protocol. This is achieved by parsing the protocol out of the path each time the stream wrapper is invoked for a creational type of method.

Closes #639

@jeremeamia @jeskew

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -91,6 +91,9 @@ class StreamWrapper
     /** @var CacheInterface Cache for object and dir lookups */
     private $cache;
 
+    /** @var string The opened protocol (e.g., "s3") */
+    private $protocol = 's3';
+
     /**
      * Register the 's3://' stream wrapper
      *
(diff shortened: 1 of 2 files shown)


jeskew:
:+1:

jeremeamia:
:+1:

PR #667, opened 2015-06-30
Cache the members list in StructureShape

jeskew:
/cc @mtdowling @jeremeamia

### DIFF
--- a/src/Api/StructureShape.php
+++ b/src/Api/StructureShape.php
@@ -6,6 +6,8 @@ namespace Aws\Api;
  */
 class StructureShape extends Shape
 {
+    protected $members;
+
     public function __construct(array $definition, ShapeMap $shapeMap)
     {
         $definition['type'] = 'structure';
@@ -24,12 +26,11 @@ class StructureShape extends Shape
      */
     public function getMembers()
     {
-        $result = [];
-        foreach ($this->definition['members'] as $name => $definition) {
-            $result[$name] = $this->shapeFor($definition);
+        if (empty($this->members)) {
+            $this->generateMembersHash();
         }
 
-        return $result;
+        return $this->members;
     }
 
     /**
@@ -54,10 +55,22 @@ class StructureShape extends Shape
      */
     public function getMember($name)
     {
-        if (!isset($this->definition['members'][$name])) {
+        $members = $this->getMembers();
+
+        if (!isset($members[$name])) {
             throw new \InvalidArgumentException('Unknown member ' . $name);
         }
 
-        return $this->shapeFor($this->definition['members'][$name]);
+        return $members[$name];
+    }
+
+
+    protected function generateMembersHash()
+    {
+        $this->members = [];
+
+        foreach ($this->definition['members'] as $name => $definition) {
+            $this->members[$name] = $this->shapeFor($definition);
+        }
     }
 }
--- a/src/Api/StructureShape.php
+++ b/src/Api/StructureShape.php
@@ -6,7 +6,10 @@ namespace Aws\Api;
  */
 class StructureShape extends Shape
 {
-    protected $members;
+    /**
+     * @var Shape[]
+     */
+    private $members;
 
     public function __construct(array $definition, ShapeMap $shapeMap)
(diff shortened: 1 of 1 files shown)


mtdowling:
Can you provide motivation for this? Does this significantly increase perf? What about memory utilization for large data structures (like recursive DynamoDB)?

src/Api/StructureShape.php
```diff
@@ -54,10 +55,22 @@ public function hasMember($name)
...
+        if (!isset($members[$name])) {
             throw new \InvalidArgumentException('Unknown member ' . $name);
         }
 
-        return $this->shapeFor($this->definition['members'][$name]);
+        return $members[$name];
+    }
+
+
+    protected function generateMembersHash()
```
mtdowling: Any reason this is protected?

src/Api/StructureShape.php
```diff
@@ -6,6 +6,8 @@
  */
 class StructureShape extends Shape
 {
+    protected $members;
```
mtdowling: Any reason this is protected? I try to avoid protected members since they expose an implementation detail of an implementation detail to subclasses.

- https://books.google.com/books?id=ka2VUBqHiWkC&pg=PA89&lpg=PA89&dq=effective+java+protected+members&source=bl&ots=yZFkOmt-NX&sig=oTFs5pu9W45xG6wOQGBnL5nZ4l8&hl=en&sa=X&ei=QdWSVdCBMc_1oASB9KOwAw&ved=0CDoQ6AEwBA#v=onepage&q=effective%20java%20protected%20members&f=false
- http://programmers.stackexchange.com/questions/162643/why-is-clean-code-suggesting-avoiding-protected-variables

jeskew:
Given a large dataset with deeply recursive structures, this change decreased the memory usage of `Aws\Api\StructureShape::getMembers` from 32MB to 23KB on a paginated scan. This also decreased the amount of memory allocated by `Aws\Api\ShapeMap::resolve` from 15.6MB to 50KB. The full backfire profile can be viewed [here](https://blackfire.io/slots/compare/780586a3-2190-4e12-ad64-b9cf058b8924...31ad6977-86a4-40b7-865b-64454e7fe495/graph?callname=main%28%29&selected=&settings[dimension]=pmu&settings[display]=focused).

Resolving the shape members as it's currently done involves allocating a new array and populating it via ShapeMap lookups on every call to `getMembers`. Blackfire showed that function allocated around 378MB of memory over the course of a DDB query scan. Memory usage by this method never exceed 32MB at any given time, so those arrays are clearly being taken care of by GC... eventually. This change also dropped total GC time by 100ms. 

Copy-on-write semantics for pass-by-value arrays means that it's safe to return the same array on every call to `getMembers` while still only keeping one master copy instead of generating hundreds of thousands for the GC to handle.

src/Api/StructureShape.php
jeskew: No reason. Changing to private.

mtdowling:
Awesome. LGTM.

src/Api/StructureShape.php
```diff
@@ -24,12 +29,11 @@ public function __construct(array $definition, ShapeMap $shapeMap)
      */
     public function getMembers()
     {
-        $result = [];
-        foreach ($this->definition['members'] as $name => $definition) {
-            $result[$name] = $this->shapeFor($definition);
+        if (empty($this->members)) {
```
GrahamCampbell: Just a micro-optimization. It's faster not to call emtpy isn't it?

jeskew: I only want to generate the hash once per instance.

mtdowling: I think Graham is saying that you can omit the `empty` function and just call `if ($this->members) {`

jeskew: Oh, right. I find `empty` to be clearer than an exclamation point when reading code, and using empty avoids the implicit cast to boolean that simply calling `if (!$this->members) {` would entail.

mtdowling:
:ship:

PR #670, opened 2015-07-02
Check the provided cache before attempting to build credentials

jeskew:
If a `credentials.cache` option is passed into the `Credentials::factory` method, that cache should be checked for an instance of `CredentialsInterface` before attempting to build one from the environment. Since `CredentialsInterface` extends `Serializable`, we can rely on implementations to properly restore themselves when retrieved from cache.

/cc @mtdowling @jeremeamia


jeremeamia:
If looks good to me. If you are confident about it, go ahead and merge.

PR #676, opened 2015-07-07
Use late static binding for self instantiating the DynamoDb SessionHandler

NSmithUK:
Otherwise fromClient() does not work correctly from objects that extend Aws\DynamoDb\SessionHandler

This change should not effect existing functionality of those who call fromClient() directly on Aws\DynamoDb\SessionHandler.

### DIFF
--- a/src/DynamoDb/SessionHandler.php
+++ b/src/DynamoDb/SessionHandler.php
@@ -66,7 +66,7 @@ class SessionHandler implements \SessionHandlerInterface
             $connection = new StandardSessionConnection($client, $config);
         }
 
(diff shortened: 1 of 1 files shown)


jeremeamia:
Looks good to me.

Not that it's terribly important, but I'm curious about why you are extending the class.

NSmithUK:
@jeremeamia thanks.

The simplest way for us to add compression and encryption support to a SessionHandler is to extend the class. Then we pull in a PHP Trait that overrides the read() and write() methods, so they filter the data however we need, before passing it to the parent class.

PR #679, opened 2015-07-07
Allow the data attribute type to be configured in DynamoDb Sessions

NSmithUK:
At present it's hard coded to be a string, which prevents session data that is compressed and/or encrypted being stored without first being converted to base64.

### DIFF
--- a/src/DynamoDb/StandardSessionConnection.php
+++ b/src/DynamoDb/StandardSessionConnection.php
@@ -22,11 +22,12 @@ class StandardSessionConnection implements SessionConnectionInterface
     {
         $this->client = $client;
         $this->config = $config + [
-            'table_name'       => 'sessions',
-            'hash_key'         => 'id',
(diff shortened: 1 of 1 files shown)


jeremeamia:
Seems like a good change, but going to do a more in-depth review before merging. Thanks.

jeskew:
Sorry to take so long to get back to you on this. Part of PHP's native [SessionHandlerInterface](http://php.net/manual/en/class.sessionhandlerinterface.php) is the stipulation that only strings will be passed to the `write` method. The means for creating those strings is controlled by the [`session.serialize_handler`](http://php.net/manual/en/session.configuration.php#ini.session.serialize-handler) configuration value, but no serialize handler should be returning binary.

If you were worried about the performance implications of needing to base64 encode your data, then this change wouldn't make a huge difference, as binary fields are base64 encoded by the SDK before being sent to DynamoDb in any case.

PR #680, opened 2015-07-07
Add credentials cache and put it at the top of the provider chain

jeskew:
This PR adds a caching credentials provider wrapper (`Aws\Credentials\CredentialsProvider::cache`) that can be used in the same way as `CredentialsProvider::chain` and `CredentialsProvider::memoize` to compose a complex provider. It's meant to be used with credentials that are expensive to instantiate, namely the `InstanceProfileProvider` from the same namespace. By default, a file-based implementation is used to wrap `CredentialsProvider::instanceProfile` in `CredentialsProfile::defaultProvider`.


mtdowling:
Can you please provide more information and documentation about this PR?

We have our own cache interface, so I think this should be updated to use it instead of Doctrine: https://github.com/aws/aws-sdk-php/blob/master/src/CacheInterface.php

I'm not sure that we need a hard dependency on Doctrine. In fact, I'd rather we just have a single implementation that writes credentials to disk instead of using Doctrine at all. If someone wants Doctrine cache integration, they would just need to write an adapter.

src/Credentials/CachingProviderTrait.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace Aws\Credentials;
+
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\PhpFileCache;
+
+trait CachingProviderTrait
+{
+    private static $configCredentialsCache = 'credentials.cache';
```
mtdowling: Why do we need the prefixes here?

src/Credentials/InstanceProfileProvider.php
```diff
@@ -2,6 +2,7 @@
 namespace Aws\Credentials;
```
mtdowling: I don't think this credential provider should have any knowledge of a cache at all. Caching should be a wrapper around other providers.

src/Credentials/CredentialProvider.php
```diff
@@ -63,8 +68,9 @@ public static function defaultProvider(array $config = [])
     {
         return self::memoize(
             self::chain(
-                self::env(),
-                self::ini(),
+                self::cache($config),
```
mtdowling: Why cache the credentials by default? This only seems valuable if you know that you are using instance profile credentials.

src/Credentials/CachingProviderTrait.php
```diff
@@ -0,0 +1,67 @@
...
+            $credentials,
+            $credentials->getExpiration() - time()
+        );
+    }
+
+    /**
+     * @param array $config
+     * @return Cache|null
+     */
+    private static function getCredentialsCache(array $config = [])
```
mtdowling: What does this do? What's the motivation for a default cache here?

src/Credentials/CachingProviderTrait.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return new PhpFileCache($cacheDir);
+    }
+
+    private static function getCacheKey(array $config = [])
+    {
+        return isset($config[self::$configCredentialsCacheKey]) ?
+            $config[self::$configCredentialsCacheKey]
+            : 'credentials_' . crc32(gethostname());
```
mtdowling: I think this carried over from v2, and I'd rather not see it repeated in V3. I don't actually think we should compute a cache key like this.

src/Credentials/CredentialProvider.php
```diff
@@ -160,16 +166,21 @@ public static function memoize(callable $provider)
...
             if ($key && $secret) {
-                return Promise\promise_for(
-                    new Credentials($key, $secret, getenv(self::ENV_SESSION))
+                $credentials = new Credentials(
+                    $key,
+                    $secret,
+                    getenv(self::ENV_SESSION)
                 );
+
+                self::tryToCache($credentials, $config);
```
mtdowling: This should not know about the cache

src/Credentials/CredentialProvider.php
```diff
@@ -225,15 +236,40 @@ public static function ini($profile = null, $filename = null)
...
-                )
+            $credentials = new Credentials(
+                $data[$profile]['aws_access_key_id'],
+                $data[$profile]['aws_secret_access_key'],
+                isset($data[$profile]['aws_security_token'])
+                    ? $data[$profile]['aws_security_token']
+                    : null
             );
+
+            self::tryToCache($credentials, $config);
```
mtdowling: This shouldn't know about the cache either. Caching should wrap other providers like the memoize function.

src/Credentials/CachingProviderTrait.php
```diff
@@ -0,0 +1,67 @@
...
+                $config[self::$configCredentialsCache]
+                : self::getDefaultCache();
+        }
+
+        return null;
+    }
+
+    private static function getDefaultCache()
+    {
+        $cacheDir = sys_get_temp_dir() .
```
mtdowling: Don't we already have a default location for AWS caching that the JSON loader uses?

src/Credentials/CredentialProvider.php
jeskew: No caching is enabled by default, but the cache promise is at the top of the provider chain so that caches are checked before credentials are instantiated. I agree with another comment you made on this PR that caching should be built into `self::memoize` instead.

src/Credentials/CachingProviderTrait.php
jeskew: Would it make sense to use the access key instead (so that it's not tied to a particular host)?

src/Credentials/CredentialProvider.php
mtdowling: I wasn't saying the memoize function should have external caching built into it, but rather that there could be another function similar to the memoize function that allows for a customizable external cache. There may be some refactoring that could be done to share code, but I think these are separate abstractions.

src/Credentials/CredentialProvider.php
```diff
@@ -154,6 +156,36 @@ public static function memoize(callable $provider)
         };
     }
 
+    public static function cache(
```
mtdowling: Can you add a docblock here?

src/FileCache.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
```
mtdowling: We usually don't put a new line here

src/FileCache.php
```diff
@@ -0,0 +1,67 @@
...
+        $key = md5($key);
+
+        return implode(DIRECTORY_SEPARATOR, [
+            $this->cacheDir,
+            substr($key, 0, 2),
+            substr($key, 2, 2),
+            "$key.cache",
+        ]);
+    }
+}
```
mtdowling: Missing newline

tests/FileCacheTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+namespace Aws\Test;
+use Aws\FileCache;
```
mtdowling: should be a line between the ns and the use statements.

src/FileCache.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function remove($key)
+    {
+        unlink($this->getCachePath($key));
+    }
+
+    /**
```
mtdowling: We don't use inheritdoc in the SDK.

src/FileCache.php
```diff
@@ -0,0 +1,67 @@
...
+     * {@inheritdoc}
+     */
+    public function get($key)
+    {
+        if ($retrieved = @file_get_contents($this->getCachePath($key))) {
```
mtdowling: I think it might be better to first do a `file_exists` and then a regular `file_get_contents` call.

src/Credentials/CredentialProvider.php
```diff
@@ -154,6 +156,36 @@ public static function memoize(callable $provider)
...
+        callable $provider,
+        CacheInterface $cache = null,
+        $cacheKey = null
+    ) {
+        $cache = $cache ?: new FileCache;
+        $cacheKey = $cacheKey ?: 'aws_cached_credentials';
+
+        if (!$cache) {
```
mtdowling: I think this conditional is unnecessary

src/Credentials/CredentialProvider.php
```diff
@@ -154,6 +156,36 @@ public static function memoize(callable $provider)
...
+            return $provider;
+        }
+
+        return function () use ($provider, $cache, $cacheKey) {
+            $found = $cache->get($cacheKey);
+            if ($found instanceof CredentialsInterface && !$found->isExpired()) {
+                return Promise\promise_for($found);
+            }
+
+            $cacheable = $provider()->wait();
```
mtdowling: This isn't what we want. We want to call `then` on the provider and return a promise that is fulfilled with the credentials. In that callback we'd then set the credentials in the cache.

jeskew:
The history for this change is pretty noisy, so I'll squash the commits once everything looks ok.

src/FileCache.php
```diff
@@ -0,0 +1,61 @@
...
+            if (isset($cached['data']) &&
+                (empty($cached['expiration']) || $cached['expiration'] > time())
+            ) {
+                return $cached['data'];
+            }
+
+            $this->remove($key);
+        }
+
+        return null;
```
GrahamCampbell: this statement isn't needed

jeskew: You're right. I had it in as documentation (to signal that null is the intentional return value for this code path), but the docblock already states that the method can return null.

PR #681, opened 2015-07-09
bugfix(S3/StreamWrapper) Init protocol from passed path. Protocols are

dstevenson:
enforced by PHP to match.

Issue #693

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -515,6 +515,7 @@ class StreamWrapper
      */
     public function rename($path_from, $path_to)
     {
+        $this->initProtocol($path_from);
         $partsFrom = $this->withPath($path_from);
         $partsTo = $this->withPath($path_to);
         $this->clearCacheKey($path_from);
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -515,6 +515,8 @@ class StreamWrapper
      */
     public function rename($path_from, $path_to)
     {
+        // PHP will not allow rename across wrapper types, so we can safely
+        // assume $path_from and $path_to have the same protocol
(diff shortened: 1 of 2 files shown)


tests/S3/StreamWrapperTest.php
```diff
@@ -409,6 +409,16 @@ public function testCanDeleteNestedFolderWithRmDir()
 
     /**
      * @expectedException \PHPUnit_Framework_Error_Warning
+     * @expectedExceptionMessage rename(): Cannot rename a file across wrapper types
+     */
+    public function testRenameEnsuresProtocolsMatch()
```
dstevenson: On the fence about this unit test since it's technically only unit testing native PHP functionality that's outside the scope of the StreamWrapper. Thoughts?

src/S3/StreamWrapper.php
```diff
@@ -515,6 +515,7 @@ private function formatKey($key)
      */
     public function rename($path_from, $path_to)
     {
+        $this->initProtocol($path_from);
```
dstevenson: This should be safe because PHP natively ensures a rename is issued on the same wrapper.

mtdowling: Ah interesting. I did not know that.

tests/S3/StreamWrapperTest.php
dstevenson: Actually, I'm beginning to like this now since if we ever forget, or behavior in PHP changes, we'll probably need to modify `rename` accordingly.

mtdowling: Yeah sounds good to me.

src/S3/StreamWrapper.php
dstevenson: Done.

PR #682, opened 2015-07-10
Calculate correct S3 Transfer key with or without prefix (fixes #653)

### LINKED ISSUE #653: 400 MalformedXML error when using S3 Transfer Manager
Using [the example found in the documentation](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-transfer.html), first with 3.0.5 and then with 3.0.6 installed via Composer:
```php
//Here the source and destination are generic values, of course...
    'region'  => 'us-west-2',
    'version' => '2006-03-01',
]);
$source = '/path/to/source/files';
$dest = 's3://bucket';
$manager = new \Aws\S3\Transfer($client, $source, $dest);
$manager->transfer();
```
Results in the Aws\S3\Exception\S3Exception:
> Error executing "PutObject" on "https://s3-us-west-2.amazonaws.com/bucket/"; AWS HTTP error: Client error: 400 MalformedXML (client): The XML you provided was not well-formed or did not validate against our published schema - MalformedXMLThe XML you provided was not well-formed or did not validate against our published schema [...]
(Again, the bucket name in the error has been replaced with a generic...)

...and nothing is being uploaded to S3.

I've investigated the trace:
>[...]/vendor/aws/aws-sdk-php/src/WrappedHttpHandler.php(76): Aws\WrappedHttpHandler->parseError(Array, Object(GuzzleHttp\Psr7\Request), Object(Aws\Command)

Though I wasn't able to figure anything out by looking at the __invoke method in WrappedHttpHandler.php (I'm a dummy).

Has something changed recently or is there an issue on my end? I am near certain this was working fine the day before yesterday.

acoulton:
Resolve an issue where S3 Transfer was attempting to upload objects with an empty key when uploading to the root of an S3 bucket.

Also ensures that there is no `//` between prefix and path, to avoid uploading objects one level deeper than expected inside a directory with empty name.

This does appear to resolve the issue reported in #653 but I am unsure whether it has other implications.

### DIFF
--- a/src/S3/Transfer.php
+++ b/src/S3/Transfer.php
@@ -323,7 +323,7 @@ class Transfer implements PromisorInterface
         );
         
         if (isset($this->s3Args['Key'])) {
-            return rtrim($this->s3Args['Key']), '/').'/'.$relative_file_path;
+            return rtrim($this->s3Args['Key'], '/').'/'.$relative_file_path;
         } else {
             return $relative_file_path;
         }
--- a/src/S3/Transfer.php
+++ b/src/S3/Transfer.php
@@ -198,7 +198,7 @@ class Transfer implements PromisorInterface
     }
 
     /**
-     * Normalize a path so that it has a trailing slash.
+     * Normalize a path so that it has UNIX-style directory separators and no trailing /
      *
      * @param string $path
      *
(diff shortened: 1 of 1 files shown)


src/S3/Transfer.php
```diff
@@ -198,7 +198,7 @@ private function determineScheme($path)
     }
 
     /**
-     * Normalize a path so that it has a trailing slash.
+     * Normalize a path so that it has UNIX-style directory separators and no trailing /
```
acoulton: This method does the opposite of the previous doc-comment. Stripping trailing / does appear to be important, however, as filesystem paths with trailing / produce unexpected results elsewhere in the class.

src/S3/Transfer.php
```diff
@@ -317,15 +317,16 @@ private function uploadMultipart($filename)
 
     private function createS3Key($filename)
     {
-        if (!isset($this->s3Args['Key'])) {
-            return '';
```
acoulton: It is not clear why this method should ever return an empty string - I can't find anything useful in the git history and there doesn't appear to be test coverage. Since this seems to be causing #653 I've assumed it's safe to remove this behaviour - I have not reviewed to see if this has unexpected results elsewhere.

acoulton:
Sorry, I pasted this into the github editor incorrectly. I don't have a local clone handy but will squash this if/when you confirm you're happy with it.

jeskew:
@acoulton I added a test and merged this in as #703. Thanks for putting this together!

PR #684, opened 2015-07-10
Use hex instead of base64 output for S3 SigV4 checksum

jeskew:
The integration test on multipart uploads of nonseekable streams was failing. Switching the checksum encoding from base64 to hex fixes that.


jeremeamia:
Looks good.

PR #685, opened 2015-07-10
Add credentials cache and wrap it around instance profiles

jeskew:
This PR adds a caching credentials provider wrapper (`Aws\Credentials\CredentialsProvider::cache`) that can be used in the same way as `CredentialsProvider::chain` and `CredentialsProvider::memoize` to compose a complex provider. It's meant to be used with credentials that are expensive to instantiate, namely the `InstanceProfileProvider` from the same namespace. By default, a file-based implementation is used to wrap `CredentialsProvider::instanceProfile` in `CredentialsProfile::defaultProvider`.

Squashed from #680


src/FileCache.php
```diff
@@ -0,0 +1,59 @@
+<?php
+namespace Aws;
+
+class FileCache implements CacheInterface
+{
+    private $cacheDir;
+
+    public function __construct()
+    {
+        $this->cacheDir = (new JsonCompiler)->getCacheDir();
```
jeremeamia: Feels like cheating.

jeremeamia:
I like this more that the previous pull request.

jeskew:
This has grown too big for a single PR, so I'm splitting it into #714 and a follow-up.

PR #688, opened 2015-07-13
Unbind middleware closures from $this

jeskew:
The implicit `$this` binding on the functions affected by this patch was leaving a circular reference and preventing some memory allocated to instances of `Aws\S3\S3Client` from being freed.


mtdowling:
shipit

jeremeamia:
:ship:

PR #690, opened 2015-07-15
Suppress error from mkdir

afrowuk:
If two or more processes execute the code, the 1st may create the
directory after the is_dir call, while the 2nd will fail with "mkdir():
File exists". Suppressing the error while adding a 2nd is_dir call is
necessary.

### DIFF
--- a/src/JsonCompiler.php
+++ b/src/JsonCompiler.php
@@ -32,16 +32,16 @@ class JsonCompiler
         $this->cacheDir = getenv(self::CACHE_ENV)
             ?: sys_get_temp_dir() . '/aws-cache';
 
-        if (!is_dir($this->cacheDir)) {
-            if (!mkdir($this->cacheDir, 0777, true)) {
-                $message = 'Unable to create cache directory: %s. Please make '
-                    . 'this directory writable or provide the path to a '
-                    . 'writable directory using the AWS_PHP_CACHE_DIR '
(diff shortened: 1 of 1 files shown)


src/JsonCompiler.php
```diff
@@ -32,16 +32,16 @@ public function __construct($useCache = true)
...
-                $message = 'Unable to create cache directory: %s. Please make '
-                    . 'this directory writable or provide the path to a '
-                    . 'writable directory using the AWS_PHP_CACHE_DIR '
-                    . 'environment variable. Note that this cache dir may need '
-                    . 'to be cleared when updating the SDK in order to see '
-                    . 'updates.';
-                throw new \RuntimeException(sprintf($message, $this->cacheDir));
-            }
+        if (!is_dir($this->cacheDir)
+            && !@mkdir($this->cacheDir, 0777, true)
```
jeskew: I think you actually want this warning to surface. It won't halt script execution and will let you know that you should proactively create this directory to avoid a race condition. (A `post-dump-autoload` call to `Aws\JsonCompiler::purge()` in your composer file will create the default directory under safe conditions.)

mtdowling:
I'm ok with this change. It's what Doctrine/cache does when creating cache directories. https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FileCache.php#L169-L173

PR #691, opened 2015-07-15
2.8

ak-incaendo:


GrahamCampbell:
?

jeskew:
`2.8` and `master` are both actively developed, the first housing version 2 of the AWS SDK for PHP and the second housing version 3. There are no plans in the short or long term to merge the two.

PR #695, opened 2015-07-16
Defer inspection of memoized credential provider

jeskew:
Enclosing over a pending call `$provider` seems to prevent variables accessible to the promise (but not the value with which it is fulfilled) from being eligible for garbage collection. In the case of `Aws\CredentialProvider::defaultProvider`, this included an instance of `Aws\HandlerList`, an instance of `Aws\Api\ShapeMap`, and all of their member properties.


mtdowling:
:ship:

Issue #687, opened 2015-07-13
locationName is not being used


duncan3dc:
It looks like when the [XmlParser](https://github.com/aws/aws-sdk-php/blob/c66b1cfa66feee34cc589a8c4583691b14f89c28/src/Api/Parser/XmlParser.php#L62) tries to get the `locationName`, the [ListShape](https://github.com/aws/aws-sdk-php/blob/c66b1cfa66feee34cc589a8c4583691b14f89c28/src/Api/ListShape.php#L28) class doesn't return it, because it's using the `member` element of the definition, and `locationName` is at the same level as `member` and not it's child.

This was discovered by calling the `S3Client::listObjectVersions()` method, and not getting any `DeleteMarkers` or `Versions` in the `Result` (because their location name is singular).

I tried to create a pull request, or even a failing test but couldn't get anything useful together sorry

jeskew:
A likely fix was merged into master. Please reopen if you encounter this error again.

PR #697, opened 2015-07-16
Sync updated protocol test cases and apply necessary fixes

jeskew:
Fix parser tests

Fix serializer tests

Use locationName of shapes even when they are flattened lists whose members have no locationName

### DIFF
--- a/src/Api/Parser/AbstractRestParser.php
+++ b/src/Api/Parser/AbstractRestParser.php
@@ -93,10 +93,10 @@ abstract class AbstractRestParser extends AbstractParser
         switch ($shape->getType()) {
             case 'float':
             case 'double':
-                $value = doubleval($value);
+                $value = (float) $value;
                 break;
             case 'long':
-                $value = intval($value);
+                $value = (int) $value;
                 break;
             case 'boolean':
                 $value = filter_var($value, FILTER_VALIDATE_BOOLEAN);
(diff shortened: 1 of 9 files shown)


src/Api/Parser/AbstractRestParser.php
```diff
@@ -89,18 +89,30 @@ private function extractHeader(
...
-                $value = new DateTimeResult($value);
-            } catch (\Exception $e) {
-                // If the value cannot be parsed, then do not add it to the
-                // output structure.
-                return;
-            }
+        switch ($shape->getType()) {
+            case 'float':
+            case 'double':
+                $value = doubleval($value);
```
GrahamCampbell: Using `(double) $value` is faster and is identical in terms of the casting behaviour.

src/Api/Parser/AbstractRestParser.php
```diff
@@ -89,18 +89,30 @@ private function extractHeader(
...
+                break;
+            case 'long':
+                $value = intval($value);
```
GrahamCampbell: `(int) $value`, etc

src/Api/Parser/AbstractRestParser.php
GrahamCampbell: Or rather, `(float) $value`.

mtdowling: :+1:  Agreed, let's use casts instead

jeskew:
Addresses (among other things) #687

mtdowling:
:shipit:

PR #698, opened 2015-07-17
Make PHP 7 failures fail the build

jeskew:
PHP 7 has moved from alpha to beta, and our test suite is currently green on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ env:
 matrix:
   allow_failures:
     - php: hhvm
-    - php: 7
   fast_finish: true
 
 install:


GrahamCampbell:
:+1:

GrahamCampbell:
Though, the version of php 7 travis use isn't the beta release, so it still might break since it's just built from the master...

mtdowling:
> Though, the version of php 7 travis use isn't the beta release, so it still might break since it's just built from the master...

Yeah, that is a bit of a risk. It doesn't matter that much though as we don't have any process that hangs off of Travis builds. I think this is fine to merge.

jeremeamia:
:ship:

PR #700, opened 2015-07-20
Make the smoke tests less dependent on account state 

jeskew:
(and make account state-related failures count as a skip instead of a fail)


jeremeamia:
LGTM.

-- Jeremy Lindblom
PHP Software Engineer, AWS

mtdowling:
:ship:

PR #707, opened 2015-07-28
Ensure that fopen calls return false when errors are encountered

jeskew:
Currently the S3 StreamWrapper will trigger an error and return true when opening an invalid `s3://` path for reading. This PR ensures that calling `fopen` will return false on failure, which will prevent `file_get_contents` from returning an empty string (as mentioned on #609).

e4f078a is unrelated but it's the only way to get Travis to actually finish a build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ php:
   - 5.6
   - hhvm
 
+sudo: false
+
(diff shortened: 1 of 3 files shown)


mtdowling:
:ship:

PR #708, opened 2015-07-30
Allow user-defined user agents

jeskew:


src/ClientResolver.php
```diff
@@ -146,7 +147,14 @@ class ClientResolver
             'doc'      => 'A handler that accepts a command object, request object and returns a promise that is fulfilled with an Aws\ResultInterface object or rejected with an Aws\Exception\AwsException. A handler does not accept a next handler as it is terminal and expected to fulfill a command. If no handler is provided, a default Guzzle handler will be utilized.',
             'fn'       => [__CLASS__, '_apply_handler'],
             'default'  => [__CLASS__, '_default_handler']
-        ]
+        ],
+        'user-agent' => [
```
jeremeamia: We used underscores instead of hyphens for other config options.

src/ClientResolver.php
```diff
@@ -441,6 +449,36 @@ public static function _apply_http_handler($value, array &$args, HandlerList $li
...
+        if (!is_array($value)) {
+            $value = [$value];
+        }
+
+        $value = array_map(
+            function ($ua) { return (string) $ua; },
+            $value
+        );
+
+        $value []= 'aws-sdk-php/' . Sdk::VERSION;
```
jeremeamia: aws-sdk-php/{version} needs to be the first part.

jeremeamia:
I'm not sure we should let people override the whole UA string. We always want `aws-sdk-php/VERSION` as the first part of the UA string.

src/ClientResolver.php
```diff
@@ -441,6 +449,36 @@ public static function _apply_http_handler($value, array &$args, HandlerList $li
         );
     }
 
+    public static function _apply_user_agent($value, array &$args, HandlerList $list)
+    {
+        if (!is_array($value)) {
+            $value = [$value];
+        }
+
+        $value = array_map(
```
mtdowling: I wonder if `$value = array_map('strval', $value);` would be faster/more clear?

mtdowling:
This needs to be documented before merging, but otherwise LGTM

jeremeamia:
:ship:

PR #709, opened 2015-07-30
Ensure that an unserialized RefreshableInstanceProfileCredentials object is still useful

jeskew:
This class was serializing with just the credentials it had at the time `serialize` was called, so any attempt to subsequently refresh the credentials would have `null` where it was expecting an instance of `Aws\Common\InstanceMetadata\InstanceMetadataClient`. Would address #710.

### DIFF
--- a/src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php
+++ b/src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php
@@ -42,6 +42,21 @@ class RefreshableInstanceProfileCredentials extends AbstractRefreshableCredentia
         $this->client = $client ?: InstanceMetadataClient::factory();
     }
 
+    public function serialize()
+    {
+        return json_encode(array(
+            'credentials' => parent::serialize(),
+            'client' => serialize($this->client),
+        ));
+    }
+
+    public function unserialize($value)
+    {
+        $serialized = json_decode($value, true);
+        parent::unserialize($serialized['credentials']);
+        $this->client = unserialize($serialized['client']);
+    }
+
     /**
      * Attempt to get new credentials from the instance profile
      *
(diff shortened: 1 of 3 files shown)


src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php
```diff
@@ -42,6 +42,21 @@ public function __construct(CredentialsInterface $credentials, InstanceMetadataC
         $this->client = $client ?: InstanceMetadataClient::factory();
     }
 
+    public function serialize()
+    {
+        return json_encode(array(
+            'credentials' => parent::serialize(),
+            'client' => serialize($this->client),
```
mtdowling: Can the client actually be serialized?

jeskew: It can. There's no custom serialization defined, so `serialize` will just capture all object properties. The output is a bit large so I could cut out whatever is unnecessary.

mtdowling: Are you sure that things like the event dispatcher of a client with its anonymous functions can be serialized?

jeskew: The anonymous functions would not be serialized. The only part of the client that would get serialized would be its config array, and unserializing it would just use that config to call the constructor. 

The other option would be to use a `__wakeup` method instead and have it call `InstanceMetadataClient::factory()`, i.e., mimic what happens when you [create a new RefreshableInstanceProfileCredentials without a client](https://github.com/aws/aws-sdk-php/blob/2.8/src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php#L42). (This is what is done by default in [`Credentials::factory()`](https://github.com/aws/aws-sdk-php/blob/2.8/src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php#L42).) Doing so would involve fewer changes but would wipe out any custom InstanceMetadataClient, but I'm not sure anyone has ever created or used a custom one of those.

Issue #712, opened 2015-08-01
Argument mis-matches between S3's putObject() and MultiPartUploader() in v3


onethumb:
The immediate problem is that the `S3Client` expects `Bucket` and `Key`, whereas the `MultiPartUploader` expect `bucket` and `key`.  

The larger problem is that important arguments aren't available via `MultiPartUploader`, like `Metadata` which is fairly critical for many workloads that use `Metadata` to set a consistent hash or identifier, given that we can't predict what the `ETag` returned from an S3 multi-part upload will be.

There's some oddness around `Content-Type`, too, where it looks like it's hard-coding it based on filename extension:  https://github.com/aws/aws-sdk-php/blob/master/src/S3/MultipartUploader.php#L190-L194 

Are there plans to bring `MultiPartUploader` to parity with `S3Client::putObject()`'s arguments?

mtdowling:
I think that we could update the MultipartUploader to accept Bucket/bucket, Key/key, ACL/acl.

`Metadata` and other options are meant to be applied using the [callback configuration options](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-multipart-upload.html#configuration). You can observe operations before they are sent and mutate them as needed. For example, `before_initiate` is a function that is called with the [CreateMultipartUpload](http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3-2006-03-01.html#createmultipartupload) command that is about to be executed.

```php
$uploader = new MultipartUploader($s3Client, $source, [
    'bucket' => 'your-bucket',
    'key'    => 'my-file.zip',
    'before_initiate' => function (Aws\Command $command) {
        $command['Metadata']['xyz'] = '123';
        echo 'Initiating';
    }
]);
```

*Note: when providing a `state` that is already initiated, the `before_initiate` callback is not invoked.*

[...]

PR #713, opened 2015-08-01
Add Metadata from config to params

onethumb:
Lots of missing arguments from `MultiPartUploader` that are present in `S3::putObject()`, but this one is mission critical for us.  

See:  https://github.com/aws/aws-sdk-php/issues/712

### DIFF
--- a/src/S3/MultipartUploader.php
+++ b/src/S3/MultipartUploader.php
@@ -187,6 +187,11 @@ class MultipartUploader extends AbstractUploader
             $params['ACL'] = $this->config['acl'];
         }
 
+	// Set our Metadata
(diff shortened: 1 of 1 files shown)


src/S3/MultipartUploader.php
```diff
@@ -187,6 +187,11 @@ protected function getInitiateParams()
             $params['ACL'] = $this->config['acl'];
         }
 
+	// Set our Metadata
+	if(isset($this->config['Metadata'])) {
```
GrahamCampbell: cs

onethumb:
Closing, since there's a better way to handle this.  See: https://github.com/aws/aws-sdk-php/issues/712

Issue #715, opened 2015-08-05
__call definition


iam-merlin:
Hi,
Simple request to drop default values for __call in Aws\Sdk.

The `__call` method's `$args` parameter is required and will always be present as an array (http://www.php.net/manual/en/language.oop5.overloading.php#object.call), adding a default value (like `$args = array()`) is not needed and even causes problems, as described in padraic/mockery#263.

Removing this allows us to manipulate the object as expected with mock.

Thank you (I can PR if you want)

iam-merlin:
For reference (same issue on a different class with merge) : https://github.com/aws/aws-sdk-php/pull/92

jeskew:
Thanks for reporting; the attached PR will go out in the next release.

PR #716, opened 2015-08-05
Update Sdk.php

iam-merlin:
It's "pointless. For __call magic methods, PHP always sets $args to an array."

By the way, it becames problematic to mock the SDK.

see : https://github.com/aws/aws-sdk-php/issues/715

### DIFF
--- a/src/Sdk.php
+++ b/src/Sdk.php
@@ -76,7 +76,7 @@ class Sdk
         }
     }
 
-    public function __call($name, array $args = [])
+    public function __call($name, array $args)
     {
         if (strpos($name, 'create') === 0) {
             return $this->createClient(


GrahamCampbell:
:-1:

GrahamCampbell:
You probably shouldn't be trying to mock this anyway.

src/Sdk.php
```diff
@@ -76,7 +76,7 @@ public function __construct(array $args = [])
         }
     }
 
-    public function __call($name, array $args = [])
+    public function __call($method, $args)
```
jeremeamia: - Changing `$name` to `$method` doesn't work at all.
- I don't believe there is a need to remove the `array` typehint.
- But yes, the ` = []` should be removed.

GrahamCampbell: :+1: @jeremeamia

iam-merlin: oups, my bad. indeed ofc.

iam-merlin:
@GrahamCampbell , it's simply to test a dumb factory :/, this is the only reason to mock aws/sdk.

jeskew:
I think you should leave the `array` typehint on `$args` and add a test case that makes sure the class is mockable.

GrahamCampbell:
> I think you should leave the array typehint on $args and add a test case that makes sure the class is mockable.

Yeh. The typehint should stay.

iam-merlin:
Hi,
SDK is already mocked using phpunit (in aws test). For mockery, it seems it's a problem with an old version and a specific test for mockery need a new dev dependency (and that is not, in my point of view, a good idea). Now it's ok for me.

I think the last dumb change we previsouly discuss can be accepted now.

jeskew:
Cool; I didn't realize the mocking issue only affected mockery. Could you squash the commits?

iam-merlin:
done ;)

jeskew:
Thanks!

PR #718, opened 2015-08-06
Allow the uploading of falsy strings to glacier

jeskew:


src/Aws/Glacier/Model/MultipartUpload/UploadPartGenerator.php
```diff
@@ -237,7 +237,7 @@ protected function generateUploadParts(EntityBodyInterface $body)
         $uploadContext = new UploadPartContext($this->partSize, $body->ftell());
 
         // Read the data from the streamed body in 1MB chunks
-        while ($data = $body->read(min($this->partSize, Size::MB))) {
+        while (strlen($data = $body->read(min($this->partSize, Size::MB))) > 0) {
```
mtdowling: I think we should make this more readable by moving the assignment and conditions to their own lines.

mtdowling:
One minor suggestion, but LGTM

jeremeamia:
Looks good.

Issue #722, opened 2015-08-10
cURL error 3: malformed URL when searching with CloudSearchDomainClient


rms2219:
I'm creating a CloudSearchDomainClient in my Laravel 5.1 app. I'm using version 3.0.1 of the AWS SDK for Laravel package. I create my client like this:

```
$cloudSearch = app('aws')->createClient('CloudSearchDomain', ['endpoint' => 'my.search.endpoint.url']);
```

I add search arguments:

```
$searchArguments = [
            'queryParser' => 'structured',
            'size' => $limit,
            'start' => $offset ?: 0
        ];
```

And then I go on to build my structured query. For brevity's sake, I'll leave that out, but it works fine using the same API version but the SDK version for Laravel 4.2.

When I call "search" on the client object, I get the following error message:

Error executing "Search" on "/2013-01-01/search?format=sdk&pretty=true&q.parser=structured&size=200&start=0&q=%28and%20%28and%20%28or%20%28term%20field%3Ddescription%20%27x%27%29%20%28prefix%20field%3Ddescription%20%27x%27%29%29%29%20isncnr%3A0%20isactive%3A1%20%29"; AWS HTTP error: cURL error 3: <url> malformed (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)

I would expect my search endpoint to be prepended to the URL listed in the error message, but I'm unsure if that's the problem.

My credentials are in my .env file and I'm able to use other AWS services (e.g. getting documents from S3 works fine)

jeskew:
You'll need to specify a protocol on the endpoint (either `http` or `https`).

jeremeamia:
Maybe you could add a check for a protocol on the provided endpoint and throw an exception if it's not there with a helpful message. Letting a bad endpoint go all the way to a cURL error seems like it is no fun. This is especially important for this service because customers tend to copy & paste their cloudsearch endpoint right from the console, which does not include a protocol.

jeskew:
@jeremeamia That's a good point. I opened a PR to do so.

PR #723, opened 2015-08-10
Ensure that custom endpoints have a defined scheme and host

jeskew:
Currently, it's very easy to copy a value from the CloudSearchDomain console and be greeted with an opaque error message ('cURL error 3'). As described in #722, this is what happens when you supply a custom `endpoint` to a client that cURL will not be able to use directly. This PR adds a validation function to the ClientResolver that ensures endpoints have a scheme and host.


jeremeamia:
Cool!

mtdowling:
LGTM

PR #724, opened 2015-08-11
Precompile all included JSON files

jeskew:
Rather than relying on the JsonCompiler, this PR moves the transpilation of the SDK's JSON data files from a runtime concern to a build-time concern.


mtdowling:
```diff
diff --git a/src/functions.php b/src/functions.php
index 0d3dd9d..234c12a 100644
--- a/src/functions.php
+++ b/src/functions.php
@@ -132,14 +132,7 @@ function or_chain()
  * Loads a compiled JSON file from a PHP file.
  *
  * If the JSON file has not been cached to disk as a PHP file, it will be loaded
- * from the JSON source file, written to disk as a PHP file, and returned. This
- * allows subsequent access of the JSON file to be read from a compiled PHP
- * script which is added to PHP's in-memory opcode cache.
- *
- * The default directory used to save compiled PHP scripts is the "aws-cache"
- * sub-directory of PHP temp directory (return value of sys_get_temp_dir()).
- * You can customize where the cache files are stored by specifying the
- * `AWS_PHP_CACHE_DIR` environment variable.
+ * from the JSON source file and returned.
  *
  * @param string $path Path to the JSON file on disk
  *
@@ -150,6 +143,10 @@ function load_compiled_json($path)
 {
     static $loader;
 
+    if ($compiled = @include("$path.php")) {
+        return $compiled;
+    }
+
```

Any reason to not use `flle_exists` here instead or using the silence operator?

mtdowling:
```diff
 /**
- * Loads JSON files and compiles them into PHP files so that they are loaded
- * from PHP's opcode cache.
+ * Loads JSON files and compiles them into PHP arrays
  *
- * @internal Please use Aws\load_compiled_json() instead.
+ * @internal Please use json_decode instead.
  */
 class JsonCompiler
 {
-    const CACHE_ENV = 'AWS_PHP_CACHE_DIR';
```

We might want to leave the constant here if someone is actually using it in their code.

```diff
-    /**
-     * Loads a JSON file.
-     *
-     * @param string $path Provided path.
-     * @param string $real Normalized path.
-     *
-     * @return array
-     * @throw \InvalidArgumentException if file does not exist.
-     */
-    private function loadJsonFromFile($path, $real)
-    {
-        if (!file_exists($real)) {
+        if (!file_exists($path)) {
             throw new \InvalidArgumentException(
-                sprintf("File not found: %s, realpath: %s", $path, $real)
+                sprintf("File not found: %s", $path)
             );
         }
 
-        return json_decode(file_get_contents($real), true);
+        return json_decode(file_get_contents($path), true);
     }
```

Perhaps we should emit a warning here?

mtdowling:
Can you describe what the git commit hook you added does?

GrahamCampbell:
```
Loads JSON files and compiles them into PHP arrays
```

Missing fullstop.

jeskew:
@mtdowling I would prefer not to trigger a call to `stat` if it can be avoided since we're counting on bypassing disk I/O and pulling these files from the PHP opcache. `opcache_is_script_cached` almost does what I would want, but it's only available on >=5.6, so the check would end up fairly convoluted:

```php
if ((function_exists('opcache_is_script_cached') && opcache_is_script_cached($path)) 
    || file_exists($path)
) {
    return include($path);
}
```

It seems cleaner to just let `include` delegate to the opcache and filesystem as needed. Maybe I should remove the silencer and let the function trigger a warning before moving on to searching out the JSON file?

mtdowling:
Just using include sounds reasonable to me. When using the SDK out of the box, calling include should be successful every time, so there's no real concern for normal SDK usage.

jeskew:
It still works, I just threw on the WIP label while the compiled JSON is not in the commit. GH was almost useless, so I'll add them back in once the generator code is cleared.

PR #725, opened 2015-08-12
Ensure that connect errors on S3 are retried as configured

jeskew:
It seems that S3 connection errors were not being retried due to a function signature mismatch.


mtdowling:
:ship:

Issue #726, opened 2015-08-13
Fatals on Waiters occasionally


onethumb:
We see intermittent PHP crashes with `Call to member function getStatusCode() on null` from this line:  https://github.com/aws/aws-sdk-php/blob/master/src/Waiter.php#L245

Since `$result->getResponse()` can return a `ResponseInterface` or `NULL`, checking the result first alleviates the problem.  Example:

```php 
$response = $result->getResponse(); // this sometimes returns NULL
    if(isset($response)) {
        return $acceptor['expected'] == $response->getStatusCode();
    }
```

jeskew:
If the response is null, that would generally mean that an error was raised by the HTTP handler. I added a separate check to make sure that exceptions without responses are retried within the waiter's maxAttempts limit.

PR #727, opened 2015-08-13
Fix/eliminate waiter npe

jeskew:
Addresses #726 and allows retries to be configured on a command instead of just on a client.

/cc @mtdowling

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -54,6 +54,11 @@ class RetryMiddleware
             ResultInterface $result = null,
             $error = null
         ) use ($maxRetries) {
+            // Allow command-level options to override this value
+            $maxRetries = null !== $command['retries'] ?
+                $command['retries']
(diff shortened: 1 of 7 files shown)


src/RetryMiddleware.php
```diff
@@ -54,6 +54,11 @@ public static function createDefaultDecider($maxRetries = 3)
             ResultInterface $result = null,
             $error = null
         ) use ($maxRetries) {
+            // Allow command-level options to override this value
+            $maxRetries = null !== $command['retries'] ?
```
mtdowling: I think this should be `@retries` as it is a meta parameter that is not part of the API. We are already using `@` to designate other parameters as SDK related rather than API related (e.g., `@http`), so I think it makes sense.

mtdowling:
Looks good other than renaming it to `@retries`. We might also need to update the docs.

mtdowling:
:shipit:

PR #728, opened 2015-08-13
Scrub STS tokens and keys, too

jeskew:


mtdowling:
LGTM.

Maybe we could make the scrubbing pattern stuff more generic as well in that it could be a hash of pattern to replacement?

PR #734, opened 2015-08-19
Map phar in stub file and set a versioned alias

### LINKED ISSUE #563: Every .phar 2.6.16 and above fails to load on the second request
My coworker and I downloaded 2.8.2, then 2.7.0, and 2.6.16. None of the phars for these versions worked, and spat out this error:

```[Thu Apr 30 15:50:21.909040 2015] [:error] [pid 2621] [client ] PHP Warning:  require(phar://aws.phar/aws-autoloader.php): failed to open stream: phar error: invalid url or non-existent phar &quot;phar://aws.phar/aws-autoloader.php&quot; in /usr/local/lib/php/aws.phar on line 3, referer: https://
[Thu Apr 30 15:50:21.909127 2015] [:error] [pid 2621] [client ] PHP Fatal error:  require(): Failed opening required 'phar://aws.phar/aws-autoloader.php' (include_path='.:/usr/share/php:/usr/share/pear:/usr/local/lib/php:/usr/local/lib/php/db:/var/www') in /usr/local/lib/php/aws.phar on line 3, referer: https://```

We then went up from 2.6 to 2.6.9, 2.6.12, 2.16.14, and 2.6.15, at that point realizing that 2.6.15 is the most up-to-date version of the .phar we can use. We are considering the use of Composer in the future (which we are actually already using), but wanted to report our issues with these phar versions.

jeskew:
Another attempt to resolve #563. Other projects that distribute a phar (including Composer, PHPUnit, and Behat) include a call to Phar::mapPhar($filename)) in their phar stubs, and its absence could explain why aws-autoloader.php cannot be found when aws.phar is loaded from opcache instead of from the filesystem.

/cc @mtdowling

### DIFF
--- a/build/Burgomaster.php
+++ b/build/Burgomaster.php
@@ -302,18 +302,20 @@ EOT
      *
      * @param $dest
      * @param string $autoloaderFilename Name of the autoloader file.
+     * @param string $alias The phar alias to use
      *
      * @return string
      */
-    private function createStub($dest, $autoloaderFilename = 'autoloader.php')
+    private function createStub($dest, $autoloaderFilename = 'autoloader.php', $alias = null)
     {
         $this->startSection('stub');
         $this->debug("Creating phar stub at $dest");
 
-        $alias = basename($dest);
+        $alias = $alias ?: basename($dest);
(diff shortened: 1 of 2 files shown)


mtdowling:
:ship:

Issue #710, opened 2015-07-31
RefreshableInstanceProfileCredentials get non-object for $client


stefandoorn:
Once in a while I notice the following error in our production logs. It happens only some (2-3) times a day.

````
Call to a member function getInstanceProfileCredentials() on a non-object in /var/www/vendor/aws/aws-sdk-php/src/Aws/Common/Credentials/RefreshableInstanceProfileCredentials.php:52
````

I tried some debugging but unfortunately I can never reproduce the problem myself and neither I really get a grip on how this would happen. Also at this moment I don't have a deeper trace, this is all is logged at the moment. I'm using version 2.8.13 of the API in production at the moment.

jeskew:
Hi @stefandoorn,

The root cause of the error you encountered is that an instance profile credentials object can't be effectively woken up from cache right now if the credentials have expired. There's an open pull request to fix this.

stefandoorn:
Hi @jeskew, 

Thanks for the clarification :) I will be happy to test it live as soon as it made it into a release.

jeskew:
A fix for this was included in today's release. Please feel free to reopen if you have any further questions.

stefandoorn:
Hi @jeskew, wanted to let you know I had some problems on/after deploy:

Aws\Ec2\Exception\Ec2Exception: AWS Error Code: AuthFailure, Status Code: 401, AWS Request ID: XXX, AWS Error Type: client, AWS Error Message: Authorization header or parameters are not formatted correctly., User-Agent: aws-sdk-php2/2.8.17 Guzzle/3.9.3 curl/7.40.0 PHP/5.5.26 [] []

Removing the cached credentials by hand solved things, error only happened once. Not sure if more people have this problem, but atleast good you know it happened I think :) Not sure if invalidating cache is possible on release, but would be a nice feature for sure :)

jeskew:
@stefandoorn Glad to hear clearing the cache worked. 

Unfortunately, I'm not sure we can reliably clear caches on upgrading, as Composer will only execute post-update scripts registered on a project's root, not on that project's dependencies. In v3, we use the version number as part of the cache key, and that might be something we could look into backporting.

PR #735, opened 2015-08-19
Add a mechanism for caching InstanceProfile credentials

jeskew:
This PR allows users to pass an instance of `Aws\CacheInterface` as the `credentials` in their client configuration. The default provider chain will still be used, but the cache will hold on to credentials returned by the Ec2 instance metadata service.

Additionally, this PR includes an adapter for instances of `Doctrine\Common\Cache\Cache`, as the only bundled implementation of `Aws\CacheInterface` is an in-memory cache.

### DIFF
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -46,7 +46,7 @@ that loads API files from the ``src/data`` folder of the SDK.
 credentials
 ~~~~~~~~~~~
 
-:Type: ``array|Aws\Credentials\CredentialsInterface|bool|callable``
+:Type: ``array|Aws\Credentials\CredentialsInterface|bool|callable|Aws\CacheInterface``
 
 If you do not provide a ``credentials`` option, the SDK will attempt to load
 credentials from your environment in the following order:
@@ -107,6 +107,21 @@ create credentials using a function.
         'version'     => 'latest',
         'region'      => 'us-west-2',
         'credentials' => $provider
+    ]);
+
+Pass an instance of ``Aws\CacheInterface`` to cache the values returned by the
+default provider chain across multiple processes.
+
+.. code-block:: php
+
+    use Aws\DoctrineCacheAdapter;
+    use Aws\S3\S3Client;
+    use Doctrine\Common\Cache\ApcCache;
+
+    $s3 = new S3Client([
+        'version'     => 'latest',
+        'region'      => 'us-west-2',
+        'credentials' => new DoctrineCacheAdapter(new ApcCache),
     ]);
 
 You can find more information about providing credentials to a client in the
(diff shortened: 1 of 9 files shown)


docs/guide/configuration.rst
```diff
@@ -46,7 +46,7 @@ that loads API files from the ``src/data`` folder of the SDK.
 credentials
 ~~~~~~~~~~~
 
-:Type: ``array|Aws\Credentials\CredentialsInterface|bool|callable``
+:Type: ``array|Aws\Credentials\CredentialsInterface|bool|callable|Aws\CacheInterface``
```
jeremeamia: These types should probably be sorted in some kind of reasonable way.

jeremeamia:
LGTM. :-) Hi guys!

docs/guide/configuration.rst
jeskew: I didn't notice they used to be in alphabetical order until I read your comment. I see no reason to break with tradition on that.

src/Credentials/CredentialProvider.php
```diff
@@ -61,11 +62,16 @@ class CredentialProvider
      */
     public static function defaultProvider(array $config = [])
     {
+        $cache = isset($config['credentials'])
```
mtdowling: Instead of two ternary checks here, could we just create an `$instanceProfileProvider` variable that is first created as a stock instance profile provider and then decorated with a cache if the cache parameter is present? I think that would make this easier to follow.

stefandoorn:
This feature would be very welcomed by me. We are working on a high traffic website, using IAM credentials for EC2, AutoScaling & DynamoDB. The more caching on these requests, the better. I'm currently upgrading v2 to v3 and in v2 it was fairly easy to cache the credentials. Before I did that we could hit some API limits under high traffic. Now when migrating to v3 I couldn't find that feature. Ofcourse it's possible to implement a solution myself (as suggested in here als: https://twitter.com/awsforphp/status/638072096497795072), but I think more people would be happy to use a common feature like this.

stefandoorn:
@jeskew. Recently we discussed something related to cache in this issue (2.8): https://github.com/aws/aws-sdk-php/issues/710. In there you mentioned about the cache keys using version names. Am I right that this PR is going to implement the feature and right now it aint implemented yet?

jeskew:
@stefandoorn This PR does not use the version number as part of the key, as the mechanism proposed is simpler than what is in v2. In v2, the object cached included an InstanceMetadataClient; in this case, all the cache will store is your key, secret, token, and expiration.

I was referring in #710 to the cache used by `Aws\JsonCompiler`, which has since been removed.

stefandoorn:
@jeskew Perfect. When is it expected for this PR to be shipped? I would be happy to continue my migration from v2 to v3 :)

jeskew:
The only thing it was waiting on was for me to clear up a merge conflict, which is now done. I'm merging into master, so it will go out with the next release.

Issue #741, opened 2015-08-25
Key as an empty string lists bucket contents...


byustephen:
When given a key as an empty string, the GetObject command lists the entire bucket contents. Isn't this a bug?

Though an interesting case a file in our DB had it's path removed, and when trying to download the file, instead of an error message, it listed the entire bucket contents. You can see our code below, and i've added in a fake line to show you what would cause the error. Regardless, is this a bug? I would expect an error if the key is missing (since it gives an error if the key is null, or if the key is wrong). 

		$command = $this->s3->getCommand('GetObject', array(
		   'Bucket' => $this->bucket_name,
		  // 'Key' => $file_data->ref,
           'Key' => '',
		   'ResponseContentDisposition' => 'attachment; filename="'.$file_name.'"'
		));

		$signedUrl = $command->createPresignedUrl('+10 minutes');
		echo $signedUrl;
		header('Location: '.$signedUrl);
		die();

jeskew:
Thanks for reporting this. 

What appears to be happening is that the key is used to construct the URL, which for a `GetObject` command looks like `https://<endpoint>/<bucket>/<key>`. The only check performed on keys is that they must be strings (there's no check to make sure the key isn't an empty string), so you're ending up with a signed URL for `https://<endpoint>/<bucket>/`. This is a distinguished from a `ListObjects` command by a trailing slash, and S3 does not appear to care about trailing slashes.

I don't believe this issue can be quickly fixed in the SDK, so if you're running into it regularly, then you might want to stick an `if (empty($file_data->ref))` check above call to `getCommand` in the code you posted.

byustephen:
Thanks for looking into this. I've already added in a if statement to see if the key (or in this case the $file_data->ref) is empty. I just thought it was a security issue that should be looked into. Cheers!

PR #742, opened 2015-08-26
Check model-defined min and max values during validation

jeskew:
When min and max values are provided in the models, payloads can be checked by the SDK instead of relying on the service to provide this validation. This would help address #741.

/cc @mtdowling


src/Api/Validator.php
```diff
@@ -91,10 +92,21 @@ private function check_structure(StructureShape $shape, $value)
     private function check_list(ListShape $shape, $value)
     {
         if (!is_array($value)) {
-            $this->addError('must be an array. Found ' . Aws\describe_type($value));
+            $this->addError('must be an array. Found '
+                . Aws\describe_type($value));
             return;
         }
 
+        if ($shape['min'] && count($value) < $shape['min']) {
```
mtdowling: One minor optimization might be to store the count in a variable and compute it once

src/Api/Validator.php
```diff
@@ -155,6 +181,19 @@ private function check_string(Shape $shape, $value)
         if (!$this->checkCanString($value)) {
             $this->addError('must be a string or an object that implements '
                 . '__toString(). Found ' . Aws\describe_type($value));
+            return;
+        }
+
+        if ($shape['min'] && strlen($value) < $shape['min']) {
```
mtdowling: Might want to store in a variable and compute strlen only once

mtdowling:
:ship:

PR #743, opened 2015-08-27
Allow getIterator() to continue to work without a *_token

radford:
Allow `getIterator()` calls to non-paginated APIs, such as

```php
$ec2->getIterator('DescribeSubnets');
```
to continue to work as they did in v2 instead of failing with
```
Call to undefined method Aws\Command::search()
```

### DIFF
--- a/src/AwsClient.php
+++ b/src/AwsClient.php
@@ -245,7 +245,7 @@ class AwsClient implements AwsClientInterface
             return $this->getPaginator($name, $args)->search($key);
         }
 
-        $result = $this->getCommand($name, $args)->search($key);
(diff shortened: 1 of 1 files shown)


jeskew:
I think this should still use the `search` method so that you could pass in something more complex than a key. The issue is that the command created by `getCommand` was not being executed.

radford:
Good idea.  I'll update the patch.

radford:
@jeskew, any update on this?

jeskew:
@radford Sorry, I thought I was waiting on your update. I can work on this today.

PR #753, opened 2015-09-04
Fixed guzzle version constraint

GrahamCampbell:
Without this fix, you're allowing ALL versions of guzzle 5.3 or greater, for example, guzzle 10.5.2 would be allowed by this version constraint.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require": {
         "php": ">=5.5",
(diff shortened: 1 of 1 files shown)


jeskew:
@GrahamCampbell can you squash out the first commit?

GrahamCampbell:
Meh, will create a new PR. CBA to clone the repo. Can do this from the web interface.

GrahamCampbell:
There you go: https://github.com/aws/aws-sdk-php/pull/755.

PR #756, opened 2015-09-05
Treat 200-coded error responses from S3 as retriable errors

jeskew:
S3 sometimes returns error with a status code of 200 (specifically their [copy operations](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html)). This PR would treat those as errors and retry the requests.

/cc @mtdowling

### DIFF
--- a/src/S3/AmbiguousSuccessParser.php
+++ b/src/S3/AmbiguousSuccessParser.php
@@ -3,7 +3,10 @@ namespace Aws\S3;
 
 use Aws\Api\Parser\AbstractParser;
 use Aws\CommandInterface;
+use Aws\Exception\AwsException;
+use Exception;
 use Psr\Http\Message\ResponseInterface;
+use SimpleXMLElement;
 
 class AmbiguousSuccessParser extends AbstractParser
 {
@@ -15,27 +18,57 @@ class AmbiguousSuccessParser extends AbstractParser
 
     /** @var callable */
     private $parser;
+    /** @var string */
+    private $exceptionClass;
 
-    public function __construct(callable $parser)
+    public function __construct(callable $parser, $exceptionClass = AwsException::class)
     {
         $this->parser = $parser;
+        $this->exceptionClass = $exceptionClass;
     }
 
     public function __invoke(
         CommandInterface $command,
         ResponseInterface $response
     ) {
-        $fn = $this->parser;
-
         if (200 === $response->getStatusCode()
             && isset(self::$ambiguousSuccesses[$command->getName()])
         ) {
-            $xml = new \SimpleXMLElement($response->getBody());
+            $xml = new SimpleXMLElement($response->getBody());
             if ('Error' === $xml->getName()) {
-                throw new \DomainException($xml->xpath('/Error/Message')[0]);
+                throw $this->parseError(
+                    $xml->xpath('/Error/Message')[0],
+                    $command
+                );
             }
         }
 
-        return $fn($command, $response);
+        try {
+            $fn = $this->parser;
+            return $fn($command, $response);
+        } catch (Exception $e) {
+            throw $this->parseError(
+                "Error parsing response for {$command->getName()}:"
+                    . " AWS parsing error: {$e->getMessage()}",
+                $command,
+                $e
+            );
+        }
+    }
+
+    private function parseError($message, CommandInterface $command, Exception $previous = null)
+    {
+        $context = [];
+        $context['connection_error'] = true;
+        if (isset($previous)) {
+            $context['exception'] = $previous;
+        }
+
+        return new $this->exceptionClass(
+            $message,
+            $command,
+            $context,
+            $previous
+        );
     }
 }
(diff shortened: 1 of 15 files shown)


src/S3/AmbiguousSuccessParser.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct(callable $parser)
+    {
+        $this->parser = $parser;
+    }
+
+    public function __invoke(
+        CommandInterface $command,
+        ResponseInterface $response
+    ) {
+        $fn = $this->parser;
```
mtdowling: Why not just try/catch around the invocation of `$this->parser`, and catch SimpleXML exceptions? After catching then check if it's one of our known ambiguous errors and throw the appropriate retryable error.

src/WrappedHttpHandler.php
```diff
@@ -95,9 +95,21 @@ private function parseResponse(
...
         $status = $response->getStatusCode();
-        $result = $status < 300
-            ? $parser($command, $response)
-            : new Result();
+        try {
+            $result = $status < 300
+                ? $parser($command, $response)
+                : new Result();
+        } catch (\Exception $e) {
+            return new Promise\RejectedPromise($this->parseDeserializationError(
```
mtdowling: Why is this code necessary? It also violates the return type as we need to return a ResultInterface here not a promise.

src/S3/AmbiguousSuccessParser.php
jeskew: @mtdowling The ambiguous errors will be valid XML, they just won't be the shape expected by the RestXML parser for a copy operation. I exposed the service's error parser to this class and used that instead of SimpleXML, but the copy errors need to be handled separately from list errors, as those will return a 200 and some fraction of a valid XML document (rather than a 200 and the wrong XML document).

jeremeamia: :disappointed:

src/S3/AmbiguousSuccessParser.php
```diff
@@ -0,0 +1,77 @@
...
+            if (isset($parsed['code']) && isset($parsed['message'])) {
+                throw $this->parseError($parsed['message'], $command);
+            }
+        }
+
+        try {
+            $fn = $this->parser;
+            return $fn($command, $response);
+        } catch (Exception $e) {
+            throw $this->parseError(
```
mtdowling: Is this try/catch necessary? Aren't we covered by the error parsing before parsing with the wrapped parser?

src/S3/AmbiguousSuccessParser.php
mtdowling: Ah I see. Makes sense.

jeskew: They are separate errors, having nothing in common besides their being error responses with a 200 status code, so they could be caught by separate parsers.

src/S3/MalformedResponseParser.php
```diff
@@ -0,0 +1,43 @@
+<?php
+namespace Aws\S3;
+
+use Aws\Api\Parser\AbstractParser;
+use Aws\CommandInterface;
+use Aws\Exception\AwsException;
+use Exception;
+use Psr\Http\Message\ResponseInterface;
+
+class MalformedResponseParser extends AbstractParser
```
mtdowling: So is this parser for when S3 returns a ListObjects response that is invalid XML?

src/S3/MalformedResponseParser.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function __invoke(
+        CommandInterface $command,
+        ResponseInterface $response
+    ) {
+        $fn = $this->parser;
+
+        try {
+            return $fn($command, $response);
+        } catch (Exception $e) {
```
mtdowling: If this is meant to catch errors parsing XML, then I think catching `\Exception` is a little too broad. In fact, we should only be catching XML parsing errors to ensure that only that specific error is being retried. If we just catch all parsing errors, then we'd be retrying everything multiple times when we fail to parse. Instead of a catch-all like this, I think we should hand-pick what parsing errors are retryable (i.e., XML errors).

jeskew: Unfortunately, SimpleXMLElement will only throw one class of exception: `Exception`. It seemed convenient, as a generic `MalformedResponseParser` could be used with any service, not just S3, but for that to be true I think I would need to define an `Aws\ParserException` and translate JSON and SimpleXML errors into that.

src/S3/MalformedResponseParser.php
```diff
@@ -0,0 +1,43 @@
+<?php
+namespace Aws\S3;
+
+use Aws\Api\Parser\AbstractParser;
+use Aws\Api\Parser\Exception\ParserException;
+use Aws\CommandInterface;
+use Aws\Exception\AwsException;
+use Psr\Http\Message\ResponseInterface;
+
+class MalformedResponseParser extends AbstractParser
```
mtdowling: Maybe this should be RetryMalformedResponseParser to better show the intent?

jeskew: @mtdowling Renamed to RetryableMalformedResponseParser.

mtdowling:
Looks good. I think there should be new unit tests for the new classes, but after that: :shipit:

PR #760, opened 2015-09-11
Feature/annotate clients

jeskew:
This PR adds back the "@method" annotations that powered autocomplete and static analysis in v2.

/cc @mtdowling


build/PhpFileLinterTrait.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+trait PhpFileLinterTrait
+{
+
```
mtdowling: Extra new line

build/PhpFileLinterTrait.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
```
mtdowling: Should this be marked `@internal`?

build/PhpFileLinterTrait.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param string $path
+     *
+     * @return bool
+     */
+    private function lintFile($path)
+    {
+        static $linter;
+
+        if (empty($linter)) {
+            $linter = !empty(opcache_get_status(false)['opcache_enabled']) ?
```
mtdowling: ? on the next line

build/annotate-clients.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
```
mtdowling: This file should at least have a comment to describe what it does and the usage.

build/annotate-clients.php
```diff
@@ -0,0 +1,43 @@
...
+$options = getopt('', ['class:', 'tag:', 'all']) + ['class' => [], 'tag' => []];
+
+array_walk($options, function (&$value) {
+    if (!is_array($value)) {
+        $value = [$value];
+    }
+});
+
+if (isset($options['all'])) {
+    $options['class'] = array_values(array_map(function (array $manifest) {
```
mtdowling: What is this code doing? Are you supposed to pass in the classes to update as an argument?

build/ClientAnnotator.php
```diff
@@ -0,0 +1,207 @@
...
+            $docLines []= ' *';
+        }
+
+        // append API @method annotations
+        $docLines = array_merge($docLines, array_map(function (array $op) {
+            return " * @method {$op['return']} {$op['signature']} {$op['doc']}";
+        }, $this->getMethodAnnotations()));
+
+        // add back the closing line
+        $docLines []= $lastLine;
```
mtdowling: Space between variable and array push token

build/ClientAnnotator.php
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        return $endpoint;
+    }
+
+    private function getDefaultDocComment()
+    {
+        return <<<EODC
+/**
+ * This client is used to interact with the **{$this->getApiDefinition()['metadata']['serviceFullName']}** service.
```
mtdowling: Can this just become something much simpler like: `{$this->getApiDefinition()['metadata']['serviceFullName']} client`

build/ClientAnnotator.php
```diff
@@ -0,0 +1,207 @@
...
+        $lastLine = array_pop($docLines);
+
+        // add a padding line if needed
+        if (' *' !== end($docLines)) {
+            $docLines []= ' *';
+        }
+
+        // append API @method annotations
+        $docLines = array_merge($docLines, array_map(function (array $op) {
+            return " * @method {$op['return']} {$op['signature']} {$op['doc']}";
```
mtdowling: What's the motivation for adding the description of the method in the method annotation? It adds a lot of extra weight to each client? I'm not even sure where this text would show up in the editor. I'd prefer the doc be used for just listing which versions of an API the method is compatible with.

mtdowling:
How does this affect documentation generation?

jeskew:
@mtdowling The new annotations don't end up in the docs.

Makefile
```diff
@@ -92,6 +92,14 @@ compile-json:
 	php -dopcache.enable_cli=1 build/compile-json.php
 	git diff --name-only | grep ^src/data/.*\.json\.php$ || true
 
+annotate-clients: clean
+	php -dopcache.enable_cli=1 build/annotate-clients.php --tag=latest
```
mtdowling: Why is enabling the opcache necessary?

mtdowling:
I see some magic happening here with using the opcode cache to lint vs linting a script using `php -l`. What's the reasoning for this? If it's for performance, can you provide the numbers on how much this influences the build time for compiling JSON scripts and generating docblocks?

build/docs/theme/class.latte
```diff
@@ -275,7 +275,7 @@
         </div>
         {/if}
 
-        {if $ownMagicMethods}
+        {if $ownMagicMethods && (!$class->parentClass || $class->parentClass->name !== 'Aws\AwsClient')}
```
mtdowling: Is there a way to comment in latte scripts? If so, this definitely warrants a comment to describe why this is needed.

Makefile
jeskew: That enables fast file linting with `opcache_compile`. Unfortunately, it turns out that you cannot lint code declaring symbols in this way, so I can remove that parameter.

build/annotate-clients.php
```diff
@@ -0,0 +1,64 @@
...
+        $namespace = Aws\manifest($endpoint)['namespace'];
+
+        return "Aws\\{$namespace}\\{$namespace}Client";
+    }, $alteredApiFiles));
+    $options['class'] = array_merge($options['class'], $clientsWithChangedApis);
+}
+
+foreach (array_unique($options['class']) as $classToUpdate) {
+    // Update the @method annotations on a client.
+    (new ClientAnnotator($classToUpdate))
```
mtdowling: Can you split this into separate statements for readability?

build/annotate-client-locator.php
```diff
@@ -0,0 +1,28 @@
...
+
+$annotations = array_map(function ($namespace) {
+    return " * @method \\Aws\\{$namespace}\\{$namespace}Client"
+        . " create{$namespace}(array \$args = [])";
+}, $namespaces);
+$previousAnnotationPattern = '/^\* @method'
+    . ' \\\\Aws\\\\(?:[a-zA-Z0-9]+)\\\\(?:[a-zA-Z0-9]+)Client'
+    . ' create(?:[a-zA-Z0-9]+)\\(array \$args = \\[\\]\\)/';
+
+(new ClassAnnotationUpdater(
```
mtdowling: This should be separate statements I think. First create, and then call. Spanning multiple lines in a ctor and then calling a method in the expression makes this hard to read.

build/ClientAnnotator.php
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public function updateApiMethodAnnotations()
+    {
+        return (new ClassAnnotationUpdater(
+            $this->reflection,
+            $this->getMethodAnnotations(),
+            $this->getDefaultDocComment(),
+            '/^\* @method \\\\Aws\\\\Result /'
+        ))
```
mtdowling: This should be two statements.

build/ClassAnnotationUpdater.php
```diff
@@ -0,0 +1,114 @@
...
+        // add a padding line if needed
+        if (' *' !== end($docBlockLines)) {
+            $docLines []= ' *';
+        }
+
+        // append API @method annotations
+        $docBlockLines = array_merge($docBlockLines, $this->linesToAppend);
+
+        // add back the closing line
+        $docBlockLines []= $lastLine;
```
mtdowling: `$docBlockLines []` -> `$docBlockLines[]`

mtdowling:
Should we also add the async magic methods for sending operations (e.g., `listBucketsAsync`)?

jeskew:
The magic was introduced with `.json.php` compilation build step, where fast linting makes a real difference:
```
a45e60c78b4f:aws-sdk-php eskewj$ time php build/compile-json.php 

real    0m9.634s
user    0m7.076s
sys     0m1.960s
a45e60c78b4f:aws-sdk-php eskewj$ time php -dopcache.enable_cli=1 build/compile-json.php 

real    0m2.051s
user    0m1.807s
sys     0m0.107s
```

Unfortunately, I ended up having to use `php -l` for docblock generation, as the opcache cannot recompile code that has already been loaded into a given process if it declares named symbols. `php -l` seems to be fast enough.
```
a45e60c78b4f:aws-sdk-php eskewj$ time php build/annotate-clients.php --all

real    0m2.900s
user    0m1.909s
sys     0m0.721s
```

jeskew:
@mtdowling I think you've got a point about the async methods. I'll add those in and remove the ini param from the makefile.

build/docs/theme/class.latte
GrahamCampbell: what if it's expected twice so the parent of the parent is Aws\AwsClient?

GrahamCampbell: you really need a loop here

mtdowling:
The return value of an async method is a Guzzle promise, not a result object. After that is changed, :ship:

Issue #759, opened 2015-09-10
Renaming large files using PHP StreamWrapper


lonormaly:
I try to "move" or "rename" large file (>5GB) from one bucket to the other.
The operation succeed on files smaller than 5GB but fails on bigger.
Does the rename operation of the StreamWrapper supports larger files?

rename('s3://'.S3_TEMP_BUCKET_MEDIA.'/'.$tempPath, 's3://'.S3_BUCKET_MEDIA.'/'.$path);

If not, what is the proper way of doing it?

P.S - I started from MultipartUpload but it doesn't seem right...

Cheers

jeskew:
Which version of the SDK are you using?

lonormaly:
Latest - 3.3.4

jeskew:
`rename` should work the way you're using it. Does the operation fail silently, or do you get an error and stack trace?

lonormaly:
Silently I guess... I don't get any notification about its success or failure; I waited more than an hour and checked at the target path and the file wasn't there so I assumed it failed.

Did you try it on large files (I tried on 18GB file)

jeskew:
`rename` would emit warnings and return `false` instead of throwing an exception (similar to what would happen when operating on local files). Did the function return `true` and not send anything to your PHP/Nginx/Apache logs?

lonormaly:
It didn't return anything. 
I waited long time.
Should the operation take few ms or should I expect to wait minutes / (hours?)

jeskew:
It's a blocking operation that always returns a boolean value. On a file system, rename operations can fail for a number of reasons (permissions, nonexistent destination directories, etc), which will cause `rename` to [return false](http://php.net/manual/en/function.rename.php).

Once the function returns, the operation will have either succeeded or failed. You wouldn't have to wait.

lonormaly:
I'm aware of that, still what I described occurred repeatedly...

Did you try to rename large files yourself?

On Friday, September 11, 2015, Jonathan Eskew <notifications@github.com>
wrote:

> It's a blocking operation that always returns a boolean value. On a file
> system, rename operations can fail for a number of reasons (permissions,
> nonexistent destination directories, etc), which will cause rename to return
> false <http://php.net/manual/en/function.rename.php>.
>
> Once the function returns, the operation will have either succeeded or
> failed. You wouldn't have to wait.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #763, opened 2015-09-14
Add a multipart copy abstraction

jeskew:
This PR allows for copying large objects. It uses the multipart abstraction set up for Glacier and S3 uploads, but I had to refactor `Aws\Multipart\AbstractUploader` to isolate interaction with streams, as no streams are involved in a multipart copy. The [first commit](https://github.com/jeskew/aws-sdk-php/commit/67fb8ee1e35161f4e8d5c8893c78d0f85216f2d7) is just refactoring all the non-stream stuff into `Aws\Multipart\AbstractUploadManager`, from which `Aws\Multipart\AbstractUploader` descends.

The [second commit](https://github.com/jeskew/aws-sdk-php/commit/55a46135a116f6695a884190516e21636bbd5260) adds `Aws\S3\MultipartCopy`, which is a descendent of  `Aws\Multipart\AbstractUploadManager`. It also adds a `copy` method to `Aws\S3\S3Client`, which calls `headObject` on a source and then performs either a `copyObject` operation or a multipart copy, depending on the source's size. `Aws\S3\StreamWrapper::rename` now calls `copy` instead of `copyObject` to support objects larger than 5GB.

This would address #759.

/cc @mtdowling


lonormaly:
This is more than great! When will the PR be merged? Can't wait to test it :)

jeskew:
@lonormaly It still needs to be reviewed. You can check out the feature branch directly if you'd like to do some exploratory testing.

jeskew:
@mtdowling The latest commit adds support for copying versions of objects, as that portion of the key [cannot be URL encoded](https://github.com/aws/aws-sdk-js/issues/727).

src/Multipart/AbstractUploadManager.php
```diff
@@ -0,0 +1,295 @@
...
+use GuzzleHttp\Promise;
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7;
+use InvalidArgumentException as IAE;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Encapsulates the execution of a multipart upload to S3 or Glacier.
+ */
+abstract class AbstractUploadManager implements Promise\PromisorInterface
```
mtdowling: Mark as `@internal`?

mtdowling:
This looks good. The only thing that concerns me is making a bunch of stuff protected, but if we mark that class as `@internal`, I think this can be :ship:'d

lonormaly:
Please! This one is really important for us :+1:

PR #769, opened 2015-09-18
Add some tests to catch memory and resource links

jeskew:
/cc @mtdowling

### DIFF
--- a/features/performance/streaming.feature
+++ b/features/performance/streaming.feature
@@ -1,0 +1,30 @@
+# language: en
+@performance @streaming
+Feature: Streaming transfers consume a fixed amount of memory
+
+  Scenario Outline: Streaming uploads are O(1) in memory usage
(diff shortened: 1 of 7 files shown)


features/performance/streaming.feature
```diff
@@ -0,0 +1,30 @@
...
+    And I take a snapshot of my resources
+    When I upload the file
+    Then I should not have leaked any resources
+
+    Examples:
+    | bytes       |
+    | 2048        |
+    | 2097152     |
+    | 20971520    |
+    | 2147483648  |
```
mtdowling: Do these examples actually cause us to try with a 2gb file?

jeskew: Absolutely!

jeskew: I'll drop it to 200MB.

mtdowling:
:ship:

PR #771, opened 2015-09-19
Fix/s3 connection error handling

jeskew:
To address some reported issues with the retry handling on the S3 client, this PR adds a few more statements to S3's custom retry handler. This should prevent 400 socket timeout errors from being infinitely retried and should ensure connection reset errors are retried within the bounds of the max retry limit.

/cc @mtdowling


till:
@jeskew Neat, that looks like it could do the trick.

PR #776, opened 2015-09-24
added functionality for chmod in S3 StreamWrapper

femans:
chmod works now too, which is very useful because there is no option yet to upload files through a frontend of a webapp to s3 and have them available for the end user, just using the streamwrapper. 
Well, that is what I needed it for.

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -320,10 +320,9 @@ class StreamWrapper
      *                        implemented, which is used for chmod.
      *                        For other permissions altering functions
      *                        other values are used
-     * @param int    $mode    Permissions. 700-range permissions map to
-     *                        ACL_PUBLIC. 600-range permissions map to
-     *                        ACL_AUTH_READ. All other permissions map to
-     *                        ACL_PRIVATE. Expects octal form.
+     * @param int    $mode    unix-style filesystem permissions in octal mode
+     *                        last digit maps to public-read or public-read-write
+     *                        all other modes map to private
      */
     public function stream_metadata( $path , $option, $mode ) {
         $this->initProtocol($path);
@@ -332,7 +331,7 @@ class StreamWrapper
 
         if( $option == STREAM_META_ACCESS ) {
             if (!isset($params['ACL'])) {
-                $params['ACL'] = $this->determineAcl($mode);
+                $params['ACL'] = $this->unixMode2Acl($mode);
             }
             return $this->boolCall(function () use ($params, $path) {
                 $this->getClient()->putObjectAcl($params);
@@ -898,6 +897,29 @@ class StreamWrapper
             case '7': return 'public-read';
             case '6': return 'authenticated-read';
             default: return 'private';
+        }
+    }
+
+    /**
+     * method to map unix filesystem permissions to ACL.
+     *
+     * @param int $mode     unix-style filesystem permissions in octal mode
+     *                      last digit determines if it is public-read or 
+     *                      public-read-write; all other modes are mapped to
+     *                      private
+     *
+     * @return string
+     */
+    private function unixMode2Acl($mode)
+    {
+        switch (substr(decoct($mode), 2, 1)) {
+        case '7': 
+        case '6':
+            return 'public-read-write';
+        case '5': 
+        case '4': 
+            return 'public-read';
+        default: return 'private';
         }
     }
 
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -312,6 +312,39 @@ class StreamWrapper
         return $this->triggerError("File or directory not found: $path", $flags);
     }
 
+    /** 
+     * Support for chmod()
+     *
+     * @param string $path    Directory which should be created.
+     * @param int    $option  Currently only STREAM_META_ACCESS has been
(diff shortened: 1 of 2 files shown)


jeskew:
I like the idea of supporting `chmod` with the S3 StreamWrapper, but I'm not sure I agree with how you're mapping permissions. Why is this looking at the user permissions instead of global permissions? I would personally be surprised if calling `chmod('s3://bucket/key', 0700)` made an object world-readable.

jeskew:
I'd prefer to have @mtdowling weigh in before proceeding.

femans:
Hi Jeskey, honestly, I think you are right. I didn´t think about it too much because I used the mapping code of mkdir, which I assumed to be right. But is that then sound, or should this too be different? 
I wouldn´t mind improving the mapping and commit it.

femans:
Ok I examined the code, and it appears indeed at least odd to me how the permissions are mapped, if not a minor security issue. I don´t know how I overlooked this to begin with, except lack of time. I only used chmod to set public-read files. I think this should be changed as well for mkdir.

My suggestion would be to map the following canned acl´s:
public-read: the third digit is 4 or 5
public-read-write: the third digit is 6 or 7
private: the rest

the rest don´t get mapped. The second digit is omitted. Its a little rudimentary, but I guess more accurate than allowing public-read for 7xx, and authenticated-read for 6xx.

I will write the code tomorrow-ish and submit another pull request.

femans:
So if the suggested mapping is agreed, I will build it in this branch.

femans:
I added the alternative mapping function for chmod, next to the one for mkdir. It isn´t very clean to have two different mapping functions for no apparent reason, but at least mkdir functionality is not compromised.

mtdowling:
Octal permissions don't cleanly map to Amazon S3 ACL permissions, and I don't think we can map octal permissions to ACLs in a way that will make sense. I think that coming up with different ways to map octal to ACL may make sense for some users, it will likely confuse many more. The current behavior is to return false, telling the user that the stream wrapper is unable to chmod the file. I think that is the appropriate behavior.

For the use case mentioned here, the best way to set an ACL is using stream context options when creating the file.

femans:
Hi Michael, I totally agree that octal permissions don´t map well.
However I don´t have the experience that the current behaviour is returning false, because I simply got a not implemented error, breaking my code, which costed me ample time debugging (which brought me to the idea of implementing it). For a large code base this makes it more complicated to use it as a drop in library, for which I think it was designed; otherwise I could just as well write my own wrapper to do filestuff, (which was my other idea). 
I think it would be more prudent to have it at least available, returning false in some or all cases, or the default acl, or do something useful, maybe depending stream context options. I agree with Jonathan on this, I also like the idea of support for chmod. 
Besides this, the mapping I designed is less confusing than the one in place for mkdir, (no offense, please). Because if you do mkdir('foo/bar', 0700), you don´t expect it to be publicly readable (now it is), but with 0655 you do (now it isn´t). However since S3 doesn´t really work with directories I don´t regard this as a big issue, although I didn´t really test it through.

jeskew:
After spending some time trying to come up with the "right" octal to ACL mapping, I have to say I agree with @mtdowling and I don't think there's any way to do so cleanly. Would it help your use case if `stream_metadata` were implemented and always returned `false`?

femans:
Hi, I think it would be an improvement over a non-implemented. And there is obviously no perfect mapping possible, I agree.
However, for us I don´t see an alternative to either keep using my fork, or creating my own streamwrapper which would be a spinoff of the one in the sdk, because there are too many private methods and variables to be able to simply subclass it. Both options are preferred to changing the aws settings in the code, because I want the codebase to stay filesystem agnostic. In the meanwhile both options are suboptimal, because staying up to date with the official sdk requires merging every time. But that is ok.

jeskew:
@femans Is there a reason why using options set on a stream context wouldn't solve your issue? You stated that you needed to upload items to S3 and have them be publicly readable, which can be accomplished as follows:
```php
$handle = fopen("s3://$bucketName/$key", 'w', null, stream_context_create([
    's3' => ['ACL' => 'public-read'],
]));
fwrite($handle, "I'm publicly visible!");
fclose($handle);
```

Having `chmod` return `false` instead of throwing an error sounds like a better experience, but I would want to be able to accurately model `0642` before adding support for a full-fledged `chmod`. It looks like this wouldn't be possible with canned ACLs.

femans:
As I said, I would like the code to remain as agnostic as possible of the filesystem. This makes it easier to deploy across different types of servers. 
This means using the common operations to do common things, most preferably leaving the code in tact that was there while migrating to s3, and not have a filesystem specific handler; otherwise I have to make a wrapper for this that can switch between different handles for different filesystems. In which case using my own fork is simply easier at this point. 
But I will look into the option of wrapping it in a different way as you suggest. Thanks!

jeskew:
Sure. You might also want to take a look at filesystem abstraction libraries like [Flysystem](http://flysystem.thephpleague.com/) and [Gauffrette](https://github.com/KnpLabs/Gaufrette). Flysystem only supports two levels (completely private or completely public), but it sounds like that would fit your use case.

I'll open a separate PR to have `stream_metadata` always return `false`.

femans:
Thanks Jonathan for the suggestions, and looking into it.

PR #781, opened 2015-09-28
fix Validator's checkCanString behaviour

MaxKlyukin:


jeskew:
Thanks for submitting this!

PR #782, opened 2015-09-28
improved acl mapping

femans:
as suggested and promised in https://github.com/aws/aws-sdk-php/pull/776#issuecomment-143505174 

I don´t expect this to get pulled without some discussion about what would be the best mapping; however imo the current mapping is erroneous and needs improving; 700 should of course be as private as possible.

I chose this simple form:
public-read: the third digit is 4 or 5
public-read-write: the third digit is 6 or 7
private: the rest

rationale:
the x-bit has no meaning in acl, so these options make the most sense. 0 or 1 is no permissions, 2 or 3 are write-but-not-read, which is rarely seen in the wild and of course has no corresponding acl, which leaves read and read+write.
The group (2nd) digit cannot really be described by canned acl policies; it is not authenticated-read, because this does not refer to the private group of the file.
The owner (1st) digit doesn´t get a lot of flexibility in canned acl´s, there is only full access for the owner, no matter which option is chosen. So maybe you might want an error or something if anything else than 6 or 7 is given, in stead of simply give full access to the owner no matter what is requested, which is my solution.

There may be better ways to improve the mapping than using canned acl, I am open to suggestions. Also on expanding the unit tests.

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -861,10 +861,14 @@ class StreamWrapper
      */
     private function determineAcl($mode)
     {
-        switch (substr(decoct($mode), 0, 1)) {
-            case '7': return 'public-read';
-            case '6': return 'authenticated-read';
-            default: return 'private';
+        switch (substr(decoct($mode), 2, 1)) {
+        case '7': 
+        case '6':
+            return 'public-read-write';
+        case '5': 
+        case '4': 
+            return 'public-read';
+        default: return 'private';
         }
     }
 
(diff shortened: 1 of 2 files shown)


jeskew:
I don't think the behavior of permissions on `mkdir` can change without a major version bump. `chmod` can use a different mapping, and maybe there can be an opt-in mechanism for mkdir. I can pick this up if you'd prefer, as this is more work than you initially signed up for with #776.

femans:
If you agree with this mapping, I can add it to the chmod branch using a different function name for the mapping method, this is not too much work. I can do this easily tomorrow. Let´s discuss this further in that pull request, before this gets more confusing :D
I can understand you are hesitant to change mkdir mapping in the same version, and I don´t immediately see how to build in this opt-in mechanism, so this is something which you could pick up, yes. Then I leave this pull request to you to decide what you want to do with it.

jeskew:
Closing in favor of #776.

PR #790, opened 2015-10-05
Add async counterparts to S3Client::upload and S3Client::copy

jeskew:
This PR adds `uploadAsync` and `copyAsync` methods to `Aws\S3\S3Client`, both of which return promises that will be fulfilled with the result of the underlying operation. `upload` and `copy` now simply call `uploadAsync` and `copyAsync` respectively and wait on the returned promise.

/cc @mtdowling

### DIFF
--- a/src/S3/MultipartCopy.php
+++ b/src/S3/MultipartCopy.php
@@ -57,6 +57,11 @@ class MultipartCopy extends AbstractUploadManager
         ]);
     }
 
+    public function copy()
+    {
+        return $this->upload();
+    }
+
     protected function loadUploadWorkflowInfo()
     {
         return [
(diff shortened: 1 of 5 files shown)


src/S3/S3Client.php
```diff
@@ -467,43 +500,76 @@ public function copy(
...
-        if ($maxCopyableSize < $objectStats['ContentLength']) {
-            return (new MultipartCopy($this, $copySource, [
-                'bucket' => $bucket,
-                'key' => $key,
-                'source_metadata' => $objectStats,
-                'acl' => $acl,
-            ] + $options))
-                ->upload();
-        }
+        return $this->headObjectAsync($headParams)
```
jeremeamia: Not sure, but maybe a coroutine would make this easier to read/understand.

fntlnz: :+1:  for the coroutine idea!

src/S3/MultipartCopy.php
```diff
@@ -57,6 +57,11 @@ public function __construct(S3Client $client, $source, array $config = [])
         ]);
     }
 
+    public function copy()
```
mtdowling: Missing docblock

src/S3/S3Client.php
```diff
@@ -467,43 +501,81 @@ public function copy(
...
+     *
+     * @see self::copy for more info about the parameters above.
+     * @return PromiseInterface     Returns a promise that will be fulfilled
+     *                              with the result of the copy.
+     */
+    public function copyAsync(
+        $fromBucket,
+        $fromKey,
+        $bucket,
+        $key,
```
mtdowling: Can you rename these two parameters, bucket and key, to something like `$toBucket`, `$toKey` or `$destBucket`, `$destKey`?

mtdowling:
Couple pieces of feedback, but otherwise, :shipit:

PR #796, opened 2015-10-16
Update the API docs in some helpful ways

jeskew:
/cc @mtdowling


build/docs/classes/DocsBuilder.php
```diff
@@ -226,6 +244,31 @@ private function updateClients(array $services)
         }
     }
 
+    private function updateAliases(array $services, array $compatibleVersions)
```
mtdowling: I think this approach is fine for now. Let's try to see if we can come up with a better approach going forward though.

mtdowling:
I gave this a ship on Friday via email, but it isn't shown here for some reason. Let me try again: :ship:

jeskew:
I didn't commit and push one final change on Friday, which was to generate the sitemap after the API docs had been written out. The previous sitemap only contained links to class, namespace, and function docs.

jeskew:
This PR:
* Adds a listing of an API version's methods underneath each version link on a client class's page,
* Adds redirects for forward-compatible APIs,
* Updates the API doc sitemap after writing out the version docs, and
* Automates the production of the quick links on the homepage.

/cc @chrisradek

PR #798, opened 2015-10-21
Respect client-supplied location constraints on createBucket commands

jeskew:
This PR allows custom location constraints to be supplied.

/cc @mtdowling @chrisradek

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -624,9 +624,13 @@ class S3Client extends AwsClient
         return static function (callable $handler) use ($region) {
             return function (Command $command, $request = null) use ($handler, $region) {
                 if ($command->getName() === 'CreateBucket') {
-                    if ($region === 'us-east-1') {
+                    $locationConstraint = isset($command['CreateBucketConfiguration']['LocationConstraint'])
+                        ? $command['CreateBucketConfiguration']['LocationConstraint']
+                        : null;
+
+                    if ($locationConstraint === 'us-east-1') {
                         unset($command['CreateBucketConfiguration']);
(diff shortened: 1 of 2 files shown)


src/S3/S3Client.php
```diff
@@ -623,7 +623,9 @@ private function getLocationConstraintMiddleware()
         $region = $this->getRegion();
         return static function (callable $handler) use ($region) {
             return function (Command $command, $request = null) use ($handler, $region) {
-                if ($command->getName() === 'CreateBucket') {
+                if ($command->getName() === 'CreateBucket'
+                    && empty($command['CreateBucketConfiguration']['LocationConstraint'])
+                ) {
                     if ($region === 'us-east-1') {
```
mtdowling: I think the behavior of removing the CreateBucketConfiguration structure should always be performed when the supplied value is `us-east-1`. Right now it's only removed in the provided value is empty. If the user supplies `us-east-1`, we should also remove that as that is the current behavior.

jeskew: How would a client configured to talk to `us-east-1` then specify a location constraint of a different region?

mtdowling: I don't see `us-east-1` listed as a valid LocationConstraint in the S3 docs: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html. You're supposed to omit the LocationConstraint altogether if you want to create a bucket in `us-east-1`. I was saying that the pull request should be updated to also remove the LocationConstraint if the value it's set to is `us-east-1`.

mtdowling:
:ship:

PR #806, opened 2015-10-28
Build out an index of terms to search

jeskew:
This PR will add API operations and class methods to the client-side search functionality in our API docs.

/cc @mtdowling @chrisradek


chrisradek:
:shipit:

PR #812, opened 2015-11-03
Let the RT keep track of session IDs

jeskew:
Rather than have the session handler keep track of the current session id, this PR punts that responsibility to the runtime. `session_id` is therefore called in the `close` handler instead of the `open` handler, as some users had reported that certain runtimes would not return a valid session ID when `session_id` was called before `session_start` had completed.

/cc @chrisradek @mtdowling

### DIFF
--- a/src/DynamoDb/SessionHandler.php
+++ b/src/DynamoDb/SessionHandler.php
@@ -32,9 +32,6 @@ class SessionHandler implements \SessionHandlerInterface
     /** @var string Stores serialized data for tracking changes. */
     private $dataRead;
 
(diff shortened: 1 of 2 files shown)


chrisradek:
:ship: :it:

dhensby:
This is going to mess with a possible resolution #821 as we need to track the changing of session ids to be able to reliably determine if session data needs to be written

jeskew:
@dhensby I don't think that's true, both because this PR was for the current 3.x branch of development and because #821 should be able to support HHVM and PHP 7.

Issue #585, opened 2015-05-21
AbstractLockingStrategy discards DynamoDbException data


benp84:
After setting up the DynamoDb session handler I noticed my app was regenerating its session data on every request. It turns out the session handler was silently failing. The culprit was these lines in AbstractLockingStrategy:

    } catch (DynamoDbException $e) {
      return false;
    }

Why is this exception data thrown away? I'd really like this exception to bubble up to my global exception handler so I can know about the problem and how to solve it. It turns out it was an IAM permission issue but there's no way to know this without a debugger.

mtdowling:
From what I understand, stream wrapper implementations are not supposed to throw exceptions on error, but rather return a boolean true or false value. We'll look into seeing if that's a hard requirement, and if not, see if throwing an exception here is possible. If it emits a warning *and* throws an exception (due to internal PHP issues), then that might still be a better experience than currently swallowing the exception.

jeskew:
This was addressed in v3 in #813. Please feel free to reopen if you have any questions or concerns.

PR #813, opened 2015-11-04
Trigger errors when the session handler fails to write or delete data

jeskew:
This PR converts exceptions thrown within the DynamoDB session handler into warning-level errors. The current behavior of the handler (returning false on failure) is not changed, but errors will become more visible (as requested in #585).

/cc @mtdowling @chrisradek


chrisradek:
:shipit:

PR #814, opened 2015-11-04
Add a CognitoIdentity credential provider

jeskew:
The added provider uses a Cognito Identity's anonymous APIs to create credentials. It also moves the `createCredentials` method from `Aws\Sts\StsClient` into a trait.

/cc @mtdowling @chrisradek

### DIFF
--- a/src/CognitoIdentity/CognitoIdentityProvider.php
+++ b/src/CognitoIdentity/CognitoIdentityProvider.php
@@ -1,0 +1,61 @@
+<?php
+namespace Aws\CognitoIdentity;
+
+use Aws\Credentials\Credentials;
+use GuzzleHttp\Promise;
+
+class CognitoIdentityProvider
+{
+    /** @var CognitoIdentityClient */
+    private $client;
+    /** @var string */
+    private $identityPoolId;
+    /** @var string|null */
+    private $accountId;
+    /** @var array */
+    private $logins;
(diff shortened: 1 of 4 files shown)


src/CognitoIdentity/CognitoIdentityClient.php
```diff
@@ -43,4 +44,7 @@
  * @method \Aws\Result updateIdentityPool(array $args = [])
  * @method \GuzzleHttp\Promise\Promise updateIdentityPoolAsync(array $args = [])
  */
-class CognitoIdentityClient extends AwsClient {}
+class CognitoIdentityClient extends AwsClient
+{
+    use CredentialCreatingClientTrait;
```
mtdowling: It seems weird to duplicate the `createCredentialsMethod` on multiple clients when they do exactly the same thing and don't really create credentials specific to the client. I wonder if we should just add a factory/creational function to Credentials that creates them from an array, mark the Sts method `createCredentials` deprecated, and make that simply proxy to the factory method.

jeskew: I was overzealous in consolidating the two methods -- the credentials shape returned by Cognito Identity is slightly different from that returned by STS, so I ended up keeping them separate.

mtdowling:
:ship:

Issue #815, opened 2015-11-05
ProvisionedThroughputExceededException not automatically retried for putItem requests


mikemherron:
It looks like ProvisionedThroughputExceededException are not being correctly retried by the SDK when invoking putItem and exceeding a tables provisioned throughput. Adding debug print statements to RetryMiddleware.php suggests that the delay is being calculated but not applied.

The below test script, placed in the root of a directory where the SDK is installed via composer, can be used to demonstrate this:

```php
<?php

include "vendor/autoload.php";

use Aws\DynamoDb\DynamoDbClient;

/*
 Add the below line of code to vendor/aws/aws-sdk-php/src/RetryMiddleware.php, around 
 line 120 to print out the retry delays that should be applied:
 
 <code>
 // Delay fn is called with 0, 1, ... so increment after the call.
 $command['@http']['delay'] = $delay($retries++);
 + print gmdate("H:i:s") . ":Retrying with delay:" . $command['@http']['delay'] . "\n";
 return $handler($command, $request)->then($g, $g);
 </code>
*/

function out( $msg ) {
    print gmdate("H:i:s") . ": " . $msg . "\n";
}

$options = [
    'region'    => "eu-west-1",
    'endpoint'  => "https://dynamodb.eu-west-1.amazonaws.com",
    'credentials' => [
        'key'       => "",
        'secret'    => ""
    ],
    'version'   => 'latest',
];

$client = new DynamoDbClient($options);

$tables = $client->listTables();

if ( !in_array( 'TestThroughputExceptions', $tables['TableNames' ] ) ) {   

    out( "Creating TestThroughputExceptions table");

    $client->createTable( [
        'TableName' => 'TestThroughputExceptions',
        'AttributeDefinitions' => [ 
            [ 'AttributeName' => 'ID',
              'AttributeType' => 'S' ] 
        ],
        'KeySchema' => [ 
            [ 'AttributeName' => 'ID',
              'KeyType' => 'HASH' ] 
        ],
        'ProvisionedThroughput' => [
            'ReadCapacityUnits'  => 1,
            'WriteCapacityUnits' => 1
        ]
    ]);
    
    $client->waitUntil( 'TableExists', [
        'TableName' => 'TestThroughputExceptions'
    ]);
}


while( true ) {
    out( "Starting putItem");
    $client->putItem( [
        'TableName'     => 'TestThroughputExceptions',
        'Item'          => [ 'ID' => ['S' => 'SomeID' ] ]
    ]);
[...]

PR #816, opened 2015-11-06
Backoff in MS (not seconds) and throw in some jitter

jeskew:
This PR converts the delay value provided from the SDK to Guzzle from seconds to milliseconds. Guzzle is expecting MS, so the previous behavior was to delay for trivial amounts of time (as reported in #815). Since I was already mucking about in the backoff code, I threw in some randomness.

One thing I'm not clear on is if the delay times are correct. [The DynamoDB docs advise delaying  (50 * (int) pow(2, $retries - 1)) between retries](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html), whereas the standard AWS guideline is to [wait pow(2, $retries - 1) * 100 milliseconds](http://docs.aws.amazon.com/general/latest/gr/api-retries.html). When adding jitter, should I aim to make those values match the maximum possible retry delay (which is what I've done below), or should I try to make those the average retry delay?

/cc @mtdowling @chrisradek

### DIFF
--- a/src/DynamoDb/DynamoDbClient.php
+++ b/src/DynamoDb/DynamoDbClient.php
@@ -78,7 +78,7 @@ class DynamoDbClient extends AwsClient
                 RetryMiddleware::createDefaultDecider($value),
                 function ($retries) {
                     return $retries
-                        ? (50 * (int) pow(2, $retries - 1)) / 1000
+                        ? RetryMiddleware::exponentialDelay($retries) / 2
                         : 0;
                 }
             ),
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -86,7 +86,7 @@ class RetryMiddleware
      */
     public static function exponentialDelay($retries)
     {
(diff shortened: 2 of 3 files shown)


chrisradek:
I'm not sure which way is preferred, but the JavaScript SDK uses the guidelines in the DynamoDB docs for calculating the delays for the DynamoDB service, and uses something close to the AWS guidelines as the default for other services (instead of multiplying by 100 ms, it multiplies by 30).

jeskew:
I believe [the DynamoDB docs](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html#APIRetries) imply that (50 * 2 ** (retries - 1)) should be a maximum value: "For example, **up to** 50 milliseconds before the first retry, **up to** 100 milliseconds before the second, **up to** 200 milliseconds before third, and so on."

This change will cause the delay to be an average of 25 ms before the first retry, but never more than 50 ms.

chrisradek:
:shipit:

Issue #821, opened 2015-11-12
DynamoDB SessionHandler does not recognised regenerated session IDs


dhensby:
When calling `session_regenerate_id()` `SessionHandler` does not recognise that the ID has changed.

`SessionHandler::write` doesn't check for a changed session ID and when the session data hasn't changed, it fails to identify that the session data needs to be re-written.

Steps to reproduce:

1. Create session.
2. Write static data
2. Store session.
3. Regenerate session id
4. Access new session
5. Find new session is empty

jeskew:
This appears to be a known limitation of using PHP's SessionHandlerInterface: there's no way to hook into `session_regenerate_id` using the object-oriented version of [`session_set_save_handler`](http://php.net/manual/en/function.session-set-save-handler.php). [This StackOverflow question](http://stackoverflow.com/questions/32555254/php-custom-session-handler-handling-session-regenerate-id) has a bit more detail, including a note that PHP < 5.5 won't be able to hook into `session_regenerate_id` even using the callback-based `session_set_save_handler`.

I think the session handler should keep track of the ID associated with the data it has read. Keeping that context means that the check to see if data has changed will be more reliable.

dhensby:
DynamoDB's `SessionHandler` writing empty session when the ID changes is not a limitation of PHP's `SessionHanderInterface` IMO. It's a limitation of the current implementation of `doWrite` on the `AbstractLockingStrategy`, which attempts to reduce data overhead by only sending session data to DynamoDB *when the session data has changed*.

If the `session_id()` changes, it will create a new record in DynamoDB, but it won't send the data.

The solution is for DynamoDB to detect if either the data OR the session ID has changed. This could be done 1 of 2 ways.

1. Track the originally read session ID and check to see if the written one has changed against it.
2. Ping Dynamo DB to see if a record with the ID exists and, if not, send the whole data set.

1 is more elegant in my opinion because if the session ID changes several times (and the session is written each time) the session data should be sent; we don't really need to know if the session ID has changed against the last regenerated ID we just need to know if it changed at all. It also has the advantage of saving reads against Dynamo DB.

jeskew:
Fixes to v2 (#825) and v3 (#826) have been merged. Thanks for all your help on finding the right solution to this, @dhensby!

dhensby:
Fantastic, thanks

PR #822, opened 2015-11-12
FIX Changed session IDs now have data written

dhensby:
This is a fix for #821; where sessions where IDs are regenerated but session data remains the same fail to have the data propagated to the new session record in DynamoDB.

### DIFF
--- a/src/Aws/DynamoDb/Session/SessionHandler.php
+++ b/src/Aws/DynamoDb/Session/SessionHandler.php
@@ -332,7 +332,7 @@ class SessionHandler implements SessionHandlerInterface
         $this->sessionWritten = $this->lockingStrategy->doWrite(
             $this->formatId($id),
             $data,
-            ($data !== $this->dataRead)
+            ($id !== $this->openSessionId) || ($data !== $this->dataRead)
         );
 
         return $this->isSessionWritten();
--- a/tests/Aws/Tests/DynamoDb/Session/SessionHandlerTest.php
+++ b/tests/Aws/Tests/DynamoDb/Session/SessionHandlerTest.php
@@ -122,6 +122,70 @@ class SessionHandlerTest extends AbstractSessionTestCase
         $this->assertTrue($this->handler->write('test', 'ANYTHING'));
     }
 
+    public function testSessionWritesDataOnChangedSessionID()
+    {
+        //set up our test session store and data
+        $store = array();
+        $mockSessionData = serialize(array(
+            'fizz' => 'buzz',
+        ));
+
+        //construct our mock client/strategy/handler
+        $client = $this->getMockedClient();
+        $strategy = $this->getMock('Aws\DynamoDb\Session\LockingStrategy\LockingStrategyInterface');
+        $handler = SessionHandler::factory(array(
+            'dynamodb_client' => $client,
+            'locking_strategy' => $strategy,
+        ));
+
+        $command = $this->getMockedCommand($client);
+        $command->expects($this->any())
+            ->method('execute')
+            ->will($this->returnValue(array('foo' => 'bar')));
+
+        $client->expects($this->any())
+            ->method('getIterator')
+            ->will($this->returnValue(array()));
+
+        //doRead function to fetch session data by ID
+        $strategy->expects($this->any())
+            ->method('doRead')
+            ->will($this->returnCallback(function ($id) use (&$store) {
+                if (isset($store[$id])) {
+                    return array(
+                        'expires' => time() + 10000,
+                        'data' => $store[$id],
+                    );
+                }
+
+                return array();
+            }));
+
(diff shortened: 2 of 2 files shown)


dhensby:
Unfortunately, I'm not sure how best to test this actually works as expected, my [attempted test](https://github.com/dhensby/aws-sdk-php/commit/d64791d046f2b46b35dcdf5604a9c4e6a9b6bae1) didn't work too well.

jeskew:
Since you're planning on changing session IDs, I think this PR needs to take account of the fact that you're writing to multiple IDs. Maybe `$this->dataRead` should be changed from a string to an array mapping IDs to session data?

dhensby:
@jeskew at the moment it works pretty flawlessly on my site without need to have an array mapping.

The idea isn't to run two session IDs side-by-side, but more when you want to simply give the visitor a new session ID. so the old session ID doesn't need to be kept.

`session_regenerate_id ([ bool $delete_old_session = false ] )` if delete old session is true then we definitely don't need to keep it in memory; though the fact it could be kept is relevant.

dhensby:
Also, I contest the idea that this is an "enhancement", it's a bug because at the moment if you regenerate your session ID whithout changing the session data the data is never sent to dynamodb and so you lose it.

jeskew:
What if you change the session ID twice?

dhensby:
Then it'll be written twice. 

What circumstances are you thinking it would cause a problem?

dhensby:
To be clear, the bug this is fixing is that if a call to `session_regenerate_id()` is made and the session data is not modified then the session will be lost because the data never gets sent.

This is a bug as it is a normal use of the built in PHP function

jeskew:
You're not updating the stored session ID (`$this->sessionId`) when you know that the open session's ID might have changed. If you'd prefer to handle this in a way other than using an array, that's fine. But if you're updating this class to knowingly handle multiple session IDs, it should do so throughout.

dhensby:
For the avoidance of doubt, this is to fix a bug where **`session_regenerate_id()` used with NO changes to the session data causes empty sessions to be written to DynamoDB.**

When using that as context consider the following: it doesn't matter how many times you call `session_regenerate_id()` during runtime, only the LAST `session_id()` generated will be stored in DynamoDB (unless you manually force a write). This is because the session is not written when `session_regenerate_id()` is called but, instead, when PHP's session engine says to write it.

ie:

```
session_start();
for ($i =0; $i < 10; ++$i) {
    session_regenerate_id(true);
}
die;
```
^ this will only cause ONE hit to DynamoDB with an ID that is different to the one that was used at the beginning

Therefore, the point of handling multiple sessions *in this instance* is not a valuable use of resource.

When evaluating whether the session should be written, we should only care if the session ID we are committing now has changed from the session ID we previously committed.

jeskew:
In the code posted, the `write` method will not be used. A single write will be triggered by the `__destruct` method, which will be called by the runtime at the end of the PHP process. The problem this introduces is that the session handler tracks the session data read (`$this->dataRead`), the open session ID (`$this->openSessionId`), and whether any data has been written to the session (`$this->sessionWritten`). Those are all scalar values that assume the session ID will not change.

If there were a way in PHP >=5.3.3 (which is the version v2 has to support) to hook into `session_regenerate_id`, that hook could be used to wipe out `$this->dataRead`, `$this->openSessionId`, and `$this->sessionWritten`. But since that hook was introduced in PHP 5.5, the tracking variables themselves need to be modified so that they can keep track of the ID, data, and write status of multiple session IDs, as those session IDs can be changed with no notification an unlimited number of times.

I get that the PR as written solves the particular issue you ran into under the circumstances you're expecting, but I think it can be made more general.

tractorcow:
> Those are all scalar values that assume the session ID will not change.

Isn't this assumption the problem? Because it seems that the user can have a new ID regenerated for them.

http://php.net/manual/en/function.session-regenerate-id.php

The docs here clearly state that the ID is changed, although the EXISTING session should still be retained, but recorded alongside this ID. If you are throwing away the session data after this change, then PHP session behaviour is not being maintained properly.

jeskew:
@tractorcow Exactly. Just checking the id provided to `write` does nothing to correct that assumption.

tractorcow:
Ok, I think I understand. :) I hope @dhensby can find a solution that solves than issue then.

jeskew:
Following the chat on gitter, I'm sold on calling this a bug instead of an enhancement, as we're making it harder to follow [OWASP session management recommendations](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renew_the_Session_ID_After_Any_Privilege_Level_Change).

jeskew:
Closing in favor of #825 and #826. Thanks for all your work on this and the fruitful discussion.

PR #825, opened 2015-11-13
Take account of the fact that session IDs can and should change.

jeskew:
This PR removes three protected members from the DynamoDb `SessionHandler`: `$dataRead`, `$openSessionId`, and `$sessionWritten`. In their place, the session handler uses `session_id` to get the open session ID and will look up session data and whether a session was written dynamically based on the current session ID in arrays at `$sessionDataRead` and `$sessionsWritten`. 

Since PHP does not allow `__get` methods to be declared `protected`, this PR has a side effect of making the `->dataRead` and `->sessionWritten` members publicly readable, but this information could always have been accessed from anywhere in the process using PHP's `session_*` functions.

The motivation for this PR is that the session handler as it is written is unable to detect when `session_regenerate_id` is called, as `session_set_save_handler` provides no session ID generation hook prior to PHP 5.5. Since changing the session ID after changing privilege levels is [an official OWASP recommendation](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renew_the_Session_ID_After_Any_Privilege_Level_Change), the Dynamo session handler should support doing so seamlessly.

/cc @dhensby @mtdowling @chrisradek

### DIFF
--- a/src/Aws/DynamoDb/Session/SessionHandler.php
+++ b/src/Aws/DynamoDb/Session/SessionHandler.php
@@ -66,20 +66,10 @@ class SessionHandler implements SessionHandlerInterface
      */
     protected $sessionName;
 
-    /**
-     * @var string Stores the serialized data that was read for tracking changes
-     */
-    protected $dataRead;
-
-    /**
-     * @var string Keeps track of the open session's ID
-     */
-    protected $openSessionId;
-
-    /**
-     * @var bool Keeps track of whether the session has been written
-     */
-    protected $sessionWritten;
+    /** @var array */
+    private $sessionDataRead = array();
+    /** @var array */
+    private $sessionsWritten = array();
 
     /**
      * Factory method to create a new DynamoDB Session Handler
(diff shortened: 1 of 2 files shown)


dhensby:
Thanks @jeskew I'll test this when I get into the office.

dhensby:
@jeskew OK - I've tested this locally and it behaves as expected now.

I also cherry-picked my test from https://github.com/dhensby/aws-sdk-php/commit/6f03bf320835beb590cc61ceb80e25b9554a73db and the test suite passed too.

jeskew:
@dhensby I added the test you wrote as well to this PR.

dhensby:
:)

chrisradek:
:shipit:

mtdowling:
It looks like the same type of logic applied in this PR can be done so without needing a hash of session IDs to session data. I think we could just track the expected session ID against the actual session ID and get the same bug fix.

dhensby:
@mtdowling that was my original approach in #822

PR #826, opened 2015-11-13
Change sessions written/read trackers from scalars to id => scalar mappings

jeskew:
This PR ports the changes from #825 to v3 to address the concern raised in #821 across all versions. There's less tomfoolery with `__get` and such because the instance members requiring modification were already private.

/cc @mtdowling @chrisradek


chrisradek:
:shipit:

mtdowling:
:ship:

Issue #823, opened 2015-11-12
S3 GetObject throwing String could not be parsed as XML


Halama:
Hello,
we are getting `String could not be parsed as XML` errors from S3 client `getObject` method.
It is happening randomly maybe on 1 of few thousands `getObject` requests to S3. It started few days ago.

Error stack trace:

![image](https://cloud.githubusercontent.com/assets/903531/11121660/61983d1a-8957-11e5-865a-035b2065cc11.png)

The downloaded file is gzipped CSV. You can notice in stack trace that response was 200 and the response body is also present. But for some reason promise was rejected and error handler was started.
On this line https://github.com/aws/aws-sdk-php/blob/7b369751845660edb983edded7e6b3d697baba4c/src/WrappedHttpHandler.php#L142 the XML response is expected.

Another part of stack trace with promise rejection:

![image](https://cloud.githubusercontent.com/assets/903531/11121739/cfb5e22a-8957-11e5-8a50-3a3d3cb72ef1.png)

AWS SDK version: **3.10.0**

thanks

jeskew:
You mention that the errors started appearing a few days ago; did this follow an upgrade? If so, what version did you upgrade from?

And have you found any way to reliably reproduce this error, or does it occur completely randomly?

ondrejhlavacek:
@jeskew It happened in versions 3.9.4 and 3.10.0. It might have started after upgrading from version 3.2.*, I don't recall this happening in 3.2.*. It occurs completely randomly. Series of following attempts to perform the same action didn't throw any error.

jeskew:
Do you have a full trace for the first screenshot above?

ondrejhlavacek:
@jeskew sure, I just blanked out some sensitive params and deleted the 2 parent exceptions that didn't have any aws-php-sdk related info
https://s3.amazonaws.com/keboola-logs/debug-files/2015/11/12/2015-11-12-12-33-35-5644790faa91a-exception-excerpt.html

jeskew:
I see that you're using the 'SaveAs' parameter on the `getObject` call -- is it possible that what's causing the promise to be rejected is a file system error? Just a hunch, but that may explain why the issue is so hard to reproduce.

jeskew:
@Halama @ondrejhlavacek Is there any chance you could try with `dev-master`? #827 should make the underlying error more visible, but it has not yet been included in a release.

ondrejhlavacek:
[...]

PR #827, opened 2015-11-13
Allow the error handler to recover from error parser errors

jeskew:
Should help with the visibility of errors when responses do not necessarily follow the expected error format (e.g., #823).

/cc @mtdowling @chrisradek

### DIFF
--- a/src/WrappedHttpHandler.php
+++ b/src/WrappedHttpHandler.php
@@ -145,6 +145,8 @@ class WrappedHttpHandler
                     . "{$parts['message']} - " . $err['response']->getBody();
             } catch (ParserException $e) {
                 $parts = [];
+                $serviceError .= ' Unable to parse error information from '
+                    . "response - {$e->getMessage()}";
             }
 
             $parts['response'] = $err['response'];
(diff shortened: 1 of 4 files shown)


src/WrappedHttpHandler.php
```diff
@@ -138,11 +139,15 @@ private function parseError(
...
             $parts = ['response' => null];
         } else {
-            $errorParser = $this->errorParser;
-            $parts = $errorParser($err['response']);
+            try {
+                $parts = call_user_func($this->errorParser, $err['response']);
+                $serviceError .= " {$parts['code']} ({$parts['type']}): "
+                    . "{$parts['message']} - " . $err['response']->getBody();
+            } catch (ParserException $e) {
+                $parts = [];
```
mtdowling: Is there any way to add some kind of mention of the parse error in the returned error?

jeskew: I added an extra phrase to the `$serviceError` that points out that the body could not be parsed.

mtdowling:
:ship:

PR #828, opened 2015-11-18
add autoload.php for those people who does not use composer to manager their code

huyanping:


jeskew:
Thanks for your work on this, but I think this use case is already covered by our [phar and zip distributions](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/getting-started/installation.html). The zip distribution contains all of the SDK, its dependencies, and an autoloader.

huyanping:
Thanks, I will talk to my colleagues about this

PR #832, opened 2015-12-02
Feature/move integ suite to cucumber

jeskew:
We currently have three test suites, one of which is almost never run. This PR moves the tests from our 'integ' suite into the cucumber suite and converts integ tests to unit tests where applicable.

/cc @mtdowling @chrisradek


jeremeamia:
Nice work. :smile:

mtdowling:
Nice! You'll need to update the Makefile: https://github.com/aws/aws-sdk-php/blob/master/Makefile#L48

tests/ConcurrencyTest.php
```diff
@@ -0,0 +1,101 @@
+<?php
```
mtdowling: I think this needs to still be an integration test. We need to ensure that we're actually able to send requests concurrently and the handler bindings actually work. I'd feel much more confident in the implementation if we still sent real requests for these basic tests.

jeskew: Since we support pluggable HTTP backends, wouldn't that be more a test of Guzzle and Guzzle\Promises than of the SDK?

mtdowling: It would be testing promises, Guzzle bindings with the SDK, SDK clients, et. al. At minimum, I think we need integration tests for the bindings (I can't remember if those exist somewhere else offhand). Since we ship the SDK with Guzzle bindings, I think we should have an integration test that makes sure those bindings work correctly. I think integration tests for this are just a really easy way to make sure something isn't super broken before a release. It's nice just to make sure that the bindings are correct I think.

jeskew:
@mtdowling I updated the makefile and moved the concurrency test back to the integ suite. There was an unrelated hiccup in Travis (var_dump overloading was set to 2 for PHP 7, causing many of the validator tests to fail), which is addressed in the last commit.

mtdowling:
:ship:

PR #834, opened 2015-12-03
Add a helper function to generate SMTP password from credentials

jeskew:
Based on [this section of the SES docs](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-credentials.html#smtp-credentials-convert).

/cc @mtdowling @chrisradek

### DIFF
--- a/src/Ses/SesClient.php
+++ b/src/Ses/SesClient.php
@@ -1,6 +1,8 @@
 <?php
 namespace Aws\Ses;
 
+use Aws\Credentials\CredentialsInterface;
+
 /**
  * This client is used to interact with the **Amazon Simple Email Service (Amazon SES)**.
  *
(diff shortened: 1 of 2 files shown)


chrisradek:
:shipit:

src/Ses/SesClient.php
```diff
@@ -83,4 +85,15 @@
  * @method \Aws\Result verifyEmailIdentity(array $args = [])
  * @method \GuzzleHttp\Promise\Promise verifyEmailIdentityAsync(array $args = [])
  */
-class SesClient extends \Aws\AwsClient {}
+class SesClient extends \Aws\AwsClient
+{
+    public static function generateSmtpPassword(CredentialsInterface $creds)
```
mtdowling: Sorry for being late, but this needs a docblock

PR #706, opened 2015-07-27
Generate shared examples in API documentation

DanielNi:
Adds functionality for generating shared documentation examples from the language-agnostic JSON format. These new examples will live alongside the JSON api documents.

The code in this PR loads the examples from the DocsBuilder.php, which passes them into the createHtmlForOperation function. From there, the SharedExampleBuilder class parses the JSON and creates a PHP example at the end of each operation in the documentation.

PR #837, opened 2015-12-07
Feature/generate shared examples

jeskew:
This PR merges in #706 and fixes a few merge conflicts.

/cc @mtdowling @chrisradek


jeskew:
The test failure is unrelated and is addressed in #832

mtdowling:
Is there any way we can add tests to the example builder?

Given that we represent date/time values as DateTime objects in responses, do you want to represent that somehow in a response? What about blob types in a response? If you do want to do that, then you'll need to walk the model at the same time as walking the example data.

jeskew:
@mtdowling I updated the PR to walk the model and insert marker values for dates and blobs. This allows for better support of structures and maps, too.

mtdowling:
:ship:

PR #838, opened 2015-12-08
Added ReleaseLabel field for EMR RunJobFlow starting for AMI version 4

pampy:
Starting with EMR release 4, the field "ApiVersion" is no longer supported. It's replaced by "ReleaseLabel", according to this announcement:
http://docs.aws.amazon.com/ElasticMapReduce/latest/ReleaseGuide/emr-release-differences.html

The current implementation does not allow to set a ReleaseLabel, since the validation will remove it before sending the request.

I added the parameter to the JSON API description, which now allows us to start new EMR 4 clusters.


jeskew:
Hi @pampy,

Thanks for putting this together! I saw that the EMR model was a little out of date and pulled in an up-to-date version in the latest release (3.11.5).

PR #844, opened 2015-12-11
Ensure errors with unparseable responses are retried when idempotent

jeskew:
This PR makes a code path in the S3 retry handler more specific by only checking error-level responses for the string "Your socket connection to the server..." The branching logic was also catching 200-level responses with reset connections, meaning that they weren't retried.

/cc @mtdowling @chrisradek

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -749,7 +749,10 @@ class S3Client extends AwsClient
             } elseif ($error instanceof AwsException
                 && $retries < $maxRetries
             ) {
-                if ($error->getResponse()) {
+                if (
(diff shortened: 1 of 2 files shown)


mtdowling:
"unparseable" meaning we can't even extract an HTTP error code?

jeskew:
The response is parseable (we can extract headers and a status code), but the body is not. S3 responses are normally XML, but that is not true when executing a `GetObject` call.

In the specific error case that prompted this PR, the HTTP status code was 200 -- the connection seems to have been reset after headers and some data had already been sent.

mtdowling:
:ship:

PR #846, opened 2015-12-13
Add a CloudFront cookie signer

jeskew:
This PR lets customers create signed cookies to access private CloudFront distributions. Signed cookies are constructed in a similar manner to signed URLs, so a significant portion of this PR is factoring the signer and policy generator into reusable components.

/cc @mtdowling @chrisradek


src/CloudFront/Policy.php
```diff
@@ -0,0 +1,61 @@
...
+     *                                      indicating when this policy will
+     *                                      become valid.
+     * @param string|null       $ip         (Optional) The IP address range
+     *                                      (expressed in standard IPv4 CIDR
+     *                                      format) of the client(s) permitted
+     *                                      to make requests under this policy.
+     *
+     * @link http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-setting-signed-cookie-custom-policy.html#private-content-custom-policy-statement-cookies
+     */
+    public function __construct($resource, $until, $from = null, $ip = null)
```
mtdowling: I could go either way on if this class should exist. My main concern if how we can maintain this class as CloudFront policies evolve. For example, if they introduce more complexity in policies (e.g, nested structures and many more top-level parameters), how will this class and the constructor evolve?

mtdowling:
Looks good, just had a comment on how we'll maintain the policy class.

src/CloudFront/Policy.php
jeskew: @mtdowling That's a good point. I could simplify the class by having it only support canned policies (resource + expiration time), as that's a much stronger contract than CloudFront's current policy spec.

jeskew:
@mtdowling I moved the policy generation to a method on the signer and removed support for generating custom policies. Anything more complicated than the canned policy would need to be passed in to the signer as a JSON string.

mtdowling:
:ship:

PR #852, opened 2015-12-17
Clean DelegationSetId in Route53Client

xgin:
The fix 1b4688c4ba19ef8431e8ba96c8ebbe573fe83d43 seems incomplete:
It works for `GetReusableDelegationSet` request, where the parameter's name is `Id`
But does not work for `CreateHostedZone` request, where the parameter is `DelegationSetId`

That is the reason I've workarounded here: https://github.com/plesk/ext-route53/blob/master/plib/library/Client.php#L158

### DIFF
--- a/src/Route53/Route53Client.php
+++ b/src/Route53/Route53Client.php
@@ -117,7 +117,7 @@ class Route53Client extends AwsClient
     {
         return function (callable $handler) {
             return function (CommandInterface $c, RequestInterface $r = null) use ($handler) {
-                foreach (['Id', 'HostedZoneId'] as $clean) {
+                foreach (['Id', 'HostedZoneId', 'DelegationSetId'] as $clean) {
                     if ($c->hasParam($clean)) {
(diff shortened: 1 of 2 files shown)


jeskew:
While I agree that this PR seems to capture the original intent of the referenced commit, I don't think this change is completely backwards compatible. What if other users are expecting IDs prefaced with '/delegationset/'?

xgin:
Take a look at the example of the request: http://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateHostedZone.html#API_CreateHostedZone_Examples
The parameter is `<DelegationSetId>NZ8X2CISAMPLE</DelegationSetId>`
But SDK constructs `<DelegationSetId>/delegationset/NZ8X2CISAMPLE</DelegationSetId>` - such request is always failed with `NoSuchDelegationSet` error.

jeskew:
Good point. I misread this at first and thought it was transforming the response, not the request. This seems good to merge but I'd like to add a test.

PR #854, opened 2015-12-18
Track next tokens as a map instead of a pair of arrays

jeskew:
This would allow the use of optional response parameters as pagination tokens.

/cc @mtdowling @chrisradek


mtdowling:
:ship:

Issue #856, opened 2015-12-21
S3 Transfer construct doesn't set ACL option. (no acl on uploadDirectory on S3)


NielsCorneille:
You can't set acl on uploadDirectory method.
You can pass options to the uploadDirectory method, which are passed on to the Transfer constructor. 
But only 'Key' and 'Bucket' are loaded when constructing MultipartUploader or making the PutObject request.

Maybe the acl option should be added in `$this->s3Args` and loaded from there?

jeskew:
You could do this by passing in a `'before'` option when creating the transfer manager. This would be a callable that takes an instance of `\Aws\Command` and mutates it as necessary. To apply the same ACL to all uploads, you could do something like the following:
```php
$transfer = new \Aws\S3\Transfer($client, $source, $dest, [
    'before' => function (\Aws\Command $command) {
        if (in_array($command->getName(), ['PutObject', 'CreateMultipartUpload'])) {
            $command['ACL'] = 'authenticated-read';
        }
    },
]);
```

While I can see the utility of specifying a single ACL to be applied to all files, I think the callback approach provides more flexibility. In many cases, you would want to apply different ACLs to files being uploaded based on their local permissions or patterns in the filename, and that's something that a callable supports.

jeskew:
Looks like the `before` callback option is undocumented. I'll open a PR to fix that.

jeskew:
@NielsCorneille does using the `before` callback resolve this issue for you?

jeskew:
Closing, as the transfer manager's documentation was updated to show how to provide an ACL via a callback. Please feel free to reopen if you have any questions or concerns.

PR #857, opened 2015-12-21
Use passed acl in MultipartUploader

geryjuhasz:


jeskew:
Similar behavior can be achieved by passing a callable to the transfer manager's constructor (see [my comment on the related issue](https://github.com/aws/aws-sdk-php/issues/856#issuecomment-166381790)). I'm not sure a blanket ACL is as useful as a callable that can selectively apply ACLs.

jeskew:
Closing in favor of #858, as I believe using a callback is a better option. Please feel free to reopen if you have any questions or concerns.

PR #858, opened 2015-12-21
Update Transfer Manager documentation to highlight before callbacks

jeskew:
The docs for the transfer manager were a bit stale. This PR updates them and adds an example for `before` callbacks.

/cc @mtdowling @chrisradek


chrisradek:
:shipit:

PR #864, opened 2015-12-31
Better document the SNS validator package. Resolves #625

### LINKED ISSUE #625: Missing Aws\Sns\MessageValidator\MessageValidator classes
The SNS MessageValidator classes did not get included in version 3 of AWS PHP SDK. So after upgrading to v3 there is no way to validate incoming SNS messages. Do you offer any alternative? The migration guide does not mention this issue at all, we discovered the problem during migration.

jeskew:
/cc @mtdowling @chrisradek


chrisradek:
:shipit:

PR #869, opened 2016-01-07
More options for DynamoDB unmarshaler

bakura10:
Hi,

By default DynamoDB wrap boolean and sets inside custom structures, which seems to bring very few benefits, instead of complicating the usage.

If this is of interest I can write tests :).

### DIFF
--- a/src/DynamoDb/Marshaler.php
+++ b/src/DynamoDb/Marshaler.php
@@ -13,6 +13,8 @@ class Marshaler
         'ignore_invalid'  => false,
         'nullify_invalid' => false,
         'wrap_numbers'    => false,
+        'wrap_booleans'   => true,
+        'wrap_sets'       => true
     ];
 
     /** @var array Marshaler options. */
@@ -280,21 +282,30 @@ class Marshaler
                     }
                     return $data;
                 }
-                // NOBREAK: Unmarshal M the same way as L, for arrays.
+            // NOBREAK: Unmarshal M the same way as L, for arrays.
             case 'L':
                 foreach ($value as &$v) {
                     $v = $this->unmarshalValue($v, $mapAsObject);
                 }
                 return $value;
             case 'B':
-                return new BinaryValue($value);
+                if ($this->options['wrap_booleans']) {
+                    return new BinaryValue($value);
+                }
+
+                return (bool) $value;
             case 'SS':
             case 'NS':
             case 'BS':
                 foreach ($value as &$v) {
                     $v = $this->unmarshalValue([$type[0] => $v]);
                 }
-                return new SetValue($value);
+
+                if ($this->options['wrap_sets']) {
+                    return new SetValue($value);
+                }
+
+                return $value;
         }
 
         throw new \UnexpectedValueException("Unexpected type: {$type}.");
--- a/src/DynamoDb/Marshaler.php
+++ b/src/DynamoDb/Marshaler.php
@@ -13,7 +13,6 @@ class Marshaler
         'ignore_invalid'  => false,
         'nullify_invalid' => false,
         'wrap_numbers'    => false,
-        'wrap_booleans'   => true,
         'wrap_sets'       => true
     ];
 
@@ -282,18 +281,14 @@ class Marshaler
                     }
                     return $data;
                 }
-            // NOBREAK: Unmarshal M the same way as L, for arrays.
+                // NOBREAK: Unmarshal M the same way as L, for arrays.
             case 'L':
                 foreach ($value as &$v) {
                     $v = $this->unmarshalValue($v, $mapAsObject);
                 }
                 return $value;
             case 'B':
-                if ($this->options['wrap_booleans']) {
-                    return new BinaryValue($value);
-                }
-
-                return (bool) $value;
+                return new BinaryValue($value);
             case 'SS':
             case 'NS':
             case 'BS':
--- a/tests/DynamoDb/MarshalerTest.php
+++ b/tests/DynamoDb/MarshalerTest.php
@@ -376,4 +376,14 @@ JSON;
         $this->assertEquals(3, count($set));
         $this->assertEquals(3, iterator_count($set));
     }
+
+    public function testCanUnmarshalSetsToArray()
+    {
+        $m = new Marshaler(['wrap_sets' => false]);
+        $result = $m->unmarshalItem([
+            'foo' => ['SS' => ['a', 'b']],
+        ]);
+
+        $this->assertEquals(['foo' => ['a', 'b']], $result);
+    }
 }


jeskew:
The `'B'` key stands for binary in this case, not boolean. Boolean values are flagged as [`'BOOL'`](https://github.com/aws/aws-sdk-php/blob/5ee0f33fafe47740c03ff38ddb73ae4f52b4da5b/src/DynamoDb/Marshaler.php#L155) and unmarshalled as [raw boolean values](https://github.com/aws/aws-sdk-php/blob/5ee0f33fafe47740c03ff38ddb73ae4f52b4da5b/src/DynamoDb/Marshaler.php#L264). 

The reason binary and set values are wrapped is because the PHP representation of their raw values are indistinguishable from strings and arrays, respectively, so the wrapper is necessary if you want to successfully write the same data you read from DynamoDb. Rather than making set wrapping an optional behavior, `Aws\DynamoDb\SetValue` could implement `ArrayAccess`, which would let you use sets as if they weren't wrapped. Would that work for you?

bakura10:
Hhmm not entirely, the problem is that I realized the mistake because one of my method had an "array" typehint. I was quite surprised, before I found that it was actually the SDK wrapping around a SetValue. So I fixed it by adding a "->toArray()" but as it comes from nowhere it's a bit confusing.

jeremeamia:
First, make sure a Set (S), not List (L), is actually the type that you want.

The `SetValue` exists because an array like following `[0 => 3, 1 => 4, 2 => 2]` is too ambiguous for the Marshaler to infer it's type (i.e., we can't tell if you want `{"NS": [3, 4, 2]}` or `{"L": [3, 4, 2]}`). In order for the data to make a roundtrip from output to input and preserve the types, the `SetValue` wrapper is required. However, if you are just reading data, I can see the usefulness of adding the new options to the `Marshaler`. However, `'wrap_booleans'` should be `'wrap_binaries'` instead.

bakura10:
I've updated the PR (only with wrap_sets as I'm actually unsure about the binary).

jeremeamia:
@jeskew the reason I didn't make `SetValue` implement `ArrayAccess` is that I did not want it to be mutable in ways that would corrupt the data. For example, `$set['foo'] = 'bar';` would be bad.

jeskew:
@jeremeamia Good point; if I went that route I would probably make offsetSet throw an exception when invoked.

I don't think I would want to disable set wrapping in any case, as that behavior could be handled by an end-user's code when they know they absolutely need arrays. Even when just reading, the distinction between a set and a list can be important.

jeremeamia:
What about disabling wrappers in a one-off way. Like as a [second argument to `unmarshalItem`](https://github.com/bakura10/aws-sdk-php/blob/unmarshal/src/DynamoDb/Marshaler.php#L244). Not sure I'm 100% sold on that idea, but it _is_ an idea.

bakura10:
I like this idea :). Should I do it as part of this PR or a new one?

jeskew:
I'm not sold on the idea. The most important thing the marshaller needs to be able to do is to round-trip data. If you need data in a different format than what is returned, you can transform the array returned to you:
```php
$data = array_walk_recursive($data, function (&$leaf) {
    if ($leaf instanceof SetValue) {
        $leaf = $leaf->toArray();
    }
});
```

The reason I'm hesitant to add `$unwrap` as an optional parameter to `unmarshalItem` is that `unmarshalValue` already has an optional second argument used to control different behavior.

bakura10:
Yeah, but this definitely adds some boilerplate and won't work recursively. I'm not sure to understand why the addition of a new "unwrap" would prevent the round-trip. This is just for the unmarshaling for allowing easier and more consistent consumption. Currently the presence of those SetValues just force us to add those toArray whenever we detect a SetValue.

jeskew:
`array_walk_recursive` will work recursively.

Once a set has been transformed into an array, it is indistinguishable from a list. That means that calling `$marshaler->unmarshalItem($data, true);` would have no inverse, thus preventing you from saving back the data you read. Both the `wrap_numbers` option and the `SetValue` class exist to make sure that `unmarshalItem` is fully reversible; the former prevents loss of precision and integer overflows on 32-bit systems (e.g., PHP 5 on Windows), and the latter prevents the conflation of multiple data types into just `array`.

Perhaps you could encapsulate walking and transforming the array into a class in your application? Something like:
```php
class LossyMarshaler extends Marshaler
{
    public function unmarshalItem(array $data, $unwrap = false)
    {
        $unmarshaled = parent::unmarshalItem($data);
        if ($unwrap) {
            $unmarshaled = array_walk_recursive($unmarshaled, function (&$leaf) {
                if ($leaf instanceof SetValue) {
                    $leaf = $leaf->toArray();
                }
            });
        }

        return $unmarshaled;
    }
}
```

mtdowling:
I'm :-1: on adding this option as wrapping sets is a requirement for roundtripping DynamoDB types in PHP (given the ambiguity between an empty array and an empty set). We also wouldn't know how to serialize the set after loading it then trying to save it back as we would only see an array of values rather than a set.

jeremeamia:
I think @bakura10 is looking for something to make the use case of "I'm just reading data and not writing it back" easier. When you are not making a roundtrip, the wrapper objects are clumsy. I agree that adding a class-level option to turn off wrappers seems unsafe, but adding a parameter on `unmarshalItem` to turn off wrappers seems fine. In that case, you make the decision to not use wrappers just for the one-off context that you are calling it in. I'm not 100% sold on whether or not you should make that change, but it seems like it could be useful and would not cause problems, IMO.

bakura10:
Definitely, we have various layer of data anyway, so what we retrieve from DynamoDB is converted to entity most of the time, and then the entity is dehydrated before being sending back to DynamoDB. The wrapper just make our life a bit less complex, and also really less intuitive when we just need to retrieve the raw data just for consumption. Let me know if I cna update this PR in any way: )

jeskew:
That can be achieved by walking the unmarshaled array or subclassing the marshaler. In that case, the customer knows they are doing something potentially unsafe. @bakura10 I understand that this PR would mean that you wouldn't need to subclass the marshaler in your application, but it's not a usage pattern that the SDK should officially support. DynamoDB treats sets and lists very differently, and I would rather see the SDK move towards reflecting Dynamo semantics.

That said, is there anything that could be done to `SetValue` that would make your data access more intuitive? Maybe implementing support for `add`, `remove`, `contains`, or `union`? If your data model requires a set and not a list, the wrapper class should provide some value, at least enough value that you would typehint as `SetValue` instead of as `array`.

bakura10:
I understand the rationale behind this and it was mostly from a usage point of view. In some of our micro-services, we just work with the array of DynamoDb, so we may have that kind of code:

```php
public function myFunction(array $info)
{
    $this->otherFunction($info['my_list']);
}

public function otherFunction(array $list)
{
}
```

From a user point of view, we just expect a simple array in `myFunction` (which is what we receive), and we were tricked multiple times when trying to access a property thinking it will just be a plain array, but actually receiving the SetValue.

But anyway, if you think this is out of the scope, we will just subclass the Marshaler :).

jeskew:
@bakura10 That actually sounds like an argument for wrapping sets -- if your data model is using the same key to store either a set or a list, the distinction sounds important.

I opened #876 to track improvements to the `SetValue` wrapper. Please feel free to chime in there if you think of anything that could be helpful.

PR #872, opened 2016-01-08
Add a cache adapter for the psr-6 interface

jeskew:
The added class allows users to create an instance of `Aws\CacheInterface` from an instance of `Psr\Cache\CacheItemPoolInterface`.

/cc @mtdowling @chrisradek

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,8 @@
         "doctrine/cache": "~1.4",
         "aws/aws-php-sns-message-validator": "~1.0",
         "nette/neon": "^2.3",
-        "andrewsville/php-token-reflection": "^1.4"
+        "andrewsville/php-token-reflection": "^1.4",
+        "psr/cache": "^1.0"
     },
     "suggest": {
         "ext-openssl": "Allows working with CloudFront private distributions and verifying received SNS messages",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -20,7 +20,8 @@
         "guzzlehttp/guzzle": "~5.3|~6.0.1|~6.1",
         "guzzlehttp/psr7": "~1.0",
         "guzzlehttp/promises": "~1.0",
-        "mtdowling/jmespath.php": "~2.2"
+        "mtdowling/jmespath.php": "~2.2",
+        "psr/cache": "^1.0"
```
mtdowling: Does it need to be required?

jeskew: No. I could move this to the suggestions and dev dependencies blocks.

mtdowling:
Sure, why not? :shipit:

GrahamCampbell:
Don't we need to add this to "suggest"?

jeskew:
@GrahamCampbell I took it out of the composer suggestions, as anyone with a PSR-6 compliant cache will already have the interfaces present. `doctrine/cache` is suggested because they provide a cache implementation that can be used, whereas `psr/cache` does not.

PR #877, opened 2016-01-13
Add support for supplying source iterators to the transfer manager

jeskew:
Iterators must yield absolute paths and must be paired with a `base_dir` option. We document that using iterators with the transfer manager is possible, but they are currently not supported.

/cc @mtdowling @chrisradek

### DIFF
--- a/src/S3/Transfer.php
+++ b/src/S3/Transfer.php
@@ -219,29 +219,25 @@ class Transfer implements PromisorInterface
         $prefix = "s3://{$parts['Bucket']}/"
             . (isset($parts['Key']) ? $parts['Key'] . '/' : '');
 
-        $commands = [];
-        foreach ($this->getDownloadsIterator() as $object) {
+        $commands = Aws\map($this->getDownloadsIterator(), function (
+            $object
+        ) use ($prefix) {
             // Prepare the sink.
-            $localKey = $object;
-            if ($prefix && strpos($localKey, $prefix) === 0) {
-                $localKey = substr($localKey, strlen($prefix));
-            }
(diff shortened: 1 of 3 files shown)


src/S3/Transfer.php
```diff
@@ -211,74 +215,48 @@ private function normalizePath($path)
...
+            // Create the directory if needed.
+            $dir = dirname($sink);
+            if (!is_dir($dir) && !mkdir($dir, 0777, true)) {
+                return Promise\rejection_for(
+                    new \RuntimeException("Could not create dir: {$dir}")
+                );
+            }
+
+            // Create the command.
+            $commands[] = $this->client->getCommand(
```
mtdowling: I don't remember why I didn't do it this way before, but what about creating a generator that yields these build up commands rather than storing them all in an array? Given that we're downloading based on listObjects, the number of GetObject commands could be practically infinite.

mtdowling:
One question about generators, but otherwise, :shipit:

mtdowling:
:ship:

src/S3/Transfer.php
```diff
@@ -211,74 +215,44 @@ private function normalizePath($path)
...
+            'before'      => $this->before,
+            'rejected'    => function ($reason, $idx, Promise\PromiseInterface $p) {
+                $p->reject($reason);
+            }
+        ]))->promise();
     }
 
     private function createUploadPromise()
     {
-        // Creates an iterator that yields promises for either upload or
```
jeremeamia: This comment should probably remain.

jeremeamia:
Looks good!

PR #884, opened 2016-01-21
Automatically request URL encoding on ListObjects

jeskew:
Also, url decode affected fields if the EncodingType was set to URL by the SDK.

/cc @mtdowling @xibz @chrisradek


mtdowling:
:ship:

PR #886, opened 2016-01-22
Follow standard conventions in class naming

jeskew:
Because class and method names are not case sensitive in PHP, I do not believe this is a breaking change. I'll have to add a couple redirects for the API doc pages, but that should be the only thing affected.

/cc @chrisradek @mtdowling


chrisradek:
:shipit:

PR #888, opened 2016-01-25
Correctly requeue throttled WriteRequestBatch items

jeskew:
This PR ensures that when unprocessed items are requeued from the response to a BatchWriteItem operation in a WriteRequestBatch flush, they keep their PutRequest/DeleteRequest identifier.

/cc @mtdowling @xibz


mtdowling:
:shipit:

PR #890, opened 2016-01-26
Update credentials cache to check for updated Doctrine behavior

jeskew:
The test was checking for the existence of something in a cache, but the item must have an expiration in the future for Doctrine 1.6 to report that it contains that key. Our 2.8 build is failing because of this.


xibz:
Looks good

PR #892, opened 2016-01-26
Missing iam access denied exception class

adam-ut:
I ran into an issue when using the IamClient getUser function. It said that the AccessDeniedException class didn't exist. This commit adds the file, following the standards of the other AWS Exception classes.


jeskew:
If this PR is for version 2, you want to open it against the 2.8 branch, not the master branch.

adam-ut:
Thought I selected that. I'll fix it. Thanks!

PR #894, opened 2016-01-27
Add support for 2015-07-27 CloudFront model

jeskew:
This PR adds back in an older CloudFront model. Some work needed to be done on the manifest builder to support this, so that's included as well.

/cc @mtdowling @xibz


mtdowling:
¯\_(ツ)_/¯  as long as this was requested by customers, then :shipit:

PR #895, opened 2016-01-29
Throw an exception when attempting to serialize an AwsClient

### LINKED ISSUE #891: Exception: Serialization of 'Closure' is not allowed
If I use `3.x` release of this SDK and create a `S3Client()` instance ALL my test cases fail with the exception mentoined above. Unfortunately it doesn't tell me where the error is. I only get the following error message:

```
383) Comments\Test\TestCase\Controller\CommentsControllerTest::testAjaxAddFailure
Exception: Serialization of 'Closure' is not allowed
```

It took me a while to isolate this issue as best as possible and it seems to be an issue in here: https://github.com/aws/aws-sdk-php/blob/255013bbf2f06a764ee3717fe75bf57e6458c134/src/ClientResolver.php#L210-L254

My configuration looks like:

```
$s3Client = new S3Client([
    'version' => 'latest',
    'region'  => 'YOUR_REGION',
    'credentials' => [
        'secret' => 'YOUR_SECRET',
        'key' => 'YOUR_KEY'
    ]
]);
```

If I use `2.x` release of this SDK with the following configuration:

```
$s3Client = S3Client::factory([
    'version' => 'latest',
    'region'  => 'YOUR_REGION',
    'credentials' => [
        'secret' => 'YOUR_SECRET',
        'key' => 'YOUR_KEY'
    ]
]);
```

everything works fine and the error is gone.

Some details about my environment:

PHP 5.6.16-2
PHPUnit 4.x but also tested with 5.x
Framework: CakePHP 3.x

jeskew:
Clients are not serializable, but the error message produced when attempting to serialize isn't particularly helpful. It's usually `Serialization of 'Closure' is not allowed` or something similar. This PR would short-circuit that attempt and immediately throw a RuntimeException. SDK consumers who need to serialize a client should implement and use a serialization proxy. This resolves #891.

### DIFF
--- a/src/AwsClient.php
+++ b/src/AwsClient.php
@@ -273,6 +273,12 @@ class AwsClient implements AwsClientInterface
         return new Waiter($this, $name, $args, $config);
     }
 
+    public function __sleep()
+    {
+        throw new \RuntimeException('Instances of ' . static::class
+            . ' cannot be serialized');
+    }
+
     /**
      * Get the signature_provider function of the client.
      *
--- a/tests/AwsClientTest.php
+++ b/tests/AwsClientTest.php
@@ -322,4 +322,14 @@ class AwsClientTest extends \PHPUnit_Framework_TestCase
             'version'      => 'latest'
         ]);
     }
+
+    /**
+     * @expectedException \RuntimeException
+     * @expectedExceptionMessage Instances of Aws\AwsClient cannot be serialized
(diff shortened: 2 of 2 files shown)


xibz:
:ship:

mtdowling:
:shipit:

chrisradek:
:boat:

PR #896, opened 2016-01-29
Prune docs and examples from the phar and zip

jeskew:
This PR updates Burgomaster's recursiveCopy method to take an optional iterator of files to copy, then uses an iterator to exclude docs-2.json.php and examples-1.json.php files from the phar and zip.

This cuts the phar down to 4.2 MB and brings the zip to under 800 KB.


mtdowling:
Nice :ship:

PR #897, opened 2016-01-29
Updating validation to be more flexible

mtdowling:
This commit updates the SDK to allow customers to configure exactly what
they want to validate client side. I've added the ability to provide
validation constraints as feature flags to a Validator and to the
validate client option. This commit also updates the SDK to no longer
validate "max" by default.

@jeskew

### DIFF
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -767,7 +767,7 @@ to the HTTP handler.
 validate
 ~~~~~~~~
 
-:Type: ``bool``
+:Type: ``bool|array``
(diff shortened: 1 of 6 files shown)


mtdowling:
It would be nice to have a similar commit made to v2, or to at least disable "max" validation in v2.

mtdowling:
Fixed the PHP 5.5 syntax error and rebased.

jeskew:
Nice! :boat: 

I don't think I can make a similar fix for v2, since validation is exclusively handled by Guzzle prior to 3.0.

PR #900, opened 2016-02-01
Add a multi-region client

jeskew:
This PR adds a generic and an S3-specific multi-region client. They proxy to a pool of regionalized clients, allowing users to specify a region on individual operations by providing an `@region` parameter. For example:
```php
$s3 = new \Aws\S3\MultiRegionClient(['version' => 'latest', 'region' => 'us-east-1']);
$s3->getObject([
    'Bucket' => 'bucket',
    'Key' => 'key',
    '@region' => 'eu-west-1',
]);
```

This is still missing documentation, but I'd appreciate a review and any feedback. Some quick notes on the implementation:
* A region must be provided when creating a multi-region client; this serves as the default region when none is specified in an operation's parameter. 
* All clients in the pool share a handler. 
* The S3 multi-region client is able to recover from sending requests to the wrong region and resends them through the appropriate regionalized client.
* The multi-region functionality is implemented as a trait rather than an alternate implementation of `Aws\AwsClientInterface`.
  * This is done so that it can be used to create clients that are subclasses of service clients.
  * This means that the S3 multi-region client can be used with the stream wrapper and multipart upload objects, which take an `Aws\S3\S3Client` in their constructor.

/cc @mtdowling @xibz @chrisradek


mtdowling:
AWESOME

High level feedback before looking at the code:

> A region must be provided when creating a multi-region client; this serves as the default region when none is specified in an operation's parameter.

What would you think about making the region optional and then just having a runtime check to ensure a region was resolved for a particular operation? I think a valid use case is wanting to create a client and not know which regions is will connect to because they would only know at runtime.

> All clients in the pool share a handler.

+1

> The S3 multi-region client is able to recover from sending requests to the wrong region and resends them through the appropriate regionalized client.

Nice.

> The multi-region functionality is implemented as a trait rather than an alternate implementation of Aws\AwsClientInterface.

This seems like a less composable approach to me. You cannot wrap a client at runtime, it has to be a compile time "copy and paste" of the trait implementation.

> This is done so that it can be used to create clients that are subclasses of service clients.

What is the advantage of that over just being an instance of `ClientInterface`?

> This means that the S3 multi-region client can be used with the stream wrapper and multipart upload objects, which take an Aws\S3\S3Client in their constructor.

Maybe we should change that class to instead check the provided client at runtime to see if it uses the "s3" service?

jeskew:
@mtdowling Thanks for the feedback! 

I updated this PR to be a slightly more expansive change. The base multi-region client just implements `AwsClientInterface`, and method implementations that are shared between a standard `AwsClient` and the new multi-region version have been moved to an `AwsClientTrait` that is used by both. Similarly, there's now an `S3ClientInterface` implemented by the standard `S3Client` and its multi-region counterpart, with code being shared between the two with a trait. S3 helper classes like the stream wrapper and the multipart uploader have been updated to take an instance of `S3ClientInterface` instead of an instance of `S3Client`.

src/MultiRegionClient.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * @param string $region
+     *
+     * @return AwsClientInterface
+     */
+    protected function getClientFromPool($region)
+    {
+        if (empty($region) || !$this->isStringable($region)) {
+            throw new \InvalidArgumentException();
```
mtdowling: Needs error message

src/S3/ObjectCopier.php
```diff
@@ -0,0 +1,87 @@
+<?php
+namespace Aws\S3;
+
+use GuzzleHttp\Promise\PromisorInterface;
+
+class ObjectCopier implements PromisorInterface
```
mtdowling: Add docs

src/S3/ObjectUploader.php
```diff
@@ -0,0 +1,126 @@
+<?php
+namespace Aws\S3;
+
+use GuzzleHttp\Promise\PromisorInterface;
+use GuzzleHttp\Psr7;
+use Psr\Http\Message\StreamInterface;
+
+class ObjectUploader implements PromisorInterface
```
mtdowling: docs

src/S3/ObjectUploader.php
```diff
@@ -0,0 +1,126 @@
...
+    private $client;
+    private $bucket;
+    private $key;
+    private $body;
+    private $acl;
+    private $options;
+    private static $defaults = [
+        'before_upload' => null,
+        'concurrency'   => 3,
+        'mup_threshold' => 16777216,
```
mtdowling: constant?

mtdowling:
Let me know when this is reviewable again (with context on what changed), and I'll take a look.

jeskew:
@mtdowling I added support for multiregion clients to the Sdk class and added some documentation.

mtdowling:
:shipit:

PR #902, opened 2016-02-02
Update .travis.yml to use HHVM 3.11

jeskew:
This PR tries to get Travis to test against HHVM >= 3.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ php:
   - 7.0
   - hhvm
 
-sudo: false
+sudo: required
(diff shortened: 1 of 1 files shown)


jeremeamia:
Interesting results. :confused:

jeskew:
@jeremeamia Yeah, that's a lot of failures. It looks like they all trace back to a usage of `Promise\coroutine`. HHVM 3.11 may have fixed the syntax bug, but their generator semantics are clearly still different from PHP.

Issue #192, opened 2013-12-06
Most effective way to enable arbitrary seeking in an S3 Stream Wrapper implementation?


coredumperror:
Hello again AWS devs! I mentioned in issue #178 that I was working on a Drupal module called [S3 File System](https://drupal.org/project/s3fs), which uses the new SDK to implement a Drupal-compatible stream wrapper around S3. @jeremeamia asked me to keep you guys up-to-date on my progress, and since I'm essentially finished now, I wanted to let you know.

However, I'm still having one small problem that I was hoping you guys could help me with.

Drupal's image handling code calls the PHP builtin function [`getimagesize()`](http://php.net/manual/en/function.getimagesize.php) on every image that gets uploaded, and _some_ images exhibit problems. These problems originate from the fact that `getimagesize()` sometimes calls `fseek()` on the image resource.

I wrote my stream wrapper to work in a very similar manner to your own `Aws\S3\StreamWrapper` class (please feel free to check out `S3fsStreamWrapper.inc` in my module's code by cloning http://git.drupal.org/project/s3fs.git), and thus it can't do the seek because the EntityBody's `seek()` function returns FALSE. Unfortunately, enabling the "seekable" option (wrapping the `EntityBody` in a `CachingEntityBody`) isn't good enough, because `getimagesize()` is attempting to seek to a not-yet-read part of the file.

As an experimental solution, I extended the Guzzle\Http\CachingEntityBody class and overode the seek() function. Now, instead of throwing a runtime exception when the caller attempts to seek past the end of the cache, it reads in enough of the remote file to let the seek work.

So my question is: is that a good way to do this? Are there drawbacks that I haven't taken into account? I'm sure you guys are much more expert in the ways of filesystem-level programming than myself, so I imagine that you'd be able to come up with an even better solution.

PR #903, opened 2016-02-04
Allowed the S3 seekable body to be configured

andythorne:
This issue stems from https://github.com/aws/aws-sdk-php/issues/192.

 #192 worked by overriding the `stream_open` method and wrapping the body in an extended `CachingStream` class. However, this does not work any more since the body is a private property and unable to be retrieved by a child class.

The simplest solution seemed to me to be to allow the seekable stream class to be configured by the stream options.

### DIFF
--- a/tests/S3/CustomCachingStream.php
+++ b/tests/S3/CustomCachingStream.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Aws\Test\S3;
+
+use GuzzleHttp\Psr7\Stream;
+
+/**
+ * Test Dummy for StreamWrapperTest to test seekable_handler
+ *
+ * @see \Aws\Test\S3\StreamWrapperTest::testOpensCustomSeekableReadStream
+ * @see \Aws\Test\S3\StreamWrapperTest::testOpensSeekableReadStreamHandlerNotCallable
+ * @see \Aws\Test\S3\StreamWrapperTest::testOpensSeekableReadStreamHandlerReturnInvalidStream
+ */
+class CustomCachingStream extends Stream
+{
+}
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -6,8 +6,8 @@ use Aws\LruArrayCache;
 use Aws\Result;
 use Aws\S3\Exception\S3Exception;
 use GuzzleHttp\Psr7;
-use GuzzleHttp\Psr7\Stream;
 use GuzzleHttp\Psr7\CachingStream;
+use GuzzleHttp\Psr7\Stream;
 use Psr\Http\Message\StreamInterface;
 
 /**
@@ -22,7 +22,9 @@ use Psr\Http\Message\StreamInterface;
  * the "seekable" stream context option true. This will allow true streaming of
  * data from Amazon S3, but will maintain a buffer of previously read bytes in
  * a 'php://temp' stream to allow seeking to previously read bytes from the
- * stream.
+ * stream. You can set the seekable stream class with a callable with the option
+ * "seekable_handler". The callable must return an instance of
+ * \Psr\Http\Message\StreamInterface.
  *
  * You may pass any GetObject parameters as 's3' stream context options. These
  * options will affect how the data is downloaded from Amazon S3.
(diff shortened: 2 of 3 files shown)


jeskew:
Instances of `CachingStream` [support seeking beyond what has already been read](https://github.com/guzzle/psr7/blob/master/src/CachingStream.php#L44). Are you seeing the same error (caused by `getimagesize` calling `fseek`) as described in #192?

andythorne:
Yep, exactly the same. I replicated the fix in #192 for master and it worked.

jeskew:
The fix for #192 has been part of v3 since it launched, as described in #587. Can you describe the error you're seeing and the steps needed to correct it? That's not clear from the test included in this PR.

andythorne:
No problem, I'll have to get back to you on Monday though as I'm away all weekend!

andythorne:
[`getimagesize()`](https://php.net/getimagesize) never sets `$whense` to `SEEK_END`, so i'm not sure how that fixed the problem. (see https://github.com/php/php-src/blob/master/ext/standard/image.c#L480)

Anyway, I've spend a long time debugging and there's certainly some strange behaviour happening, but I can only replicate it with 1 image atm (https://gist.github.com/andythorne/d636142a777407073695)

This is the current trace flow with the current CachingStream:

* image size (from AWS `ContentLength: 24167` header)
* call `getimagesize()` (with `seekable => true`)
* read 8192 bytes from remote stream and added to local stream
* `eof()` called, returns false
* seek on 9392 bytes with `$whence = SEEK_SET`
  * read in 1200 bytes (9392 - 8192 bytes)
  * S3 returns 397 bytes (seems odd, like it's reading to a certain block size)
* `eof()` called, returns false
* read 8192 bytes
* ends and `getimagesize()` returns false

The test doesn't reflect the fix for the seekable length - this  PR will allow you to override the CachingStream class.

We are using https://gist.github.com/andythorne/d1d036abfd2dbdc5e60c#file-seekablecachingstream-php-L70 atm to do this. The key lines are 70-72. Where in the above flow the stream only reads 397 bytes, this code sees it's still behind and carries on reading. The next read correctly fetches all the request bytes. 

So, maybe the proper fix is in the Guzzle `CachingStream` class that needs to loop the read in seek to check we have forward-seeked enough?

jeskew:
Thanks for the thorough investigation! From the testing I've done, this issue appears to only affect certain JPEG images. Since it's caused by inconsistent `fseek` behavior, I would prefer to fix it in Guzzle or the SDK rather than forcing affected users to fix it themselves. 

That said, I'm not opposed to allowing some customizability on how caching stream wrappers are instantiated, but I think it would make more sense to pass in a callable that takes a stream as an argument and returns a seekable stream. Just passing in a class name forces all custom caching streams to support calling the constructor with a single positional argument.

andythorne:
Yeah I agree, the more I investigated the more sense it made to patch Guzzle.

I'll modify the PR to use a callable tomorrow.

andythorne:
Updated, you can now set `seekable_handler` to a callable to return a instance of `StreamInterface`

tests/S3/CustomCachingStream.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Aws\Test\S3;
+
+use GuzzleHttp\Psr7\Stream;
+
+/**
+ * Class CustomCachingStream
+ *
+ * @package S3
```
GrahamCampbell: The package annotation shouldn't be used. It's only there to replace namespaces in php 5.2, and you've missmatched it with the namespace here anyway.

tests/S3/CustomCachingStream.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Aws\Test\S3;
+
+use GuzzleHttp\Psr7\Stream;
+
+/**
+ * Class CustomCachingStream
```
GrahamCampbell: this phpdoc adds no value

andythorne:
Updated

jeskew:
@andythorne Is this still necessary with the upstream change to Guzzle?

andythorne:
It offers the ability to customise the seekable stream, so it offers some value - but it's no longer directly tied to the guzzle changes.

mtdowling:
I don't think this PR is necessary now and would complicate the API as it currently stands. I don't see the need to have the ability to swap out what stream is used to provide the exact same functionality of caching+seeking. This should, IMO, be an implementation detail of the stream wrapper that users do not have to worry about.

andythorne:
@mtdowling Yeah I think that's fair. The original bug was a guzzle one, so I'll close this one down. Thanks for your help @jeskew

PR #904, opened 2016-02-04
Change formatId() to protected in the DynamoDB SessionHandler

NSmithUK:
Allowing it to be overridden in classes that extend Aws\DynamoDb\SessionHandler.

We have a use case for this where we need to use a different table key value to PHP's session name. Whist this can be achieved with most methods by extending SessionHandler, then amending $id before passing it to parent::, this does not work with close() as it directly access session_id().

### DIFF
--- a/src/DynamoDb/SessionHandler.php
+++ b/src/DynamoDb/SessionHandler.php
@@ -220,7 +220,7 @@ class SessionHandler implements \SessionHandlerInterface
      *
      * @return string Prepared session ID.
      */
-    private function formatId($id)
+    protected function formatId($id)
     {
         return trim($this->sessionName . '_' . $id, '_');
     }


jeskew:
This would [break any code](https://3v4l.org/IKgjo) that extends the session handler and has a different, unrelated method named 'formatId.'

Would it make sense in your application to modify the session key directly? You can do so by calling `session_id` with the new ID or by registering a session handler with [a defined `create_sid` handler](http://php.net/manual/en/function.session-set-save-handler.php).

jeskew:
Another option would be to provide a custom [`SessionConnectionInterface`](https://github.com/aws/aws-sdk-php/blob/master/src/DynamoDb/SessionConnectionInterface.php) when creating the SessionHandler. You could extend one of the existing implementations and overwrite the [`formatKey` method](https://github.com/aws/aws-sdk-php/blob/f716fa21f61983e6ce619e49acf0fa77966bc52c/src/DynamoDb/StandardSessionConnection.php#L133), which is not private.

NSmithUK:
Overwriting formatKey in my own SessionConnection is a splendid idea. I've gone with that. Thank you.

PR #908, opened 2016-02-08
Update HHVM issue referred to in FAQ

jeskew:
HHVM fixed the syntax issue with yield, but the semantics are still different. Updating the docs to reflect the new issue number.

/cc @chrisradek @xibz


xibz:
:ship:

PR #914, opened 2016-02-19
Add transfer stats

### LINKED ISSUE #913: Getting retry information from a command
In v2 of the SDK, we had the ability to attach a Guzzle logger that would get run whenever a request had to retry. There currently doesn't exist this functionality in v3, nor any hack to get anything like that. This is actually a hard blocker for us to migrate to v3.

In a Gitter chat, @jeskew brought up the possibility of adding metadata to a response that showed how many retries a particular request took. He also said that an "on_retry" callable could be injected that would allow a user to action off of every retry. The former would be useful, for example, for DynamoDB users as it can assist in autoscaling and giving an idea on the user experience. I'm not sure what the latter would be helpful for, though.

jeskew:
This PR binds the number of retries attempted to exceptions and results returned by `Aws\RetryMiddleware`. This will resolve #913.

/cc @xibz @chrisradek @mtdowling

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -110,6 +110,13 @@ class ClientResolver
             'fn'      => [__CLASS__, '_apply_credentials'],
             'default' => [CredentialProvider::class, 'defaultProvider'],
         ],
+        'stats' => [
+            'type'  => 'value',
+            'valid' => ['bool', 'array'],
+            'default' => ['retries' => true, 'http' => false],
+            'doc'   => 'Set to true to gather transfer statistics on requests sent. Alternatively, you can provide an associative array with the following keys: retries: (bool) Set to false to disable reporting on retries attempted; http: (bool) Set to false to disable collecting statistics from lower level HTTP adapters (e.g., values returned in GuzzleHttp\TransferStats).',
+            'fn'    => [__CLASS__, '_apply_stats'],
+        ],
         'retries' => [
             'type'    => 'value',
             'valid'   => ['int'],
@@ -332,7 +339,10 @@ class ClientResolver
     {
         if ($value) {
             $decider = RetryMiddleware::createDefaultDecider($value);
-            $list->appendSign(Middleware::retry($decider), 'retry');
+            $list->appendSign(
+                Middleware::retry($decider, null, $args['stats']['retries']),
+                'retry'
+            );
         }
     }
 
@@ -412,6 +422,18 @@ class ClientResolver
         }
     }
 
+    public static function _apply_stats($value, array &$args)
+    {
+        $defaults = [
+            'retries' => $value === true ? true : false,
+            'http' => $value === true ? true : false,
+        ];
+
+        $args['stats'] = is_array($value)
+            ? array_replace($defaults, $value)
+            : $defaults;
+    }
+
     public static function _apply_profile($_, array &$args)
     {
         $args['credentials'] = CredentialProvider::ini($args['profile']);
@@ -443,7 +465,8 @@ class ClientResolver
             default_http_handler(),
             $args['parser'],
             $args['error_parser'],
-            $args['exception_class']
+            $args['exception_class'],
+            $args['stats']['http']
         );
     }
 
@@ -453,7 +476,8 @@ class ClientResolver
             $value,
             $args['parser'],
             $args['error_parser'],
-            $args['exception_class']
+            $args['exception_class'],
+            $args['stats']['http']
         );
     }
 
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -424,6 +424,9 @@ class ClientResolver
 
     public static function _apply_stats($value, array &$args, HandlerList $list)
     {
+        // Create an array of stat collectors that are disabled (set to false)
+        // by default. If the user has passed in true, enable all stat
+        // collectors.
         $defaults = array_fill_keys(
             ['http', 'retries', 'timer'],
             $value === true
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -113,7 +113,7 @@ class ClientResolver
         'stats' => [
             'type'  => 'value',
             'valid' => ['bool', 'array'],
-            'default' => ['retries' => true, 'http' => false, 'timer' => false],
+            'default' => false,
             'doc'   => 'Set to true to gather transfer statistics on requests sent. Alternatively, you can provide an associative array with the following keys: retries: (bool) Set to false to disable reporting on retries attempted; http: (bool) Set to true to enable collecting statistics from lower level HTTP adapters (e.g., values returned in GuzzleHttp\TransferStats). HTTP handlers must support an __on_transfer_stats option for this to have an effect; timer: (bool) Set to true to enable a command timer that reports the total wall clock time spent on an operation in seconds.',
             'fn'    => [__CLASS__, '_apply_stats'],
         ],
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -114,7 +114,7 @@ class ClientResolver
             'type'  => 'value',
             'valid' => ['bool', 'array'],
             'default' => false,
-            'doc'   => 'Set to true to gather transfer statistics on requests sent. Alternatively, you can provide an associative array with the following keys: retries: (bool) Set to false to disable reporting on retries attempted; http: (bool) Set to true to enable collecting statistics from lower level HTTP adapters (e.g., values returned in GuzzleHttp\TransferStats). HTTP handlers must support an __on_transfer_stats option for this to have an effect; timer: (bool) Set to true to enable a command timer that reports the total wall clock time spent on an operation in seconds.',
+            'doc'   => 'Set to true to gather transfer statistics on requests sent. Alternatively, you can provide an associative array with the following keys: retries: (bool) Set to false to disable reporting on retries attempted; http: (bool) Set to true to enable collecting statistics from lower level HTTP adapters (e.g., values returned in GuzzleHttp\TransferStats). HTTP handlers must support an http_stats_receiver option for this to have an effect; timer: (bool) Set to true to enable a command timer that reports the total wall clock time spent on an operation in seconds.',
             'fn'    => [__CLASS__, '_apply_stats'],
         ],
         'retries' => [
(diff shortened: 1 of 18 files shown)


Sazpaimon:
I can also see usefulness in having the total amount of time that was spent sleeping as well

jeskew:
@Sazpaimon The SDK randomizes the time between retries, so I'm not sure how useful that data point would be.

Sazpaimon:
I feel it'd useful for giving an idea on how long a particular command affected a user, specifically for user-facing apps that need to communicate with, say DynamoDB tables.

mtdowling:
I could see that as a separate, more integrated feature of the SDK rather than a one-off addition. I would prefer to table that particular aspect for now and evaluate if there is a better way to more cohesively provide statistics about a transfer to a user.

Sazpaimon:
That would be fine. I'm more thinking about ways to maintain feature parity with v2 of the SDK (which allowed you to put the sleep time for an individual retry in the client.backoff.logger.template config option), but something that would give more accurate stats would probably be a better idea.

Sazpaimon:
The current version seems to omit setting the stats config in the `_applyRetryConfig` method in `DynamoDbClient`

src/ClientResolver.php
```diff
@@ -412,6 +422,18 @@ public static function _apply_debug($value, array &$args, HandlerList $list)
         }
     }
 
+    public static function _apply_stats($value, array &$args)
+    {
+        $defaults = [
+            'retries' => $value === true ? true : false,
```
xibz: No need for the ternary
```php
'retries' => $value === true;
```

src/ClientResolver.php
```diff
@@ -412,6 +422,18 @@ public static function _apply_debug($value, array &$args, HandlerList $list)
...
+            'http' => $value === true ? true : false,
```
xibz: same here

xibz:
Overall, it looks good. Have a couple comments.

src/ClientResolver.php
```diff
@@ -110,6 +110,13 @@ class ClientResolver
             'fn'      => [__CLASS__, '_apply_credentials'],
             'default' => [CredentialProvider::class, 'defaultProvider'],
         ],
+        'stats' => [
+            'type'  => 'value',
+            'valid' => ['bool', 'array'],
+            'default' => ['retries' => true, 'http' => false],
+            'doc'   => 'Set to true to gather transfer statistics on requests sent. Alternatively, you can provide an associative array with the following keys: retries: (bool) Set to false to disable reporting on retries attempted; http: (bool) Set to false to disable collecting statistics from lower level HTTP adapters (e.g., values returned in GuzzleHttp\TransferStats).',
```
mtdowling: I think these should be opt-in by default instead of opt-out by default (meaning things default to off and must be turned on). This allows us to add more ``stats`` features in the future without automatically opting users into each new check. If we keep this opt-out by default, then each time we add a new ``stats`` feature, users will need to update their configuration to opt-out of the new feature.

Presumably, if someone is utilizing the array form of the stats config, they explicitly configured it for a reason.

src/ClientResolver.php
```diff
@@ -110,6 +110,13 @@ class ClientResolver
             'fn'      => [__CLASS__, '_apply_credentials'],
             'default' => [CredentialProvider::class, 'defaultProvider'],
         ],
+        'stats' => [
+            'type'  => 'value',
+            'valid' => ['bool', 'array'],
+            'default' => ['retries' => true, 'http' => false],
```
mtdowling: What about making the default should be ``false`` by default?

src/WrappedHttpHandler.php
```diff
@@ -66,18 +72,41 @@ public function __invoke(
...
+        return Promise\promise_for($fn($request, $options))
             ->then(
-                function (ResponseInterface $res) use ($command, $request) {
-                    return $this->parseResponse($command, $request, $res);
+                function (ResponseInterface $response) use (
+                    $command,
+                    $request,
+                    $statsPromise
+                ) {
+                    return $statsPromise->then(function (array $stats) use (
```
mtdowling: Given that this class is used to power almost every transaction, it might be better to only use the additional layer of promise if the user requires transfer stats.

I also don't know of a case where you would get the `then` callback invoked and the on_stats callback would not have been completed. I wonder if there's a different way to get at this data without a promise.

src/RetryMiddleware.php
```diff
@@ -100,26 +105,107 @@ public function __invoke(
...
+            $resultStats = isset($value['@metadata']['transferStats']['http'][0])
+                ? $value['@metadata']['transferStats']['http'][0]
+                : [];
+            $stats['http'] []= $resultStats;
+        }
+    }
+
+    private function bindStatsToReturn($return, array $stats)
+    {
+        if (isset($stats['http']) && empty(array_filter($stats['http']))) {
```
mtdowling: Can you comment on why this is necessary?

src/RetryMiddleware.php
```diff
@@ -100,26 +105,107 @@ public function __invoke(
...
+        if ($return instanceof ResultInterface) {
+            if (!isset($return['@metadata'])) {
+                $return['@metadata'] = [];
+            }
+
+            $return['@metadata']['transferStats'] = $stats;
+            return $return;
+        } elseif ($return instanceof AwsException) {
+            $klass = get_class($return);
+            return new $klass(
```
mtdowling: Isn't this making the assumption that all instances of AwsException have the same constructor? I wonder if there's a more efficient or different way of achieving this.

src/Handler/GuzzleV6/GuzzleHandler.php
```diff
@@ -40,20 +41,40 @@ public function __invoke(Psr7Request $request, array $options = [])
...
-        );
+    private function parseOptions(array $options)
+    {
+        if (isset($options['__on_transfer_stats'])) {
+            $fn = $options['__on_transfer_stats'];
+            $prev = isset($options['on_stats'])
+                ? $options['on_stats']
+                : function () {};
+
+            $options['on_stats'] = function (TransferStats $stats) use ($fn, $prev) {
```
mtdowling: static function?

src/Handler/GuzzleV6/GuzzleHandler.php
```diff
@@ -40,20 +41,40 @@ public function __invoke(Psr7Request $request, array $options = [])
...
+    }
 
-                return new Promise\RejectedPromise($error);
-            }
-        );
+    private function parseOptions(array $options)
+    {
+        if (isset($options['__on_transfer_stats'])) {
+            $fn = $options['__on_transfer_stats'];
+            $prev = isset($options['on_stats'])
```
mtdowling: The extra function decoration, while minor, is unnecessary overhead for such a core component of the SDK.

src/Handler/GuzzleV6/GuzzleHandler.php
```diff
@@ -40,20 +41,40 @@ public function __invoke(Psr7Request $request, array $options = [])
...
                 }
+            );
+    }
 
-                return new Promise\RejectedPromise($error);
-            }
-        );
+    private function parseOptions(array $options)
+    {
+        if (isset($options['__on_transfer_stats'])) {
```
mtdowling: Why not move the `__on_transfer_stats` decoration code to the layers of the SDK that actually set the value of `on_stats`? Otherwise, you're introducing a new configuration option that might not be necessary.

mtdowling:
Can you also add documentation to the configuration docs so that we can better understand how this will work for end-users?

src/RetryMiddleware.php
jeskew: This assumption is commonly made throughout the SDK. The other options would be to make the transfer statistics modifiable on an instantiated AwsException or to break through the visibility restriction with closure binding or reflection.

src/Handler/GuzzleV6/GuzzleHandler.php
jeskew: Currently, no layer of the SDK is setting `on_stats`, though users of the Guzzle 6 handler are free to do so. `__on_transfer_stats` is set up as an option that is understood by both Guzzle handlers, and it has a different name so that no user code that is currently collecting Guzzle 6 stats would be affected.

mtdowling: What I'm suggesting is that you don't need this abstraction. Instead, you could do any necessary function wrapping in the places that you've added that currently set `__on_transfer_stats`. Using `__on_transfer_stats` doesn't really solve the issue either: what if there are several internal components of the SDK that need to augment `__on_transfer_stats`? You'd then need the same type of function wrapping.

src/RetryMiddleware.php
mtdowling: What about allowing the context of an exception to be mutable?

src/Handler/GuzzleV6/GuzzleHandler.php
jeskew: The only blocker on that approach is that Guzzle 5 does not support `on_stats` or provide a `GuzzleHttp\TransferStats` object that could fulfill the callback signature required for Guzzle 6 `on_stats` callables. Just using `on_stats` would simplify this feature but would make it incompatible with Guzzle 5.

src/ClientResolver.php
```diff
@@ -412,6 +422,21 @@ public static function _apply_debug($value, array &$args, HandlerList $list)
         }
     }
 
+    public static function _apply_stats($value, array &$args, HandlerList $list)
+    {
+        $defaults = array_fill_keys(
```
mtdowling: These are still defaulting to all on. If I specify `['timer' => true]`, then you'd also add in retries and http as true unless I explicitly set them to `false`. I think the less surprising thing to do here would be to require users to explicitly opt-in to stats features when they provide an associative array.

jeskew: The value passed in is either an associative array or a boolean. The individual collectors only default to `true` if `$value === true`, not if it's an array. In all other cases collectors default to `false`.

mtdowling: Ohhh, I see what you're doing here. Since it's really clever code, please add a comment to describe what's happening here. I also don't see any explicit tests for this.

mtdowling:
I had a request for a code comment and adding of missing tests. After that, :shipit: !

PR #918, opened 2016-02-24
Use an array of commands instead of mapping an iterator

jeskew:
This PR reverts 4b69d2a068b2c1b07017220cedffb4b63358240a, which can cause issues on large buckets.


xibz:
LGTM, :shipit:

PR #930, opened 2016-03-11
Updated Api/Validator Class to adhere to PHP array-keys casting

thePanz:
The current code of `Validator::checkAssociativeArray($value)` is not correctly identifying associative arrays where keys are strings containing valid integer numbers.
See http://php.net/manual/en/language.types.array.php#example-97 for an example of array-keys casting.

As an example, take the following associative array:
```
$a = [
  '0' => 'a',
  'two' => 'b',
];
```

The current function is wrongly report that it is not an associative array, given that $a[0] is set.

### DIFF
--- a/src/Api/Validator.php
+++ b/src/Api/Validator.php
@@ -241,7 +241,11 @@ class Validator
 
     private function checkAssociativeArray($value)
     {
-        if (!is_array($value) || isset($value[0])) {
+        // String array-keys containing valid integers will be cast to the integer type. E.g. the key "8" will actually
+        // be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
+        // @see http://php.net/manual/en/language.types.array.php#example-97
+        // Here we can not simply check `isset($value[0])` to distinguish between associative or sequential arrays.
+        if (!is_array($value) || isset($value[NULL])) {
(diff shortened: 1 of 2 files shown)


jeskew:
Checking if `$array[0]` is set is a quick heuristic to determine if a given array is being used as a list or a dictionary. More comprehensive checks typically involve copying all or part of an array, which is something the SDK tries to avoid. [A popular StackOverflow question](http://stackoverflow.com/questions/173400/how-to-check-if-php-array-is-associative-or-sequential) has suggestions to check if `array_keys($array) === range(0, count($array) - 1)` or even simply if `array_values($array) === $array`, both of which are also just heuristics, as array modifiers like `array_filter` will generally not modify keys, meaning that the result of an expression like `array_filter(range(1,10), function ($v) { return $v % 2 === 0; })` will appear to all the above methods to be an associative array, even though it's just a filtered range.

Even though there's no perfect solution, we do need to know in certain cases if an array is acting as a list or a map, as those tend to take different forms when serialized and sent to a service. `isset($array[0])` is the most efficient check that works for the data the SDK is expecting. This PR only checks if `is_array($array)` (I couldn't find a test case where `isset($array[null])` returned `true`), which means that the modified `checkAssociativeArray` method would cease distinguishing between lists and dictionaries.

PR #935, opened 2016-03-17
Return a rejection from Middleware::timer rejection handler.

jeskew:
The timer middleware released recently included a bug where the rejection handler returned an exception rather than throwing it or returning a rejected promise.

/cc @cjyclaire @xibz @chrisradek


chrisradek:
:shipit:

PR #937, opened 2016-03-23
Make retry logic consistent for php sdk

cjyclaire:
1. Update error code handling
2. Make Consistent Retry Scheme

/cc @jeskew

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -186,7 +186,29 @@ final class Middleware
             return new RetryMiddleware($decider, $delay, $handler, $stats);
         };
     }
-
+    /**
+     * Middleware wrapper function that adds an invocation id header to
+     * requests, which is only applied after the build step.
+     *
+     * This is a uniquely generated UUID to identify initial and subsequent
+     * retries as part of a complete request lifecycle.
+     *
+     * @return callable
+     */
+    public static function invocationId()
+    {
+        return function (callable $handler) {
+            return function (
+                CommandInterface $command,
+                RequestInterface $request
+            ) use ($handler){
+                return $handler($command, $request->withHeader(
+                    'aws-sdk-invocation-id',
+                    md5(uniqid(gethostname(), true))
+                ));
+            };
+        };
+    }
     /**
      * Middleware wrapper function that adds a Content-Type header to requests.
      * This is only done when the Content-Type has not already been set, and the
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -19,11 +19,13 @@ class RetryMiddleware
     ];
 
     private static $retryCodes = [
+        // Throttling error
         'RequestLimitExceeded'                   => true,
         'Throttling'                             => true,
         'ThrottlingException'                    => true,
         'ProvisionedThroughputExceededException' => true,
         'RequestThrottled'                       => true,
+        'BandwidthLimitExceeded'                 => true,
     ];
 
     private $decider;
@@ -85,13 +87,15 @@ class RetryMiddleware
     /**
      * Delay function that calculates an exponential delay.
      *
+     * Exponential backoff with jitter, 100ms base, 20 sec ceiling
+     *
      * @param $retries
      *
      * @return int
      */
     public static function exponentialDelay($retries)
     {
-        return mt_rand(0, (int) pow(2, $retries - 1) * 100);
+        return mt_rand(0, min(20000, (int) pow(2, $retries - 1) * 100));
     }
 
     /**
@@ -109,6 +113,8 @@ class RetryMiddleware
         $handler = $this->nextHandler;
         $decider = $this->decider;
         $delay = $this->delay;
+
+        $request = $this->addRetryHeader($request, 0, 0);
 
         $g = function ($value) use (
             $handler,
@@ -141,10 +147,18 @@ class RetryMiddleware
                 $this->updateStats($retries, $delayBy, $requestStats);
             }
 
+            // Update retry header with retry count and delayBy
+            $request = $this->addRetryHeader($request, $retries, $delayBy);
+
             return $handler($command, $request)->then($g, $g);
         };
 
         return $handler($command, $request)->then($g, $g);
+    }
+
+    private function addRetryHeader($request, $retries, $delayBy)
+    {
+        return $request->withHeader('aws-sdk-retry', "{$retries}/{$delayBy}");
     }
 
     private function updateStats($retries, $delay, array &$stats)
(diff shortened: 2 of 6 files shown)


cjyclaire:
Rework instrumentation, tests added.

@jeskew

src/Middleware.php
```diff
@@ -186,7 +186,24 @@ public static function retry(
...
+    * @return callable
+    */
+    public static function invocationID()
+    {
+        return function (callable $handler) {
+            return function (
+                CommandInterface $command,
+                RequestInterface $request = null
+            ) use ($handler) {
+                $request = $request->withHeader('aws-sdk-invocation-id', uniqid(gethostname(), true));
```
jeskew: The host name should be normalized to a fixed length and character set. Maybe you could use the md5 or crc32 of the hostname instead?

src/RetryMiddleware.php
```diff
@@ -19,11 +19,20 @@ class RetryMiddleware
...
         'RequestThrottled'                       => true,
+        'BandwidthLimitExceeded'                 => true,
+        // Clock skew error
+        'RequestTimeTooSkewed'                   => true,
+        'RequestExpired'                         => true,
+        'InvalidSignatureException'              => true,
+        'SignatureDoesNotMatch'                  => true,
+        'AuthFailure'                            => true,
+        'RequestInTheFuture'                     => true,
     ];
```
jeremeamia: I'm curious why these are all considered retry-able.

tests/RetryMiddlewareTest.php
```diff
@@ -54,7 +62,20 @@ public function testDeciderRetriesWhenAwsErrorCodeMatches()
...
+        $err = new AwsException('e', $command, ['code' => 'ThrottlingException']);
+        $err = new AwsException('e', $command, ['code' => 'ProvisionedThroughputExceededException']);
+        $err = new AwsException('e', $command, ['code' => 'RequestThrottled']);
+        $err = new AwsException('e', $command, ['code' => 'BandwidthLimitExceeded']);
+        $err = new AwsException('e', $command, ['code' => 'RequestTimeTooSkewed']);
+        $err = new AwsException('e', $command, ['code' => 'RequestExpired']);
+        $err = new AwsException('e', $command, ['code' => 'InvalidSignatureException']);
+        $err = new AwsException('e', $command, ['code' => 'SignatureDoesNotMatch']);
+        $err = new AwsException('e', $command, ['code' => 'AuthFailure']);
+        $err = new AwsException('e', $command, ['code' => 'RequestInTheFuture']);
```
jeremeamia: Confused by this as well. Isn't this overwriting `$err` over and over?

src/Middleware.php
```diff
@@ -186,7 +186,24 @@ public static function retry(
...
+    * This is a uniquely generated UUID to identify initial and subsequent retries as part of a complete request lifecycle.
+    *
+    * @return callable
+    */
+    public static function invocationID()
+    {
+        return function (callable $handler) {
+            return function (
+                CommandInterface $command,
+                RequestInterface $request = null
```
jeskew: This middleware will throw an error if `$request` is null. You could make `$request` non-nullable and mention that it must be added after the build step in the documentation block, as we do in [`Aws\S3\ApplyChecksumMiddleware`](https://github.com/aws/aws-sdk-php/blob/master/src/S3/ApplyChecksumMiddleware.php#L9).

src/Middleware.php
```diff
@@ -186,7 +186,24 @@ public static function retry(
...
         };
     }
-
+    /**
+    * Middleware wrapper function that adds an invocation id header to requests.
+    * This is a uniquely generated UUID to identify initial and subsequent retries as part of a complete request lifecycle.
+    *
+    * @return callable
+    */
+    public static function invocationID()
```
jeskew: This is a nitpick, but our naming structure would probably render this as `invocationId` instead of `invocationID`.

tests/RetryMiddlewareTest.php
jeskew: You might want to use a dataProvider instead.

src/RetryMiddleware.php
mtdowling: Yeah, I'm not sure retrying these would ever result in a success.

cjyclaire:
Add and fix test case, rework error code.

@jeskew

src/Middleware.php
```diff
@@ -186,7 +186,24 @@ public static function retry(
...
+    * @return callable
+    */
+    public static function invocationId()
+    {
+        return function (callable $handler) {
+            return function (
+                CommandInterface $command,
+                RequestInterface $request
+            ) use ($handler){
+                $request = $request->withHeader('aws-sdk-invocation-id', md5(uniqid()));
```
jeskew: This will need to be more unique than just `uniqid`. I think using `gethostname` was a good choice but would prefer to see the hostname normalized. How about `uniqid(md5(gethostname(), true)`?

src/Middleware.php
```diff
@@ -186,7 +186,24 @@ public static function retry(
             return new RetryMiddleware($decider, $delay, $handler, $stats);
         };
     }
-
+    /**
+    * Middleware wrapper function that adds an invocation id header to requests.
+    * This is a uniquely generated UUID to identify initial and subsequent retries as part of a complete request lifecycle.
```
jeskew: This documentation block should point out that the middleware can only be used after the `build` step in a handler stack. Also, please limit all lines to 80 or fewer characters.

tests/MiddlewareTest.php
```diff
@@ -73,6 +73,22 @@ public function testAddsRetrySubscriber()
...
+            $req = $request;
+            return \GuzzleHttp\Promise\promise_for(
+                new Result(['@metadata' => ['statusCode' => 200]])
+            );
+        };
+        $list->setHandler($mock);
+        $list->prependSign(Middleware::invocationId());
+        $handler = $list->resolve();
+        $handler(new Command('foo'), new Request('GET', 'http://exmaple.com'));
+        $this->assertTrue($req->hasHeader('aws-sdk-invocation-id'));
```
jeskew: I think this assertion should be in the `$mock` function rather than inspecting a symbol that is updated by reference.

tests/RetryMiddlewareTest.php
```diff
@@ -18,6 +18,23 @@
...
+            return new RejectedPromise(
+                new AwsException('e', $command, ['connection_error' => true])
+            );
+        };
+        $retryMW = new RetryMiddleware(
+            RetryMiddleware::createDefaultDecider($retries = 3),
+            [RetryMiddleware::class, 'exponentialDelay'],
+            $nextHandler,
+            true
+        );
```
jeskew: `$retryMW` is not being invoked in this test, so no assertions are being made.

cjyclaire:
Rework Test & Make changes as commented.

@jeskew

cjyclaire:
fix 80 wrap

@jeskew

jeskew:
:sheep: :it:

PR #940, opened 2016-03-28
Fix non-sequential keys when SetValue with duplicates is de-duped

Sazpaimon:
When marshaling a SetValue, the set is de-duped, but the process of de-duping leaves the keys in a state where json_encode will treat the set as an associative array making DynamoDB queries fail. Wrapping around array_values fixes this.


jeskew:
Thanks!

PR #942, opened 2016-03-29
Fix wrong tree hash if content is a single zero

sebcode:
When you tried to calculate the tree hash of the string "0" (without quotes), you got the hash `TreeHash::EMPTY_HASH` instead of the expected `5feceb66ffc86f38d952786c6d696c79c2dbc239dd4e91b46729d73a27fb57e9`.


jeskew:
Thanks!

PR #943, opened 2016-03-30
Make the input for an API example optional

jeskew:
The ElasticBeanstalk examples cover operations that do not have an input shape defined. This PR makes the example generator use an empty array when no example input is defined.

/cc @cjyclaire


cjyclaire:
Looks good to me.

:shipit:

mtdowling:
👍  🏈

PR #947, opened 2016-03-31
Refresh credentials as a set when signing requests

jeskew:
This PR addresses the same issue as https://github.com/boto/botocore/pull/764. In v2 of the SDK, if refreshable credentials expire during the signing process, the key, secret, and token may be pulled from different credential sets, as a call to `->refresh()` can be triggered by credentials being expired during calls to `->getAccessKeyId()`, `->getSecretKey()`, `->getSecurityToken()`. This PR updates the signature listener to instead get credentials using an atomic `->getCredentials()` method, which provides only a single opportunity for a refresh.

/cc @mtdowling @cjyclaire


cjyclaire:
:shipit:

Issue #946, opened 2016-03-31
S3 presigned POST in eu-central-1


Halama:
Hello,
I am receiving following error when submiting form generated by http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-presigned-post.html

```<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRequest</Code><Message>The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.</Message><RequestId>B24FC900AC875905</RequestId><HostId>zqsNhuEcWZvRR0FkCWkTQlH4nTofq3gCaosWA55JYJae3iS4lAJF12u2pVRTW/G/uiJV0LpB+zQ=</HostId></Error>```

The same code works in us-east-1 region but it doesn' work in eu-central-1.

I think this https://github.com/aws/aws-sdk-php/blob/156f73feaee26deb172ab1a3806a2bb18952d62b/src/S3/PostObject.php#L146 should be updated to signature V4
http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html

thanks

cjyclaire:
#971

PR #949, opened 2016-04-01
Update PostObject to SigV4

cjyclaire:
Update S3 post object with signature v4.
aws doc related:
http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html

issue related:
https://github.com/aws/aws-sdk-php/issues/946

@jeskew


src/S3/PostObject.php
```diff
@@ -147,15 +149,42 @@ protected function getPolicyAndSignature(CredentialsInterface $creds)
...
-                'sha1',
-                $jsonPolicy64,
-                $creds->getSecretKey(),
-                true
-            ))
+            'signature' => $signature
         ];
     }
+
+    private function getSigningKey($shortDate, $region, $service, $secretKey)
```
jeskew: This looks very similar to what's in `Aws\Signature\SignatureV4`. Could you use the same code in both (maybe via a trait)?

src/S3/PostObject.php
```diff
@@ -147,15 +149,42 @@ protected function getPolicyAndSignature(CredentialsInterface $creds)
...
+    {
+        $dateKey = hash_hmac('sha256', $shortDate, "AWS4{$secretKey}", true);
+        $regionKey = hash_hmac('sha256', $region, $dateKey, true);
+        $serviceKey = hash_hmac('sha256', $service, $regionKey, true);
+
+        return hash_hmac('sha256', 'aws4_request', $serviceKey, true);
+    }
+
+    private function gmdate($format, $ts = null)
```
jeskew: This looks like test-specific code. I don't think it should be in the `src` folder.

cjyclaire:
Make changes as comments, use traits, make src clean.

@jeskew

tests/S3/sig_hack.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace Aws\S3;
+
+// Hack gmdate() to returned the canned result.
+function gmdate($format, $ts = null)
+{
+    return isset($_SERVER['aws_time'])
```
jeskew: Why not just require the file from the signature namespace and proxy to those functions, e.g., have `gmdate` in the `Aws\S3` namespace call `\Aws\Signature\gmdate`? That way, updates only need to occur in a single location.

src/S3/SignatureTraitS3.php
```diff
@@ -0,0 +1,12 @@
...
+namespace Aws\S3;
+
+use Aws\Signature\SignatureTrait;
+
+/**
+ * Provides signature calculation for SignatureV4.
+ */
+trait SignatureTraitS3
+{
+    use  SignatureTrait;
```
jeskew: I'm not sure what the point of this trait is. Couldn't classes just use `Aws\Signature\SignatureTrait` directly?

cjyclaire:
fixed.

/cc @jeskew

cjyclaire:
Add cucumber test, signature doesn't match error
@jeskew

src/Signature/SignatureV4.php
```diff
@@ -257,10 +258,7 @@ private function getSigningKey($shortDate, $region, $service, $secretKey)
                 $this->cache = [];
                 $this->cacheSize = 0;
             }
-            $dateKey = hash_hmac('sha256', $shortDate, "AWS4{$secretKey}", true);
-            $regionKey = hash_hmac('sha256', $region, $dateKey, true);
-            $serviceKey = hash_hmac('sha256', $service, $regionKey, true);
-            $this->cache[$k] = hash_hmac('sha256', 'aws4_request', $serviceKey, true);
+            $this->cache[$k] = $this->getSigningKeyV4($shortDate, $region, $service, $secretKey);
```
jeskew: `getSigningKey` (*with* the cache) should be part of the trait.

src/Signature/SignatureTrait.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace Aws\Signature;
+
+/**
+ * Provides signature calculation for SignatureV4.
+ */
+trait SignatureTrait
+{
+    public function getSignatureV4(
```
jeskew: `public` and `protected` methods should be avoided for anything that is called from within the class. Methods should be private by default (especially in a trait).

jeskew:
We discussed the backwards compatibility concerns of this change in person (i.e., that the same policy cannot be applied to a V4 form without modification). Let me know when this is updated and ready for another review.

cjyclaire:
I believe it's now ready for another review. @jeskew

features/bootstrap/Aws/Test/Integ/S3Context.php
```diff
@@ -0,0 +1,169 @@
...
+            'expiration' => gmdate('Y-m-d\TH:i:s\Z', strtotime('+1 hour')),
+            'conditions' => [
+                ["bucket" => self::getResourceName()],
+                ["starts-with", '$key', ""],
+                ["acl" => "public-read"],
+                ["x-amz-credential" => $credentails],
+                ["x-amz-algorithm" => "AWS4-HMAC-SHA256"],
+                ["x-amz-date" => $ldt],
+            ],
+        ];
```
jeskew: Since a V4 policy needs to include conditions about the `x-amz-date`, `x-amz-credential`, and `x-amz-algorithm` fields, I don't believe this is a backwards compatible change. Any policy that is currently valid for a SigV2 POST upload would fail with a message like 'Invalid according to Policy: Extra input fields: x-amz-date' if the PostObject class suddenly started using V4 signatures.

features/s3/postObject.feature
```diff
@@ -0,0 +1,7 @@
+@s3 @integ
+Feature: POST object uploads
+
+    Scenario: Upload and object via POST
+        Given I have an "s3" client
+        When I do a POST upload
```
jeskew: This step will need to be broken up into several steps that each do a single thing explicitly. Currently, the logic of how a POST upload is completed is contained in the context class instead of in this feature file.

PR #953, opened 2016-04-05
Add support for new endpoints.json format and make PartitionEndpointProvider the default

jeskew:
This PR adds support for the new endpoints file format.

/cc @mtdowling @cjyclaire @chrisradek @xibz


mtdowling:
We talked about this PR in person. Let me know when this has changed and is ready for another review.

jeremeamia:
This looked "fun" to do. ;-)

jeskew:
@mtdowling I believe I covered what we discussed, and this PR is ready for another review.

src/ClientResolver.php
```diff
@@ -412,6 +427,12 @@ public static function _apply_endpoint_provider(callable $value, array &$args)
             if (isset($result['signatureVersion'])) {
                 $args['config']['signature_version'] = $result['signatureVersion'];
             }
+            if (isset($result['signingRegion'])) {
+                $args['config']['signing_region'] = $result['signingRegion'];
```
mtdowling: Is the user provides an explicit signing region in the client constructor, it should override what comes back from the endpoint provider. Is that case handled?

src/CommandPool.php
```diff
@@ -58,7 +58,7 @@ public function __construct(
                 if ($before) {
                     $before($command, $key);
                 }
-                yield $client->executeAsync($command);
+                yield $key => $client->executeAsync($command);
```
mtdowling: Is this an unrelated change?

src/Endpoint/Partition.php
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Determine if this partition contains the provided region.
+     *
+     * @param string $region
+     *
+     * @return bool
+     */
+    public function matchesRegion($region)
+    {
+        if (isset($this->definition['regionRegex'])) {
```
mtdowling: Couldn't you first check if the list of regions for the partition matches the given region, and only then fall back to regex based results?

src/Endpoint/Partition.php
```diff
@@ -0,0 +1,163 @@
...
+        $data += isset($this->definition['defaults'])
+            ? $this->definition['defaults']
+            : [];
+
+        return $data;
+    }
+
+    private function getSignatureVersion(array $data)
+    {
+        static $supportedBySdk = [
```
mtdowling: This is good, but on a side note: I wonder if we should add "sdkPreferredSignatureVersion" or something similar to the config file to remove this from each implementing SDK.

src/Endpoint/PartitionProviderInterface.php
```diff
@@ -0,0 +1,14 @@
...
+interface PartitionProviderInterface
+{
+    /**
+     * Gets the partition containing a given region.
+     * 
+     * @param string $region
+     *
+     * @return Partition
+     */
+    public function getPartition($region);
```
mtdowling: In order to allow this interface to be more extensible, perhaps this should be a more targeted method name... Like getPartitionFromRegion($region) or something.

tests/CommandPoolTest.php
```diff
@@ -35,6 +35,19 @@ public function testEnsuresBeforeIsCallable()
         new CommandPool($client, [], ['before' => 'foo']);
     }
 
+    public function testYieldsKey()
```
mtdowling: Unrelated change?

src/CommandPool.php
jeskew: Yes, but the `mapRegions` function doesn't make much sense without it. If the iterator provided to the command pool yields meaningful keys, they are only currently passed to the `before` callback, not `fulfilled` or `rejected`.

src/ClientResolver.php
jeskew: I believe that case would be handled, but I agree that there should be an explicit test for overriding signing_region and signing_name.

src/MultiRegionClient.php
```diff
@@ -45,6 +51,25 @@ public static function getArguments()
...
+                'doc'     => 'AWS partition to connect to. Valid partitions include "aws," "aws-cn," and "aws-us-gov."',
+                'default' => function (array $args) {
+                    $region = isset($args['region']) ? $args['region'] : '';
+                    return PartitionEndpointProvider::defaultProvider()
+                        ->getPartition($region);
+                },
+                'fn'      => function ($value, array &$args) {
+                    if (is_string($value)) {
+                        $value = PartitionEndpointProvider::defaultProvider()
+                            ->getPartition($value);
```
mtdowling: Isn't this trying to get a partition by partition name, but the method only works with region names?

src/MultiRegionClient.php
```diff
@@ -113,6 +139,52 @@ public function getEndpoint()
     }
 
     /**
+     * @param $commandName
```
mtdowling: Without a description, it's not clear what this method does.

src/MultiRegionClient.php
```diff
@@ -113,6 +139,52 @@ public function getEndpoint()
...
+        $commandGen = function () use ($regions, $commandName, $args) {
+            foreach ($regions as $region) {
+                yield $region => $this->getClientFromPool($region)
+                    ->getCommand($commandName, $args);
+            }
+        };
+        $resultHandler = static function ($result, $region) use (&$results) {
+            $results[$region] = $result;
+        };
+        $pool = new CommandPool($this, $commandGen(), [
```
mtdowling: I wonder if exposing any of the CommandPool options is something we should support? What if someone doesn't want to do these operations concurrently or wants to control the concurrency level?

src/MultiRegionClient.php
```diff
@@ -45,6 +51,25 @@ public static function getArguments()
                     };
                 },
             ],
+            'partition' => [
```
mtdowling: Can you add tests for configuring multi-region clients?

src/Signature/SignatureProvider.php
```diff
@@ -109,6 +109,7 @@ public static function version()
     {
         return function ($version, $service, $region) {
             switch ($version) {
+                case 's3v4':
```
mtdowling: It might be a good idea to add a test to ensure s3v4 can be resolved. While it's dead simple right now, it would help in case of a future refactor.

tests/Endpoint/PartitionTest.php
```diff
@@ -0,0 +1,403 @@
...
+            // partition endpoint setting
+            [$partition, 'us-east-1', 'baz', 's3v4'],
+            // service setting
+            [$partition, 'buzz', 'fizz', 'anonymous'],
+            // service endpoint setting
+            [$partition, 'pop', 'fizz', 's3v4'],
+            // no overrides
+            [$partition, 'puff', 'quux', 'v4'],
+            // unknown service
+            [$partition, 'us-east-1', 'iot', 'v4'],
```
mtdowling: How is this one resolving? "us-east-1" doesn't match the region explicitly nor do I see a region regex that it would match? Is this a test for an unknown service and and unknown region (e.g., fallback to assuming the aws partition as a last resort)? If so, maybe you should add a test that is just for unknown service, and one that is just for unknown region.

mtdowling:
Looking good! Have you tested out the new endpoint resolution against the old pattern based system to make sure you're resolving them in the same way? I did that in Botocore by finding every permutation endpoints from the new system and compared it against what the old system would have resolved.

src/CommandPool.php
mtdowling: Ah ok

jeskew:
@mtdowling I didn't add a test for each service in each region, but the trickier cases were covered by `Aws\Test\PatternEndpointProviderTest::testResolvesEndpoints`. I moved those cases to the PartitionEnpointProviderTest.

jeskew:
This PR has been updated with the latest version of endpoints.json. In addition, I've removed the `mapRegions` method to keep this change focussed.

/cc @mtdowling @xibz @cjyclaire

src/Endpoint/Partition.php
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        return false;
+    }
+
+    public function getRegionsForService($service)
+    {
+        if (isset($this->definition['services'][$service]['isRegionalized'])
+            && false === $this->definition['services'][$service]['isRegionalized']
+        ) {
```
cjyclaire: This first if statement seems to be the same with function resolveRegion($service, $region) despite that array is returned. While I can see that these 2 functions have different purpose, can you resolve the duplicates? Since $region parameter is only returned when the if check fails?

src/Endpoint/Partition.php
```diff
@@ -0,0 +1,171 @@
...
+        ];
+
+        $possibilities = array_intersect(
+            $supportedBySdk,
+            isset($data['signatureVersions'])
+                ? $data['signatureVersions']
+                : ['v4']
+        );
+
+        return array_pop($possibilities);
```
cjyclaire: While the logic look right, I'm little confused about the array operations going on there, is the $possibilities just has single element? If not, why just pop the last element of the result? If so, is there any privilege for using array_intersect for checking existence of an element in an array?

src/Endpoint/Partition.php
```diff
@@ -0,0 +1,171 @@
...
+        $this->definition = $definition;
+    }
+
+    public function toArray()
+    {
+        return $this->definition;
+    }
+
+    public function offsetGet($offset)
+    {
```
cjyclaire: It is just a personal thing that I feel it is more nature and common to name the getter and setter like 'getOffset', 'setOffset', 'unsetOffset'.

tests/Endpoint/PartitionEndpointProviderTest.php
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * @param string $region
+     * @param string $partition
+     */
+    public function testResolvesPartitionsByRegion($region, $partition)
+    {
+        // Use the default endpoints file
+        $p = PartitionEndpointProvider::defaultProvider();
+        $this->assertSame($partition, $p->getPartitionFromRegion($region)['partition']);
+    }
```
cjyclaire: In consideration for completeness and consistency, how about adding the test for unknown region and check whether default region is added as well?

src/Endpoint/Partition.php
jeskew: That's a good point. I'll factor that out into a separate method.

src/Endpoint/Partition.php
jeskew: I'm using `array_intersect` here to make sure that only signature versions that are supported by the SDK are pulled from the endpoints file. For example, Import/Export's signature versions are: `['v2', 'v4']`; the intersection of that against the SDK-supported signature versions will return an array with just one element: `v4`
```php
$supported = ['s3v4', 'v4', 'anonymous'];
$valid = array_intersect($supported, ['v2', 'v4']);
// $valid is [1 => 'v4']; 
```

If the service and SDK support more than one signature version, we should use them in the order of preference provided by the `$supportedBySdk` variable. I should be using `array_shift` instead of `array_pop` to enforce that.

src/Endpoint/Partition.php
jeskew: These method signatures are defined by [`ArrayAccess`](http://php.net/manual/en/class.arrayaccess.php), so we're stuck with the names they have, unfortunately.

src/Endpoint/Partition.php
cjyclaire: I see, make sense to me.

src/Endpoint/Partition.php
cjyclaire: Okay, then it should be alright.

tests/Endpoint/PartitionEndpointProviderTest.php
jeskew: Will do!

cjyclaire:
Looks good to me! :shipit:

mtdowling:
There's one last test case to write to validate the endpoints against a list of known endpoints. Then 🚢

jeskew:
@mtdowling I see your 🚢 and raise you a pile of documentation and tests.

PR #960, opened 2016-04-11
Update ES data plane guide

jeskew:
… to reflect integrated signers in Elastica and Elasticsearch-PHP plugins.

/cc @cjyclaire @xibz


cjyclaire:
Sweet. :shipit:

PR #961, opened 2016-04-12
Document limitations of copy method wrt stream wrapper

jeskew:
`copy` cannot accurately report failure when the target is an S3 path. This PR adds a note to the stream wrapper guide to that effect.

/cc @cjyclaire @xibz


xibz:
LGTM, :shipit:

PR #966, opened 2016-04-13
Presign without signing

### LINKED ISSUE #954: S3 V3 Presigned GET requests - always adds X-Amz-Content-Sha256
Hello,
first of all thanks for closing my issue regarding the host header.

The SDK is adding X-Amz-Content-Sha256 even if i want to have a presigned GET request where the content is not known for the client yet.

Your docs are telling me this:
```
You don't include a payload hash in the Canonical Request, because when you create a presigned URL, you don't know the payload content because the URL is used to upload an arbitrary payload. Instead, you use a constant string UNSIGNED-PAYLOAD. 
```

And this:
```
 Unsigned payload option – You include the literal string UNSIGNED-PAYLOAD when constructing a canonical request, and set the same value as the he x-amz-content-sha256 header value when sending the request to S3.
```

Okay, so if i have no payload i should set the URL to: `x-amz-content-sha256=UNSIGNED-PAYLOAD` if im right, or just leave it out but never use it to callculate other hashes.

So i call `createPresignedRequest` from the class `S3Client` like this:
```php
        $command = $client->getCommand('GetObject', [
            'Bucket' => 'myfavoritepics123',
            'Key'    => '5XXXXXXa.jpg'
        ]);
        $request = $client->createPresignedRequest($command, '+20 minutes');
```

Using this, X-Amz-Content-Sha256 will be added to the url regardless if im using GET/PUT/PATCH/POST. But it shouldn't.

```
   /**
     * Always add a x-amz-content-sha-256 for data integrity.
     */
```

As far as i understood your coding, if im using S3 its using the `S3SignatureV3` class and the function is being called from `Middleware.php` which is causing the problem. Please investigate.

jeskew:
When creating presigned URLs, requests are first signed with `Middleware::signer` and then presigned. This PR updates the `createPresignedRequest` method to remove the signing middleware before serializing the command into a request. It also adds some presigned URL integ tests to make sure everything's hunky-dory.

Resolves #954.

/cc @mtdowling @xibz @cjyclaire

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -219,6 +219,8 @@ class S3Client extends AwsClient implements S3ClientInterface
 
     public function createPresignedRequest(CommandInterface $command, $expires)
     {
+        $command->getHandlerList()->remove('signer');
+
         /** @var \Aws\Signature\SignatureInterface $signer */
         $signer = call_user_func(
             $this->getSignatureProvider(),
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -16,6 +16,7 @@ use Aws\CommandInterface;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Promise;
 use GuzzleHttp\Psr7;
+use Psr\Http\Message\RequestInterface;
 
 /**
  * Client used to interact with **Amazon Simple Storage Service (Amazon S3)**.
@@ -219,6 +220,7 @@ class S3Client extends AwsClient implements S3ClientInterface
 
     public function createPresignedRequest(CommandInterface $command, $expires)
     {
+        $command = clone $command;
         $command->getHandlerList()->remove('signer');
 
         /** @var \Aws\Signature\SignatureInterface $signer */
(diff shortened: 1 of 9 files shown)


src/S3/S3Client.php
```diff
@@ -219,6 +219,8 @@ public static function isBucketDnsCompatible($bucket)
 
     public function createPresignedRequest(CommandInterface $command, $expires)
     {
+        $command->getHandlerList()->remove('signer');
```
mtdowling: Does it matter that we've mutated the command in this method? Maybe we can clone the command or, a more involved change, we could return the removed handler when calling remove, and add it back at the end of the method.

jeskew: I looked into this, and commands are used up when executed or presigned. I can push up a change that clones the command, but I wasn't sure if it was in scope for this PR.

src/AwsClientTrait.php
```diff
@@ -61,6 +61,7 @@ public function execute(CommandInterface $command)
 
     public function executeAsync(CommandInterface $command)
     {
+        $command = clone $command;
```
mtdowling: Is this necessary?

jeskew: Not strictly, but executing a command consumes it in the same way presigning does. Without cloning here, you can only execute a command once.

mtdowling:
:shipit:

PR #967, opened 2016-04-13
Clear stat cache on stream_flush

jeskew:
This PR updates the StreamWrapper wipe the cached stats on a path when writing new data to S3 in `stream_flush`. Previously, the stat would be cached in PHP stat cache and in the internal cache.

/cc @xibz @cjyclaire @mtdowling


mtdowling:
:ship:

PR #970, opened 2016-04-20
Document stream wrapper permissioning limitations.

jeskew:
This PR adds a note to the stream wrapper docs explaining what permissions a client is expected to have vis-a-vis the objects accessed with the stream wrapper.

/cc @mtdowling

### DIFF
--- a/docs/service/s3-stream-wrapper.rst
+++ b/docs/service/s3-stream-wrapper.rst
@@ -23,6 +23,15 @@ This allows you to access buckets and objects stored in Amazon S3 using the
 bucket name followed by a forward slash and an optional object key or prefix:
 ``s3://<bucket>[/<key-or-prefix>]``.
 
+.. note::
+
+    The stream wrapper is designed for working with objects and buckets on which
+    you have at least read permission. This means that your user should have
+    permission to execute ``ListBucket`` on any buckets and ``GetObject`` on any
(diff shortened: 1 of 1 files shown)


docs/service/s3-stream-wrapper.rst
```diff
@@ -23,6 +23,12 @@ This allows you to access buckets and objects stored in Amazon S3 using the
 bucket name followed by a forward slash and an optional object key or prefix:
 ``s3://<bucket>[/<key-or-prefix>]``.
 
+.. note::
+
+    The stream wrapper is designed for working with objects and buckets on which
+    you have at least read permission. It is recommended that you use S3 client
```
mtdowling: Do we also need list permissions, or is that lumped into read permissions?

jeskew: I was thinking of [the S3 ACL 'read' grant](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html), which corresponds to `ListBucket`, `ListBucketVersions`, and `ListBucketMultipartUploads` for buckets and `GetObject`, `GetObjectVersion`, and `GetObjectTorrent` for objects. I'll call those out here, as that's a very specific definition of 'read.'

mtdowling:
:shipit:

PR #971, opened 2016-04-21
Update S3 POST object to Signature V4

cjyclaire:
Update S3 post object to SigV4.

cc:\ @jeskew

### DIFF
--- a/features/s3/postObject.feature
+++ b/features/s3/postObject.feature
@@ -1,0 +1,29 @@
+@s3 @integ
+Feature: POST object uploads
+
+    Scenario: Upload an object via POST object with Signature V2
+        Given I have an s3 client and I have a file
+        And I have an array of form inputs as following:
+        |key              |value              |
+        |acl              |public-read        |
+        And I provide a json string policy as following:
+        """
+        {"expiration":"2026-04-25T22:53:16Z",
+        "conditions":[{"acl":"public-read"}]}
+        """
+        When I create a POST object SigV2 with inputs and policy
+        And I make a HTTP POST request
+        Then the file called "file.ext" is uploaded
+
+    Scenario: Upload an object via POST object with Signature V4
+        Given I have an s3 client and I have a file
+        And I have an array of form inputs as following:
+        |key              |value              |
+        |acl              |public-read        |
+        And I provide an array of policy conditions as following:
+        |key              |value              |
+        |acl              |public-read        |
+        And I want the policy expires after "+ 2 hours"
+        When I create a POST object SigV4 with inputs and policy
+        And I make a HTTP POST request
+        Then the file called "file.ext" is uploaded
--- a/src/S3/PostObjectV4.php
+++ b/src/S3/PostObjectV4.php
@@ -3,12 +3,21 @@ namespace Aws\S3;
 
 use Aws\Credentials\CredentialsInterface;
 use GuzzleHttp\Psr7\Uri;
+use Aws\Signature\SignatureTrait;
+use Aws\Common\Enum\DateFormat;
+use Aws\Signature\SignatureV4 as SignatureV4;
+use Aws\Api\TimestampShape as TimestampShape;
 
 /**
- * @deprecated
+ * Encapsulates the logic for getting the data for an S3 object POST upload form
+ *
+ * @link http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html
+ * @link http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html
  */
-class PostObject
+class PostObjectV4
 {
+    use SignatureTrait;
+
     private $client;
     private $bucket;
     private $formAttributes;
@@ -18,38 +27,49 @@ class PostObject
     /**
      * Constructs the PostObject.
      *
+     * The options array accepts the following keys:
+     * @link http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
+     *
      * @param S3ClientInterface $client     Client used with the POST object
      * @param string            $bucket     Bucket to use
      * @param array             $formInputs Associative array of form input
      *                                      fields.
-     * @param string|array      $jsonPolicy JSON encoded POST policy document.
-     *                                      The policy will be base64 encoded
-     *                                      and applied to the form on your
-     *                                      behalf.
+     * @param array             $options    Policy condition options
+     * @param mixed             $expiration Upload expiration time value. By
+     *                                      default: 1 hour vaild peroid.
      */
     public function __construct(
         S3ClientInterface $client,
         $bucket,
         array $formInputs,
-        $jsonPolicy
+        array $options = [],
+        $expiration = '+1 hours'
     ) {
         $this->client = $client;
         $this->bucket = $bucket;
 
-        if (is_array($jsonPolicy)) {
-            $jsonPolicy = json_encode($jsonPolicy);
-        }
-
-        $this->jsonPolicy = $jsonPolicy;
+        // setup form attributes
         $this->formAttributes = [
             'action'  => $this->generateUri(),
             'method'  => 'POST',
             'enctype' => 'multipart/form-data'
         ];
 
+        // setup basic policy
+        $policy = [
+            'expiration' => TimestampShape::format($expiration, 'iso8601'),
+            'conditions' => $options,
+        ];
+
+        // setup basic formInputs
         $this->formInputs = $formInputs + ['key' => '${filename}'];
-        $credentials = $client->getCredentials()->wait();
-        $this->formInputs += $this->getPolicyAndSignature($credentials);
+
+        // finalize policy and signature
+        $credentials = $this->client->getCredentials()->wait();
+        $this->formInputs += $this->getPolicyAndSignature(
+            $credentials,
+            $policy
+        );
     }
 
     /**
@@ -114,16 +134,6 @@ class PostObject
         $this->formInputs[$field] = $value;
     }
 
-    /**
-     * Gets the raw JSON policy.
-     *
-     * @return string
-     */
-    public function getJsonPolicy()
-    {
-        return $this->jsonPolicy;
-    }
-
     private function generateUri()
     {
         $uri = new Uri($this->client->getEndpoint());
@@ -141,19 +151,37 @@ class PostObject
         return (string) $uri;
     }
 
-    protected function getPolicyAndSignature(CredentialsInterface $creds)
-    {
-        $jsonPolicy64 = base64_encode($this->jsonPolicy);
+    protected function getPolicyAndSignature(
+        CredentialsInterface $credentials,
+        array $policy
+    ){
+        $ldt = gmdate(SignatureV4::ISO8601_BASIC);
+        $sdt = substr($ldt, 0, 8);
+        $policy['conditions'][] = ['X-Amz-Date' => $ldt];
+
+        $region = $this->client->getRegion();
+        $scope = $this->createScope($sdt, $region, 's3');
+        $creds = "{$credentials->getAccessKeyId()}/$scope";
+        $policy['conditions'][] = ['X-Amz-Credential' => $creds];
+
+        $policy['conditions'][] = ['X-Amz-Algorithm' => "AWS4-HMAC-SHA256"];
+
+        $jsonPolicy64 = base64_encode(json_encode($policy));
+        $key = $this->getSigningKey(
+            $sdt,
+            $region,
+            's3',
+            $credentials->getSecretKey()
+        );
 
         return [
-            'AWSAccessKeyId' => $creds->getAccessKeyId(),
-            'policy'    => $jsonPolicy64,
-            'signature' => base64_encode(hash_hmac(
-                'sha1',
-                $jsonPolicy64,
-                $creds->getSecretKey(),
-                true
-            ))
+            'X-Amz-Credential' => $creds,
+            'X-Amz-Algorithm' => "AWS4-HMAC-SHA256",
+            'X-Amz-Date' => $ldt,
+            'Policy'           => $jsonPolicy64,
+            'X-Amz-Signature'  => bin2hex(
+                hash_hmac('sha256', $jsonPolicy64, $key, true)
+            ),
         ];
     }
 }
--- a/src/Signature/SignatureTrait.php
+++ b/src/Signature/SignatureTrait.php
@@ -1,0 +1,49 @@
+<?php
+namespace Aws\Signature;
+
+/**
+ * Provides signature calculation for SignatureV4.
+ */
+trait SignatureTrait
+{
+    /** @var array Cache of previously signed values */
+    private $cache = [];
+
+    /** @var int Size of the hash cache */
+    private $cacheSize = 0;
+
+    private function createScope($shortDate, $region, $service)
+    {
+        return "$shortDate/$region/$service/aws4_request";
+    }
+
+    private function getSigningKey($shortDate, $region, $service, $secretKey)
+    {
+        $k = $shortDate . '_' . $region . '_' . $service . '_' . $secretKey;
+
+        if (!isset($this->cache[$k])) {
+            // Clear the cache when it reaches 50 entries
+            if (++$this->cacheSize > 50) {
+                $this->cache = [];
+                $this->cacheSize = 0;
+            }
+
+            $dateKey = hash_hmac(
+                'sha256',
+                $shortDate,
+                "AWS4{$secretKey}",
+                true
+            );
+            $regionKey = hash_hmac('sha256', $region, $dateKey, true);
+            $serviceKey = hash_hmac('sha256', $service, $regionKey, true);
+            $this->cache[$k] = hash_hmac(
+                'sha256',
+                'aws4_request',
+                $serviceKey,
+                true
+            );
+        }
+
+        return $this->cache[$k];
+    }
+}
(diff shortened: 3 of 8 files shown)


features/s3/postObject.feature
```diff
@@ -0,0 +1,9 @@
+@s3 @integ
+Feature: POST object uploads
+
+    Scenario: Upload and object via POST
+        Given I have an "s3" client
+        And I have a file
+        When I create a POST object
```
jeskew: This needs to be a bit more explicit. Cucumber features are both tests and documentation for how to interact with the code. You might want to change this to either a series of steps or something like "I create a POST object with the following conditions:" followed by a a string that you can JSON-decode.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,214 @@
...
+        }
+        $policy = [
+            'expiration' => $expiration,
+            'conditions' => [],
+        ];
+        // policy condition updates
+        $starts_with = []; // handle start-with conditions differently
+        $start_with_key = '';
+        $pool = ['x-amz-credential', 'x-amz-algorithm', 'x-amz-date'];
+        foreach ($options as $option) {
```
jeskew: All these loops and branches make a large surface area for bugs. Could this be made simpler by just taking a list of conditions, using [Boto3's `generate_presigned_post`](https://botocore.readthedocs.org/en/latest/reference/services/s3.html#S3.Client.generate_presigned_post) as an example?

Instead of taking an array of options in the constructor, take `$conditions` as an array of arrays. Some of them will be a single key => value pair like `['acl' => 'public-read']` , while others will be a three-member tuple like `['starts-with', '$Content-Type', 'image/']`.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Constructs the PostObject.
+     *
+     * The options array accepts the following keys:
+     * @link http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
+     *
+     * @param S3ClientInterface $client     Client used with the POST object
+     * @param string            $bucket     Bucket to use
+     * @param array             $options    Policy condition options
+     * @param string            $expiration Upload expiration date. By
```
jeskew: This should be polymorphic. Usually the SDK will take an integer, `DateTime`, or string when a date value is needed.

src/Signature/SignatureTrait.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /** @var int Size of the hash cache */
+    private $cacheSize = 0;
+
+    private function createScope($shortDate, $region, $service)
+    {
+        return "$shortDate/$region/$service/aws4_request";
+    }
+
+    public function getSigningKey($shortDate, $region, $service, $secretKey)
```
jeskew: I don't think this should be a public method. It was private in the `SignatureV4` class.

tests/Integ/S3Context.php
```diff
@@ -6,19 +6,195 @@
...
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
 
 class S3Context implements Context, SnippetAcceptingContext
 {
     use IntegUtils;
 
+    const MB = 1048576;
+    const RESOURCE_POSTFIX = 'aws-test-integ-s3-context';
```
jeskew: This constant is only used once and is unlikely to be used in another method. You could just put this string directly in the `getResourceName` method.

tests/Integ/S3Context.php
```diff
@@ -6,19 +6,195 @@
...
+        return $bucketName;
+    }
+
+    /**
+     * @BeforeSuite
+     */
+    public static function createTempFile()
+    {
+        self::$tempFile = tempnam(sys_get_temp_dir(), self::getResourceName());
+        file_put_contents(self::$tempFile, str_repeat('x', 10 * self::MB + 1024));
```
jeskew: Is there a reason the file is so large? This could take a couple of seconds to upload when running the tests on spotty wifi. It's not a huge deal, but I think a ~128K file exercises the same functionality. Also, there's no real need for a constant here: just `128 * 1024` is fine.

features/s3/postObject.feature
```diff
@@ -0,0 +1,9 @@
+@s3 @integ
+Feature: POST object uploads
+
+    Scenario: Upload and object via POST
```
jeskew: This one scenario is testing both V2 and V4 signature POST uploads. They should be split into separate scenarios.

cjyclaire:
I think it's now ready for another review.

cc:/ @jeskew

src/S3/PostObjectV4.php
```diff
@@ -54,46 +54,28 @@ public function __construct(
...
             'enctype' => 'multipart/form-data'
         ];
+
         // setup basic policy documents
-        if (is_null($expiration)) {
-            $expiration = gmdate('Y-m-d\TH:i:s\Z', strtotime('+1 hours'));
-        }
+        $expiration = date (
+            'Y-m-d\TH:i:s\Z',
+            strtotime(gmdate(self::ISO8601_BASIC) . '+' . $expires .' hours')
```
jeskew: I think this is too restrictive. What if somebody needs to use a specific expiration that is not an even number of hours from the current time? If this argument instead just represents a specific time, customers will have a lot more flexibility in what expiration they can specify.

Take a look at [how the SDK serializes timestamps sent to services](https://github.com/aws/aws-sdk-php/blob/master/src/Api/TimestampShape.php#L25) for an example of taking a specific time as a `DateTime`, string, or integer.

tests/Integ/S3Context.php
```diff
@@ -280,4 +145,120 @@ public function iChangeTheBodyOfThePreSignedRequestToBe($body)
...
+    }
+
+    /**
+     * @When I create a POST object with policy condition :version:
+     */
+    public function iCreateAPostObjectWithPolicyCondition(
+        $version,
+        TableNode $table
+    ){
+        date_default_timezone_set('UTC');
```
jeskew: I don't think this is relevant to the test.

tests/Integ/S3Context.php
```diff
@@ -280,4 +145,120 @@ public function iChangeTheBodyOfThePreSignedRequestToBe($body)
...
+                ["starts-with", '$key', ""],
+            ],
+        ];
+
+        $args = [];
+        foreach ($table as $row) {
+            $args[$row['key']] = $row['value'];
+        }
+
+        if ($version == 'v2') {
```
jeskew: Rather than branching here, why not just create two different methods? V2 and V4 signing have different interfaces.

src/S3/PostObjectV4.php
```diff
@@ -54,46 +54,28 @@ public function __construct(
...
-                    } else {
-                        $this->formInputs[$key] = $value;
-                    }
+            $policy['conditions'][] = $option;
+            if (sizeof($option) > 2) { // starts_with
+                if ($option[1] == '$key' && strlen(trim($option[2])) >= 1){
+                    $this->formInputs['key'] = $option[2];
                 }
+            } else {
+                $this->formInputs += $option;
```
jeskew: I'm not sure adding these values to the form inputs is the right thing to do. What if it's already in the array supplied by the user? How does S3 respond to duplicate fields?

I would just trust that the user has added what they need to the form inputs and the conditions arguments.

features/s3/postObject.feature
```diff
@@ -1,9 +1,19 @@
...
+        |condition  |{"acl":"public-read"}    |
+        And I make a HTTP POST request with postObject "v2"
+        Then a file is uploaded with signature "v2" and its filename should be "filev2.ext"
+
+    Scenario: Upload an object via POST object with Signature V4
+        Given I have an s3 client and I have a file
+        When I create a POST object with policy condition "v4":
+        |key              |value              |
+        |acl              |public-read        |
+        |X-Amz-Algorithm  |AWS4-HMAC-SHA256   |
```
jeskew: Users should not have to specify the algorithm as part of their policy. That needs to be added by the signer.

src/S3/PostObjectV4.php
```diff
@@ -193,7 +175,7 @@ private function generateUri()
 
     protected function getPolicyAndSignature(CredentialsInterface $creds)
```
jeskew: This method needs to modify the policy to add the appropriate credential, algorithm, and date to the list of conditions.

tests/Integ/S3Context.php
```diff
@@ -280,4 +145,120 @@ public function iChangeTheBodyOfThePreSignedRequestToBe($body)
...
+            $this->preparePostData($postObject, 'v2');
+
+        } else { // v4
+            $ldt = gmdate(PostObjectV4::ISO8601_BASIC);
+            $policy['conditions'][] = ["X-Amz-Date" => $ldt];
+
+            $sdt = substr($ldt, 0, 8);
+            $scope = "$sdt/{$this->s3Client->getRegion()}/s3/aws4_request";
+            $credentials = "{$this->s3Client->getCredentials()->wait()->getAccessKeyId()}/$scope";
+            $policy['conditions'][] = ["X-Amz-Credential" => $credentials];
```
jeskew: The 'X-Amz-Credential' condition should be handled by the signer. As you mentioned when we discussed this in person, the condition field names are case sensitive, and that complexity should be encapsulated in the signer.

cjyclaire:
Fix date issue and rework test.

cc:/ @jeskew

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+     * @param array             $options    Policy condition options
+     * @param mixed             $expiration Upload expiration time value. By
+     *                                      default: 1 hour vaild peroid.
+     */
+    public function __construct(
+        S3ClientInterface $client,
+        $bucket,
+        array $formInputs,
+        array $options = [],
+        $expiration = NULL
```
jeskew: `null` would to be lowercase. The SDK follows the [PSR-2 coding style](http://www.php-fig.org/psr/psr-2/#2-5-keywords-and-true-false-null).

But I see below that you're interpreting a null as meaning `'+1 hours'`. Why not just make that string the default value?

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+
+        // setup form attributes
+        $this->formAttributes = [
+            'action'  => $this->generateUri(),
+            'method'  => 'POST',
+            'enctype' => 'multipart/form-data'
+        ];
+
+        // setup basic policy
+        $expiration = is_null($expiration)
```
jeskew: This could be simplified to `TimestampShape::format(is_null($expiration) ? '+1 hours' : $expiration, 'iso8601');` It could be simplified even further if the default value of `$expiration` were '+1 hours'; all validation would then be handled by `TimestampShape::format`.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+            'expiration' => $expiration,
+            'conditions' => [],
+        ];
+        foreach ($options as $option) {
+            $policy['conditions'][] = $option;
+        }
+
+        $this->formInputs = $formInputs + ['key' => '${filename}'];
+        // add v4 conditions for policy and formInputs
+        foreach ($this->getV4Conditions() as $key => $value) {
```
jeskew: There's a race condition here. You're resolving credentials in `getV4Conditions` and later on line 87. Credentials can be refreshable, which means that you might end up creating a policy with one set of credentials and then signing it with another. Resolve the credentials once.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+     *                                      default: 1 hour vaild peroid.
+     */
+    public function __construct(
+        S3ClientInterface $client,
+        $bucket,
+        array $formInputs,
+        array $options = [],
+        $expiration = NULL
+    ) {
+        $this->ldt = gmdate(self::ISO8601_BASIC);
```
jeskew: I don't think these values need to be bound to the class. Why not just resolve them when signing?

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        $this->formInputs[$field] = $value;
+    }
+
+    /**
+     * Gets the raw JSON policy.
+     *
+     * @return string
+     */
+    public function getJsonPolicy()
```
jeskew: I'm not sure this should be exposed as a public method on a V4 signer class. The policy isn't final until it's signed.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,209 @@
...
+            // Use virtual-style URLs
+            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+        }
+
+        return (string) $uri;
+    }
+
+    protected function getV4Conditions()
+    {
+        $scope = "{$this->sdt}/{$this->client->getRegion()}/s3/aws4_request";
```
jeskew: You can just call `createScope`; it's part of the `SignatureTrait`.

Maybe it would make sense to resolve the region differently if a user injected a multi-region client. You could call `determineBucketRegion` to find out the appropriate region.

features/s3/postObject.feature
```diff
@@ -0,0 +1,18 @@
+@s3 @integ
+Feature: POST object uploads
+
+    Scenario: Upload an object via POST object with Signature V2
+        Given I have an s3 client and I have a file
+        When I create a POST object in SigV2 with:
```
jeskew: The V2 signer just takes a string. Would it make more sense to use a [pyString](http://docs.behat.org/en/v3.0/guides/1.gherkin.html#pystrings) instead of a table?

features/s3/postObject.feature
```diff
@@ -0,0 +1,18 @@
...
+        |key              |value                |
+        |condition        |{"acl":"public-read"}|
+        And I make a HTTP POST request
+        Then a file is uploaded and its filename should be "file.ext"
+
+    Scenario: Upload an object via POST object with Signature V4
+        Given I have an s3 client and I have a file
+        When I create a POST object in SigV4 with:
```
jeskew: Both tests should be more fleshed out. It's still not clear from this cucumber test what inputs a PostObject is expecting. Maybe a separate step should add the form fields?

src/S3/PostObjectV4.php
cjyclaire: Since the jsonPolicy is set after the policy has been fully updated, and in signer, it just use an base64_encode version of the policy, so I do think the raw json version policy is final. 
Yet it also make sense to me if it's private, because now it's called in junit test for checking policy before signer. Thus in case of customer usage, this function doesn't matter that much to be exposed as public.

cjyclaire:
Rework cucumber test and timestamp etc.

cc:/ @jeskew

cjyclaire:
Now it's review ready.
cc:/ @jeskew

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,191 @@
...
+            // Use virtual-style URLs
+            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+        }
+
+        return (string) $uri;
+    }
+
+    protected function getPolicyAndSignature(
+        CredentialsInterface $credentials,
+        $policy
```
jeskew: This should be type hinted as an array.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,191 @@
...
+            'enctype' => 'multipart/form-data'
+        ];
+
+        // setup basic policy
+        $expiration = TimestampShape::format($expiration, 'iso8601');
+
+        $policy = [
+            'expiration' => $expiration,
+            'conditions' => [
+                ['X-Amz-Algorithm' => "AWS4-HMAC-SHA256"],
```
jeskew: This is being added in `getPolicyAndSignature`, isn't it? `$policy` here can just be defined as:
```php
$policy = [
    'expiration' => TimestampShape::format($expiration, 'iso8601'),
    'conditions' => $options,
];
```

That would capture your intention here more concisely by communicating the role of `$options` here.

src/S3/PostObjectV4.php
```diff
@@ -0,0 +1,191 @@
...
+        $policy
+    ){
+        $ldt = gmdate(SignatureV4::ISO8601_BASIC);
+        $sdt = substr($ldt, 0, 8);
+        $policy['conditions'][] = ['X-Amz-Date' => $ldt];
+
+        $region = $this->client->getRegion();
+        $scope = $this->createScope($sdt, $region, 's3');
+        $creds = "{$credentials->getAccessKeyId()}/$scope";
+        $policy['conditions'][] = ['X-Amz-Credential' => $creds];
```
jeskew: For multiple array additions, you can use the union operator (`+`) instead of pushing each one on individually:
```php
$policy['conditions'] += [
    'X-Amz-Date' => $ldt,
    'X-Amz-Credential' => $creds,
    'X-Amz-Algorithm' => 'AWS-HMAC-SHA256',
];
```

cjyclaire:
cc:\ @jeskew

jeskew:
Looking good! The last step is to make sure all the documentation uses this new class. Users new to this feature should be guided to the V4-signing PostObject, so the [pre-signed post documentation](https://github.com/aws/aws-sdk-php/blob/master/docs/service/s3-presigned-post.rst) should be updated, and the v2 signer should be marked with an `@deprecated` annotation.

cjyclaire:
Docs updated and deprecated tag marked

cc:\ @jeskew

jeskew:
:shipit:

PR #974, opened 2016-04-26
Do not attempt to parse the response body when no output shape is defined in JSON-RPC

jeskew:
We already do this for REST-JSON and REST-XML.

/cc @mtdowling @cjyclaire


cjyclaire:
looks good. :shipit:

PR #979, opened 2016-05-02
Ensure booleans are serialized as strings (not ints) in querystrings

jeskew:
`True` and `false` should serialize to "true" and "false," respectively. They were previously being rendered as "1" and "", which is not how [other SDKs](https://github.com/aws/aws-sdk-go/blob/master/private/protocol/rest/build.go#L244) are handling these values.

/cc @cjyclaire @xibz @mtdowling


cjyclaire:
Looks sweet. :shipit:

PR #987, opened 2016-05-09
Document how to disable content decoding

### LINKED ISSUE #986: V3 SDK error not seen with v2: "cURL error 61: Error while processing content unencoding: invalid code lengths set"
Hello,

I am running into a curious error with the v3 SDK when trying to download certain objects from S3. Interestingly, the problem does not occur with the v2 SDK.

So -- as mentioned, I'm using the v3 PHP SDK (using the .phar file) to download files from S3. Most files download without issue, but I receive a cURL 61 error when trying to download certain .bin files encoded as follows:
Content-Type: 'text/plain'
Content-Encoding: 'gzip'

The files have the wrong headers, yes -- that's a separate issue that I need to fix. However -- the cURL error I receive with the v3 SDK is:

    AWS HTTP error: cURL error 61: Error while processing content unencoding: invalid code lengths set (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)  (server): 200 OK

--

To be clear -- this error does not occur when using the AWS CLI, OR with the v2 PHP SDK. It DOES occur with these methods in the v3 SDK:

Aws\S3\S3Client::getObject()
Aws\S3\Transfer::transfer()
    
----

Interestingly, I can "work around" the issue by specifying the "ResponseContentEncoding" parameter in the getObject() function. The following call is successful in v3:

    $result = $s3->getObject([
        'Bucket' => 'my-bucket',
        'Key' => 'my-key.bin',
        "ResponseContentEncoding" => 'binary/octet stream',
    ]);
   

So, my question: is there a reason the v3 SDK runs into problems in downloading these files with incorrect headers, but the v2 SDK does not? I'd very much like to use v3 and the S3\Transfer::transfer() method for ease of use, but I do not wish to strip all of the encodings off my S3 files.

Please let me know if I can provide more information, including debug logs of the call.

Thanks!

jeskew:
~~This PR documents Guzzle's `decode_content` option as part of the HTTP handler configuration and defaults it to `false` in the S3 client. This should take care of #986 and any similar issues related to content encoding in S3.~~

This PR documents Guzzle's `decode_content` option as part of the HTTP handler configuration and documents the need to set it to `false` when dealing with compressed objects in S3. I do not believe the
default can be changed due to backwards compatibility concerns.

/cc @xibz @mtdowling @cjyclaire

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -199,3 +199,16 @@ was written for a different version of PHP.
 
 If you encounter a parsing error, please check your system and make sure it
 fulfills the SDK's :doc:`/getting-started/requirements`.
+
+
+Why is the S3 client decompressing gzipped files?
+-------------------------------------------------
+
+Some HTTP handlers -- including the default Guzzle 6 HTTP handler -- will
+inflate compressed response bodies by default. This behavior can be overridden
+by setting the :ref:`http_decode_content` HTTP option to ``false``. For
+backwards compatibility reasons, this default cannot be changed, but it is
+recommended that you disable content decoding at the S3 client level.
+
+See :ref:`http_decode_content` for an example of how to disable automatic
+content decoding.
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -426,6 +426,44 @@ information provided by different HTTP handlers will vary.
   specific PHP stream resource.
 
 
+.. _http_decode_content:
+
+decode_content
+^^^^^^^^^^^^^^
+
+:Type: ``bool``
+
+Instructs the underlying HTTP handler to inflate the body of compressed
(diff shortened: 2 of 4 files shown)


cjyclaire:
Looks good to me. :shipit:

cjyclaire:
Looks good. :shipit:

xibz:
Im assuming there's already a test for this`http_decode_content` option. If not, I'd add that, otherwise it looks good.

cjyclaire:
There is a test for `decode_content` set to `false`, then test for overwrite that to `true` could be added.

jeskew:
Good point. I added a test to ensure the value is passed to http handlers when specified on a client or a command.

cjyclaire:
Sweet! :shipit:

Issue #997, opened 2016-05-18
3.18.9 breaks S3 signature generation


okdewit:
Creating presigned URIs using the following method does not work in 3.18.9, the generated signatures are invalid.
https://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-presigned-url.html

Rolling back to our previously installed version (3.18.6) works.

We are using aws-sdk-php through aws-sdk-php-symfony.

Halama:
Same for me. I've just updated to 3.18.9 from 3.18.6 and presigned urls are broken.

nodefortytwo:
It appears 3.18.8 actually broke the generation, 3.18.7 confirmed working for me.

cjyclaire:
Thanks for notice, working on a fix release right now.

cjyclaire:
Thanks to all! The fixed release is out.

Issue #996, opened 2016-05-18
Typo added on S3SignatureV4 ?


danguer:
Hi,

I'm getting an error with presigned urls on S3 with last version, I noticed the last change:
https://github.com/aws/aws-sdk-php/commit/0377ccb6988ee12d6a4c86eb1312f3059a63a861
Changed the value to a constant, but the previous value is: 
'UNSIGNED-PAYLOAD';
And the new value is:
'UNSIGNED PAYLOAD';

cjyclaire:
Thanks for notice! New version has been released with fix. Free free to open if you have further questions.

PR #998, opened 2016-05-18
Fix S3 presign request

cjyclaire:
Fix the S3 presign  #996, #997 

cc: @mtdowling @xibz @jeskew

### DIFF
--- a/src/Signature/S3SignatureV4.php
+++ b/src/Signature/S3SignatureV4.php
@@ -41,6 +41,11 @@ class S3SignatureV4 extends SignatureV4
                 'X-Amz-Content-Sha256',
                 $this->getPresignedPayload($request)
             );
+        } else {
+            $request = $request->withHeader(
+                'X-Amz-Content-Sha256',
+                $request->getHeaderLine('x-amz-content-sha256')
+            );
         }
 
         return parent::presign($request, $credentials, $expires);
--- a/src/Signature/S3SignatureV4.php
+++ b/src/Signature/S3SignatureV4.php
@@ -41,11 +41,6 @@ class S3SignatureV4 extends SignatureV4
                 'X-Amz-Content-Sha256',
                 $this->getPresignedPayload($request)
             );
-        } else {
-            $request = $request->withHeader(
-                'X-Amz-Content-Sha256',
-                $request->getHeaderLine('x-amz-content-sha256')
-            );
         }
 
         return parent::presign($request, $credentials, $expires);
--- a/src/Signature/S3SignatureV4.php
+++ b/src/Signature/S3SignatureV4.php
@@ -9,7 +9,7 @@ use Psr\Http\Message\RequestInterface;
  */
 class S3SignatureV4 extends SignatureV4
 {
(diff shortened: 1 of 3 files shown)


xibz:
@cjyclaire - This change looks good. When modifying the .feature files, the cucumber test still pass, correct?

cjyclaire:
@xibz  That's true.

xibz:
@cjyclaire :shipit:

mtdowling:
If the test was run when the change was made originally, then there's either something wrong with the test or the test is inadequate to detect the failure. How could this test have worked both before this change and after this change?

xibz:
@mtdowling - No idea. I am confused there as well. Which is why I asked the question. However, I know the previous value `UNSIGNED-PAYLOAD` was originally used. So, the constant must be of that value. However, I think we need more testing or at least fix the testing to ensure the proper things are tested. We need a mock signature test with `UNSIGNED-PAYLOAD`.

mtdowling:
Do these tests hit the actual service or are they mocked? We should be integration testing here to ensure it actually works.

xibz:
These are integration tests. Either S3 is accepting the space character as valid or something else is wrong with the test itself. 

@cjyclaire - Can you confirm that you are getting expected responses from S3 when signing `UNSIGNED PAYLOAD` and sending `UNSIGNED-PAYLOAD`?

jeskew:
Looking at the previous implementation of [`S3SignatureV4`](https://github.com/aws/aws-sdk-php/blob/a8b62d6990fd8ea97d1ae4ce370779db7fd247c5/src/Signature/S3SignatureV4.php#L53), the value of 'ContentSHA256' was never used as the presigned payload. The server must have ignore the value in a `x-amz-content-sha-256` query parameter when the canonical request was signed with a body checksum of 'UNSIGNED-PAYLOAD.'

cjyclaire:
@jeskew make sense! Thank you to all!

xibz:
@jeskew - A test should have thrown an error. A sufficient test should be added to prevent this from happening again. Thanks for chiming in :)!

jeskew:
@xibz Those tests are currently failing on master.

xibz:
@jeskew - Thanks for the help and clarification!

cjyclaire:
It seems that cucumber test have a typo in command passed in as well, fix has been updated.

src/Signature/S3SignatureV4.php
```diff
@@ -41,6 +41,11 @@ public function presign(
                 'X-Amz-Content-Sha256',
                 $this->getPresignedPayload($request)
             );
+        } else {
+            $request = $request->withHeader(
```
jeskew: This is a no-op. You're assigning the current value of the header to the be the new value of the header.

cjyclaire: I see, are you suggesting it is the "ContentSHA256" causing the trouble only?

jeskew: I believe so. The test must be discarding the supplied value and using what's returned by `getPresignedPayload` for both scenarios.

cjyclaire:
Fix updated

xibz:
These changes look good, :shipit:

PR #1000, opened 2016-05-19
Update build-manifest

cjyclaire:
This updates service build with directory name as key instead of endpointPrefix.

cc: @jeskew @xibz @mtdowling

### DIFF
--- a/build/build-manifest.php
+++ b/build/build-manifest.php
@@ -51,8 +51,10 @@ foreach (scandir(__DIR__ . '/../src') as $dir) {
 $manifest = [];
 foreach (glob(__DIR__ . '/../src/data/**/**/api-2.json') as $file) {
     $model = json_decode(file_get_contents($file), true);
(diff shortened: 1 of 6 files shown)


build/build-manifest.php
```diff
@@ -51,8 +51,10 @@
 $manifest = [];
 foreach (glob(__DIR__ . '/../src/data/**/**/api-2.json') as $file) {
     $model = json_decode(file_get_contents($file), true);
+    preg_match('@src/data/([^/]+)/[0-9]{4}-[0-9]{2}-[0-9]{2}/api-2.json$@', $file, $matches);
+    $identifier = $matches[1];
```
xibz: Should we handle when the regex doesn't match?

cjyclaire: That's true, I'm also thinking about testing it throughly, will follow up with a new commit.

cjyclaire:
Inject `serviceIdentifier` to metadata to make sure that Paginators, Waiters and docs loaded right.

xibz:
LGTM, :shipit:

Issue #1002, opened 2016-05-20
Mix of float and int returned in exponentialDelay function breaks mt_rand in PHP 7.0


davzie:
Offending line:

https://github.com/aws/aws-sdk-php/blob/master/src/RetryMiddleware.php#L100

Traceback from Laravel queue worker:

```code

ErrorException: mt_rand() expects parameter 2 to be integer, float given
1
File "/sites/foodim.com/releases/20160520143427/vendor/aws/aws-sdk-php/src/RetryMiddleware.php", line 100 in handleError
+ 1 non-project frame 
3
File "/sites/foodim.com/releases/20160520143427/vendor/aws/aws-sdk-php/src/RetryMiddleware.php", line 100 in exponentialDelay
4
File "/sites/foodim.com/releases/20160520143427/vendor/aws/aws-sdk-php/src/RetryMiddleware.php", line 146 in Aws\{closure}
5
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 203 in callHandler
6
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 156 in GuzzleHttp\Promise\{closure}
7
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/TaskQueue.php", line 61 in run
8
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/guzzle/src/Handler/CurlMultiHandler.php", line 96 in tick
9
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/guzzle/src/Handler/CurlMultiHandler.php", line 123 in execute
10
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 246 in invokeWaitFn
11
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 223 in waitIfPending
12
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 266 in invokeWaitList
13
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 225 in waitIfPending
14
File "/sites/foodim.com/releases/20160520143427/vendor/guzzlehttp/promises/src/Promise.php", line 269 in invokeWaitList
15
[...]

PR #1003, opened 2016-05-20
Fix for mixed return from min causing mt_rand issues

davzie:
For issue #1002


cjyclaire:
Looks good to me! Thanks!

PR #1005, opened 2016-05-27
PR-01 - Nette Framework AWS SDK Extension

JakubKontra:
Hi,
could you please add our new AWS SDK Extension for Nette framework to related projects list?

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ try {
 * [AWS SDK ZF2 Module][mod-zf2]
 * [AWS Service Provider for Silex][mod-silex]
 * [AWS SDK Bundle for Symfony][mod-symfony]
+* [AWS SDK Extension for Nette Framework][mod-nette]
 * [Guzzle Version 6][guzzle-docs] – PHP HTTP client and framework
 * Other [AWS SDKs & Tools][aws-tools] (e.g., js, cli, ruby, python, java, etc.)
 
@@ -156,3 +157,4 @@ try {
 [mod-zf2]: https://github.com/aws/aws-sdk-php-zf2
 [mod-silex]: https://github.com/aws/aws-sdk-php-silex
 [mod-symfony]: https://github.com/aws/aws-sdk-php-symfony
+[mod-nette]: https://github.com/ublaboo/aws-sdk-nette-extension


marekjalovec:
AWS SDK works fine with Nette FW without custom extension for DI container (I'm using it that way every day on one project), it's just one line to create a new AWS Client and this even forces you to checkout your key to the repository, which is not something you should do at all.

paveljanda:
@marekjalovec What you wrote is completely against modern coding standards.

1, It is not necessary to create more instances of S3Client. (or other asw sdk clients)
2, Nette DIC takes care of singletons perfectly
3, Configuration parameters should stay in config (*.neon) files.
4, Nette Extensions are perfect way how to distribute these configurations / classes registration along other projects / teams.

You should use it, really. :)

paveljanda:
@marekjalovec And what do you mean by "checkout your key to the repository, which is not something you should do at all"?

This is just an extension, config parameters stays in project. Or did you mean something else?

marekjalovec:
@paveljanda I am not saying I create more clients, just the one singleton. but this extension encourages users to checkout aws credentials into config.neon file in the project repository. you should never have such stuff in your repo. ever.

JakubKontra:
@marekjalovec I'm probably blind. I can't see problem there. You don't need put your credentials into config.neon. You can use for example config.local.neon / config.production.neon to specify it.

paveljanda:
Where do store the secret cay?

The thing is, you should leave all configuration parameters (credentials etc) at one place. Nette Framework gives you the ability to use perfect `neon` format. So put it in there.

Another thing is being earnest and storing these data in for example `config.local.neon` only on your server..

juniwalk:
@marekjalovec If you think that it forces you to commit your secret keys to git then you are using it wrong in the first place.

You are supposed to create separate `*.neon` file and fill params there and then include in in `config.neon` or in your bootstrap.php. You of course have to add it into `.gitignore` file. That is how you use it.

marekjalovec:
You should maybe emphasize that in the docs, so people won't screw up :-)

fprochazka:
Improving the documentation with the usage pattern is definitely a good idea.

The package may be simple, but isn't that an advantage really?
Imho even a single class deservers a package, if it makese sense. I see no problem with this.

I see no problem with the package.

cjyclaire:
Thanks for the pull request (and conversation :) ). The intent is to list the AWS-provided projects there; I've updated the README to make that clearer. PR: #1006

paveljanda:
It's pretty bad luck for Nette framework then. Also I don't see any advantage of aws taking care of these extensions (the symfony bundle requires aws/aws-sdk-php in version ^3.2.6, so it's not that aws provides support for latest versions of aws-sdk-php).

Anyway, thanks for the reply.

PR #1012, opened 2016-06-05
configurable backoff retries

hanoii:
Debugging an odd socket idle timeout error I wanted to increase the backoff number of retries. Although it's possible to create your own backoff I thought of this simple change to allow configuration of the backoff num of retries.

### DIFF
--- a/tests/Aws/Tests/Common/Client/ClientBuilderTest.php
+++ b/tests/Aws/Tests/Common/Client/ClientBuilderTest.php
@@ -397,6 +397,27 @@ class ClientBuilderTest extends \Guzzle\Tests\GuzzleTestCase
         $this->assertFalse($client->getConfig('client.backoff'));
     }
 
+    public function testDefaultBackoffMaxRetries()
+    {
+        $config = array(
+            'service' => 'dynamodb',
+            'region'  => 'us-east-1',
+            'service.description' => array(
+                'signatureVersion' => 'v2',
+                'regions' => array('us-east-1' => array('https' => true, 'hostname' => 'foo.com'))
+            ),
+        );
+
+        $client = ClientBuilder::factory('Aws\\DynamoDb')->setConfig($config)->build();
+        $this->assertNull($client->getConfig(Options::BACKOFF_RETRIES));
+        $plugin = $client->getConfig(Options::BACKOFF);
+        $this->assertInstanceOf('Guzzle\Plugin\Backoff\BackoffPlugin', $plugin);
+        $strategy = $this->readAttribute($plugin, 'strategy');
+        $this->assertInstanceOf('Guzzle\Plugin\Backoff\TruncatedBackoffStrategy', $strategy);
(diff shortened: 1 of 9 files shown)


cjyclaire:
Thanks for the PR! It looks good to me, yet I'll take a second look at the issue before ship it :)

cjyclaire:
@hanoii , since this PR introduces new logic in configuration, could you add some tests as well?

hanoii:
I actually attempted to look into that, but I saw the whole backoff functionality not that much tested, but will try to look at it.

hanoii:
@cjyclaire I added the test, although here I am gonna need feedback/help/guidance if you are not happy with it. I am basically testing that the config is properly set and then that it actually reached the truncated strategy of the default builder, however, each "client" has its own builder, so I am not testing each of those. Awaiting your comments and or directions into properly adding a test to it.

cjyclaire:
Great thanks for your efforts in contributing the test! 
Could you also add another test for default retry behavior without `'client.backoff.retries'` configured? 
It's true that each "client" has its own builder, yet S3 and DynamoDb are configured with `$retries` beside default behavior, thus it would be nice to have both of them tested.

tests/Aws/Tests/Common/Client/ClientBuilderTest.php
```diff
@@ -396,6 +397,28 @@ public function testAllowsBackoffDisabling()
...
+            'region'  => 'us-east-1',
+            'service.description' => array(
+                'signatureVersion' => 'v2',
+                'regions' => array('us-east-1' => array('https' => true, 'hostname' => 'foo.com'))
+            ),
+            'client.backoff.retries' => 6
+        );
+
+        $client = ClientBuilder::factory('Aws\\DynamoDb')->setConfig($config)->build();
+        $this->assertEquals(6, $client->getConfig(Options::BACKOFF_RETRIES));
```
cjyclaire: I'd suggest mock parameter in testing instead of constant, which feels more nature in mocking client behavior.
$client->getConfig('client.backoff');
$client->getConfig('client.backoff.retries'));

hanoii: Actually I use the  constant by looking at other tests, and in a way it made sense as you are also testing that the constant matches the actual string parameter, don't you think?

hanoii:
@cjyclaire what about now? I removed the constant as it really seems to be used everywhere on the tests.

tests/Aws/Tests/Common/Client/ClientBuilderTest.php
cjyclaire: Make sense.

hanoii:
Just added two more tests for also testing the setting of the retries on those clients that have non-default backoff plugins.

cjyclaire:
Awesome, it looks good to me. I'll merge it.

Issue #1015, opened 2016-06-14
Recovering from MultipartUploadException during S3ClientTrait::uploadDirectory()


joecwallace:
Hello! Thank you so much for this library. It's been very helpful and a pleasure to work with so far, but I've hit a bit of a snag...

I'm trying to upload a directory to a bucket. The directory contains some large files that result in a multipart upload. Sometimes, this throws a `MultipartUploadException`.

At that point, how do I recover from the exception, since I don't have an explicit source path? (My current `composer.lock` version of the SDK is `3.18.9`, if that's helpful at all.)

Code sample of what I am trying / would like to do:
```php
$directory = '/path/to/my/dir';
$client = new S3Client($config);

try {
    $client->uploadDirectory($directory, 'my-bucket');
} catch (MultipartUploadException $e) {
    $source = '???'; // Is there some way to get this from $client or $e?

    $uploader = new MultipartUploader($client, $source, ['state' => $e->getState()]);
    $uploader->upload();
}
```

cjyclaire:
@joecwallace Thanks for the information!

`$source` (except streams) passed to a `MultipartUploader` will be automatically rewound before uploading. The `UploadState` object can be used to resume an upload that failed to complete. It will only attempt to upload parts that are not already uploaded. 

When tracing the file when exception occurs, you could try `$e->getFile()`

Meanwhile, since SDK V3 is used, I'd suggest using [Transfer Manager](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-transfer.html) instead of `uploadDirectory()`(Though we still support it ;) ).  [Configuration](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-transfer.html#configuration) for `Transfer Manager` is more flexible in many ways.

joecwallace:
@cjyclaire - Thank you for the reply. :-)

I think my question is still open though: how do I recover from the `MultipartUploadException` in the `catch` block above?

To provide a more concrete example:

1. I have a directory `/foo/bar`
1. In `/foo/bar` are 3 files: `x.txt`, `y.txt`, & `z.txt`
1. I execute `$client->uploadDirectory('/foo/bar', 'my-bucket')`
1. A `MultipartUploadException` is thrown while uploading `y.txt`
1. I catch the exception and wish to call `(new MultipartUploader($client, '/foo/bar/y.txt', ['state' => $e->getState()]))->upload()`

[...]

PR #1017, opened 2016-06-24
Provide file path information when S3 Transfer() fails

cjyclaire:
Feature request from #1015.
By providing file information, customer would have better troubleshooting experience.

cc:\ @mtdowling @xibz


src/S3/Transfer.php
```diff
@@ -2,6 +2,8 @@
 namespace Aws\S3;
```
mtdowling: The changes made to this file are untested.

src/S3/Transfer.php
```diff
@@ -305,6 +324,8 @@ private function upload($filename)
         $args = $this->s3Args;
         $args['SourceFile'] = $filename;
         $args['Key'] = $this->createS3Key($filename);
+        $this->inProgress['FileSource'] = $args['Bucket'] . "/" . $args['Key'];
```
mtdowling: There are actually N number of transfers in progress at any given time. When a failure occurs, the transfer stops at the point of failure, leaving the transfer in a state in which there are N number of transfers that are incomplete. I believe that when you wait on the Transfer it will throw when the first error is encountered, leaving the other currently in-progress transfers incomplete. You might want to double check the behavior here. If that's the case, then you would need to provide context on all of the in-progress transfers *and* the failed transfer when throwing the exception.

cjyclaire: That's true, make sense.

src/S3/Transfer.php
cjyclaire: I thought about this, will add tests for tracking `InProgress`

tests/Exception/TransferExceptionTest.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+    
+    public function getTestCases()
+    {
+        return [
+            [ 'foo/bar/a.txt', 1, [] ],
+            [ 'foo/bar/a.txt', 0, [ 'foo/bar/b.txt'] ]
+        ];
+    }
+}
```
mtdowling: missing new line

cjyclaire:
Use PR #1038 to track instead

PR #1024, opened 2016-06-30
[Doc only] typo fix & add documentation for retry scheme in FAQ & caching credentials

cjyclaire:
There are several issues asking for more documentation for SDK retry behavior.
Typo fix for `timeout` behavior, default timeout is 60s, which is configured via `default_socket_timeout` php.ini setting.

cc:\ @mtdowling @xibz @petermoon

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -215,7 +215,7 @@ content decoding.
 
 
 How do I disable body signing in S3?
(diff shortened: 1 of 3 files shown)


cjyclaire:
I'll follow up with new commit with more documentation for caching credentials.

PR #1025, opened 2016-07-01
Fix ci build failure

cjyclaire:
With newer `guzzlehttp/psr7` version upgrade, ci build is failing.

This PR fix the failure corresponds to `Psr7\parse_request()` behavior.
Removing hard-code duplicate headers signing test case.

:cc\ @mtdowling @xibz

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=5.5",
         "guzzlehttp/guzzle": "~5.3|~6.0.1|~6.1",
-        "guzzlehttp/psr7": "~1.0",
+        "guzzlehttp/psr7": "~1.3.1",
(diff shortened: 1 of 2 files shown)


xibz:
:Ship:

composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=5.5",
         "guzzlehttp/guzzle": "~5.3|~6.0.1|~6.1",
-        "guzzlehttp/psr7": "~1.0",
+        "guzzlehttp/psr7": "~1.3.1",
```
Tobion: `~1.3.1` is probably not what you really want as it will not allow 1.4 anymore. You should consider using `^1.3.1`

Issue #1031, opened 2016-07-11
SDK createClient method merges args twice


cjunge-work:
See: https://github.com/aws/aws-sdk-php/blob/master/src/Sdk.php#L199-L203

cjyclaire:
@cjunge-work , thanks for catching that!

PR #1032, opened 2016-07-12
Merge args only once when creating client with SDK

cjyclaire:
Currently [Sdk](https://github.com/aws/aws-sdk-php/blob/master/src/Sdk.php#L194) merges args[] twice in creating clients. This PR removes the extra merge. 
issue: #1031 

cc:\ @xibz @mtdowling


mtdowling:
:shipit:

PR #1037, opened 2016-07-13
Fix CI Warning caused by PHPUnit Api deprecation

cjyclaire:
[PHPUnit](https://github.com/sebastianbergmann/phpunit) Api `getMock()` Api is tagged as [deprecated](https://github.com/sebastianbergmann/phpunit/blob/5.4/src/Framework/TestCase.php#L1557) since PHPUnit 5.4.0, and there is announcement a month ago about its [`backward-compatibility-break`](https://github.com/sebastianbergmann/phpunit/issues/2194) in PHPUnit 6.

To maintain backward-compatibility in SDK, with composer dependency ` "phpunit/phpunit": "4.0|5.0"`, replacing `getMock()` with `getMockBuilder()`.

Referring original [`getMock()`](https://github.com/sebastianbergmann/phpunit/blob/5.4/src/Framework/TestCase.php#L1536) method, and [`MockBuilder`](https://github.com/sebastianbergmann/phpunit-mock-objects/blob/master/src/Framework/MockObject/MockBuilder.php) in PHPUnit.

cc:\ @mtdowling @xibz

### DIFF
--- a/tests/Api/Parser/JsonRpcParserTest.php
+++ b/tests/Api/Parser/JsonRpcParserTest.php
@@ -47,7 +47,7 @@ class JsonRpcParserTest extends \PHPUnit_Framework_TestCase
 
         $instance = new JsonRpcParser($service, $parser);
         $result = $instance(
-            $this->getMock(CommandInterface::class),
+            $this->getMockBuilder(CommandInterface::class)->getMock(),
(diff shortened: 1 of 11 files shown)


mtdowling:
:shipit:

PR #1038, opened 2016-07-14
Add S3 TransferException to track failed file source

cjyclaire:
Feature request from #1015.
By providing file information, customer would have better troubleshooting experience.

Besides having `getFileSouce()` to get the source file path that cause error, also add `getUncompletedFiles()` to retrieve remaining files that haven't been transferred.

cc:\ @mtdowling @xibz

### DIFF
--- a/src/S3/Transfer.php
+++ b/src/S3/Transfer.php
@@ -252,8 +252,10 @@ class Transfer implements PromisorInterface
             'concurrency' => $this->concurrency,
             'before'      => $this->before,
             'fulfilled'   => function ($value, $idx, Promise\PromiseInterface $p) {
-                $uri = $value->get('@metadata')['effectiveUri'];
-                $this->clearTaskDone((new Uri($uri))->getPath());
+                if ($value !== null && $value instanceof Aws\ResultInterface) {
+                    $uri = $value->get('@metadata')['effectiveUri'];
+                    $this->clearTaskDone((new Uri($uri))->getPath());
+                }
             },
             'rejected'    => function ($reason, $idx, Promise\PromiseInterface $p) {
                 $cmd = $reason->getCommand();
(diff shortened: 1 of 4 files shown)


src/S3/Transfer.php
```diff
@@ -233,18 +239,30 @@ private function createDownloadPromise()
...
         }
 
         // Create a GetObject command pool and return the promise.
         return (new Aws\CommandPool($this->client, $commands, [
             'concurrency' => $this->concurrency,
             'before'      => $this->before,
+            'fulfilled'   => function ($value, $idx, Promise\PromiseInterface $p) {
+                if ($value !== null && $value instanceof Aws\ResultInterface) {
+                    $uri = $value->get('@metadata')['effectiveUri'];
+                    $this->clearTaskDone((new Uri($uri))->getPath());
```
mtdowling: This would then be updated to just remove the object from the SplObjectStorage

src/S3/Transfer.php
```diff
@@ -233,18 +239,30 @@ private function createDownloadPromise()
...
+                }
+            },
             'rejected'    => function ($reason, $idx, Promise\PromiseInterface $p) {
-                $p->reject($reason);
+                $cmd = $reason->getCommand();
+                $this->handleTransferException(
```
mtdowling: I don't think we should be changing the exception here. All we need to do here is to remove the command from the inFlightTransfers object.

src/S3/Transfer.php
```diff
@@ -233,18 +239,30 @@ private function createDownloadPromise()
...
+                    $reason,
+                    "/" . $cmd['Bucket'] . "/" . $cmd['Key']
+                );
             }
         ]))->promise();
```
mtdowling: This is where you'd wrap the exception by adding a `then` method on the `promise()` call that decorates the given exception, adds the information about the failed command, and the list of inFlightTransfers / remainingTransfers.

cjyclaire:
We have talked about the PR offline, closing, tracking with #1140 instead

PR #1039, opened 2016-07-14
Feature/signature validation

timothy-r:
Adds a validate method to SignatureV4 class, to allow applications to validate signed requests.


mtdowling:
What's the motivation for this feature?

timothy-r:
We want to make signed requests from a lambda function to a PHP application and would rather use the AWS signature library than our own version.

cjyclaire:
@timothy-r , Thanks for your efforts in this PR, the intention for reusing the signature library is totally understandable, however, from SDK and security perspective in maintaining the AWS repository, it's not appropriate to allow validation from other application operations other than AWS operations. Enabling certain features with future maintenance is beyond AWS SDK scope.

Since the signature v4 algorithm is public with AWS [docs](http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html), feel free to look at our implementation or other third party libraries in terms of signature only.

skyzyx:
Lame. :(

@timothy-r: If you are simply looking to generate a cryptographic signature that can be signed with a secret key and passed over the public Internet to another trusted server, take a look at [skyzyx/signer](https://github.com/skyzyx/signer). It’s an implementation of the AWS Signature v4 that was extracted from the SDK and the AWS-specific pieces were either removed or repurposed.

Alternatively, you may also be able to use a JWT library to accomplish this since there’s a tremendous amount of overlap between the two use cases.

Issue #1049, opened 2016-07-26
SNS times out if simplexml extension not available


stevenmunro:
This is a precautionary issue as I currently have a ticket open with Amazon regarding a simple script designed to send a text message to my phone resulted in over 166,000 messages being sent through the system. Unfortunately I have been stung with a massive bill.

PHP7
SDK v3

If the simplexml extension is absent from your PHP, your script may never complete, thus bombarding messages until it times out.

I have disabled/enabled simplexml multiple times and my testing has been rather conclusive.

The AWS SDK for PHP should never had proceeded with the transaction unless it had properly detected that simplexml was available in the first place. Particularly when there is a potential high cost involved.

cjyclaire:
@stevenmunro apology for the trouble, and thank you for reporting. Yet may I ask the error message when `simplexml` is disabled? When I was trying to reproducing the issue, lacking `simplexml` would result in `400 Bad Request` which will cause the script failure and this 400 Error will not be retried by SDK. And since the request is failed, it won't arrive at server-side. 
So, could you provide more details about your testing case then we could dig more into this issue?

stevenmunro:
Hi @cjyclaire 
While it was timing out the only error message that was dumped to my PHP-error.log is the following;

[26-Jul-2016 00:00:22 UTC] PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /var/www/html/vendor/guzzlehttp/psr7/src/MessageTrait.php on line 147
[26-Jul-2016 00:00:22 UTC] PHP Stack trace:
[26-Jul-2016 00:00:22 UTC] PHP   1. {main}() /var/www/html/sms.php:0
[26-Jul-2016 00:00:22 UTC] PHP   2. send_sms() /var/www/html/sms.php:35
[26-Jul-2016 00:00:22 UTC] PHP   3. Aws\AwsClient->publish() /var/www/html/sms.php:28
[26-Jul-2016 00:00:22 UTC] PHP   4. Aws\AwsClient->__call() /var/www/html/sms.php:28
[26-Jul-2016 00:00:22 UTC] PHP   5. Aws\AwsClient->execute() /var/www/html/vendor/aws/aws-sdk-php/src/AwsClientTrait.php:78
[26-Jul-2016 00:00:22 UTC] PHP   6. GuzzleHttp\Promise\Promise->wait() /var/www/html/vendor/aws/aws-sdk-php/src/AwsClientTrait.php:59
[...]

PR #1053, opened 2016-07-27
When PHP Fatal Error is not surfaced, disable retry on certain errors

cjyclaire:
Fix issue #1049 

When using some customized PHP installation approaches in certain OS environments, PHP Fatal Error sometimes couldn't be surfaced at first place. Current retry mechanism checks exceptions and ignores error, this caused trouble for silent failing and retrying succeed requests when dependency is missing in parsing responses.

cc: @mtdowling @xibz

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -130,7 +130,7 @@ class RetryMiddleware
         ) {
             $this->updateHttpStats($value, $requestStats);
 
-            if ($value instanceof \Exception) {
+            if ($value instanceof \Exception || $value instanceof \Error) {
                 if (!$decider($retries, $command, $request, null, $value)) {
                     return \GuzzleHttp\Promise\rejection_for(
                         $this->bindStatsToReturn($value, $requestStats)
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -130,7 +130,7 @@ class RetryMiddleware
         ) {
             $this->updateHttpStats($value, $requestStats);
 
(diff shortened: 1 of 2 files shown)


cjyclaire:
Notice that the failing is due to `\Error` class is available in PHP7 only, working on work-arounds.

src/RetryMiddleware.php
```diff
@@ -130,7 +130,7 @@ public function __invoke(
         ) {
             $this->updateHttpStats($value, $requestStats);
 
-            if ($value instanceof \Exception) {
+            if ($value instanceof \Exception || $value instanceof \Error) {
```
mtdowling: Maybe create a helper method for this that only checks for `\Error` if it exists.

cjyclaire: Make sense, will do!

cjyclaire:
CI is passing now and ready for another review. @mtdowling

src/RetryMiddleware.php
```diff
@@ -130,7 +130,7 @@ public function __invoke(
         ) {
             $this->updateHttpStats($value, $requestStats);
 
-            if ($value instanceof \Exception) {
+            if ($value instanceof \Exception || $this->isErrorObject($value)) {
```
mtdowling: Why not just look for Throwable?

http://php.net/manual/en/class.throwable.php

cjyclaire: That makes perfect sense to me, changes is made in the latest commit.

mtdowling:
Interestingly, it appears that `::class` works even on undefined classes...

:shipit:

Issue #1055, opened 2016-07-29
PHP5.5 and Throwable and your Retry implementation


bestis:
Hi,

AWS SDK PHP says >=5.5 but then you have this in RetryMiddleware.php:
`
    private function isThrowable($value)
    {
        return class_exists(\Throwable::class)
            ? $value instanceof \Throwable
            : $value instanceof \Exception ;
    }
`
What do you think happens if we have class loader that throws errors if you try that? It floods the error log with Failed to load class, Throwable errors :(

Could you do this in any other way that would not need loading the class? If not, then at least add autoload false to it?
`
    private function isThrowable($value)
    {
        return class_exists(\Throwable::class, false)
            ? $value instanceof \Throwable
            : $value instanceof \Exception ;
    }
`

bestis:
Or should it even be:
`class_exists('Throwable', false)`

cjyclaire:
@bestis thank you for pointing this out, I just open a PR contains the fix, a new release will be out containing once the PR is merged.

GrahamCampbell:
You don;t need to check if a class exists. PHP won;t try to load it.

cjyclaire:
Closing, fix PR is merged and will be tagged in tomorrow release.

PR #1056, opened 2016-07-29
Disable autoload when checking class existence

cjyclaire:
Issue related: #1055 
Disable autoload when checking class existence.

cc:\ @mtdowling @xibz

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -130,7 +130,7 @@ class RetryMiddleware
         ) {
             $this->updateHttpStats($value, $requestStats);
 
-            if ($this->isThrowable($value)) {
+            if ($value instanceof \Exception || $value instanceof \Throwable) {
                 if (!$decider($retries, $command, $request, null, $value)) {
                     return \GuzzleHttp\Promise\rejection_for(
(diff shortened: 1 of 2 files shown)


src/RetryMiddleware.php
```diff
@@ -209,7 +209,7 @@ private function bindStatsToReturn($return, array $stats)
 
     private function isThrowable($value)
     {
-        return class_exists(\Throwable::class)
+        return interface_exists('Throwable', false)
```
GrahamCampbell: Calling instanceof on a non-exitence class is fine. PHP won;t try to autoload it because it's clever enough to see that it's clearly `false` if the definition hasn't even been loaded yet.

GrahamCampbell:
👎 See my above comment. Just check `instanceof Throwable` and `Exception`. No need for the conditional logic.

cjyclaire:
Interestingly, in original commit for the changes #1053 , using `instanceof` causing CI failing for PHP version < 7, that's why checking condition is added. I'm trying to figure out why.

cjyclaire:
This PR has been updated and ready for another review. @mtdowling @xibz

GrahamCampbell:
👍

PR #1067, opened 2016-08-09
DynamoDB Marshaler mapAsObject argument

Pierozi:
Hello,
It seems the argument `mapAsObject` has been forgot to the method `unmarshalItem` as the
method description already reference it and the `unmarshalValue` method implementation.


cjyclaire:
@Pierozi Thanks for catching! Looks good to me, I'll merge it.

Issue #1071, opened 2016-08-17
Issue with ELB client API (Could not resolve hostname)


fideloper:
Hello!

Since [methods on the ElasticLoadBalancer client are not documented](http://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.ElasticLoadBalancing.ElasticLoadBalancingClient.html), and the Console in the web app appears to use the ElasticLoadBalancerV2 API internally ([which are documented](http://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.ElasticLoadBalancingV2.ElasticLoadBalancingV2Client.html)), **I'm assuming we should be using the V2 ELB client**.

> Of course let me know if I'm wrong - perhaps V2 is meant for ALB instead of ELB? The lack of documentation for the "v1" client makes it hard to know how to get around this.

## Error:

When I use that, I receive the following error:

> Aws\ElasticLoadBalancingV2\Exception\ElasticLoadBalancingV2Exception with message 'Error executing "DescribeLoadBalancers" on "https://elasticloadbalancingv2.us-east-1.amazonaws.com"; AWS HTTP error: cURL error 6: Could not resolve host: elasticloadbalancingv2.us-east-1.amazonaws.com (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)'

## Info / Code:

**Package Tried:** 

* `aws/aws-sdk-php` @ `3.18.39`, commit `85f1fddaeb40b95106b2a2764268e9c89fc258ce`
* `aws/aws-sdk-php` @ `3.19.0`, commit `204337efe872a0207781089833de0f8594fc5fd7`

**Code**

```php
<?php

require_once('vendor/autoload.php');

$elb = new Aws\ElasticLoadBalancingV2\ElasticLoadBalancingV2Client([
    'version' => 'latest', // Also tried '2015-12-01'
    'region' => 'us-east-1',
]);

// http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-elasticloadbalancingv2-2015-12-01.html#describeloadbalancers
$balancers = $elb-> describeLoadBalancers();

// Throws Aws\ElasticLoadBalancingV2\Exception\ElasticLoadBalancingV2Exception 
```

Let me know if I'm using the SDK incorrectly for ELB!

Thanks

fideloper:
Using the latest AWS CLI tool (which has elb and elbv2 commands):

This is on my account which has one "classic" ELB (no ALB's):

```bash
[...]

PR #1072, opened 2016-08-17
Fix ElasticLoadBalancingV2 endpoints resolving issue

cjyclaire:
Addressing issue #1071 

Resolving ElasticLoadBalancingV2 with ElasticLoadBalancing endpoints.

cc:\ @mtdowling @xibz

### DIFF
--- a/src/ElasticLoadBalancingV2/ElasticLoadBalancingV2Client.php
+++ b/src/ElasticLoadBalancingV2/ElasticLoadBalancingV2Client.php
@@ -2,6 +2,9 @@
 namespace Aws\ElasticLoadBalancingV2;
 
 use Aws\AwsClient;
+use Aws\Command;
+use Aws\CommandInterface;
+use Psr\Http\Message\RequestInterface;
 
 /**
  * This client is used to interact with the **Elastic Load Balancing** service.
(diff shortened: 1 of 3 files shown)


src/ElasticLoadBalancingV2/ElasticLoadBalancingV2Client.php
```diff
@@ -64,4 +67,38 @@
...
+    {
+        $region = $this->getRegion();
+        return static function (callable $handler) use ($region) {
+            return function (
+                CommandInterface $command,
+                RequestInterface $request = null
+            ) use ($handler, $region) {
+                $request = $request->withUri(
+                    $request->getUri()
+                        ->withHost("elasticloadbalancing.{$region}.amazonaws.com")
```
mtdowling: I think this should be configured in the constructor, not each time a request is created. This just overwrites the host header, so if you wanted to hit a custom endpoints (like a test endpoint for example), this would not allow that.

cjyclaire: That's true, changes are made in the latest commit.

xibz:
:shipit:

Issue #1065, opened 2016-08-08
Error executing "HeadBucket" with v3 SDK


mdevine82:
So I'm having an issue with moving my code from using an old version of the SDK to the latest v3. When I run through the following code, it fails with the following error:

_Error executing "HeadBucket" on "https://s3.amazonaws.com/My_Bucket_Name"; AWS HTTP error: Client error: `HEAD https://s3.amazonaws.com/My_Bucket_Name` resulted in a `403 Forbidden` response: (client): 403 Forbidden (Request-ID: 4471096CBF8C37F2)_

Here is the code I'm using, any help would be appreciated. I've boiled it down to a simple example.

```
$s3Client = new Aws\S3\S3MultiRegionClient([
        'version'     => 'latest',
        'region'      => 'us-east-1',
        'credentials' => [
            'key'    => MY KEY,
            'secret' => MY SECRET,
        ],
    ]);
 
    $bucket = My_Bucket_Name;
    $region = $s3Client->determineBucketRegion($bucket);
 
    $response = $s3Client->putObject([
        'Body' => '12345',
        'Bucket' => $bucket,
        'Key' => rand(1,10000) . '.txt',
        '@region' => $region
    ]);
```

I also posted this on the forums but didn't get a response, so I just wanted to mirror that post here which I'll also do on the flip side when we figure it out.

https://forums.aws.amazon.com/message.jspa?messageID=734219

cjyclaire:
@mdevine82 , in using `Aws\S3\S3MultiRegionClient`, you don't have to include `region` parameter in constructing the client. 

More docs about [`S3MultiRegionClient`](https://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-multiregion-client.html).

mdevine82:
Alright, so if I remove the region altogether from both the constructor and the putObject.  Will it do the lookup automatically and store in that internal cache?

cjyclaire:
@mdevine82 In caching the regions, you need to do something like:
```
use Aws\DoctrineCacheAdapter;
use Aws\Sdk;
use Doctrine\Common\Cache\ApcuCache;

$sdk = new Aws\Sdk([
    'version' => 'latest',
    'region' => 'us-east-1',
    'S3' => [
        'bucket_region_cache' => new DoctrineCacheAdapter(new ApcuCache),
    ],
]);
```

cjyclaire:
[...]

PR #1075, opened 2016-08-24
Adding a rejected promise handler to determineBucketRegionAsync

cjyclaire:
Addressing issue #1065 

Adding a rejected promise handler to `determineBucketRegionAsync` to surface exception immediately when making a `HEAD` bucket request.

cc:\ @mtdowling @xibz @jeskew

### DIFF
--- a/src/S3/S3ClientTrait.php
+++ b/src/S3/S3ClientTrait.php
@@ -2,6 +2,7 @@
 namespace Aws\S3;
 
 use Aws\CommandInterface;
+use Aws\Exception\AwsException;
 use Aws\HandlerList;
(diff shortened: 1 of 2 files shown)


jeskew:
You might want to add a test with an exception that *does* contain a response, as would be generated by the situation discussed in #1065. (An `x-amz-bucket-region` header would be present on a 4XX response, and no exception would be surfaced.)

cjyclaire:
@jeskew Make sense! This test case is added in the latest commit.

jcchavezs:
Will this be merged soon?

mtdowling:
:shipit:

PR #1077, opened 2016-09-02
Add shared config file & assumeRole credential into credential chain

cjyclaire:
#### For an unified credential behavior across SDKs and CLI, this PR reworks current default credential provider logic, which contains:
1. Add AssumeRoleCredentialProvider support
2. Add Shared config file support (`~/.aws/config`)
3. Fix ECS credential throws exception outside of a `RejectedPromise`
4. Add `config_profile` option for client configuration
#### Unified default credential provider chain:
1. hard coded credentials
2. environment variable credentials
3. assume role profile
4. `~/.aws/credentials` static profile
5. `~/.aws/config` static profile
6. `Ecs` credentials
7.  `Ec2` instance Metadata credentials
#### Remaining works:
1. When this PR becomes more mature, I'll add more detailed credential guide in our documentation.
2. In consideration of backwards-compatibility, when this PR gets final merge, a minor version bump will happen.
3. Adds to 2., in terms of maintaining backwards-compatibility, whether an environment variable should be introduced for this unified credential behavior. ()
   - Scenario that may cause a break: 
     
     having assume role profile living inside same `~/.aws/credentials`, assume role credentials will be sourced before static credentials in that file. (Yet considering assume role profile is never introduced into PHP SDK, this case won't happen normally)

CC:\ @mtdowling @xibz


eric-tucker:
Does this include MFA support?  Or is that in a future release?

cjyclaire:
@eric-tucker yes it does. With `AssumeRoleCredentialProvider`, by passing `assume_role_params` with [MFA configurations](http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sts-2011-06-15.html#assumerole) will do the work.

eric-tucker:
+1

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
      *
      * @return callable
      */
-    public static function ini($profile = null, $filename = null)
+    public static function ini($profile = null, $filename = null, $config_file = false)
     {
-        $filename = $filename ?: (self::getHomeDir() . '/.aws/credentials');
-        $profile = $profile ?: (getenv(self::ENV_PROFILE) ?: 'default');
+        $filename = self::getFileName($filename, $config_file);
+        $profile = $profile ?: (
```
mtdowling: This line really needs a comment to explain what's going on

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+            if (!is_array($args)) {
+                // Rejected promise
+                return $args;
+            }
+            $provider = new AssumeRoleCredentialProvider($args);
+            return $provider();
+        };
+    }
+
+    public static function loadConfigurationForAssumeRoleCredential(
```
mtdowling: Can this be private?

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+        $data,
+        $assume_role_profile,
```
mtdowling: Please use camelCase for consistency with the rest of the SDK

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+                    self::ini($source_profile, $filename),
+                    self::ini($source_profile),
+                    self::ini($source_profile, null, true)
+                );
+            }
+        }
+        unset($data[$assume_role_profile]['source_profile']);
+
+        if (isset($data[$assume_role_profile]['region'])
+        ) {
```
mtdowling: Why is this on a new line?

src/Credentials/CredentialProvider.php
```diff
@@ -78,7 +79,10 @@ public static function defaultProvider(array $config = [])
         return self::memoize(
             self::chain(
                 self::env(),
+                self::assumeRole(),
```
mtdowling: Shouldn't you pass in the $config array? Otherwise the config is not being used

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
     }
 
     /**
+     * Credential provider that creates credentials using assume role
+     * credentials from ini profile
+     *
+     * @param string|null $assume_role_profile assume_role_profile name to be used
```
mtdowling: camelCase

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+     *                                         By default: assumes_role for credentials file
+     *                                         profile assumes_role for config file.
+     * @param string|null $filename            If provided, uses a custom filename rather
+     *                                         than looking in the home directory.
+     * @param bool|false  $config_file         the type of file the assume_role_profile
```
mtdowling: camelCase

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+     * Credential provider that creates credentials using assume role
+     * credentials from ini profile
+     *
+     * @param string|null $assume_role_profile assume_role_profile name to be used
+     *                                         By default: assumes_role for credentials file
+     *                                         profile assumes_role for config file.
+     * @param string|null $filename            If provided, uses a custom filename rather
+     *                                         than looking in the home directory.
+     * @param bool|false  $config_file         the type of file the assume_role_profile 
+     *                                         lives in. Use credentials file by default.
```
mtdowling: /s/Use/Uses/

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+     *
+     * @return callable
+     */
+    public static function assumeRole(
+        $assume_role_profile = null,
+        $filename = null,
+        $config_file = false
+    ){
```
mtdowling: Missing space

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+    }
+
+    public static function loadConfigurationForAssumeRoleCredential(
+        $data,
+        $assume_role_profile, 
+        $filename, 
+        $config_file
+    ){
+        $args = [];
+        if (!isset($data[$assume_role_profile]['source_profile'])) {
```
mtdowling: This doesn't belong as part of the assume role provider. This should be moved into the default provider

src/Credentials/CredentialProvider.php
```diff
@@ -267,32 +271,136 @@ public static function ecsCredentials(array $config = [])
...
+            // No source_profile specified, jump to next in chain after ini
+            $args['credentials'] = self::chain(
+                self::ecsCredentials(),
+                self::instanceProfile()
+            );
+        } else {
```
mtdowling: What is going on here? This seems overly complex

src/ClientResolver.php
```diff
@@ -118,6 +118,12 @@ class ClientResolver
             'doc'   => 'Allows you to specify which profile to use when credentials are created from the AWS credentials file in your HOME directory. This setting overrides the AWS_PROFILE environment variable. Note: Specifying "profile" will cause the "credentials" key to be ignored.',
             'fn'    => [__CLASS__, '_apply_profile'],
         ],
+        'config_profile' => [
```
cjyclaire: Since there is a `profile` parameter already (which checks for credential file only), I thought it could be handy when customer knows clearly what they are doing, they don't need to create a provider. It doesn't overlap with `AWS_PROFILE`, it just make it quicker in deciding which type of file is looking at.

For other SDKs, I know Ruby just has `profile` parameter only, and it use credential chain to check credential file before checking config file.

src/Credentials/CredentialProvider.php
cjyclaire: I thought about that, I didn't pass in the `$config` array because, I'd like to have the `defaultProvider()` plays the role of our SDK default credential chain instead of a provider where any customization could happen. I think it's a general good practice for our customer to be aware of what credential provider they are using when they really wants to apply some customization. Thoughts?

The current `$config` living there is serving for those instance credential providers.

src/ClientResolver.php
mtdowling: I'd like to see if we can limit the API surface area here and utilize a single option. I think it would be a good idea to see how all of the other SDKs are handling this.

src/Credentials/CredentialProvider.php
mtdowling: It's possible that people want to use the default credential chain, but also need to provide custom values to each chain depending on whether or not that chain is actually utilized... This might be a more complex problem that would need to be solved in a separate update, so maybe this can be left out for now.

src/Credentials/CredentialProvider.php
mtdowling: I don't see a reason to publicly expose this. If you can't figure out a good way to test this if it's private, then let's add the `@internal` annotation to the docblock.

cjyclaire:
We have talked about this PR offline, and a lot of refractor is done regarding the complexity there. 

I create a separate `AssumeRoleCredentialResolver` to resolve some chaining logic when applying assumeRole credential from profile. This also make it more flexible if customer wants to utilize part of resolving logic in credential chain.

I think this PR is ready for a second review now. @mtdowling

src/Credentials/CredentialTrait.php
```diff
@@ -0,0 +1,61 @@
...
+            return "Cannot read credentials from $fileName";
+        }
+        $data = parse_ini_file($fileName, true);
+        if ($data === false) {
+            return "Invalid credentials file: $fileName";
+        }
+        if (!isset($data[$profile])) {
+            return "'$profile' not found in $fileName";
+        }
+        return '';
```
mtdowling: Could just return nothing (i.e., omit the `return ''` so that null is returned.

src/Credentials/CredentialTrait.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Sourcing credential profile data from file helpers.
+ */
+trait CredentialTrait
+{
+    /**
+     * Check profile availability in a given file.
+     * Return empty string or error message string
+     *
+     * @return string
```
mtdowling: return null|string Returns a string describing any error that could occur.

src/Credentials/CredentialProvider.php
```diff
@@ -267,77 +280,73 @@ public static function ecsCredentials(array $config = [])
...
-            }
-            $data = parse_ini_file($filename, true);
-            if ($data === false) {
-                return self::reject("Invalid credentials file: $filename");
-            }
-            if (!isset($data[$profile])) {
-                return self::reject("'$profile' not found in credentials file");
+        return function () use ($profile, $fileName) {
+            $msg = self::checkProfile($profile, $fileName);
+            if (!empty($msg)) {
```
mtdowling: This could just be `if ($msg) {`

src/Credentials/CredentialProvider.php
```diff
@@ -267,77 +280,73 @@ public static function ecsCredentials(array $config = [])
...
+                return self::reject($msg);
             }
-            if (!isset($data[$profile]['aws_access_key_id'])
```
mtdowling: The previous approach was used so that it would be able to handle an invalid profile name that couldn't be found in the ini file. The current implementation would throw a warning when the profile can't be found because you're accessing an array key that may not exist.

Can you add a test for this too?

src/Credentials/CredentialProvider.php
```diff
@@ -267,77 +280,73 @@ public static function ecsCredentials(array $config = [])
...
             }
 
-            if (empty($data[$profile]['aws_session_token'])) {
-                $data[$profile]['aws_session_token']
-                    = isset($data[$profile]['aws_security_token'])
-                        ? $data[$profile]['aws_security_token']
-                        : null;
+            if (empty($data['aws_session_token'])) {
+                $data['aws_session_token']
+                    = isset($data['aws_security_token']) ?
```
mtdowling: Can you make the indentation here look like it used to?

src/Credentials/CredentialProvider.php
```diff
@@ -75,10 +78,20 @@ public static function defaultProvider(array $config = [])
...
         return self::memoize(
             self::chain(
                 self::env(),
+                self::assumeRole(),
+                self::assumeRole(
+                    'profile assumes_role',
+                    $defaultConfigFileName
+                ),
                 self::ini(),
+                self::ini(
```
mtdowling: This new chain setup looks really expensive -- we are potentially reading and ini parsing both the credentials files and config files twice just in this chain, and looking at the assumeRole method, that number could climb to 6.

Have you benchmarked this? How does this affect the common usage of the SDK with environment variables and instance profile credentials?

src/Credentials/AssumeRoleCredentialResolver.php
```diff
@@ -0,0 +1,178 @@
...
+                    CredentialProvider::assumeRole(
+                        $sourceProfile,
+                        $homeDir . self::DEFAULT_CONFIG_FILE
+                    ),
+                    CredentialProvider::ini($sourceProfile, $this->fileName),
+                    CredentialProvider::ini(
+                        $sourceProfile,
+                        $homeDir . self::DEFAULT_CREDENTIAL_FILE
+                    ),
+                    CredentialProvider::ini(
```
mtdowling: Would all these chain calls really load the ini file and ini parse it this many times?

src/Credentials/AssumeRoleCredentialProvider.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Loads assume role credentials.
+     *
+     * @return PromiseInterface
+     */
+    public function __invoke()
+    {
+        $client = $this->client;
+        return Promise\promise_for($client->assumeRoleAsync($this->assumeRoleParams))
```
mtdowling: You don't need a promise_for when you know the return value of `assumeRoleAsync` will always be a promise.

src/Credentials/CredentialProvider.php
cjyclaire: Actually, the behavior stays the same. Because `checkProfile` is called first before `getProfileData`, it would have thrown exception already for this situation. Current test covers (this](https://github.com/aws/aws-sdk-php/blob/master/tests/Credentials/CredentialProviderTest.php#L235) already. Let me know if I misunderstood what you are suggesting :)

cjyclaire:
We have talked offline with the behavior there. Refractor is done and documentation for this behavior is added. Ready for another review :) @mtdowling

cjyclaire:
Ecs fix has been separated from this PR and this PR has been rebased and up-to-date.

cjyclaire:
We have discussed this offline, considering `Resolver` for profiles could introduce fair latency in fetching credentials, this part is separated from PR #1137

PR #1079, opened 2016-09-07
Automate release note

cjyclaire:
This PR is aimed at moving one step further in automated release process.

Known limitations:
- this automated release note script doesn't count new service release

cc\ @mtdowling @xibz

### DIFF
--- a/build/build-changelog.php
+++ b/build/build-changelog.php
@@ -1,0 +1,33 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$UPDATE = " - Added support for latest API updates.";
+$path = __DIR__ . '/../CHANGELOG.md';
+
+$diff = shell_exec("git status --porcelain");
+$diff_array = explode("\n", $diff);
+
+$entries = [];
+foreach ($diff_array as $line) {
+    $entry = explode('/', $line);
+    if (sizeof($entry) < 5) continue;
+
+    $service = \Aws\manifest($entry[2])["namespace"];
+    if (strpos($entry[4], 'docs') === 0 ||
+        strpos($entry[4], 'examples') === 0 ||
+        in_array($service, $entries)
+    ) {
+        continue;
+    }
+
+    $entries[] = $service;
+}
+
+//update CHANGELOG.md before tag the release
+$update_change_log = "sed -i '' -e 's/CHANGELOG/CHANGELOG\\\n\\\n## next release\\\n";
+foreach ($entries as $service) {
+    $update_change_log .= "\\\n* `Aws\\\\{$service}`{$UPDATE}";
+}
+$update_change_log .= "/' {$path}";
+shell_exec($update_change_log);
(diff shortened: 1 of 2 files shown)


build/build-changelog.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$UPDATE = " - Added support for latest API updates.";
+$path = __DIR__ . '/../CHANGELOG.md';
+
+$diff = shell_exec("git status --porcelain");
```
xibz: It looks like `--porcelain` is only supported in versions `2.10.0` and above. This may be an issue with the git versions we use. May want to check on that.

cjyclaire: Thanks for notice! It actually works at my local v2.8.1, and its [original commit](https://github.com/git/git/commit/6f15787181a163e158c6fee1d79085b97692ac2f) merged in 2009, I'll find our the initial version that this feature comes out.

xibz: Okay, that's good. When I was running the command it was erroring out with `porcelain` not defined. Must've misspelled it!

mtdowling:
Can you provide some example outputs of this?

> this automated release note script doesn't count new service release

Are you going to add support for this before this is merged?

cjyclaire:
@mtdowling Of course, here are some examples:
after model changes introduced, `git status`:
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   src/data/application-autoscaling/2016-02-06/api-2.json
        modified:   src/data/cognito-identity/2014-06-30/api-2.json
        modified:   src/data/dynamodb/2012-08-10/docs-2.json
        ...
```
`git status --porcelain` gives:
```
 M src/data/application-autoscaling/2016-02-06/api-2.json
 M src/data/cognito-identity/2014-06-30/api-2.json
 M src/data/dynamodb/2012-08-10/docs-2.json
 ...
```
run `make changelog` then `CHANGELOG.md` will look like:
```
# CHANGELOG

## next release

* `Aws\ApplicationAutoScaling` - Added support for latest API updates.
* `Aws\CognitoIdentity` - Added support for latest API updates.
 ...
```
This `make changelog` command will automated generate release notes for API related changes.

I won't add support for detect new service at this point, because new service model change requires separate build command, which also need manual inputs. I'd like to have this merged as a start point of testing some basic release automation.

mtdowling:
The changelog, in this proposed format, doesn't reveal much information about the change. Couldn't we automate detecting what operations are added, service version bumps, addition of new fields, deprecation of fields, new services, etc? All of that could be automated in a way that would be close to what we are currently providing to customers, but still providing about as much information as our current changelog.

cjyclaire:
@mtdowling Actually, I agree I'd like to see that happen instead of general message here (how other SDK handle their release notes currently). I think I'll work on that before merging this PR :)

cjyclaire:
Closing, now the information could be accessible in a different way.

PR #1093, opened 2016-09-23
Properly set security token for S3 direct upload when using temporary credentials

bakura10:
Hi,

I've been using locally the PostObjectV4 to create form inputs that allowed to directly upload to S3.

This worked perfectly while using access keys. However once deployed and using temporarily credentials, this didn't work.

After a lot of research, it appears that you need to pass an additional "x-amz-security-token" to the policy, and this token must also be set in the HTML form. cf: http://stackoverflow.com/questions/18884683/browser-uploads-to-s3-with-instance-roles/19024044

I've tested this fix and it works :).

### DIFF
--- a/src/S3/PostObjectV4.php
+++ b/src/S3/PostObjectV4.php
@@ -54,6 +54,14 @@ class PostObjectV4
             'enctype' => 'multipart/form-data'
         ];
 
+        $credentials   = $this->client->getCredentials()->wait();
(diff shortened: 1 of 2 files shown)


cjyclaire:
@bakura10 Thank you for the PR! The addition and test case both looks good to me. I'll merge it and tag it in next release.

bakura10:
Awesome, thanks a lot! :)

PR #1094, opened 2016-09-25
Remove exec bits from JSON files

siwinski:
I'm not sure if this pull request will be accepted or not based on the following from the guidelines for contributing: 

> We may choose not to accept pull requests that change files in the src/data directory, since we generate these files based on our internal knowledge of the AWS services. Please check in with us ahead of time if you find a mistake or missing feature that would affect those files.

If this pull request is not acceptable, please take this as a request to update the data files generator to remove the exec bit for generated data files.


cjyclaire:
@siwinski It's true that this PR cannot be merged in directly, thanks for noticing! I'll take a look.

cjyclaire:
Closing, this is taken care of in the latest release, thanks!

siwinski:
THANKS @cjyclaire

PR #1095, opened 2016-09-28
[Doc Fix] Duplicate "EC2" in Doc quicklink

cjyclaire:
In current [API doc](http://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html), there exists 2 EC2 QuickLinks leads to 2 versions of `Ec2Client`.

This bug is introduced in PR #1072 that intends to fix the problem of ELB and ELBV2 sharing same service title by indexing them with `title` and `shortTitle` separately. In `quickLink` part, since now each service version has different key in `services` array, the latest version of the service needs to be kept.

@mtdowling

### DIFF
--- a/build/docs/classes/DocsBuilder.php
+++ b/build/docs/classes/DocsBuilder.php
@@ -148,6 +148,17 @@ class DocsBuilder
             return array_shift($versions);
         }, $services);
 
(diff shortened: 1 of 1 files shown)


mtdowling:
🚢

PR #1111, opened 2016-10-21
Add DynamoDb keywords

remipelhate:
Hi there! I thought it'd be quite handy to be able check whether or not a certain word is a reserved keyword in DynamoDB, so... I defined these as constants to easily reference them when assembling a query.

I got the list of http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html, though the `REMOVE` keyword seems to be missing. I added it to this class, but I presume the documentation should be updated as well :)


cjyclaire:
@remipelhate Appreciate your thoughts for this PR! 
It appears as a reasonable request at first glance, yet taking a second thought, given that those `Reserved Words` in DynamoDB doesn't bound to their service interface, and will fail API after a request has been made, you have also notice that there is also discrepancy between docs and latest API behavior. It looks like a change that requires continuous maintenance from SDK side with manual update only. Looking at this huge list, it adds up redundancy to SDK as well considering that this list might keep appending. Furthermore, if this was merged, people might depend on this functionality for validation words, it could easily break people if the const list has been updated or hasn't been updated in time. Thus I think it would be a more clean way to let it fall naturally. Thoughts? :)

mtdowling:
Agreed with @cjyclaire. This is too large of a change to be able to maintain in the SDK. Perhaps a separate repository meant for specifically building DynamoDB queries would be a more appropriate place for this.

remipelhate:
Alright, makes sense. It does indeed rather belong a the repository offering the ability to build DynamoDB queries. That's in fact the reason I needed it in the first place. I developed a DynamoDB QueryBuilder that automatically generates a placeholder for an attribute name when it's a reserved word. Thanks for the insights guys!

PR #1122, opened 2016-11-21
Fix DynamoDB session handler sleep time

pspiller:
Between attempts to access a locked session, the handler is meant to
wait for a configurable number of microseconds. The handler code was
previously multiplying the wait time by 10^-6 before passing it to
usleep, even though the values were already in microseconds. For the
default min and max wait time values of 10000 and 50000, the usleep
argument would always be less than one and so the handler wouldn't
wait at all between attempts.

This commit removes the multiplication by 10^-6, so the handler sleeps
for the intended amount of time between requests.


cjyclaire:
@pspiller Thanks for the PR fix! It helps keep consistency with our public doc [guide](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/dynamodb-session-handler.html#configuration)

jeremeamia:
Doh! 😦  Nice catch @pspiller

PR #1127, opened 2016-11-28
change us-standard to us-east-1 in doc example

EronHennessey:
I received a complaint that 'us-standard' is not a valid region. Changed to 'us-east-1'.


cjyclaire:
Thanks for catching!

Issue #1129, opened 2016-12-04
PostObjectV4 overwrites virtual-style URLs


jaypea:
Hi, 

we are trying to upgrade the sdk from v2 to v3 in our big legacy app. Our complete system is based on virtual host style URLs for S3. 
So we changed the S3Client Config to like this:
```
$client = new S3Client([
    'credentials' => [
        'key' => 'sdfsdf',
        'secret' => 'sdsds',
    ],
    'region' => 'eu-west-1',
    'version' => '2006-03-01',
    'endpoint' => '<bucketName>.s3-eu-west-1.amazonaws.com',
    'bucket_endpoint' => true,
]);
```

This is the suggested workaround mentioned in this issue (https://github.com/aws/aws-sdk-php/issues/1118#issuecomment-260515622). This comment also states that the sdk would use path-style instead of virtual-style to reduce dns lookups and increase performance. 

However, inside of  `\Aws\S3\PostObjectV4` we found this:
```
    private function generateUri()
    {
        $uri = new Uri($this->client->getEndpoint());

        if ($uri->getScheme() === 'https'
            && strpos($this->bucket, '.') !== false
        ) {
            // Use path-style URLs
            $uri = $uri->withPath($this->bucket);
        } else {
            // Use virtual-style URLs
            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
        }

        return (string) $uri;
    }
```
which would use virtual-style URLs for every non-ssl request.

In our case, it adds the bucket to the endpoint again, which was already provided in the s3client config.
So when using PostObjectV4, the resulting URL is
`http://<bucketname>.<bucketname>.s3-eu-west-1.amazonaws.com`.

Is this a bug or is there something we could configure differently?

As a workaround we could create 2 different s3client objects with different configurations, but this feels wrong somehow.

Switching to https is not an option currently, because our internal test system relies on fake-s3 , which does not do ssl.

cjyclaire:
@jaypea Thanks for the information and your efforts in upgrading, appreciate that!

So as you have observed, the discussion in #1118 refers to a general S3 workaround, which is not same as `PostObjectV4` case. In your case, you would like to have `virtual-host` style URL with `http` scheme. Then you could simply do:
``` php
    $s3 = new \Aws\S3\S3Client([
    'region' => 'us-west-2',
    'version' => 'latest',
    'scheme' => 'http'
   ]);
```
[...]

PR #1131, opened 2016-12-05
Fix PostObjectV4 ignoring customized virtual-style endpoint

cjyclaire:
Fixing issue #1129 

The bug: When virtual-style endpoint is hard coded by `endpoint` configuration, `PostObjectV4` doesn't detect it, will still trying to convert it to virtual-style, which causes 2 buckets in the url looks like: `https://bucket.bucket.s3.amazonaws.com`.

@mtdowling @xibz

### DIFF
--- a/src/S3/PostObjectV4.php
+++ b/src/S3/PostObjectV4.php
@@ -151,8 +151,10 @@ class PostObjectV4
             // Use path-style URLs
             $uri = $uri->withPath($this->bucket);
         } else {
-            // Use virtual-style URLs
-            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+            // Use virtual-style URLs if haven't been set up already
+            if (strpos($uri->getHost(), $this->bucket) === false) {
+                $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+            }
         }
 
         return (string) $uri;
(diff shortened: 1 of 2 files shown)


src/S3/PostObjectV4.php
```diff
@@ -151,8 +151,10 @@ private function generateUri()
             // Use path-style URLs
             $uri = $uri->withPath($this->bucket);
         } else {
-            // Use virtual-style URLs
-            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+            // Use virtual-style URLs if haven't been set up already
+            if (strpos($uri->getHost(), $this->bucket) === false) {
```
xibz: Shouldn't this check only for prefix? Because what if they used a field with the same bucket name?

cjyclaire: @xibz That's true, thanks for catching, I'll fix it!

cjyclaire:
Fixed, ready for another review :) @xibz

xibz:
Just need to add a . against the host check

src/S3/PostObjectV4.php
```diff
@@ -151,8 +151,10 @@ private function generateUri()
             // Use path-style URLs
             $uri = $uri->withPath($this->bucket);
         } else {
-            // Use virtual-style URLs
-            $uri = $uri->withHost($this->bucket . '.' . $uri->getHost());
+            // Use virtual-style URLs if haven't been set up already
+            if (strpos($uri->getHost(), $this->bucket) !== 0) {
```
xibz: Should be
```php
if (strpos($uri->getHost(), $this->bucket . '.') !== 0) {
}
```

Issue #1132, opened 2016-12-06
EcsCredentialProvider introduces 1sec latency for each request on EC2 instances


saintberry:
I think there is a bug with the logic in the `CredentialProvider::defaultProvider()` chain on EC2 instances.

If no credentials are passed into the client _and_ there is no credential cache configured, I believe 1sec (a default timeout) of latency is being introduced due to the order of the chain on EC2 instances.

If no credentials are found in the environment or ini the EcsCredentialProvider is called [before the InstanceProfileProvider](https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/CredentialProvider.php#L78). The issue here is that in the EcsCredentialProvider a request is made to [http://169.254.170.2 with a 1sec timeout](https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/EcsCredentialProvider.php#L45). On EC2 instances this request is just going to hang.

I think this bug was introduced in [this pull request](https://github.com/aws/aws-sdk-php/pull/1126/files) which removed a check to see if `getenv(self::ENV_URI)` was empty or not. I think this check was trying to determine if the instance was ECS or EC2.

Can someone else confirm this is the case?

I think it would be good for `EcsCredentialProvider` to fail early if the instance is not ECS and avoid that Guzzle request.

cjyclaire:
@saintberry Thanks for noticing, so that PR doesn't remove the check, it lets the request to fall when wrapped in promise instead of throw exception immediately. Appreciate pointing out the latency there, `timeout` is configurable via provider now. However since the ecs credential is not cached correctly, it causes the latency. The  PR #1133 looks like making the correct fix by caching the ecs credentials properly. I'll take a deeper look into that and deliver the fix shortly. Thank you for your patients.

anniballo:
Hello,

We are using the DynamoDB Session Handler in our PHP application and just noticed this same 1sec delay in every request to our application. Doing some digging we also found that the DynamoDB client when using the EC2 IAM role (instead of directly passing credentials in the config) was the culprit. We just moved our production infrastructure to hardcoded credentials but would like to move back to using the IAM role (for obvious security and simplicity) when fixed. Do you have any more information on this?

Thanks

Diego

cjyclaire:
[...]

PR #1133, opened 2016-12-06
Add remote credentials provider

vasanthperiyasamy:
Issue: Default credential provider instantiates both InstanceProfileProvider and EcsCredentialProvider. This results in 1 sec latency for Instance profile security-credentials calls (as it has to wait for the EcsCredentialProvider->__invoke() to timesout).

Fix: Instantiate only one remote credentials provider based on the context(EC2/ECS) 

Fixes issue #1132


src/Credentials/CredentialProvider.php
```diff
@@ -267,6 +265,27 @@ public static function ecsCredentials(array $config = [])
...
+     */
+    public static function remoteCredentialsProvider(array $config = [])
+    {
+        $ecsCredentialsUri = getenv(EcsCredentialProvider::ENV_URI);
+
+        if (!$ecsCredentialsUri) {
+            return self::instanceProfile($config);
+        }
+
+        return self::ecsCredentials($config);
```
cjyclaire: There is a problem with the logic here, even with `ENV` existing, there is possibility that the ecs credential fails when making the request, in that case, instanceProfile needs to be retrieved.

cjyclaire: That's true, however, for a default credential provider, it's aimed at a general credential chain that makes most sense instead of catching all possible scenarios. Especially, this a unified behavior across all SDKs for the situation that failing fetching ECS credentials will be followed with an Ec2 metadata credential attempt. Rather than return `ecsCredentials` solo, I'd prefer something like below:

``` php
return self::chain(
    self::ecsCredentials($config),
    self::instanceProfile($config)
);
```

src/Credentials/CredentialProvider.php
```diff
@@ -267,6 +265,27 @@ public static function ecsCredentials(array $config = [])
...
+     *
+     * @return InstanceProfileProvider|EcsCredentialProvider
+     * @see Aws\Credentials\InstanceProfileProvider for $config details.
+     * @see Aws\Credentials\EcsCredentialProvider for $config details.
+     */
+    public static function remoteCredentialsProvider(array $config = [])
+    {
+        $ecsCredentialsUri = getenv(EcsCredentialProvider::ENV_URI);
+
+        if (!$ecsCredentialsUri) {
```
cjyclaire: How about  changing this to one line like:
``` php
if(empty(getenv(EcsCredentialProvider::ENV_URI))){
    ...
```
In this way, if the `ENV` present with an empty string, it will fail quickly as well.

cjyclaire:
@vasanthperiyasamy Thanks for the work! Looks good to me, appreciate your efforts on this! Just merged, will be tagged in next release :)

PR #1135, opened 2016-12-07
add SQS FIFO Support

glopez620:
added required (MessageGroupId) field for SQS fifo support

### DIFF
--- a/src/Aws/Sqs/Resources/sqs-2012-11-05.php
+++ b/src/Aws/Sqs/Resources/sqs-2012-11-05.php
@@ -99,7 +99,6 @@ return array (
                     'location' => 'aws.query',
                 ),
                 'MessageGroupId' => array(
-                    'required' => true,
                     'type' => 'string',
                     'location' => 'aws.query',
                 ),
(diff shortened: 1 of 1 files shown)


src/Aws/Sqs/Resources/sqs-2012-11-05.php
```diff
@@ -98,6 +98,11 @@
                     'type' => 'string',
                     'location' => 'aws.query',
                 ),
+                'MessageGroupId' => array(
+                    'required' => true,
```
cjyclaire: Reference latest SQS model: https://github.com/aws/aws-sdk-php/blob/master/src/data/sqs/2012-11-05/api-2.json#L266, I didn't see `MessageGroupId` to be a required parameter?

src/Aws/Sqs/Resources/sqs-2012-11-05.php
```diff
@@ -98,6 +98,10 @@
                     'type' => 'string',
                     'location' => 'aws.query',
                 ),
+                'MessageGroupId' => array(
```
cjyclaire: Adding to last comment, I'm confused since this is not a member for the operation either? Could you tell me more about how you find adding this helpful?

cjyclaire:
Closing to due ages, and I don't see this is a member for the correct operation, feel free to reopen with further details :)

PR #1137, opened 2016-12-09
Adds support for AssumeRole and ./aws/config ini credentials

cjyclaire:
Referencing PR #1077 

This PR covers:
1. Introducing `AssumeRoleCredentialProvider`
2. Supports source ini credentials from `./aws/config` file in `defaultProvider`
3. Adding documentation regrading current behavior.

@mtdowling


src/Credentials/CredentialTrait.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param string $filename
+     *
+     * @return null|string
+     */
+    private static function checkProfile($profile, $filename)
+    {
+        if (!is_readable($filename)) {
+            return "Cannot read credentials from $filename";
+        }
+        $data = parse_ini_file($filename, true);
```
mtdowling: Why are you not returning the loaded `$data`? Right now, you are running this method twice, loading and parsing the same INI file. This method should return the parsed ini file and we should not re-parse the file again.

src/Credentials/CredentialTrait.php
```diff
@@ -0,0 +1,68 @@
...
+        if ($data === false) {
+            return "Invalid credentials file: $filename";
+        }
+        if (!isset($data[$profile])) {
+            return "'$profile' not found in $filename";
+        }
+    }
+
+    private static function getProfileData($profile, $filename)
```
mtdowling: This method doesn't really work... Maybe checkProfile should be combined with this method so that you're only loading the ini file once.

src/Credentials/CredentialProvider.php
```diff
@@ -47,12 +47,15 @@ class CredentialProvider
...
 
+    use CredentialTrait;
+
     /**
      * Create a default credential provider that first checks for environment
      * variables, then checks for the "default" profile in ~/.aws/credentials,
-     * then tries to make GET Request to fetch credentials if Ecs environment
-     * variable is presented, and finally checks for EC2 instance profile
-     * credentials.
+     * then checks for "profile default" profile in ~/.aws/config,
```
mtdowling: Can you document what this is (e.g., the default profile in the AWS CLI)? There's some documentation about the default credential chain that I think needs to be updated to reflect this change as well.

cjyclaire:
Addressing feedback, removing trait file, fix documentation.
Ready for review. @mtdowling :)

docs/guide/credentials.rst
```diff
@@ -77,6 +79,66 @@ been configured with an IAM role.
...
+you would need to provide ``'region'`` information, ``'credentials'`` for a ``StsClient`` and
+``'assume_role_params'`` information. Or you could simply pass in ``'client'`` parameter with a ``StsClient``.
+
+For more information regarding ``'assume_role_params'``, see `AssumeRole <http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sts-2011-06-15.html#assumerole>`_.
+
+.. code-block:: php
+
+    $assumeRoleCredentials = new AssumeRoleCredentialProvider([
+        'region' => 'us-west-2',
+        'credentials' => CredentialProvider::ini(),
```
mtdowling: I'm weary of supporting this use case. In order to do this, we'd need to create an STS client on the developer's behalf, correct? Which means we'd need to hardcode a dependency on the "latest" version of STS, correct? That would be a risky contract considering STS could ship a major version at some point that could break this credential provider. Are we doing anything to mitigate this? If not, how you feel about requiring a client be passed in and removing the 'credentials' + 'region' option?

docs/guide/credentials.rst
```diff
@@ -292,18 +354,77 @@ credentials from Amazon EC2 instance profiles.
     ]);
 
 
+ecsCredential provider
```
mtdowling: should this be plural?

src/Credentials/CredentialProvider.php
```diff
@@ -50,9 +50,10 @@ class CredentialProvider
     /**
      * Create a default credential provider that first checks for environment
      * variables, then checks for the "default" profile in ~/.aws/credentials,
-     * then tries to make GET Request to fetch credentials if Ecs environment
-     * variable is presented, and finally checks for EC2 instance profile
-     * credentials.
+     * then checks for "profile default" profile in ~/.aws/config (which is
```
mtdowling: Did you update the documentation that talks about default credential resolution to talk about this new behavior?

src/Credentials/CredentialProvider.php
```diff
@@ -64,7 +65,11 @@ class CredentialProvider
      */
     public static function defaultProvider(array $config = [])
     {
-        $localCredentialProviders = [self::env(), self::ini()];
+        $localCredentialProviders = [
```
mtdowling: Should this be its own method just like how the "removeCredentialProviders" method works?

docs/guide/credentials.rst
cjyclaire: I'm in favored of having a `client` passed in actually, having `credentials` and `region` makes it appears more convenient at first glance but actually introduces more constraints. I'll change that :)

docs/guide/credentials.rst
cjyclaire: Yes, thanks for catching!

src/Credentials/CredentialProvider.php
cjyclaire: Yes I have:  https://github.com/aws/aws-sdk-php/pull/1137/files#diff-f3a01c5f1a23795bd23ca728685e3449R385

cjyclaire:
Changes have been made, including requires `client` parameter etc.. @mtdowling

virgofx:
Any chance this can get merged? This is badly needed!

src/Credentials/AssumeRoleCredentialProvider.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $client = $this->client;
+        return $client->assumeRoleAsync($this->assumeRoleParams)
+            ->then(function (Result $result) {
+                return $this->client->createCredentials($result);
+            })->otherwise(function (AwsException $exception) {
+                $reason = $exception->getMessage();
+                throw new CredentialsException(
+                    "Error in retrieving assume role credentials: $reason."
+                );
```
mtdowling: Any reason to not add the caught exception as a previous exception when creating the CredentialsException?

mtdowling:
:shipit: from me.

Can you squash this to cleanup the history?
Can you also get a shipit from @jeskew before merging?

src/Credentials/AssumeRoleCredentialProvider.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @return PromiseInterface
+     */
+    public function __invoke()
+    {
+        $client = $this->client;
+        return $client->assumeRoleAsync($this->assumeRoleParams)
+            ->then(function (Result $result) {
+                return $this->client->createCredentials($result);
+            })->otherwise(function (AwsException $exception) {
```
virgofx: This was throwing a `CredentialsException` if you do the following:

Update local `~/.aws/config` or `~/.aws/credentials` to be invalid INI from PHP perspective but valid INI from other formats.

Basically add a `#` as a comment in the beginning of a line in the credentials file. This will cause PHP to bug out when parsing (even though this is still valid INI) and then the wrong exception type is thrown here. Need a higher level Exception or change of type?

cjyclaire: Hmm that's true, I was thinking about the API call scenario there ..., I should have made it accept a more general exception there.

src/Credentials/AssumeRoleCredentialProvider.php
cjyclaire: I can make it wrapped as a previous exception there, this makes more sense if the exception caught is not only scoped to an `AwsException`.

cjyclaire:
@mtdowling Of course, I'll definitely squash them before merge :) 

@jeskew It would be great if you could help review this PR as well :)

@virgofx Appreciate your patience and feedback on this :)

virgofx:
@cjyclaire Looking forward in getting this integrated. Any news?

cjyclaire:
@virgofx Just merged, will be tagged in our next release coming in one day or two :)

PR #1140, opened 2016-12-13
Add filePath Info in MultipartUploadException

cjyclaire:
Addressing issue #1015
Making `Bucket/Key` information accessible from `MultipartUploadException` instead of introducing complexity into `S3Transfer`. This introduces a more user-friendly way in getting file path information instead of parsing `objectUrl` in the error message.

@mtdowling

### DIFF
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -11,9 +11,8 @@ class MultipartUploadException extends \RuntimeException
     /**
      * @param UploadState      $state    Upload state at time of the exception.
      * @param \Exception|array $prev     Exception being thrown.
-     * @param array            $params   Configuration option array for MultipartUploadException
      */
-    public function __construct(UploadState $state, $prev = null, array $params = []) {
+    public function __construct(UploadState $state, $prev = null) {
         $msg = 'An exception occurred while performing a multipart upload';
 
         if (is_array($prev)) {
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -63,10 +63,13 @@ class MultipartUploadException extends \RuntimeException
 
     /**
      * Get file path of the transfer
+     * Note:
+     * - returns null when the bucket and key could not be
+     *   ascertained from the exception
      *
-     * @return array
+     * @return string|null
      */
-    public function getFilePaths()
+    public function getFilePath()
     {
         return $this->filePath;
     }
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -65,7 +65,7 @@ class MultipartUploadException extends \RuntimeException
      * Get file path of the transfer
      * Note:
      * - returns null when the bucket and key could not be
-     *   ascertained from the exception
+     *   ascertained from the exception. (e.g., Glacier)
      *
      * @return string|null
      */
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -1,7 +1,6 @@
 <?php
 namespace Aws\Exception;
 
-use Aws\CommandInterface;
 use Aws\Multipart\UploadState;
 
 class MultipartUploadException extends \RuntimeException
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -1,12 +1,15 @@
 <?php
 namespace Aws\Exception;
 
+use Aws\CommandInterface;
 use Aws\Multipart\UploadState;
 
 class MultipartUploadException extends \RuntimeException
 {
     /** @var UploadState State of the erroneous transfer */
     private $state;
+    /** @var string File path of transfer object */
+    private $filePath;
 
     /**
      * @param UploadState      $state Upload state at time of the exception.
@@ -21,8 +24,10 @@ class MultipartUploadException extends \RuntimeException
             /** @var $error AwsException */
             foreach ($prev as $part => $error) {
                 $msg .= "- Part {$part}: " . $error->getMessage(). "\n";
+                $this->collectPathInfo($error->getCommand());
             }
         } elseif ($prev instanceof AwsException) {
+            $this->collectPathInfo($prev->getCommand());
             switch ($prev->getCommand()->getName()) {
                 case 'CreateMultipartUpload':
                 case 'InitiateMultipartUpload':
@@ -54,5 +59,27 @@ class MultipartUploadException extends \RuntimeException
     public function getState()
     {
         return $this->state;
+    }
+
+    /**
+     * Get file path of the transfer
+     *
+     * @return array
+     */
+    public function getFilePaths()
+    {
+        return $this->filePath;
+    }
+
+    /**
+     * Construct file path
+     *
+     * @param CommandInterface $cmd
+     */
+    private function collectPathInfo(CommandInterface $cmd)
+    {
+        if (empty($this->filePath) && isset($cmd['Bucket']) && isset($cmd['Key'])) {
+            $this->filePath = $cmd['Bucket'] . '/' . $cmd['Key'];
+        }
     }
 }
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -9,10 +9,11 @@ class MultipartUploadException extends \RuntimeException
     private $state;
 
     /**
-     * @param UploadState      $state Upload state at time of the exception.
-     * @param \Exception|array $prev  Exception being thrown.
+     * @param UploadState      $state    Upload state at time of the exception.
+     * @param \Exception|array $prev     Exception being thrown.
+     * @param array            $params   Configuration option array for MultipartUploadException
      */
-    public function __construct(UploadState $state, $prev = null) {
+    public function __construct(UploadState $state, $prev = null, array $params = []) {
         $msg = 'An exception occurred while performing a multipart upload';
 
         if (is_array($prev)) {
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -81,7 +81,7 @@ class MultipartUploadException extends \RuntimeException
      */
     private function collectPathInfo(CommandInterface $cmd)
     {
-        if (empty($this->filePath) && isset($cmd['Bucket']) && isset($cmd['Key'])) {
+        if ($this->filePath === null && isset($cmd['Bucket']) && isset($cmd['Key'])) {
             $this->filePath = $cmd['Bucket'] . '/' . $cmd['Key'];
         }
     }
--- a/src/Exception/MultipartUploadException.php
+++ b/src/Exception/MultipartUploadException.php
@@ -8,8 +8,6 @@ class MultipartUploadException extends \RuntimeException
 {
     /** @var UploadState State of the erroneous transfer */
     private $state;
-    /** @var string File path of transfer object */
-    private $filePath;
 
     /**
      * @param UploadState      $state Upload state at time of the exception.
@@ -24,10 +22,8 @@ class MultipartUploadException extends \RuntimeException
             /** @var $error AwsException */
             foreach ($prev as $part => $error) {
                 $msg .= "- Part {$part}: " . $error->getMessage(). "\n";
-                $this->collectPathInfo($error->getCommand());
             }
         } elseif ($prev instanceof AwsException) {
-            $this->collectPathInfo($prev->getCommand());
             switch ($prev->getCommand()->getName()) {
                 case 'CreateMultipartUpload':
                 case 'InitiateMultipartUpload':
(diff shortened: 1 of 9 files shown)


src/Exception/MultipartUploadException.php
```diff
@@ -55,4 +60,26 @@ public function getState()
     {
         return $this->state;
     }
+
+    /**
+     * Get file path of the transfer
+     *
+     * @return array
```
mtdowling: Shouldn't this be `string|null`?

src/Exception/MultipartUploadException.php
```diff
@@ -55,4 +60,26 @@ public function getState()
...
+     */
+    public function getFilePaths()
```
mtdowling: Singular, right? `getFilePath`

src/Exception/MultipartUploadException.php
```diff
@@ -21,8 +24,10 @@ public function __construct(UploadState $state, $prev = null) {
             /** @var $error AwsException */
             foreach ($prev as $part => $error) {
                 $msg .= "- Part {$part}: " . $error->getMessage(). "\n";
+                $this->collectPathInfo($error->getCommand());
             }
         } elseif ($prev instanceof AwsException) {
```
mtdowling: What happens if this if/else branch and the branch above do not evaluate? The filePath value is not set. What exceptions would come through here that aren't an array or AwsException? Any random exception thrown in a promise I guess?

src/Exception/MultipartUploadException.php
```diff
@@ -1,12 +1,15 @@
...
 
 class MultipartUploadException extends \RuntimeException
 {
     /** @var UploadState State of the erroneous transfer */
     private $state;
+    /** @var string File path of transfer object */
+    private $filePath;
 
     /**
      * @param UploadState      $state Upload state at time of the exception.
```
mtdowling: Side note: It's really weird that `$prev` might be an array. We should look at this and see if we can make $prev always an exception in a backwards compatible way so it's a bit more predictable.

src/Exception/MultipartUploadException.php
mtdowling: Maybe add some docs to state that the filePath may be null if the bucket and key could not be ascertained from the previous exception. Which, by the way... This class is coupling itself to S3 now even though it's meant to be generically used for both S3 and Glacier.

src/Exception/MultipartUploadException.php
```diff
@@ -55,4 +60,26 @@ public function getState()
...
+    }
+
+    /**
+     * Construct file path
+     *
+     * @param CommandInterface $cmd
+     */
+    private function collectPathInfo(CommandInterface $cmd)
+    {
+        if (empty($this->filePath) && isset($cmd['Bucket']) && isset($cmd['Key'])) {
```
mtdowling: I don't think this will work... This is now tying this class to S3, but it's also used in Glacier: https://github.com/aws/aws-sdk-php/blob/master/src/Glacier/MultipartUploader.php#L6. Maybe the filePath needs to be injected into the exception in the constructor instead of extracted from a previous exception.

mtdowling: This is still coupling the generic MultipartUploadException to S3. Can you refactor this as previously suggested?

src/Multipart/AbstractUploadManager.php
```diff
@@ -92,8 +93,11 @@ public function promise()
         if ($this->promise) {
             return $this->promise;
         }
+        $exception = $this->client->getApi()->getServiceName() === 's3'
```
mtdowling: This is coupling concrete exceptions to the abstract class. This isn't a good way to achieve what you want.

Considering this class is marked as internal, why not make this an argument passed in to the constructor instead?

src/Multipart/AbstractUploadManager.php
```diff
@@ -13,6 +13,7 @@
 use GuzzleHttp\Psr7;
 use InvalidArgumentException as IAE;
 use Psr\Http\Message\RequestInterface;
+use Aws\S3\Exception\S3MultipartUploadException;
```
mtdowling: This looks like it isn't needed

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,50 @@
...
+    public function __construct(UploadState $state, $prev = null) {
+        if (is_array($prev)) {
+            foreach ($prev as $part => $error) {
+                $this->collectPathInfo($error->getCommand());
+            }
+        } elseif ($prev instanceof AwsException) {
+            $this->collectPathInfo($prev->getCommand());
+        }
+        parent::__construct($state, $prev);
+
```
mtdowling: Extra new line

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,50 @@
...
+use Aws\Multipart\UploadState;
+
+class S3MultipartUploadException extends \Aws\Exception\MultipartUploadException
+{
+    /** @var string File path of transfer object */
+    private $filePath;
+
+    /**
+     * @param UploadState      $state Upload state at time of the exception.
+     * @param \Exception|array $prev  Exception being thrown.
```
mtdowling: This seems to be missing information about when and why an array would be provided.

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,50 @@
...
+            $this->collectPathInfo($prev->getCommand());
+        }
+        parent::__construct($state, $prev);
+
+    }
+
+    /**
+     * Get file path of S3 transfer
+     *
+     * @return string|null
```
mtdowling: Please document when or why would this be null.

src/Multipart/AbstractUploadManager.php
cjyclaire: Agreed, I added documentation and changed this to the constructor in the latest commit :)

src/Multipart/AbstractUploadManager.php
```diff
@@ -49,16 +48,21 @@
...
+    /** @var string Exception type thrown for the upload */
+    protected $exception;
+
     /**
      * @param Client $client
+     * @param string $exception
      * @param array  $config
      */
-    public function __construct(Client $client, array $config = [])
+    public function __construct(Client $client, array $config = [], $exception = null)
```
mtdowling: Why not use `$config` instead? It's a bit awkward to put a scalar argument after an associative array argument. Can you also call it `exceptionClass` if an argument or `exception_class` if a config value?

cjyclaire: Yeah, I also feel weird about leaving the scalar argument there. I hesitated to put it in `config` array originally because all options for `config` is documented [publicly](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-multipart-upload.html#configuration), and I'm not that sure whether `exception_class` parameter is appropriate (or make sense?) to document there.

I moved it to the `config` class with renaming in the latest commit :)

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,54 @@
...
+        parent::__construct($state, $prev);
+    }
+
+    /**
+     * Get file path of S3 transfer
+     *
+     * @return string|null Returns null when 'Bucket' or 'Key' information
+     *                     are unavailable.
+     */
+    public function getFilePath()
```
mtdowling: Is there a reason to not just have a getBucket and getKey method instead of returning these as a single concatenated string?

cjyclaire: Actually, no. Taking a second thought, providing `Bucket` and `Key` is actually more useful, they could be concatenated easily anyways. I changed those to `Bucket` and `Key` in the last commit :)

mtdowling:
It looks like from the referenced issue, the customer wants to know which file that was being uploaded failed to upload. It's nice to also provide the bucket and key that we were attempting to upload to, but we should also try to capture the filename of the upload. You might be able to get this from [stream metadata](https://github.com/guzzle/psr7/blob/master/src/Stream.php#L243).

cjyclaire:
@mtdowling Totally agreed on providing the filename there, and thanks for the suggestion. I made changes in the latest commit.

Yet I kind feel I'm "polluting" the constructor or `config` array in the way I'm doing right now, since `determineSource` is private within `AbstractUploader`, do you have any suggestions to make these better?

src/Multipart/AbstractUploader.php
```diff
@@ -19,6 +19,8 @@
     public function __construct(Client $client, $source, array $config = [])
     {
         $this->source = $this->determineSource($source);
+        $config['exception_params']['file_name'] = $this->source->getMetadata('uri');
```
mtdowling: I'd just push this responsibility off onto instances of MultipartUploadException. Have them pull it out of the `$prev` array just like how they're pulling key and bucket out.

`$prev` should always be full of UploadPart operations which, if I recall correctly, have a `Body` parameter that must be a stream. You could use that if present, pull out the metadata, and populate the filename.

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,87 @@
...
+     * @return string|null Returns null when 'Key' information
+     *                     is unavailable.
+     */
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    /**
+     * Get the filename of the transfer object
```
mtdowling: Maybe call this getSourceFilename to better clarify that this is the filename of the data that was being uploaded.

src/Multipart/AbstractUploader.php
cjyclaire: That's correct, and makes perfect sense to me, thanks Michael. I fixed it and did renaming in the latest commit.

I have also done some manually testing with "real" Exceptions and checked this work as expected :)

cjyclaire:
@mtdowling I have updated this PR with S3 namespace exception, it will be great if you could have another look at it when you have time : )

src/S3/Exception/S3MultipartUploadException.php
```diff
@@ -0,0 +1,86 @@
...
+     * @param \Exception|array $prev  Exception being thrown. Could be an array of
+     *                                AwsExceptions being thrown when uploading parts
+     *                                for one object, or an instance of AwsException
+     *                                for a specific Multipart error being thrown in
+     *                                the MultipartUpload process.
+     */
+    public function __construct(UploadState $state, $prev = null) {
+        if (is_array($prev)) {
+            foreach ($prev as $part => $error) {
+                $this->collectPathInfo($error->getCommand());
```
jeskew: `$this->collectPathInfo` overwrites the `$bucket`, `$key`, and `$filename` each time it is called. Since exceptions encountered when performing a multipart upload would always pertain to the same bucket and key, I don't think this is a bug so much as it is an inefficiency.

Would it make sense just have a compound condition for the if statement that ensures the array is not empty and collects path info from either the head or the tail of that array, whichever represents the most recently encountered error? E.g.:

```php
if (is_array($prev) && $error = array_pop($prev) {
    $this->collectPathInfo($error->getCommand);
}
```

cjyclaire: Make sense, I used to make it exactly like `MultipartException` scenario, taking a second look, it does provides enough info. in this way. I'll fix that : )

cjyclaire: @jeskew Actually I thought about the this behavior inefficiency there, so I did [empty check](https://github.com/aws/aws-sdk-php/pull/1140/files#diff-6eb1eb47f835189bafd547f58182f903R74) for those values in case I overwrote them. Though this still doesn't avoid unnecessary iterates.

Taking a look at [`array_pop()`](), it will remove the element from `prev`, which will leave incomplete messaging. Looks like php doesn't have something like `array_peek`? I didn't find a clean way to do `peek()`. Thoughts? : )

src/S3/MultipartUploader.php
```diff
@@ -59,6 +59,7 @@ public function __construct(
         parent::__construct($client, $source, array_change_key_case($config) + [
             'bucket' => null,
             'key'    => null,
+            'exception_class' => 'Aws\S3\Exception\S3MultipartUploadException'
```
cjyclaire: Maybe my naming here is kind of confusing? What I'm trying to use the class namespace directly like (this)[https://github.com/aws/aws-sdk-php/pull/1140/files#diff-1f37a346f0cdf4ae8ca2d6aaa33bc7a7R127], using `S3MultipartUploadException::class` doesn't seem to work.

Maybe you are suggesting better way that I did get? Could you explain a bit about it? Or perhaps I can do better naming for the parameter there? : )

src/S3/Exception/S3MultipartUploadException.php
jeskew: Ah, I missed that `$prev` gets passed on to the parent constructor. You could use [`each($prev)[1]`](https://3v4l.org/nVNSH).

src/S3/MultipartUploader.php
jeskew: If you import the exception with a `use` statement, `::class` should just expand the classname into a fully-qualified one. [Example](https://3v4l.org/LiANN)

src/S3/Exception/S3MultipartUploadException.php
cjyclaire: Awesome, thanks! I'll fix this

src/S3/MultipartUploader.php
cjyclaire: Good to know, thanks! I'll fix it

cjyclaire:
@jeskew All fixed : ) Ready for another review. Online PHP editor is fun : )

PR #1150, opened 2017-01-06
Fix sed to work on GNU

imshashank:
```
sed -i  '' -e "s/VERSION = '.*'/VERSION = '3.20.15'/" src/Sdk.php
```
The above command would fail on GNU. 

The update works fine on both GNU and OSX 10.9+

Reference:
- http://www.grymoire.com/Unix/Sed.html#uh-62h
- http://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ check-tag:
 tag: check-tag
 	@echo Tagging $(TAG)
 	chag update $(TAG)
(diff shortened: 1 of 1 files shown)


cjyclaire:
:shipit:

PR #1152, opened 2017-01-17
Fix example json

imshashank:
Fix errors generated while create PHP Docs for aws php Sdk.


cjyclaire:
:shipit: Notice: it is important that this should be fixed from root.

PR #1155, opened 2017-01-17
Skip invalid example snippets instead of breaking the doc

cjyclaire:
Considering that manually maintain and fix example file might not be a best practice, which would introduce burden into CI process as well, since having out-dated documentation is bad, this PR skips invalid example code snippets with helpful message instead of letting it stops doc generation.

@imshashank @mtdowling
#1152 

Added notes:
I do have concerns that this solution will make it easy to ignore existing example discrepancy.

Edited:
Personally, I think this PR should only be merged when there is no better option when integrate with full CI


imshashank:
This is a merge to prevent documentation from being broken. We are working on finding a solution or broken example files in the meantime.

PR #1163, opened 2017-01-27
Update s3 Client Traits to fix broken docs

imshashank:
The update will fix the documentation for s3ClientTraits.


imshashank:
@jeskew I have fixed the `@see` links. Also what other approach would you suggest, the trait and interface have all similar methods.

jeskew:
The trait and interface have *the same* methods, since the trait is used by classes implementing the interface. [PhpDocumentor's advice] is to add a `@see` annotation so that users who land on the trait documentation page can click over to the canonical documentation in the interface.

If you would prefer to duplicate the documentation in the built API docs, could you do so during the doc build process rather than by duplicating the source?

PR #1168, opened 2017-01-31
Fix Marshaler leaving dangling references

Sazpaimon:
Some PHP versions will create references to keys instead of copying them
after doing a foreach-by-reference, which can cause unexpected state
mutations. Remove those foreach-by-reference calls to prevent that from
happening

Example: https://3v4l.org/jNSeb

### DIFF
--- a/src/DynamoDb/Marshaler.php
+++ b/src/DynamoDb/Marshaler.php
@@ -283,8 +283,8 @@ class Marshaler
                 }
                 // NOBREAK: Unmarshal M the same way as L, for arrays.
             case 'L':
-                foreach ($value as &$v) {
-                    $v = $this->unmarshalValue($v, $mapAsObject);
+                foreach ($value as $k => $v) {
+                    $value[$k] = $this->unmarshalValue($v, $mapAsObject);
                 }
                 return $value;
             case 'B':
@@ -292,8 +292,8 @@ class Marshaler
             case 'SS':
             case 'NS':
             case 'BS':
-                foreach ($value as &$v) {
-                    $v = $this->unmarshalValue([$type[0] => $v]);
+                foreach ($value as $k => $v) {
+                    $value[$k] = $this->unmarshalValue([$type[0] => $v]);
                 }
                 return new SetValue($value);
         }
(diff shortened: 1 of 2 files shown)


Sazpaimon:
I normally wouldn't suggest making a change due to a bug in the language itself, but this is now the second time this code has bitten me; both times due to issues in PHP itself:

The first was when you do an array_column on a DynamoDB item after unmarshaling it (https://bugs.php.net/bug.php?id=71660)
The second was the issue from the example (https://bugs.php.net/bug.php?id=74020)

I think if I, as an individual developer, had to report two different issues to the PHP team triggered by the exact same code, there might be other landmines hiding somewhere so I figured best to avoid them altogether.

cjyclaire:
@Sazpaimon Appreciate the contribution and information! I fully understand that those language bites are unpleasant, and I heard you for those situation described. 

You code fix looks good with tests, yet I'm wondering how/ or whether should we take this, definitely `DynamoDb Marshaler` is not the only place that uses this (Such as some protocol request serializer body etc.) And I'm feeling like merging temporary language fix one by one doesn't sound like a maintainable and sustainable solution.

Thoughts?

@imshashank @jeskew @mtdowling

Sazpaimon:
Agreed, working around language bugs in userland never smells quite right, and I debated in my head for a while whether or not to patch this since the first time I got bit by this. I decided in favor of making this PR because:
1. Switching from a foreach-by-reference to a regular foreach is a trivial change.
2. I never liked foreach-by-refererence in the first place 😄 

I understand that those in and of themselves aren't compelling reasons and would be fine if you guys decide against merging this as I can still work around this bug in my own code at least until I upgrade to PHP 7.1 where this bug has a much smaller likelihood of cropping up.

mtdowling:
I tried responding to this via email, but it didn't work for some reason... Anyways, I think we can take this PR (assuming it's been reviewed by @cjyclaire), and fix the other instances of this problem as a followup to this PR. That we can unblock @Sazpaimon now, and then fix other instances of this problem.

cjyclaire:
Thank you all for inputs! I merged this and will add task in our backlog to track following fixes.

PR #1169, opened 2017-01-31
Automate changelog

imshashank:


imshashank:
@cjyclaire @jeskew Please have a look at the scripts, it's a PHP script to automate writing changelog.

All features, doc-update etc are put inside a folder ".changes/nextrelease". 

the script creates a new tag and create a json file with name x.y.z.json in folder " .changes"

It also updates the CHANGELOG.md file and fixes the endpoints file.

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+*/
+
+$dir = "../.changes/";
+$newServiceFlag = False;
+
+
+function readChangelog($dir){
+	$dir = $GLOBALS['dir']."/nextrelease/";
+	$CHANGELOG=[];
+	if (is_dir($dir)){
```
cjyclaire: Are we going to remove 'nextrelease' directory sometime? When can it not be a dir?
If that's the case, could we throw exception let release to stop instead of passing silently? (Maybe you are thinking doc only releases)?

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+	    closedir($dh);
+	  }
+	}
+	return $CHANGELOG;
+}
+
+function cleanJSON($arr){
+	$cleanedJSON=[];
+	foreach ($arr as $x){
+		if ($x->type != "docupdate"){
```
cjyclaire: Could you change those nested `if` to a cleaner way? Such as 'docupdate' continue, set flag to `true` when there is new service, and add entry for changes?

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+	      $str = file_get_contents($dir."/".$file);
+	      $CHANGELOG= array_merge($CHANGELOG, cleanJSON(json_decode($str)) );
+	    }
+	    closedir($dh);
+	  }
+	}
+	return $CHANGELOG;
+}
+
+function cleanJSON($arr){
```
jeskew: The indentation, bracket placement, and spacing should follow the [PSR-2](http://www.php-fig.org/psr/psr-2/) standard.

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+			if ($x->type == "newservice"){
+				$GLOBALS['newServiceFlag']=True;
+			}
+			array_push($cleanedJSON,$x);
+		}
+	}
+	return $cleanedJSON;
+}
+
+function createTAG($newServiceFlag){
```
jeskew: `createTag`

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+function cmp($a, $b)
+{
+    return strcmp($a->category, $b->category);
+}
+
+function generateChangelogString($CHANGELOG){
+	usort($CHANGELOG, "cmp");
```
jeskew: The `cmp` function could be an inline anonymous function.

scripts/build-changelog.php
```diff
@@ -0,0 +1,144 @@
...
+	              "http",
+	              "https"
+	            ]
+	          }';
+	$str=file_get_contents('../src/data/endpoints.json');
+	$str=str_replace("$oldMessage", "$deletedFormat",$str);
+	file_put_contents('../src/data/endpoints.json', $str);
+}
+
+buildChangelog();
```
jeskew: Per [PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#23-side-effects), code that defines symbols (like functions) should not be in the same file as code that causes side effects.

scripts/build-changelog.php
imshashank: nextrelease directory will be inside a " .changes" folder, it will have release notes for next release. I just added checks to make sure that the directory is there and it exists.  even for a doc only update the release notes will be inside the nextrelease folder.

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * @internal
+ */
+
+class ChangelogBuilder 
+{
+  
+  public $newServiceFlag = False;
+
```
cjyclaire: Instead putting all methods into this `ChangelogBuilder` with same order as script file, I'm actually suggesting having a `ChangelogBuilder` class that could be used to initialize `ChangelogBuilder` objects, perhaps a class that takes directory parameter or some option array for constructor, and then have some method to be called. You can take a look at other builder class we have in the repo, for example [this](https://github.com/aws/aws-sdk-php/blob/master/build/ServiceBuilder.php).

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+  public function readChangelog($dir)
+  {
+    $dir = $dir."nextrelease/";
```
cjyclaire: Some formatting suggestions, there should be space around '.'

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+      if ( empty($files) ) {
+        throw new \Exception("No release notes files found in $dir folder");
+      }
+      foreach ( $files as $file ) {
+        $str = file_get_contents($dir."/".$file);
+        $changelogEntries= array_merge($changelogEntries, $this->cleanJSON(json_decode($str)));
+      }
+      closedir($dh);
+    }
+    return $changelogEntries;
```
cjyclaire: For example, if we have an OO class, we could have `$changelogEntries` as a private field, once initialize, could be easily called by other methods.

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+  public function readChangelog($dir)
+  {
+    $dir = $dir."nextrelease/";
+    $changelogEntries = [];
+    if ( !is_dir($dir) || !$dh = opendir($dir) ) {
+        throw new \Exception("nextrelease directory doesn't exists or is not readable at location $dir");
+    } else {
+      //Ignore any files starting with a (.) dot 
+      $files=preg_grep('/^([^.])/', scandir($dir));
+      if ( empty($files) ) {
```
cjyclaire: Sorry perhaps my example mislead you there,  I'm actually suggesting `if (empty($files)) {`

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+
+  public function readChangelog($dir)
+  {
+    $dir = $dir."nextrelease/";
+    $changelogEntries = [];
+    if ( !is_dir($dir) || !$dh = opendir($dir) ) {
+        throw new \Exception("nextrelease directory doesn't exists or is not readable at location $dir");
+    } else {
+      //Ignore any files starting with a (.) dot 
+      $files=preg_grep('/^([^.])/', scandir($dir));
```
cjyclaire: Space around '='

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+      foreach ( $files as $file ) {
+        $str = file_get_contents($dir."/".$file);
+        $changelogEntries= array_merge($changelogEntries, $this->cleanJSON(json_decode($str)));
+      }
+      closedir($dh);
+    }
+    return $changelogEntries;
+  }
+
+  public function cleanJSON($arr)
```
cjyclaire: For those helper method, perhaps we could have it private?

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
+<?php
+namespace Aws\Test\Build\Changelog;
+
+use Aws\Build\Changelog\ChangelogBuilder;
+
```
cjyclaire: Could you move documentation block here?

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	 *
+	 */
+
+	private $RESOURCE_DIR= "tests/Build/Changelog/resources";
+
+	private function getChangelogBuilder()
+	{
+		return new ChangelogBuilder();
+	}
+
```
cjyclaire: Testing expected exceptions, you could follow: https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.expectedException

Some example in our repo: https://github.com/aws/aws-sdk-php/blob/master/tests/JsonCompilerTest.php#L25

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	public function testreadChangelogNoDirectory()
+	{
+		$this->setExpectedException(\Exception::class);
+		$obj =$this->getChangelogBuilder();
+		$result = $obj->readChangelog("wrong-folder");
+	}
+
+	public function testreadChangelogNoReleaseNotes()
```
cjyclaire: This might be same, but could you make sure that those function naming also follows camelCases?

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	{
+		$obj = $this->getChangelogBuilder();
+		$this->setExpectedException(\Exception::class);
```
cjyclaire: Same @expectedException tag suggestions as above

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+class ChangelogBuilderTest extends \PHPUnit_Framework_TestCase
+{
+	/**
+	 * @covers ChangelogBuilder
+	 *
+	 */
+
+	private $RESOURCE_DIR= "tests/Build/Changelog/resources";
+
+	private function getChangelogBuilder()
```
jeskew: Is this necessary? You could just instantiate a new ChangelogBuilder directly in the test cases.

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	{
+		return new ChangelogBuilder();
+	}
+
+	public function testreadChangelogNoDirectory()
+	{
+		$this->setExpectedException(\Exception::class);
```
jeskew: I would recommend using a more granular exception so that you don't accidentally accept an unrelated exception as a test success. I also find the `@expectedException` annotation more clear than calling `$this->setExpectedException`, but that's an optional change.

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+
+	private $RESOURCE_DIR= "tests/Build/Changelog/resources";
+
+	private function getChangelogBuilder()
+	{
+		return new ChangelogBuilder();
+	}
+
+	public function testreadChangelogNoDirectory()
```
jeskew: s/testread/testRead/

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	public function testreadChangelogNoReleaseNotes()
+	{
+		$obj = $this->getChangelogBuilder();
+		$this->setExpectedException(\Exception::class);
+		$obj->readChangelog($this->RESOURCE_DIR);
+	}
+
+	public function testreadChangelogValid()
```
jeskew: s/testread/testRead/

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	public function testcreateTagInvalidChangelog()
+	{
+		$this->setExpectedException(\Exception::class);
+		$obj = $this->getChangelogBuilder();
+		$obj->createTag(False,$this->RESOURCE_DIR . "/CHANGELOG-invalid.md");
+	}
+
+	public function testcreateTagValid()
```
jeskew: s/testcreate/testCreate/

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	{
+		$obj = $this->getChangelogBuilder();
+		$result = $obj->createTag(False,$this->RESOURCE_DIR . "/CHANGELOG-valid.md");
+		$this->assertEquals("3.21.7",$result);
+	}
+
+	public function testcreateTagValidServiceVersionBump()
```
jeskew: s/testcreate/testCreate/

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	{
+		$obj = $this->getChangelogBuilder();
+		$result = $obj->createTAG(True,$this->RESOURCE_DIR . "/CHANGELOG-valid.md");
+		$this->assertEquals("3.22.0",$result);
+	}
+
+	public function testcreateChangelogJsonValid()
```
jeskew: s/testcreate/testCreate/

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+	{
+		if ( file_exists($this->RESOURCE_DIR . "/3.21.7") ) {
+			unlink($this->RESOURCE_DIR . "/3.21.7");
+		}
+		$obj = $this->getChangelogBuilder();
+		$obj->dir = $this->RESOURCE_DIR;
+		$CHANGELOG=$obj->readChangelog($this->RESOURCE_DIR . "/.changes/");
```
jeskew: All caps should really only be used with constants

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+                "protocols": [
+                  "http",
+                  "https"
+                ],
+                "signatureVersions": [
+                  "v2"
+                ]
+              }';
+
+    $deletedFormat = '"sdb": {
```
jeskew: Did you test with https://github.com/aws/aws-sdk-php-v3-bridge to make sure the SDB compatibility package still works with this change?

build/changelog/ChangelogBuilder.php
jeskew: I agree. The base directory should remain the same throughout a changelog build, so passing it as an argument to the class's constructor would be preferable.

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+namespace Aws\Build\Changelog;
+
+/**
+ * @internal
+ */
+
+class ChangelogBuilder 
+{
+  
+  public $newServiceFlag = False;
```
jeskew: `false` should be lower-cased. You might want to run a style fixer (e.g., http://cs.sensiolabs.org/) on this file to get everything PSR-2 compliant.

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,141 @@
...
+use Aws\Build\Changelog\ChangelogBuilder;
+
+class ChangelogBuilderTest extends \PHPUnit_Framework_TestCase
+{
+	/**
+	 * @covers ChangelogBuilder
+	 *
+	 */
+
+	private $RESOURCE_DIR= "tests/Build/Changelog/resources";
```
jeskew: This test shouldn't occur within the SDK's package on disk. (The test runner might not have permission to write to this directory on all systems.) 

Use [`sys-get-temp-dir`](http://php.net/manual/en/function.sys-get-temp-dir.php) to get the system's temporary directory and then create a subdirectory therein.

build/changelog/ChangelogBuilder.php
jeskew: Also, this should be `private`

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+
+  public function readChangelog($dir)
```
jeskew: It looks like only `buildChangelog` and `fixEndpointFile` are called from outside the class. Every other method should be `private`.

build/changelog/ChangelogBuilder.php
imshashank: Any idea on how can we test private methods?

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
+<?php
+namespace Aws\Build\Changelog;
+
+/**
+ * @internal
+ */
+
```
mtdowling: No newline should be between the docblock and the class.

build/changelog/ChangelogBuilder.php
cjyclaire: Unit test will based on behaviors, not necessarily goes to one per methods. You could test different behaviors against public methods that will be called for a `ChangelogBuilder` object. Perhaps I'm missing something but I don't think `cleanJSON` is a method that have to be used publicly as a separate call every time ?

build/changelog/ChangelogBuilder.php
mtdowling: And you should use single quotes for strings that don't use interpolation.

build/changelog/ChangelogBuilder.php
mtdowling: I suggest switching all this code over to use [DirectoryIterator](http://php.net/manual/en/class.directoryiterator.php) and http://php.net/manual/en/directoryiterator.isdot.php

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  public function createTag($newServiceFlag,$changelogFile)
+  {
+    if ( !file_exists($changelogFile) ) {
+      throw new \Exception("Changelog File Not Found", 2);
+    }
+    $lines = file($changelogFile);  
+    $TAG = explode(".",explode(" ",$lines[2])[1]);
```
mtdowling: Why is this "$TAG" and not "$tag"?

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  public function cleanJSON($arr) 
+  {
+    if ( empty($arr) || !is_array($arr) ) {
+      throw new \Exception("Invalid Input", 2);
+    } else {
+        $cleanedJSON=[];
+        foreach ( $arr as $x ) {
+          if ( $x->type == "NEW_SERVICE" ) {
```
mtdowling: Seems like a switch would work better here.

build/changelog/ChangelogBuilder.php
imshashank: Wouldn't make sense as of now to use this variable as its only used one during reading the changelog from files.

tests/Build/Changelog/ChangelogBuilderTest.php
imshashank: This was made so that in future if we want to change the constructor or add more variables then we will have only one place to change.

tests/Build/Changelog/ChangelogBuilderTest.php
imshashank: Thanks for the suggestion.

However the directory will be part of the repo and hold resources like CHANGELOG-valid.md & CHANGELOG-invalid.md files. I will move all writing or editing to temp folder.

tests/Build/Changelog/ChangelogBuilderTest.php
jeskew: If the constructor signature changes, then the signature of this method likely would too, as you would need to test different values for those new constructor parameters. You would need to change all code that calls this function anyway.

build/changelog/ChangelogBuilder.php
cjyclaire: If the read changelog is only called once across all those methods, you can have it as a variable there. Yet if it will be called for multiple methods, I'd recommend having a field for it.
Of course it's up to you to take my suggestions or not.

build/changelog/ChangelogBuilder.php
imshashank: Good Idea

imshashank: The NEW_SERVICE flag is set when a new service is being released  Besides that, all other updates except "DOC_UPDATE" are pushed to the changelog. 

Please, let me know why do u thing switch case would do better here.

build/changelog/ChangelogBuilder.php
imshashank: I havent tested it with the SDB package yet. Do u think this will break that? Since this is only for changelog Automation.

jeskew: No, it looks like the output would match what's already in the endpoints.json file. Is there a reason we need to strip this out, though? The SDK will already hard-codes the signature versions it supports in the [partition file](https://github.com/aws/aws-sdk-php/blob/master/src/Endpoint/Partition.php#L136), and omitting the 'signatureVersions' block would normally mean that the default value of `['v4']` should be used, which is not accurate for SDB.

tests/Build/Changelog/ChangelogBuilderTest.php
imshashank: Done

build/changelog/ChangelogBuilder.php
imshashank: Moved the changes to build folder and passed a dir variable in the constructor

build/changelog/ChangelogBuilder.php
imshashank: Using single codes now and fixed the formatting.

tests/Build/Changelog/ChangelogBuilderTest.php
imshashank: Thanks for the suggestion. Implemented this

build/changelog/ChangelogBuilder.php
imshashank: we need to remove the signatureversion "v2" since php sdk doesn't support it. Also having this value results in bunch of tests to fail

imshashank:
@cjyclaire @mtdowling @jeskew Can u please review the code, I have made the changes requested. 

Also, this script will only be used during build only.

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,165 @@
...
+            closedir($dh);
+        }
+        return $changelogEntries;
+    }
+
+    private function cleanJSON($arr)
+    {
+        if (empty($arr) || !is_array($arr)) {
+            throw new \Exception('Invalid Input', 2);
+        } else {
```
cjyclaire: same, might not necessarily needs to be inside the else block, if exception occurs, should have been stopped anyway

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,165 @@
...
+                }
+            }
+            return $cleanedJSON;
+        }
+    }
+
+    private function createTag($changelogFile)
+    {
+        if (!file_exists($changelogFile)) {
+            throw new \Exception('Changelog File Not Found', 2);
```
cjyclaire: Same more specific exception suggestion

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,165 @@
+<?php
+namespace Aws\Build\Changelog;
+
+/**
+ * @internal
+ */
+class ChangelogBuilder
+{
+
```
cjyclaire: Could you add a bit more documentation of fields descriptions and their types?

build/changelog/build-changelog.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+require __DIR__ . '/ChangelogBuilder.php';
+
+use Aws\Build\Changelog\ChangelogBuilder;
+
+$changelogBuilder = new ChangelogBuilder("", "", true);
```
cjyclaire: Ah, interestingly parameters inputs looks like this, it's a bit weird and not self-explanatory.

You could have a param array for the constructor, which would allow more flexible changes in the future.

Or, a easier short term way, change the order for constructer, to have `$verboseFlag` go first, then others could have default value as empty string. This will make constructor looks cleaner like: `new ChangelogBuilder(true);`

build/changelog/build-changelog.php
```diff
@@ -0,0 +1,11 @@
...
+
+$changelogBuilder->buildChangelog();
+$changelogBuilder->fixEndpointFile();
+$changelogBuilder->cleanNextReleaseFolder();
```
cjyclaire: needs a new line here

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @covers ChangelogBuilder
+ *
+ */
+class ChangelogBuilderTest extends \PHPUnit_Framework_TestCase
+{
+
+    private $RESOURCE_DIR = "tests/Build/Changelog/resources/";
+
```
cjyclaire: an extra line

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,84 @@
...
+            mkdir($tempDir . ".changes/nextrelease/", 0777, true);
+        }
+        $obj = new ChangelogBuilder($tempDir, '', false);
+        touch($tempDir . ".changes/nextrelease/temp.json");
+        $this->assertEquals(1, count(preg_grep('/^([^.])/', scandir($tempDir . ".changes/nextrelease/"))));
+        $obj->cleanNextReleaseFolder();
+        $this->assertEquals(0, count(preg_grep('/^([^.])/', scandir($tempDir . ".changes/nextrelease/"))));
+    }
+}
+
```
cjyclaire: Looks like have extra new lines?

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,176 @@
...
+    /** @var boolean */
+    private $newServiceFlag = false;
+
+    /**
+     *  The constructor requires following configure parameters:
+     * - base_dir: (String) Path to the base directory where the `.changes` folder is located. Default is empty string.
+     * - release_notes_output_dir: (String) Path to `.changes` folder where new release json will be put and content of
+     *                              `nextrelease` folder will be deleted
+     * - verbose_flag: (Boolean) Flag to enable(true)/disable(false) verbose mode
+     *
```
cjyclaire: an extra line : )

build/changelog/ChangelogBuilder.php
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public function __construct($params)
```
cjyclaire: Maybe `(array $params)`?

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+namespace Aws\Test\Build\Changelog;
+
+use Aws\Build\Changelog\ChangelogBuilder;
+
+/**
+ * @covers ChangelogBuilder
+ *
```
cjyclaire: An extra line could be removed

imshashank:
@cjyclaire Made the requested changes.

cjyclaire:
LGTM :shipit:

build/changelog/ChangelogBuilder.php
```diff
@@ -6,62 +6,72 @@
...
+    public function __construct(array $params)
     {
-        $this->baseDir = $baseDir;
-        $this->releaseNotesOutputDir = $releaseNotesOutputDir;
-        $this->verboseFlag = $verboseFlag;
+        $this->params['base_dir'] = isset($params['base_dir']) ? $params['base_dir'] : '';
+        $this->params['release_notes_output_dir'] = isset($params['release_notes_output_dir'])
+            ? $params['release_notes_output_dir']
+            : '';
+        $this->params['verbose_flag'] = isset($params['verbose_flag']) ? $params['verbose_flag'] : false;
```
jeskew: Style nit: it would be better to just call this `verbose`.

build/changelog/build-changelog.php
```diff
@@ -4,8 +4,14 @@
 
 use Aws\Build\Changelog\ChangelogBuilder;
 
-$changelogBuilder = new ChangelogBuilder("", "", true);
+$params = [
+    'verbose_flag' => true
+];
+
+$changelogBuilder = new ChangelogBuilder($params);
```
jeskew: Does the build script not read any parameters from argv?

imshashank: No, doesnt need any parameters from argv

build/changelog/ChangelogBuilder.php
imshashank: But it's a flag and not a variable. So i thought it would be better naming.

jeskew: It's a flag on the command line. Inside the class, the behavior is verbose or it is not verbose.

build/changelog/build-changelog.php
jeskew: Since the builder allows users to customize its behavior from the constructor, I think the CLI should accept those inputs from the command line. It sounds like this script will currently only work if run from the correct folder, which means automation scripts would need to `cd` into the right place before executing this command and then `cd` out afterwards.

Ideally, the CLI should honor `-v`, `--outputDir`/`-o`, and `--inputDir`/`-i`.

imshashank: The default behavior is to run it from the base directory. The option for outputDir was added for unit test to test the methods and that Changelog is written at the correct place. 
The script will be called from base directory and will not need any variables to run normally.

build/changelog/ChangelogBuilder.php
```diff
@@ -6,62 +6,72 @@
...
+     * - release_notes_output_dir: (String) Path to `.changes` folder where new release json will be put and content of
+     *                              `nextrelease` folder will be deleted
+     * - verbose_flag: (Boolean) Flag to enable(true)/disable(false) verbose mode
+     */
+    public function __construct(array $params)
     {
-        $this->baseDir = $baseDir;
-        $this->releaseNotesOutputDir = $releaseNotesOutputDir;
-        $this->verboseFlag = $verboseFlag;
+        $this->params['base_dir'] = isset($params['base_dir']) ? $params['base_dir'] : '';
```
imshashank: All array properties are set in the constructor and none of the values are REQUIRED parameters here. The constructor makes sure that each key has a value when a object is instantiated .

build/changelog/build-changelog.php
jeskew: I think it's fine to make that the default behavior, but it seems odd not to expose this configurability to the command line.

build/changelog/ChangelogBuilder.php
jeskew: I think it's much clearer to bind values pulled from the parameters array to instance properties instead of putting them in an internal parameters array. Is there any advantage to using an array instead of instance properties?

build/changelog/build-changelog.php
imshashank: I can make that a todo for future. But for not we don't need this functionality as the configs are only & only used by unit tests.

build/changelog/ChangelogBuilder.php
imshashank: Using array allows to pass valued as one variable to the constructor. Also, none of these values are required and goes with the other comment on why we don't have `outputdir` etc on command line. The bind values will be set in the constructor only and constructor design will make sure that the array values are all set always.

So the values were only passed for making the unit tests work and I don't see a lot of advantage in having these values binded.

build/changelog/ChangelogBuilder.php
imshashank: Done

build/changelog/build-changelog.php
jeskew: Is adding a todo any less work than adding a call to `get_opt`? At the very least, you should accept a `-v` flag instead of always setting `verbose` to `true`. Additionally, if no option to configure inputs and outputs is offered, the todo comment should specifically call out from which directory this script expects to be run.

build/changelog/ChangelogBuilder.php
jeskew: That's an excellent reason for the constructor accepting arguments as a hash to the constructor, but there's nothing preventing you from accepting an array and saving data to instance variables. Saving this data as a `params` array makes the data this class requires less clear (i.e., you need to inspect the constructor to figure it out; you can't just look at the names of the instance variables) and also breaks IDE autocompletion.

imshashank: May be I am missing something here but binding would not make things any clearer. The way I am thinking this is that the params array will still be passed and you can look at the constructor to see what values are being set which is the same thing you would do for checking binded parameters. So I don't see any advantage in binding parameters over passing an array and then setting values.

jeskew: Think of it this way: if there are no distinct advantages to using an internal parameters array, then why do it if you know it will prevent IDEs from autocompleting these fields? I personally think it's much clearer to see:
```php
class X {
    /**
     * @var string
     */
    private $property1;

    /**
     * @var int
     */
    private $property2;

    /**
     * @var bool
     */
    private $property3;
}
```

than it is to see:
```php
class X {
    /**
     * @var array
     */
    private $properties;
}
```

In the first case, you can explicitly call out the data type and name of each expected property, and IDEs will be able to warn on type mismatches. In the second case, you're forced to inspect the constructor to see what properties are available, and no type checking can be provided automatically. We accept community contributions to all PHP files in this repository (though build tool contributions are vanishingly rare), so we shouldn't complicate the developer experience unless there's a compelling reason to do so.

Additionally, automated tooling will flag dynamically declaring a new instance property as a code smell, but cannot do so if a method adds something new to the properties array. I just do not see any advantage to using an array of properties inside the class but can identify several disadvantages. It would also be a departure from standard practice in the SDK (e.g., [AwsClient](https://github.com/aws/aws-sdk-php/blob/master/src/AwsClient.php#L153)) which is why I'd like to understand the motivation.

build/changelog/build-changelog.php
```diff
@@ -4,9 +4,15 @@
 
 use Aws\Build\Changelog\ChangelogBuilder;
 
-$params = [
-    'verbose' => true
-];
+$params = array();
+
+$options = getopt('v:');
```
jeskew: The colon is used for options that take a value. This should be `getopt('v')`

build/changelog/build-changelog.php
```diff
@@ -4,9 +4,15 @@
 
 use Aws\Build\Changelog\ChangelogBuilder;
 
-$params = [
-    'verbose' => true
-];
+$params = array();
```
jeskew: `[]` instead of `array()`

build/changelog/build-changelog.php
```diff
@@ -4,9 +4,15 @@
...
+
+$options = getopt('v:');
+
+if (isset($option['v'])) {
```
jeskew: Optional: a ternary `$params['verbose'] = isset($option['v']) ? $option['v'] : true` would be more concise.

build/changelog/ChangelogBuilder.php
```diff
@@ -132,19 +140,19 @@ private function generateChangelogString($changelog)
...
+        $changelogFile = $this->base_dir . 'CHANGELOG.md';
         $newChangelog = $this->readChangelog();
         $tag = $this->createTag($changelogFile);
         putenv('TAG=$tag');
-        if ($this->params['verbose']) {
+        if ($this->verbose) {
             echo 'Tag for next release ' . $tag . "\n";
         }
         $this->createChangelogJson($newChangelog, $tag);
         $ChangelogUpdate = $this->generateChangelogString($newChangelog);
```
jeskew: variables should be lowerCamelCase

build/changelog/ChangelogBuilder.php
```diff
@@ -6,7 +6,15 @@
  */
 class ChangelogBuilder
 {
-    /** @var array */
+    /** @var boolean */
+    private $verbose;
+
+    /** @var string */
+    private $release_notes_output_dir;
```
jeskew: properties should be lowerCamelCase

build/changelog/ChangelogBuilder.php
```diff
@@ -6,7 +6,15 @@
...
+
+    /** @var string */
+    private $base_dir;
```
jeskew: lowerCamelCase

build/changelog/build-changelog.php
imshashank: doesnt array() looks more neat

PR #1170, opened 2017-02-02
Get method fix

moulderr:


jeskew:
Hi @moulderr,

Thanks for your contribution, but it looks like the proposed change wouldn't alter the behavior of the SDK. `Aws\Result` [implements `ArrayAccess` by way of `Aws\ResultInterface`](https://github.com/aws/aws-sdk-php/blob/master/src/ResultInterface.php#L7), which means that `$this[$key]` is equivalent to calling `$this->offsetGet($key)`, [which is defined in `Aws\HasDataTrait`](https://github.com/aws/aws-sdk-php/blob/master/src/HasDataTrait.php#L26).

Did you encounter any unexpected behavior that prompted this pull request?

cjyclaire:
Closing due to ages, yet feel free to reopen with further question or details : )

PR #1171, opened 2017-02-03
Fix appstream endpoint prefix

cjyclaire:
Appstream has a customized `endpointPrefix` as `appstream2`.
This PR works as a quick fix to address the issue.

Similar issue happened with `elasticloadbalancingv2` in the past. In the long run, we might perhaps need to consider injecting `endpoint_prefix` metadata into our endpoint sourcing chain.

@imshashank @jeskew

### DIFF
--- a/src/Appstream/AppstreamClient.php
+++ b/src/Appstream/AppstreamClient.php
@@ -42,4 +42,16 @@ use Aws\AwsClient;
  * @method \Aws\Result updateStack(array $args = [])
  * @method \GuzzleHttp\Promise\Promise updateStackAsync(array $args = [])
  */
-class AppstreamClient extends AwsClient {}
+class AppstreamClient extends AwsClient {
+
+    public function __construct(array $args)
+    {
+        if (!isset($args['endpoint'])) {
+            $scheme = isset($args['scheme'])? $args['scheme'] : 'https';
+            $args['endpoint'] =
+                "{$scheme}://appstream2.{$args['region']}.amazonaws.com";
+        }
+
+        parent::__construct($args);
+    }
+}


src/Appstream/AppstreamClient.php
```diff
@@ -42,4 +42,16 @@
  * @method \Aws\Result updateStack(array $args = [])
  * @method \GuzzleHttp\Promise\Promise updateStackAsync(array $args = [])
  */
-class AppstreamClient extends AwsClient {}
+class AppstreamClient extends AwsClient {
+
+    public function __construct(array $args)
+    {
+        if (!isset($args['endpoint'])) {
```
jeskew: I think `service` rather than `endpoint` should be injected, e.g., `if (empty($args['service'])) { $args['service'] = 'appstream2'; }`. That should take care of the endpoint and transport scheme resolution.

cjyclaire: Yeah, that's true,, I also noticed the co-exist "appstream" and "appstream2", they cleared up the "appstream" 6 days ago, that's why changes are not synced. I try remove "appstream" and keep "appstream2" only in endpoint.json file. Yet it still doesn't work because we use service identifier.

Make sense, I'll make the `service` change instead.

cjyclaire: @jeskew Actually... perhaps not, when I try to change service identifier, it will fail at resolving api models:
```
Aws\Exception\UnresolvedApiException: The appstream2 service does not have version: ..
```
This is because their api model haven't been renamed to `appstream2` ...

cjyclaire:
Closing and track #1192 instead

PR #1183, opened 2017-02-16
Typo changelog builder

imshashank:
Fixed a small typo


jeskew:
This PR should be limited to the last commit. Can you create a checkout a new branch from the tip of master and cherry-pick the last commit onto it?

PR #1188, opened 2017-02-18
update from base repo, s3 service support v2  signature

4406arthur:


cjyclaire:
@4406arthur Quick question, as from the doc link, S3 doesn't listed in the services that support sigv2. So may I ask more information? Such as how you find having sigv2 in v3 useful? 

Generally, Signature V4 is highly recommend across AWS services, and we intentionally avoid V2 signing in V3, so I'm curious and happy to hear your feedback on this : )

4406arthur:
@cjyclaire Hi, I know sigv4 is great improvement, but I found lots of storage service provider which adopted s3 api gateway did not change yet..., I hope it could support downward compatibility, maybe lots of developer meet same issue.

jeskew:
@4406arthur Is there any reason this shouldn't be released as a separate package that depends on the AWS SDK? Just FYI: we have [a separate package] for SimpleDB compatibility with V3 that contains [an implementation of SigV2](https://github.com/aws/aws-sdk-php-v3-bridge/blob/master/src/Signature/SignatureV2.php). You could create a package that depends on that and adds the `S3SignatureV2` class from this PR.

I see a lot of code specific to Google Cloud Storage in this PR, which would just be added bloat for Amazon S3 users.

README.md
```diff
@@ -6,6 +6,14 @@
 [![Apache 2 License](https://img.shields.io/packagist/l/aws/aws-sdk-php.svg?style=flat)](http://aws.amazon.com/apache-2-0/)
 [![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/aws/aws-sdk-php?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
+
+##Diff
+for some endpoint not supported Signature-v4 yet.
```
jeskew: I don't believe this is accurate for any S3 endpoint. Also, this is not the appropriate place for a changelog diff.

src/Signature/S3SignatureV2.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+    public function presign(
+        \Psr\Http\Message\RequestInterface $request,
+        \Aws\Credentials\CredentialsInterface $credentials,
+        $expires
+    ) {
+        $parsed = $this->createPresignedRequest($request, $credentials);
+        //$params = Psr7\parse_query($request->getBody());
+        //Support Google Cloud Storage
+        if($parsed['uri']->getHost() === 'storage.googleapis.com'){
```
jeskew: This looks specific to Google Cloud Storage and should live in a separate package.

PR #1194, opened 2017-02-21
Chore/sync protocol tests

jeskew:
This PR syncs the SDK protocol tests. In order to get the idempotency token tests passing, I moved the `idempotency_auto_fill` option from an Ec2/Ssm customization to one that can be used with any client.

/cc @cjyclaire && @imshashank

### DIFF
--- a/src/Api/Serializer/XmlBody.php
+++ b/src/Api/Serializer/XmlBody.php
@@ -38,7 +38,7 @@ class XmlBody
         $xml = new XMLWriter();
         $xml->openMemory();
         $xml->startDocument('1.0', 'UTF-8');
-        $this->format($shape, $shape['locationName'], $args, $xml);
+        $this->format($shape, $shape['locationName'] ?: $shape['name'], $args, $xml);
         $xml->endDocument();
 
         return $xml->outputMemory();
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -185,6 +185,13 @@ class ClientResolver
             'fn'       => [__CLASS__, '_apply_user_agent'],
             'default'  => [],
         ],
+        'idempotency_auto_fill' => [
+            'type'      => 'value',
+            'valid'     => ['bool', 'callable'],
+            'doc'       => 'Set to false to disable SDK to populate parameters that enabled \'idempotencyToken\' trait with a random UUID v4 value on your behalf. Using default value \'true\' still allows parameter value to be overwritten when provided. Note: auto-fill only works when cryptographically secure random bytes generator functions(random_bytes, openssl_random_pseudo_bytes or mcrypt_create_iv) can be found. You may also provide a callable source of random bytes.',
+            'default'   => true,
+            'fn'        => [__CLASS__, '_apply_idempotency_auto_fill']
+        ],
     ];
 
     /**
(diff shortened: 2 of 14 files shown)


cjyclaire:
@jeskew Nice, I'll take a look shortly.
A quick question, initially, when introducing `idempotency_auto_fill`, I thought about having that as client wide option, yet didn't do so due to performance concerns. Since this will require additional iterates for all members. Especially when it defaults to `true`.
Ruby has this enabled as default across all client though, yet I'm not sure will this hurts the performance for PHP? Any thoughts?

jeskew:
That's a fair point, but the middleware only iterates over the top-level members of an input shape. My worry in not adding the option to `Aws\ClientResolver` is that, per the spec, the SDK should be able to support idempotency tokens in any service. Having to add a custom parameter declaration to each service client that adds idempotency tokens doesn't seem like a sustainable approach in the long run.

src/Api/Serializer/XmlBody.php
```diff
@@ -38,7 +38,7 @@ public function build(Shape $shape, array $args)
         $xml = new XMLWriter();
         $xml->openMemory();
         $xml->startDocument('1.0', 'UTF-8');
-        $this->format($shape, $shape['locationName'], $args, $xml);
+        $this->format($shape, $shape['locationName'] ?: $shape['name'], $args, $xml);
```
cjyclaire: Has this change be covered in tests?

src/ClientResolver.php
```diff
@@ -185,6 +185,13 @@ class ClientResolver
             'fn'       => [__CLASS__, '_apply_user_agent'],
             'default'  => [],
         ],
+        'idempotency_auto_fill' => [
```
cjyclaire: Maybe also some tests in `ClientResolver`?

src/IdempotencyTokenMiddleware.php
```diff
@@ -27,18 +27,30 @@ class IdempotencyTokenMiddleware
      *  - openssl_random_pseudo_bytes (requires 'openssl' module enabled)
      *  - mcrypt_create_iv (requires 'mcrypt' module enabled)
      *
+     * You may also supply a custom bytes generator as an optional second
+     * parameter.
+     *
      * @param \Aws\Api\Service $service
+     * @param callable|null $bytesGenerator
```
cjyclaire: This is great  : ) Thanks for adding this

src/IdempotencyTokenMiddleware.php
```diff
@@ -52,11 +64,12 @@ public function __invoke(
             $operation = $this->service->getOperation($command->getName());
             $members = $operation->getInput()->getMembers();
             foreach ($members as $member => $value) {
-                if (!empty($value->toArray()['idempotencyToken'])) {
+                if (isset($value->toArray()['idempotencyToken'])) {
```
cjyclaire: I thought about using `isset` originally, then it could potentially allow an empty string as input. Thoughts?

jeskew: I'll change this to just check `$value['idempotencyToken'] === true`.

src/Api/Serializer/XmlBody.php
jeskew: We don't have specific unit tests for XmlBody; this would be covered by the protocol test cases.

src/ClientResolver.php
jeskew: Will add.

PR #1199, opened 2017-03-01
Update version constraint on guzzlehttp/psr7 to allow 1.4.1

jeskew:
This change allows the SDK to pull in new versions of the Guzzle PSR7 package other than 1.4.0. It also addresses deprecation notices that arise when using `Aws\S3\PostObject` and `Aws\S3\PostObjectV4`.

Resolves #1195 

/cc @cjyclaire && @imshashank


cjyclaire:
LGTM :shipit:

PR #1204, opened 2017-03-06
Update composer.json

### LINKED ISSUE #1205: Wrong operator in composer.json
According to [the composer documentation](https://getcomposer.org/doc/articles/versions.md#range), a **double pipe** must be used to express a logical OR in version constraints.

Current version constraint for guzzle in the composer.json is

`"guzzlehttp/guzzle": "^5.3.1|^6.2.1"`

(With that single pipe, the version constraint will never resolve to ^6.2.1)

Christoph-Harms:
According to [the composer documentation](https://getcomposer.org/doc/articles/versions.md#range), a **double pipe** must be used to express a logical OR in version constraints.

Fixes #1205

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require": {
         "php": ">=5.5",
-        "guzzlehttp/guzzle": "^5.3.1|^6.2.1",
+        "guzzlehttp/guzzle": "^5.3.1||^6.2.1",
         "guzzlehttp/psr7": "^1.3.1, !=1.4.0",
         "guzzlehttp/promises": "~1.0",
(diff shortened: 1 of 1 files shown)


kstich:
Given that a single pipe still works and the associated pain points mentioned by @mtdowling in #1205, I'd move to close this as won't fix but would like to hear other thoughts.

This could also be backlogged for some future release where the composer minimum version update is less of a sticking point.

kstich:
This PR is being closed due to the discussion in #1205. It will be considered for updating in future SDK releases where requiring a newer minimum version of Composer is an easier ask for SDK users. Thank you for the contribution @Christoph-Harms!

PR #1206, opened 2017-03-06
Sig4 with Unsigned Payload

imshashank:
@mtdowling @jeskew @cjyclaire 

Can someone please review the changes.


src/Signature/SignatureV4.php
```diff
@@ -14,21 +14,27 @@ class SignatureV4 implements SignatureInterface
 {
     use SignatureTrait;
     const ISO8601_BASIC = 'Ymd\THis\Z';
+    const UNSIGNED_PAYLOAD = 'UNSIGNED-PAYLOAD';
```
jeskew: This constant is also available on [`S3SignatureV4`](https://github.com/aws/aws-sdk-php/blob/master/src/Signature/S3SignatureV4.php#L12), which extends this class. You could remove the constant from S3SignatureV4.

src/Signature/SignatureV4.php
```diff
@@ -146,6 +152,9 @@ protected function getPayload(RequestInterface $request)
 
     protected function getPresignedPayload(RequestInterface $request)
     {
+        if ($this->unsigned){
```
jeskew: This will only have an effect on presigned requests. I think you meant to put this block inside the `getPayload` method.

src/Signature/SignatureProvider.php
```diff
@@ -114,6 +114,8 @@ public static function version()
                     return $service === 's3'
                         ? new S3SignatureV4($service, $region)
                         : new SignatureV4($service, $region);
+                case 'v4-unsigned-body':
+                    return new SignatureV4($service, $region, $unsigned = true);
```
jeskew: I think the `$unsigned = ` here is unnecessary. If you feel an additional positional parameter makes the constructor less clear, it would be more in line with the rest of the SDK to add an `$options` array (containing an `'unsigned_payload' => true` member here) as the final parameter. That would allow more options to be added over time.

tests/Signature/SignatureV4Test.php
```diff
@@ -277,6 +277,32 @@ public function testProvider()
     /**
      * @dataProvider testProvider
      */
+    public function testSignRequestUnsignedPayload($req, $sreq, $creq)
```
jeskew: Could you also add a black-box test that specifies a date, credentials, service, region, and expected signature with the unsigned option set?

src/Signature/SignatureV4.php
```diff
@@ -14,21 +14,27 @@ class SignatureV4 implements SignatureInterface
...
+    /** @var bool */
+    private $unsigned;
+
     /**
      * @param string $service Service name to use when signing
      * @param string $region  Region name to use when signing
+     * @param bool $unsigned  Flag used to set unsigned or signed payload
      */
-    public function __construct($service, $region)
+    public function __construct($service, $region, $unsigned = false)
```
mtdowling: I like the idea of using an optional associative array for configuration options... Something like:

```
public function __construct($service, $region, array $options = [])
```

Where it accepts an "unsigned" key-value pair.

tests/Signature/SignatureV4Test.php
imshashank: Will this also include adding a new testprovider with 'UNSIGNED-PAYLOAD'  for testSignRequestUnsignedPayload() test and then just do a blackbox test?

jeskew: A new provider would be one way to do it. You could also modify the existing testProvider to provide an additional parameter indicating if the request should be signed.

imshashank: I am going to do this with a new provider since all signatures will need to be changed as well.

src/Signature/SignatureV4.php
imshashank: Removed it from the s3class

imshashank:
@mtdowling @jeskew @cjyclaire Made the changes requested, please have a look.

src/Signature/SignatureV4.php
```diff
@@ -14,21 +14,27 @@ class SignatureV4 implements SignatureInterface
...
      * @param string $service Service name to use when signing
      * @param string $region  Region name to use when signing
+     * @param array $options  Array of configuration options used when signing
      */
-    public function __construct($service, $region)
+    public function __construct($service, $region, array $options = [])
     {
         $this->service = $service;
         $this->region = $region;
+        $this->unsigned = isset($options['unsigned']) ? $options['unsigned'] : false;
```
jeskew: Optional: `$options['unsigned']` should always be a boolean, so you could be more explicit in what values are accepted (e.g., `$this->unsigned = isset($options['unsigned']) && $options['unsigned'] === true`)

tests/Signature/SignatureProviderTest.php
```diff
@@ -13,6 +13,7 @@ public function versionProvider()
         return [
             ['v4', 'Aws\Signature\SignatureV4', 'foo'],
             ['v4', 'Aws\Signature\S3SignatureV4', 's3'],
+            ['v4-unsigned-body', 'Aws\Signature\SignatureV4', 'foo'],
```
jeskew: Is there an `s3v4-unsigned-body` that should be checked?

imshashank:
@mtdowling @jeskew @cjyclaire Please have a look at the updated code. I have tested that the code will work.

src/Signature/SignatureProvider.php
```diff
@@ -114,6 +114,8 @@ public static function version()
                     return $service === 's3'
                         ? new S3SignatureV4($service, $region)
                         : new SignatureV4($service, $region);
+                case 'v4-unsigned-body':
+                    return new SignatureV4($service, $region, $options = ['unsigned' => 'true']);
```
cjyclaire: The `'unsigned'` parameter name sound a bit vague, is might cause the confusion that it's trying to unsign request completely. Could you rename it perhaps something like `unsign-body`?

src/Signature/SignatureV4.php
```diff
@@ -14,21 +14,27 @@ class SignatureV4 implements SignatureInterface
...
     /** @var string */
     private $region;
 
+    /** @var bool */
+    private $unsigned;
+
     /**
      * @param string $service Service name to use when signing
      * @param string $region  Region name to use when signing
+     * @param array $options Array of configuration options used when signing
```
cjyclaire: Perhaps also add some documentation for those possible values and a bit description?

tests/Signature/SignatureProviderTest.php
mtdowling: No, we are going to switch from things like s3v4 and consolidate them into just v4 and v4-unsigned-body instead.

src/Signature/SignatureV4.php
imshashank: Added more documentation

imshashank:
Updated the PR, please have a look. @mtdowling @jeskew @cjyclaire

imshashank:
@cjyclaire Can you please elaborate?

imshashank:
@cjyclaire Added a check to unsign payload only for https requests and also updated the tests.

src/Signature/SignatureV4.php
```diff
@@ -43,9 +50,13 @@ public function signRequest(
         if ($token = $credentials->getSecurityToken()) {
             $parsed['headers']['X-Amz-Security-Token'] = [$token];
         }
-
         $cs = $this->createScope($sdt, $this->region, $this->service);
         $payload = $this->getPayload($request);
+
+        if ($this->unsigned) {
```
cjyclaire: This works, yet the check here feels a bit weird, since `getPayload` has already checked both `$this->unsigned` and `http` scheme, could we check whether `$payload` is `self::UNSIGNED_PAYLOAD` instead?

tests/Signature/SignatureV4Test.php
```diff
@@ -220,6 +220,83 @@ public function testEnsuresContentSha256CanBeCalculatedWhenSeekFails()
         $signature->signRequest($request, $credentials);
     }
 
+    public function testUnsignedPayloadProvider()
```
cjyclaire: these are all test cases with `https`, maybe also add test for `http` and verify it follows regular V4

tests/AwsClientTest.php
```diff
@@ -340,6 +340,35 @@ public function testDoesNotSignOperationsWithAnAuthTypeOfNone()
...
+                    ],
+                ],
+            ],
+            [
+                'handler' => function (
+                    CommandInterface $command,
+                    RequestInterface $request
+                ) {
+                    foreach (['Authorization','X-Amz-Content-Sha256', 'X-Amz-Date'] as $signatureHeader) {
+                        $this->assertTrue($request->hasHeader($signatureHeader));
```
cjyclaire: Besides verify the header exists, perhaps also verifys it's using 'UNSIGNED-PAYLOAD'

src/Signature/SignatureV4.php
cjyclaire: Looks good, could you also point out that this relates to APIs with streaming behaviors?

src/Signature/SignatureV4.php
imshashank: This was added so that we only add the header `X-Amz-Content-Sha256` to v4-unsigned requests.
Checking if the payload is `UNSIGNED_PAYLOAD` is same as checking for the boolean flag. 

Can you please help me in understanding why would it be better to check for `self::UNSIGNED_PAYLOAD` and not the simple boolean flag.

tests/Signature/SignatureV4Test.php
imshashank: There are already cased with http in the test class.

src/Signature/SignatureV4.php
cjyclaire: using `$this->unsigned` appears simple at first glance, yet as you have also mentioned, this check in targeting at whether a request is valid for unsigned payload, so a complete check is checking the boolean and scheme as same time, which is already done so in the `getPayload`. So even this works, this check is incomplete, and a complete check will duplicate the check [here](https://github.com/aws/aws-sdk-php/pull/1206/files#diff-50459fdb298219f1d103da728ab4afb2R142). That's why I'm suggesting checking payload value directly.

tests/AwsClientTest.php
imshashank: Added additional check

imshashank:
@jeskew @cjyclaire @mtdowling 

Please have a look, have updated the code and added additional tests.

imshashank:
@jeskew @cjyclaire @mtdowling Thanks a lot for the valuable feedback and helping me in making the code better. Merging this now :-)

PR #1210, opened 2017-03-08
Replace deprecated Guzzle Uri::resolve calls

jmarien:
To avoid the "GuzzleHttp\Psr7\Uri::resolve is deprecated since version 1.4" error. Replaced with GuzzleHttp\Psr7\UriResolver::resolve.


jmarien:
I cannot get the phpunit tests running - no docs on this and no time to investigate, so closing.

Issue #1208, opened 2017-03-07
Allow start time to be specified for pre-signed URLs


rairlie:
Aws\Signature\SignatureV4 allows creation of a pre-signed URL that expires up to 7 days from now. It would be useful if rather than assuming the validity starts now, I could specify a specific time, for example so I could create a URL that's only valid from tomorrow, or valid for 24 hours from 00:00 hrs today. This would mean the returned URL would be constant for the period specified, rather than generating an entirely unique URL every time, which would allow upstream caching.

Would there be interest in this change?

imshashank:
@rairlie Please clarify is you are talking about S3 objects presigned URL.

If yes then that is the limitation of AWS S3 and the SDK just uses their API to create the presign URL. You can submit this as a feature request with the S3 team.  

Please let me know if you need help with anything.

rairlie:
@imshashank 

I'm talking about **aws-sdk-php/src/Signature/SignatureV4.php:**
```php
    public function presign(
        RequestInterface $request,
        CredentialsInterface $credentials,
        $expires
    ) {
    ...
        $httpDate = gmdate(self::ISO8601_BASIC, time());
    ...
        $parsed['query']['X-Amz-Date'] = gmdate('Ymd\THis\Z', time());
```
And also the method:
```php
    private function convertExpires($expires)
    {
    ...
        $duration = $expires - time();
```
Notice they both use `time()`. This means the URL can only be generated with a start time of 'now'. I can't create one valid from yesterday or tomorrow. This seems a restriction imposed by the SDK. There's no reason to have this limitation?

What I'm proposing is changing it so time can be passed as a parameter to presign(). I've tried this locally and it works expected - would you consider a PR?

imshashank:
@rairlie Sorry for the misunderstanding. Yes please send us a PR. 

Thanks

imshashank:
@rairlie Also I would suggest having a look at what values the AWS S3 takes.

For presign we can only provide a value of "expires-in" and that is what the PHP SDK provides.

I dont see value in providing a value which says expire 24 hours from sometime on yesterday. Also with the S3 API, there is no way to create a URL with expiration starting from tomorrow or some time in future.

http://docs.aws.amazon.com/cli/latest/reference/s3/presign.html

I hope I understood you correctly.

rairlie:
@imshashank 

> I dont see value in providing a value which says expire 24 hours from sometime on yesterday.

[...]

PR #1211, opened 2017-03-10
Allow start time to be specified when generating pre signed URL

rairlie:
This PR implements #1208

### DIFF
--- a/src/Signature/SignatureV4.php
+++ b/src/Signature/SignatureV4.php
@@ -67,19 +67,22 @@ class SignatureV4 implements SignatureInterface
     public function presign(
         RequestInterface $request,
         CredentialsInterface $credentials,
-        $expires
+        $expires,
+        $startsAt = null
     ) {
+        $startTimestamp = $startsAt ? $this->convertToTimestamp($startsAt) : time();
+
         $parsed = $this->createPresignedRequest($request, $credentials);
         $payload = $this->getPresignedPayload($request);
-        $httpDate = gmdate(self::ISO8601_BASIC, time());
+        $httpDate = gmdate(self::ISO8601_BASIC, $startTimestamp);
         $shortDate = substr($httpDate, 0, 8);
         $scope = $this->createScope($shortDate, $this->region, $this->service);
         $credential = $credentials->getAccessKeyId() . '/' . $scope;
         $parsed['query']['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256';
         $parsed['query']['X-Amz-Credential'] = $credential;
-        $parsed['query']['X-Amz-Date'] = gmdate('Ymd\THis\Z', time());
+        $parsed['query']['X-Amz-Date'] = gmdate('Ymd\THis\Z', $startTimestamp);
         $parsed['query']['X-Amz-SignedHeaders'] = 'host';
-        $parsed['query']['X-Amz-Expires'] = $this->convertExpires($expires);
+        $parsed['query']['X-Amz-Expires'] = $this->convertExpires($expires, $startTimestamp);
         $context = $this->createContext($parsed, $payload);
         $stringToSign = $this->createStringToSign($httpDate, $scope, $context['creq']);
         $key = $this->getSigningKey(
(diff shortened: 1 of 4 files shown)


src/Signature/SignatureV4.php
```diff
@@ -67,19 +67,22 @@ public function signRequest(
     public function presign(
         RequestInterface $request,
         CredentialsInterface $credentials,
-        $expires
+        $expires,
+        $startsAt = null
```
jeskew: Instead of an optional positional argument, you could use an optional array of named arguments:
```php
public function presign(<args>, array $options = [])
{
     $startTimestamp = isset($options['start_time']) ? $options['start_time'] : time();
```

imshashank:
@rairlie Have you tested a presigned URL with a start time set in future?

src/Signature/SignatureV4.php
rairlie: No problem, now changed

rairlie:
> Have you tested a presigned URL with a start time set in future?

@imshashank Yes, I've tested it. It generates the URL, and the URL becomes valid when the start time arrives. I haven't added a test for it as its no different to any other start date in that respect.

Note the V4 Signature documentation says "requests are valid within 15 minutes of the timestamp in the request", which is what I see when using this. If I generate a URL valid from 14 minutes from now, it will be immediately usable. If I generate one valid 16 mins from now, I will have to wait 1 minute before I can use it. When trying to use it before that (an S3 presigned URL), I receive:
```
<Error>
    <Code>AccessDenied</Code>
    <Message>Request is not yet valid</Message>
...
```
Which seems right to me?

Let me know if you've any other questions I've missed.

tests/Signature/S3SignatureV4Test.php
```diff
@@ -115,6 +115,18 @@ public function testAddsSecurityTokenIfPresent()
         $this->assertContains('X-Amz-Expires=518400', $url);
     }
 
+    public function testUsesStartDateIfSpecified()
+    {
+        $options = ['start_time' => $_SERVER['aws_time']];
```
jeskew: The intent of the test would be clearer if you just used a literal value here instead of reading an environment variable. You need to go up to the setup method to see that `$_SERVER['aws_time']` is set to `strtotime('December 5, 2013 00:00:00 UTC')`, which is a layer of indirection that this PR makes unnecessary.

tests/Signature/SignatureV4Test.php
```diff
@@ -132,6 +144,42 @@ public function testAddsSecurityTokenIfPresentInPresigned()
...
+        list($request, $credentials, $signature) = $this->getFixtures();
+        $credentials = new Credentials('foo', 'bar', '123');
+        $url = (string) $signature->presign($request, $credentials, 1386720000, $options)->getUri();
+        $this->assertContains('X-Amz-Date=20131205T000000Z', $url);
+        $this->assertContains('X-Amz-Expires=518400', $url);
+    }
+
+    public function testUsesStartDateFromUnixTimestampIfPresent()
+    {
+        $options = ['start_time' => $_SERVER['aws_time']];
```
jeskew: This should also just be `strtotime(...)`

src/Signature/SignatureV4.php
```diff
@@ -67,19 +67,26 @@ public function signRequest(
     public function presign(
         RequestInterface $request,
         CredentialsInterface $credentials,
-        $expires
+        $expires,
+        array $options = []
     ) {
+        if (isset($options['start_time'])) {
```
jeskew: Optional: conditional assignments like this are clearer as ternary expressions than as a series of statements.

tests/Signature/S3SignatureV4Test.php
rairlie: No problem, now changed.

src/Signature/SignatureV4.php
rairlie: Felt conditional was clearer due to line length, but happy to change.

imshashank:
LGTM

rairlie:
Thanks :thumbsup:

PR #1213, opened 2017-03-13
Replace deprecated Guzzle Uri::resolve calls

eh-steve:
Same idea as #1210 , with unit tests passing

Checks for existence of UriResolver::resolve at runtime and uses it preferentially.

A workaround until guzzle/psr7 PR is merged (https://github.com/guzzle/psr7/pull/140)

(I realise @jeskew argued against resorting to this runtime check, but some people just need this to be working again)

### DIFF
--- a/src/Api/Serializer/RestSerializer.php
+++ b/src/Api/Serializer/RestSerializer.php
@@ -193,7 +193,7 @@ abstract class RestSerializer
         // Expand path place holders using Amazon's slightly different URI
         // template syntax.
         if(class_exists('GuzzleHttp\Psr7\UriResolver')){
-            return Psr7\UriResolver::resolve($this->endpoint, $relative);
+            return Psr7\UriResolver::resolve($this->endpoint, Psr7\uri_for($relative));
         }
         return Psr7\Uri::resolve($this->endpoint, $relative);
     }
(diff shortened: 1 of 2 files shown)


eh-steve:
For the record, I agree with @jeskew that `trigger_error` is totally unnecessary for a deprecation warning which doesn't depend on detection at runtime ([given PHP's crazy handling of errors in certain contexts](https://github.com/guzzle/psr7/pull/140#issuecomment-285797624)), and that `@deprecated` is sufficient, but I also think that dependant packages have a responsibility to handle 3rd party deprecations as soon as possible.

jeskew:
What would you think about just using UriResolve and requiring version ^1.4.1 of guzzlehttp/psr7? `Aws\Api\Serializer\RestSerializer` is used by every API operation for a number of services, and I'd like to avoid runtime checks if possible.

eh-steve:
That's a solution which would work great for me personally, and keeps things more up to date, although it might result in some unresolvable composer conflicts for certain consumers (i.e. other packages which deliberately chose the opposite approach of requiring Psr7 1.3.x because 1.4.x broke things for them due to `trigger_error`).

I totally understand why we want to avoid runtime checks like this - it would be a lot easier if Psr7 just merged your PR...

Overall I'd be happy with requiring version ^1.4.1

albertocubeddu:
Hopefully this pull request will be merged ASAP.

cjyclaire:
PR #1219 containing the fix has been merged, closing. Thanks to all : )

PR #1221, opened 2017-03-20
Fix listing of root directory of bucket

helhum:
When I use `scandir('s3://bucket.name/')` the first result is an empty string,
representing the root directory of the bucket. This should rather be a dot.

This is especially required, when using `DirectoryIterator`, as an empty result
will yield this iterator to be empty as well, leading to an empty directory listing on root level.

### DIFF
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -493,9 +493,11 @@ class StreamWrapper
         $this->objectIterator->next();
 
         // Remove the prefix from the result to emulate other stream wrappers.
-        return $this->openedBucketPrefix
+        $dir = $this->openedBucketPrefix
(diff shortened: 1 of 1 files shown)


helhum:
This might not be the right place to fix it, but it illustrates what needs to be fixed.

Please guide me whether this should be adapted elsewhere of if tests need to be added to cover this case. Thanks.

kstich:
Hi @helhum, I'm having trouble reproducing the issue either through scandir or passing the path to a DirectoryIterator. Could you please provide a reproduction or additional details so we can attempt?

kstich:
It has been 30 days without a response to the question of reproduction on the issue this PR attempts to fix; the PR is being closed. If this fix is still necessary, please file an issue or resubmit the PR with more details. We hope to hear from you soon!

PR #1229, opened 2017-03-23
Json value header trait

imshashank:


imshashank:
@jeskew @mtdowling @cjyclaire Please have a look at this PR

src/Api/Parser/AbstractRestParser.php
```diff
@@ -93,6 +93,12 @@ private function extractHeader(
     ) {
         $value = $response->getHeaderLine($shape['locationName'] ?: $name);
 
+        if (!empty($shape->offsetGet('jsonvalue'))
```
cjyclaire: Any reason why not use `if($shape->offsetGet('jsonvalue'))` directly?  [`offsetGet`](https://github.com/aws/aws-sdk-php/blob/33e8f7083fcb54fd926a8569cb81a0795e8c2633/src/Api/AbstractModel.php#L30) returns value when the key is set or it will returns null, php `if` takes care of these.

src/Api/Serializer/RestSerializer.php
```diff
@@ -127,6 +127,12 @@ private function applyHeader($name, Shape $member, $value, array &$opts)
             $value = TimestampShape::format($value, 'rfc822');
         }
 
+        if (!empty($member->offsetGet('jsonvalue'))
```
cjyclaire: Same

src/Api/Serializer/RestSerializer.php
```diff
@@ -127,6 +127,12 @@ private function applyHeader($name, Shape $member, $value, array &$opts)
...
+            && $member->offsetGet('jsonvalue')
+        ) {
+            $value = base64_encode(json_encode(json_decode($value)));
```
cjyclaire: Why `decode` then `encode` there? Perhaps you are trying to expect customer provide an valid json string input? If so why not use it directly with `base64_encode`? Also, if we are relying on customers to provide a vaild JSON string, could you somehow raise error if they are providing invalid inputs? Or maybe better documentation for what's expected for those inputs?

jeskew: Why decode and then encode?

src/Api/Parser/AbstractRestParser.php
imshashank: I just wanted to make sure that the value is properly checked. WE have had bugs before where a value was set to empty strings and it resolved as true. I believe it never hurts to be too careful.

src/Api/Serializer/RestSerializer.php
imshashank: I can wrap this around an exception, is that what you meant? This would throw an exception as of now if the json string is not valid.

Also, this is the most concise way to remove all newline characters, tabs etc and make sure that the string we are encoding can be decoded back.

src/Api/Parser/AbstractRestParser.php
imshashank: I added this additional check to make sure that this would fail if for some reason the variable was set to empty string.

src/Api/Serializer/RestSerializer.php
imshashank: The same reason described above. This I found to be a concise way of removing any new line characters etc.

src/Api/Parser/AbstractRestParser.php
jeskew: Empty strings are falsy. http://php.net/manual/en/language.types.boolean.php#language.types.boolean.casting

src/Api/Serializer/RestSerializer.php
jeskew: I don't think this safely handles all data types. https://3v4l.org/L64A8

imshashank:
Updated the code, please have a look.

src/Api/Serializer/RestSerializer.php
```diff
@@ -127,6 +127,16 @@ private function applyHeader($name, Shape $member, $value, array &$opts)
             $value = TimestampShape::format($value, 'rfc822');
         }
 
+        if ($member['jsonvalue']) {
+            $value = json_encode(json_decode($value));
```
jeskew: This assumes that jsonvalue headers are being passed in as JSON-encoded strings. I don't think that's a correct assumption.

tests/Api/test_cases/protocols/input/rest-json.json
```diff
@@ -1154,6 +1154,53 @@
...
+            "shape": "InputShape"
+          },
+          "http": {
+            "method": "POST",
+            "requestUri": "/path"
+          },
+          "name": "OperationName"
+        },
+        "params": {
+          "Foo": "{\"a\":\"b\",\"c\":{\"d\":\"e\"}}"
```
jeskew: This doesn't look like the protocol test that was added to [other](https://github.com/aws/aws-sdk-js/blob/master/test/fixtures/protocol/input/rest-json.json#L1305) [SDKs](https://github.com/aws/aws-sdk-go/blob/cbdd5fa277e256b883fa15c9733f2134209ba0d0/models/protocol_tests/output/rest-json.json#L609).

src/Api/Validator.php
```diff
@@ -189,6 +189,13 @@ private function check_boolean(Shape $shape, $value)
 
     private function check_string(Shape $shape, $value)
     {
+        if ($shape['jsonvalue']
+            && !is_resource($value)
+            && (is_object($value) || is_array($value))
```
imshashank: @mtdowling Do you have any suggestions on what other checks can be added.

imshashank:
@jeskew @cjyclaire @mtdowling Please review the code. Thanks

src/Api/Validator.php
mtdowling: This isn't properly validating jsonvalue. If the shape has the jsonvalue trait, then the provided value MUST be something that can be json_encoded. Therefore, you should have a branch here that is only conditional on if the jsonvalue trait is present. Then inside of that branch, you ensure that the value can be json encoded. If it cannot, than you must add an error via `addError`. Otherwise, if the value can be json encoded, then return early.

I did a quick benchmark to see the fastest method to check if a value can be json_encoded. A function that check for the appropriate types or interfaces is almost always faster than attempting to json_encode and look for errors. Here's my gist with the example functions: https://gist.github.com/mtdowling/ed5ea633ec8358e6d2d399a8c0aa1220. I essentially whitelist specific types and have a special case for object.

src/Api/Serializer/RestSerializer.php
```diff
@@ -126,6 +126,13 @@ private function applyHeader($name, Shape $member, $value, array &$opts)
         if ($member->getType() == 'timestamp') {
             $value = TimestampShape::format($value, 'rfc822');
         }
+        if ($member['jsonvalue']) {
+            $value = json_encode($value);
+            if ($value === null && json_last_error() !== JSON_ERROR_NONE) {
+                throw new \InvalidArgumentException('Unable to encode the provided value as JSON');
```
cjyclaire: Perhaps we could call out the fact that we are using `json_encode`, which could help debugging if this cause troubles for some customer. Such as:
`Unable to encode the provided value with 'json_encode'.`

src/Api/Validator.php
cjyclaire: I'm a bit confused with those check, maybe we could just simply skip the check like: 
``` php
if ($shape['jsonvalue']) {
    return;
}
```
I think, since you already have error handling in `RestSerializer` than throws error for those objects that couldn't be JSON serialized, and the encode value would be string anyways if succeed, what's the purpose for verifying the actually object type there?

mtdowling: Validation should be isolated from serialization. If I run my input data through the validator, then it should ensure that the values I provided are valid based on the model. This is a separate process that runs before serialization. You can also get much better error messages from validation than from a serialization error because the validator tracks error context whereas serialization does not.

src/Api/Serializer/RestSerializer.php
imshashank: [PayloadParserTrait](https://github.com/aws/aws-sdk-php/blob/master/src/Api/Parser/PayloadParserTrait.php) has methods only for decoding JSON and not for encoding.

src/Api/Validator.php
```diff
@@ -257,4 +265,19 @@ private function addError($message)
             . ' '
             . $message;
     }
+
+    private function json_encode_type_check($data)
```
mtdowling: As discussed over email, "object" can also be encoded regardless of if it is JsonSerializable. So this method could really just be simplified into `!is_resource`. JSON encoding a callable just always generates "{}", so I'd be fine with allowing it.

```php
private function canJsonEncode($data)
{
    return !is_resource($data);
}
```

imshashank: Update the code.

src/Api/Serializer/RestSerializer.php
```diff
@@ -126,6 +126,15 @@ private function applyHeader($name, Shape $member, $value, array &$opts)
         if ($member->getType() == 'timestamp') {
             $value = TimestampShape::format($value, 'rfc822');
         }
+        if ($member['jsonvalue']) {
+            $value = json_encode($value);
+            if ($value === null && JSON_ERROR_NONE !== json_last_error()) {
+                throw new \ParserException('Unable to encode the provided value'
```
jeskew: Do you have any tests exercising this error path? There's no standard `ParserException` in PHP, and I don't think the SDK declares any exceptions in the root namespace.

imshashank: Have updated the exception type to: InvalidArgumentException.

This is only added as a safety net as the error will only be thrown if the value is a resource.

jeskew: If the value is a resource, `json_last_error()` will return `JSON_ERROR_UNSUPPORTED_TYPE`, but users could also encounter other errors specific to the act of encoding like `JSON_ERROR_DEPTH` or `JSON_ERROR_RECURSION`.

My point was that this PR could add a test to make sure providing invalid data for a jsonValue header throws a catchable exception. Prior to https://github.com/aws/aws-sdk-php/pull/1229/commits/93abe6baf75b7b73e1c2f7973c541fc5bb337d01, the code would have raised an uncatchable fatal error, but the tests were still green. The tests currently only exercise the happy path and not any expected error conditions.

imshashank:
@jeskew The exception in catch in serializer is only done for a sanity check. The errors are always caught up way before anything reached serializer. Also, json_encode will only fail when the value is a resource.

If this looks good, I can go ahead and merge this.

jeskew:
@imshashank There are some cases that would slip past the validator that would still cause errors in `json_encode`. You should add test cases for the error conditions this PR is explicitly handling.

imshashank:
@jeskew Added more tests to cover the error paths

tests/Api/Serializer/RestJsonSerializerTest.php
```diff
@@ -5,6 +5,8 @@
 use Aws\Command;
 use Aws\Api\Serializer\RestJsonSerializer;
 use Aws\Test\UsesServiceTrait;
+use Doctrine\Instantiator\Exception\InvalidArgumentException;
+use Nette\Neon\Exception;
```
jeskew: I don't think either of these imports are used anywhere in this file.

tests/Api/Serializer/RestJsonSerializerTest.php
```diff
@@ -91,6 +108,48 @@ public function testPreparesRequestsWithBlobButNoForcedContentType()
...
+        $jsonValueArgs = '';
+        $request = $this->getRequest('foobar', ['baz' => $jsonValueArgs]);
+        $this->assertEquals('IiI=', $request->getHeaderLine('baz'));
+        $this->assertEquals('POST', $request->getMethod());
+        $this->assertEquals('http://foo.com/', (string) $request->getUri());
+        $this->assertEquals('', $request->getHeaderLine('Content-Type'));
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
```
jeskew: Maybe this should be explicitly `\InvalidArgumentException`? I believe PHPUnit evaluates expected exceptions in the global namespace, but per the imports  this is actually expecting `\Doctrine\Instantiator\Exception\InvalidArgumentException`, which I assume your IDE imported automatically when searching for `InvalidArgumentException`;

imshashank:
@jeskew Sorry that imports were added by the IDE. Removed it and updated the code

imshashank:
Thanks for all the help and for reviewing the code. @mtdowling @cjyclaire @jeskew

PR #1231, opened 2017-03-27
Update MultiRegionClient and S3MultiRegionClient to honor middleware

### LINKED ISSUE #1207: Multipart Upload with S3MultiRegionClient empty parts
`aws/aws-sdk-php` version `3.20.3`

I have some problems with using multipart upload with multi-region client. Steps to reproduce:

1. Use `upload` method for file which > than file part option which calls multipart methods.
2. It will successfully call `createMultipartUpload` 
3. Then calling a bunch of `uploadPart` methods with success response with ETags
4. Then calling `completeMultipartUpload` with empty parts parameter and throw exception about failed validation of XML.

Why parts parameter is empty?

I fix it with `getClientFromPool` for region method `upload` wrapper above upload method of `S3MultiRegionClient`

jeskew:
This PR would resolve #1207.

The root cause of the issue was that 'regionalizing' commands within the `S3MultiRegionClient` was occurring inside the client's `executeAsync` method, where the command would be executed and, on failure, recreated using the appropriate region. The trouble with that approach was that the multipart uploader was keeping track of completed and failed parts by adding middleware to the UploadPart command's handler list, and this list was not shared between different regionalized commands.

This PR removes the custom implementation of `executeAsync` in favor of applying regionalization via a 'determine_region' middleware. MultiRegion clients now have and execute their own (usually empty) middleware stacks, the core handler of which executes the operation on the appropriate regionalized client.

/cc @mtdowling @cjyclaire @imshashank

### DIFF
--- a/src/Endpoint/Partition.php
+++ b/src/Endpoint/Partition.php
@@ -1,14 +1,16 @@
 <?php
 namespace Aws\Endpoint;
 
+use ArrayAccess;
+use Aws\HasDataTrait;
 use InvalidArgumentException as Iae;
 
 /**
  * Default implementation of an AWS partition.
  */
-final class Partition implements PartitionInterface
+final class Partition implements ArrayAccess, PartitionInterface
 {
-    private $data;
+    use HasDataTrait;
 
     /**
      * The partition constructor accepts the following options:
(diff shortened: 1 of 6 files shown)


jeskew:
@cjyclaire Yes, I have verified with a manual test. I could add a functional test to the multi-region client to verify that the `CompleteMultipartUpload` operation is sending the right data.

imshashank:
LGTM

cjyclaire:
@jeskew Sounds good to me, :shipit:

Issue #1232, opened 2017-03-28
Can't specify RequestPayer option in S3\ObjectCopier


AntoineWattier:
### Description

I came across an issue when trying to use the `ObjectCopier` with the `RequestPayer`option or any options in general.

### How to reproduce
 - one `source` bucket with the request payer option turned on and a random `file`on it
 - one `destination` bucket 

Try to copy one file from `source` to `destination` with the `ObjectCopier` class:

```php
$promise = new \Aws\S3\ObjectCopier(
    $s3Client,
    array(
        'Bucket' => 'source',
        'Key' => 'file',
    ),
    array(
        'Bucket' => 'destination',
        'Key' => 'file',
    ),
    'private',
    array(
        'RequestPayer' => 'requester',
    )
);

$promise->copy();
```
This will result in: 
```
Error executing "HeadObject" on "https://s3.amazonaws.com/source/file";
AWS HTTP error: Client error: `HEAD https://s3.amazonaws.com/source/file` resulted in a `403 Forbidden` response
```

Without the `RequestPayer` option the copy works perfectly.

### Why is it happening? 
The issue come from the fact that `ObjectCopier` doesn't specify options in the `HeadObject` call to retrieve the file metadata cf. https://github.com/aws/aws-sdk-php/blob/master/src/S3/ObjectCopier.php#L80

### Other related issue
Going further if we specify the options in the `HeadObject` call a simple copy will work, but not the Multipart one: 

```php
$promise = new \Aws\S3\ObjectCopier(
    $s3Client,
    array(
        'Bucket' => 'source',
        'Key' => 'largefile', // 10M file
    ),
    array(
        'Bucket' => 'destination',
        'Key' => 'largefile',
    ),
    'private',
    array(
         'RequestPayer' => 'requester',
         'mup_threshold' => 8388608 // 8M threshold
    )
);

$promise->copy();
```
This one will result in:
```
ERROR: An exception occurred while uploading parts to a multipart upload. The following parts had errors:
- Part 1: Error executing "UploadPartCopy" on "https://s3.amazonaws.com/destination/largefile?partNumber=1&uploadId=foo"; 
AWS HTTP error: Client error: `PUT https://s3.amazonaws.com/destination/largefile?partNumber=1&uploadId=foo` resulted in a `403 Forbidden` response
```
One more time options are not included in the commands (`UploadPartCopy` this time)
[...]

PR #1238, opened 2017-04-03
Allow request options in S3/ObjectCopier

AntoineWattier:
This PR attempts to fix https://github.com/aws/aws-sdk-php/issues/1232

`ObjectCopier` doesn't handle any option you specify.

A simple use case to reproduce this issue is to use `IfUnmodifiedSince` option with a big value(`-1000 days` for example) and see the SDK doesn't fail at the first Head request like he should do.

The changes brought by this PR are pretty simple and shouldn't affect anything else.
I decided to not validate the options to because I don't think we really need it.

Let me know what I need to do any modifications

### DIFF
--- a/src/S3/MultipartCopy.php
+++ b/src/S3/MultipartCopy.php
@@ -13,6 +13,8 @@ class MultipartCopy extends AbstractUploadManager
     private $source;
     /** @var ResultInterface */
     private $sourceMetadata;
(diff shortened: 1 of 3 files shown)


kstich:
Closing this PR due to age since response. Additionally, an update to functionality for #1232 will be included with the updates for #1312.

PR #1240, opened 2017-04-05
Preserve iterator keys in CommandPool

soukicz:
CommandPool was preserving iterator keys only in "before" section but not the others.

### DIFF
--- a/src/CommandPool.php
+++ b/src/CommandPool.php
@@ -58,7 +58,7 @@ class CommandPool implements PromisorInterface
                 if ($before) {
                     $before($command, $key);
                 }
-                yield $client->executeAsync($command);
+                yield $key => $client->executeAsync($command);
             }
         };
 
--- a/tests/CommandPoolTest.php
+++ b/tests/CommandPoolTest.php
@@ -70,6 +70,24 @@ class CommandPoolTest extends \PHPUnit_Framework_TestCase
         $this->assertSame($results, $called);
     }
 
+    public function testInvokesFulfilledKeys()
+    {
+        $results = ['A' => new Result(), 'B' => new Result()];
+        $client = $this->getTestClient('s3');
+        $this->addMockResults($client, $results);
+        $iter = [
+            'A' => $client->getCommand('HeadBucket', ['Bucket' => 'Foo']),
+            'B' => $client->getCommand('HeadBucket', ['Bucket' => 'Foo'])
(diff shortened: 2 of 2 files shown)


imshashank:
@soukicz Hey, thanks for opening this PR. can you please explain the reasoning for the PR. Was it causing you some errors?

soukicz:
If I want to (for example) download 100 000 files from S3 and process them, I can identify them by my key. It is concept from Guzzle (https://github.com/guzzle/guzzle/pull/1203) - I can process huge amounts of requests with low memory footprint.
Currently I would have to mantain second array as map from my keys to numbered array.

```php
$client = new \Aws\S3\S3Client();

$commandGenerator = function () use ($client) {
    foreach (some_query() as $data) {
        $s3Path = 'prefix/' . $data['id'];
        $command = $client->getCommand('GetObject', [
            'Bucket' => 'bucket',
            'Key' => $s3Path
        ]);
        yield $s3Path => $command;
    }
};
$commands = $commandGenerator();
$pool = new \Aws\CommandPool($client, $commands, [
    'concurrency' => 15,
    'fulfilled' => function (
        \Aws\ResultInterface $result,
        $iterKey
    ) {
        // $iterKey contains S3 path and can be used
    },
]);

$promise = $pool->promise();
$promise->wait();
```

imshashank:
@soukicz I am just concerned if this might break some of our users if they are using the integer indexes and with this change, the array would have keys instead.

soukicz:
@imshashank yeah, I thought that too. It might be one of those bugs, that are around for so long that they are considered feature :)
But it is currently inconsistent with Guzzle Promises and with is own *before* callback - https://github.com/aws/aws-sdk-php/commit/aeb44703298d0868d7a6970158ac6314d9417060#diff-2c9e9a3c5ebdbfa8c9268af99ad83f3a If someone was using *before* callback, it has different key than resolved promise.

kstich:
@soukicz Due to the long standing nature of the non-key-retaining generator, would you mind changing this to an optional `$config['preserve_iterator_keys']` that defaults to false (old behavior) and documenting the behavior to the CommandPool constructor?

rquadling:
That's a great option. I've always found it annoying to get a set of data and have to iterate to find by a key.

kstich:
Resolving in favor of #1376

PR #1241, opened 2017-04-05
Add method to get service error message if available (like v2)

imshashank:
Had a request from a customer to get a concise message like PHP SDK v2.

This is a simple PR to add a new method `getErrorMessage()`. 

Please review the PR.

@cjyclaire @jeskew

### DIFF
--- a/src/Exception/AwsException.php
+++ b/src/Exception/AwsException.php
@@ -87,7 +87,7 @@ class AwsException extends \RuntimeException
     /**
      * Get the concise error message if any.
      *
-     * @return string
+     * @return string|null
      */
     public function getErrorMessage()
     {
--- a/src/Exception/AwsException.php
+++ b/src/Exception/AwsException.php
@@ -21,6 +21,7 @@ class AwsException extends \RuntimeException
     private $errorCode;
     private $connectionError;
     private $transferInfo;
+    private $errorMessage;
 
     /**
      * @param string           $message Exception message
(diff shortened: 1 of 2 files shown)


src/Exception/AwsException.php
```diff
@@ -81,6 +85,16 @@ public function getCommand()
     }
 
     /**
+     * Get the concise error message if any.
+     *
+     * @return string|null
+     */
+    public function getErrorMessage()
```
jeskew: I think this is too close in spirit to the `getMessage` method this class is inheriting from `\Exception`. A name like `getServiceMessage` (or something similar) would be better, as it would call out why this is different from the regular exception message and why it may or may not be present on an instance of `AwsException`.

tests/Exception/AwsExceptionTest.php
```diff
@@ -55,6 +55,13 @@ public function testProvidesResponse()
         $this->assertSame($response, $e->getResponse());
     }
 
+    public function testProvidesErrorMessage()
+    {
+        $command = new Command('foo');
+        $e = new AwsException('Foo', $command, ['message' => "test error message"]);
+        $this->assertSame("test error message", $e->getErrorMessage());
```
jeskew: Might be overkill, but what you're explicitly fixing is that the concise message is not the same as the `$message` argument passed to the `AwsException` constructor ('Foo' in this case); you could add an assertion that the value returned by the new method is not `'Foo'`.

src/Exception/AwsException.php
mtdowling: Agreed. I suggest `getAwsErrorMessage()` to match the other methods.

imshashank: Thanks for the suggestions. Have changed the method name to `getAwsErrorMessage`

PR #1244, opened 2017-04-06
Update Changelog to include Docs Only Release

imshashank:
Also removed method to fix endpoints.json file as it's no longer needed.

### DIFF
--- a/build/changelog/ChangelogBuilder.php
+++ b/build/changelog/ChangelogBuilder.php
@@ -34,6 +34,11 @@ class ChangelogBuilder
         $this->verbose = isset($params['verbose']) ? $params['verbose'] : false;
     }
 
+    public function isNewService()
+    {
+        return $this->newServiceFlag;
+    }
+
     private function readChangelog()
     {
         $releaseDir = $this->baseDir . '.changes/nextrelease/';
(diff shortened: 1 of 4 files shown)


imshashank:
@cjyclaire @jeskew Can you please quick review this.

build/changelog/ChangelogBuilder.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
cjyclaire: Could you remove the extra line?

tests/Build/Changelog/ChangelogBuilderTest.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
cjyclaire: Also here?

build/changelog/ChangelogBuilder.php
```diff
@@ -67,9 +67,7 @@ private function cleanJSON($arr)
             if ($x->type == 'NEW_SERVICE') {
```
jeskew: This `for` loop isn't really doing anything anymore. You could just set `$this->newServiceFlag` to `count(array_filter($arr, function($change) { return $change->type === 'NEW_SERVICE'; }) > 0`  and return the array that was passed in.

imshashank: Good suggestion. Implemented this. Thanks

PR #1246, opened 2017-04-11
Fix dangling references

kstich:
Follow up to #1168 to clean up other foreach-by-reference locations.


imshashank:
LGTM :shipit: 

P.S. Please rebase and squash before merge.

PR #1248, opened 2017-04-12
Improve phpdoc for functions

stof:
This uses the new `iterable` pseudo-type to document the argument rather than `mixed`.

While type-hinting it is not an option (it requires PHP 7.1 as min version), using it in the phpdoc is fine (and will make the support better in IDEs supporting PHP 7.1 features as they would understand it).
I also added the phpdoc to document the variadic argument

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -24,7 +24,7 @@ function constantly($value)
 /**
  * Filters values that do not satisfy the predicate function $pred.
  *
- * @param mixed    $iterable Iterable sequence of data.
+ * @param iterable $iterable Iterable sequence of data.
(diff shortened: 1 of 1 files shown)


kstich:
While making these updates results in buildable documentation, the `iterable` pseudo-type was added in PHP 7.1. Since the SDK is compatible with PHP >= 5.5, this could cause issues for users on versions < 7.1 who don't have `iterable`.

stof:
This is only phpdoc, so it is not affecting their code. IDEs will still understand that `iterable` means `array|\Traversable`.

Issue #1255, opened 2017-04-19
Wrong Queue ARN for FIFO Queues


sleipi:
Hi, 

When I try create a new FIFO Queue in SQS with a Redirve Policy i am getting a issue with the ARN provided from the Dead Letter Queue. 

I'am Using AWS-SDK 3.25.7

```
// create Deadletter Queue
$result = $sqsClient->createQueue([
    'QueueName' => 'foo_dl.fifo',
    'Attributes' => [
        'FifoQueue' => 'true',
        'ContentBasedDeduplication' => 'false'
    ]
]);
$dlurl = $result->get('QueueUrl');
$arn = $sqsClient->getQueueArn($dlurl);
var_dump($arn);
// "arn:aws:sqs:us-east-2:************:foo_dl:fifo"

// this seems to fix the issue
// $arn = str_replace(':fifo', '.fifo', $arn);

// create queue
$result = $sqsClient->createQueue([
    'QueueName' => 'foo.fifo',
    'Attributes' => [
        'FifoQueue' => 'true',
        'ContentBasedDeduplication' => 'false',
        'RedrivePolicy' =>
            '{"deadLetterTargetArn":"' . $arn . '","maxReceiveCount":"10"}'
    ]
]);
```

It seems like the `:fifo` suffix is wrong and need to be an `.fifo`

Kind Regards

kstich:
Hi @sleipi, I've confirmed that this is a reproducible bug and will be working to solve it soon. Thank you for the detailed report!

kstich:
This fix has been merged in and will be built in to our next release, thanks for the report @sleipi!

PR #1256, opened 2017-04-19
Fix SQS' fifo Queue ARN Generation

kstich:
Cope with SQS' .fifo/:fifo ARN inconsistency, with included test.

Fix for #1255

### DIFF
--- a/src/Sqs/SqsClient.php
+++ b/src/Sqs/SqsClient.php
@@ -66,7 +66,7 @@ class SqsClient extends AwsClient
      */
     public function getQueueArn($queueUrl)
     {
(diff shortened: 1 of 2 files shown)


src/Sqs/SqsClient.php
```diff
@@ -66,13 +66,19 @@ public function __construct(array $config)
      */
     public function getQueueArn($queueUrl)
     {
-        return strtr($queueUrl, array(
+        $arn = strtr($queueUrl, array(
```
imshashank: just for clarity can you rename `arn` to `queueArn`

kstich:
@imshashank Completed that minor change.

PR #1258, opened 2017-04-20
Add CodeCov to repo

imshashank:
Adding CodeCov to the repo.

This update deprecated phpunit 4 in favor of phpunit 4.8

Please review the code @kstich @mtdowling @jeskew @cjyclaire

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "ext-json": "*",
         "ext-dom": "*",
         "ext-simplexml": "*",
-        "phpunit/phpunit": "~4.0|~5.0",
+        "phpunit/phpunit": "^4.8.35|^5.4.0",
         "behat/behat": "~3.0",
         "doctrine/cache": "~1.4",
         "aws/aws-php-sns-message-validator": "~1.0",
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ docs/venv
 docs/_themes/flask/flask_theme_support.pyc
 docs/_ext/aws/__init__.pyc
 src/package.xml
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -29,7 +29,7 @@
         "ext-json": "*",
         "ext-dom": "*",
         "ext-simplexml": "*",
-        "phpunit/phpunit": "~4.0|~5.0",
+        "phpunit/phpunit": "~4.8",
```
jeskew: 4.8 is pretty far out of date at this point and isn't guaranteed to be compatible with PHP 7 or 7.1. To make sure the tests are still compatible with all of the versions we test in Travis, I would recommend `^4.8.35|^5.4.0|^6.0.0`

jeskew: Looks like the underscore-namespaced class names were dropped from PHPUnit 6. `^4.8.35|^5.4.0` should avoid the failures that were reported for PHP 7 and 7.1. Alternatively, you could migrate to the new test class naming convention (e.g., replace `PHPUnit_Framework_TestCase` with `\PHPUnit\Framework\TestCase`).

imshashank: I'll add this task to backLog to replace PHPUnit_Framework_TestCase with \PHPUnit\Framework\TestCase.
Have updated the composer.json file with values: ^4.8.35|^5.4.0",

imshashank:
@jeskew @kstich @mtdowling @cjyclaire Please let me know if this requires any more changes?

kstich:
If we're okay with the changes to the required minimum versions of PHPUnit, this looks good to me.

Issue #1257, opened 2017-04-20
ClientResolver causes failing behaviors if globally namespaced functions have the same name as a default


bpicolo:
[Here's the line of code in question.](https://github.com/aws/aws-sdk-php/blob/514b86e97aa84cc7e30a218df4531ea758b9ba26/src/ClientResolver.php#L254)

The issue is mostly just that PHP is an insane language. Strings that are the name of some function in scope are is_callable==true, because php is happy to use strings as a function reference.

The case here is, say we've imported some file that's entirety is this:

```
<?php

function https($whatever) {
    return 'doesntmatter';
}
```

After this, trying to use SQS will hit the 'scheme' parameter, see that scheme['default'] == 'https'.
is_callable(scheme['default']) will return true, and the ClientResolver will proceed to call the globally namespaced function. And after this, SQS put URLs are suddenly `doesntmatter://aws....`.

This could feasibly (maybe?) introduce a security vulnerability if some suddenly-malicious external package decided to override https like so (either through a composer vulnerability or whatever):

```
<?php

function https($args) {
    return 'https://malicious-sqs-queue.com/?';
} 
```

That assumes the sdk isn't doing any certificate pinning, which I would imagine is the case purely because generally only browsers bother.

I'm not sure what the right fix is, because I have no idea where the sdk expects defaults to be callable. If they're actual php Closures, I suspect you could get away with either `$a['default'] instanceof Closure` or `is_callable($a['default']) && !is_string($a['default'])`.

If they're intended to be string references to functions, might be good to check if they're within the Aws namespace.

imshashank:
@bpicolo Thanks for reporting this. I will create a PR soon to fix this probable flaw. I am trying to see if there are more places where the method is_callable is used.

imshashank:
@bpicolo Hi, I have investigated this issue and there are a lot of fo places where this keyword is called. I'll apply a fix to the line specified. We are just taking the time to make sure that we don't break any of our customer's code.

imshashank:
The fix has been merged. Closing this pull request.

PR #1266, opened 2017-04-25
Validation check for is_callable

imshashank:
Fix for Issue : https://github.com/aws/aws-sdk-php/issues/1257

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -210,7 +210,9 @@ class ClientResolver
      * - default: (mixed) The default value of the argument if not provided. If
      *   a function is provided, then it will be invoked to provide a default
      *   value. The function is provided the array of options and is expected
-     *   to return the default value of the option.
+     *   to return the default value of the option. The default value can be a
+     *   closure and can not be a callable string that is not  part of the
+     *   defaultArgs array.
      * - doc: (string) The argument documentation string.
      * - fn: (callable) Function used to apply the argument. The function
      *   accepts the provided value, array of arguments by reference, and an
@@ -235,6 +237,7 @@ class ClientResolver
 
     /**
      * Resolves client configuration options and attached event listeners.
+     * Check for missing keys in passed arguments
      *
      * @param array       $args Provided constructor arguments.
      * @param HandlerList $list Handler list to augment.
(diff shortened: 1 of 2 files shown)


kstich:
Strings can be valid callables, we should see if we can perform a better check for this.

src/ClientResolver.php
```diff
@@ -252,6 +252,7 @@ public function resolve(array $args, HandlerList $list)
                 if (isset($a['default'])) {
                     // Merge defaults in when not present.
                     $args[$key] = is_callable($a['default'])
+                    && !is_string($a['default'])
```
kstich: What about `in_array('callable', $a['valid'])` instead for this check?

imshashank: That seems like a more appropriate fix.

Makefile
```diff
@@ -1,165 +0,0 @@
-help:
```
mtdowling: Why is this being removed?

src/ClientResolver.php
```diff
@@ -252,7 +252,9 @@ public function resolve(array $args, HandlerList $list)
                 if (isset($a['default'])) {
                     // Merge defaults in when not present.
                     $args[$key] = is_callable($a['default'])
-                    && !is_string($a['default'])
+                    && (in_array('callable', $a['valid'])
```
mtdowling: Let's break this out from a ternary expression into some more readable like if statements.

imshashank:
@mtdowling @kstich Please check this and let me know if this looks good. Thanks

tests/ClientResolverTest.php
```diff
@@ -167,6 +167,19 @@ public function testValidatesCallables()
...
+            'foo' => [
+                'type'   => 'value',
+                'valid'  => ['string'],
+                'default' => function (){return 'callable_test';}
+            ]
+        ]);
+        $res = $r->resolve([], new HandlerList());
+        $this->assertEquals('callable_test',$res['foo']);
+    }
+
```
kstich: Should also add a test that a string default that is a parameter-less function (say, `gc_collect_cycles`) is not invoked.

imshashank: Added more test

mtdowling:
:ship:

Issue #1242, opened 2017-04-05
Question: Why is src/functions.php excluded from PHPUnit coverage?


bradynpoulsen:
This observation was made when looking at this following configuration in the phpunit.xml.dist file https://github.com/aws/aws-sdk-php/blob/master/phpunit.xml.dist#L28

imshashank:
@bradynpoulsen I just ran a coverage with the functions.php included and the report showed 0% coverage for functions.php. 

Although all of the method from functions.php are called in other classes and test classes.

I suspect this to be the reason that the file was excluded as it is incorrectly represented in coverage report.

bradynpoulsen:
That's interesting. I added 2 functions and tests for them in PR #1243 and ran coverage with functions.php included and coverage was reported for the ones I added.

imshashank:
@bradynpoulsen Thanks a lot for your work. I see the PR and will check the code.

kstich:
It looks like many of the tests in FunctionsTest.php don't properly map to coverage, as @imshashank found, even though they have passing tests written. We could re-introduce the functions file, update the tests with `@covers` tags, and follow-up by increasing coverage with more tests. Thoughts?

bradynpoulsen:
I agree, anytime coverage can be tracked for code that would be executed in the distributed package sounds like a winner.

kstich:
This was merged a short while ago. Thanks again @bradynpoulsen!

PR #1268, opened 2017-04-25
PHPUnit coverage for functions.php

kstich:
Add PHPUnit coverage for functions.php. Clean up + mark coverage on existing tests and add several more for increased coverage.

Resolution of issue #1242

### DIFF
--- a/tests/FunctionsTest.php
+++ b/tests/FunctionsTest.php
@@ -8,6 +8,9 @@ use Aws\S3\S3Client;
 
 class FunctionsTest extends \PHPUnit_Framework_TestCase
 {
+    /**
+     * @covers Aws\recursive_dir_iterator()
+     */
     public function testCreatesRecursiveDirIterator()
     {
         $iter = Aws\recursive_dir_iterator(__DIR__);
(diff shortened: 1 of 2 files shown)


imshashank:
@kstich I can not see the modified files in the coverage report. Is there a reason that the coverage dropped? And can it be fixed?

kstich:
I'm unsure as to why codecov didn't load those; that seems like an issue on its end as opposed to with the file changes here.

The coverage itself dropped because we re-introduced a file that was previously uncovered (functions.php) which has a slightly lower coverage than the overall repository currently. There are a few small tests that I've just added, but the majority of the missing coverage is due to a tracking issue with PHPUnit and the static manifest variable in \Aws\manifest().

tests/FunctionsTest.php
```diff
@@ -72,35 +92,109 @@ public function testUsesPhpCompilationOfJsonIfPossible()
...
+
+    /**
+     * @covers Aws\default_http_handler()
+     */
+    public function testGuzzleV5HttpHandler()
+    {
+        if (!class_exists('GuzzleHttp\Ring\Core')) {
+            $this->markTestSkipped();
+        }
+        $this->assertInstanceOf(Aws\Handler\GuzzleV5\GuzzleHandler::class, Aws\default_http_handler());
```
imshashank: Split in multiple lines

PR #1279, opened 2017-05-03
Update integration scenario to not use PHP7+ functionality

kstich:
This scenario relied on the use of error_clear_last for its Given, which is only available in PHP 7.0+. Given/And clauses of the scenario have been removed, along with their context functions, and the scenario has been renamed.


imshashank:
LGTM :shipit:

Issue #1148, opened 2016-12-30
Validate SQS message attributes


bakura10:
Hi !

Currently, the SQS SDK properly validates MD5 of message payloads to make sure the message has been received correctly (https://github.com/aws/aws-sdk-php/blob/master/src/Sqs/SqsClient.php#L119).

However, since the introduction of SQS message attributes, SQS also returns back a MD5OfMessageAttributes (https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sqs-2012-11-05.html#sendmessage) which is simply not validated in the SDK (http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html#sqs-attrib-md5)

It would be nice that this validation is carried on as well :).

cjyclaire:
@bakura10 Appreciate the feedback, to clarify the feature request there, current `validateMd5()` only works for `ReceiveMessage` validation, with recent `MessageAttributes` introduced to `sendMessage`, SDK could help calculate the MD5 message digest for SQS message attributes there and helps with the validation. (Feel free to add more clarification there :))

I'll tag this as a feature request and have it in our backlog, it will be in our implementation queue soon :)

bakura10:
Hey.

I couldn't make it any clearer :D.

kstich:
Hi @bakura10, please see #1284 for an implementation of this request. Thanks!

kstich:
@bakura10 We're pleased to say that this feature request has been integrated in to the SDK, [you can find the release here.](https://github.com/aws/aws-sdk-php/releases/tag/3.28.1)

PR #1284, opened 2017-05-11
SQS Message Attribute Validation

kstich:
Validate SQS Message Attributes via MD5 with encoded input string, supplemented with tests. Checks are done in the same `validateMd5` function so users who unregister the `sqs.md5` receive no new exceptions.

PR to complete #1148

### DIFF
--- a/src/Sqs/SqsClient.php
+++ b/src/Sqs/SqsClient.php
@@ -105,7 +105,7 @@ class SqsClient extends AwsClient
     }
 
     /**
-     * Constructs the expected md5 hash of message attributes according to the encoding
+     * Calculates the expected md5 hash of message attributes according to the encoding
      * scheme detailed in SQS documentation.
      *
      * @param array $message Message containing attributes for validation.
@@ -117,7 +117,7 @@ class SqsClient extends AwsClient
      *                     attributes.
      * @link http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html#message-attributes-items-validation
      */
-    public static function getExpectedMessageAttributesMd5($message)
+    private static function calculateMessageAttributesMd5($message)
     {
         if (empty($message['MessageAttributes'])
             || !is_array($message['MessageAttributes'])) {
@@ -127,22 +127,27 @@ class SqsClient extends AwsClient
         ksort($message['MessageAttributes']);
         $attributeValues = "";
         foreach ($message['MessageAttributes'] as $name => $details) {
-            $attributeValues .= SqsClient::getEncodedStringPiece($name);
-            $attributeValues .= SqsClient::getEncodedStringPiece($details['DataType']);
+            $attributeValues .= self::getEncodedStringPiece($name);
+            $attributeValues .= self::getEncodedStringPiece($details['DataType']);
             if (substr($details['DataType'], 0, 6) === 'Binary') {
                 $attributeValues .= pack('c', 0x02);
-                $attributeValues .= SqsClient::getEncodedBinaryPiece(
+                $attributeValues .= self::getEncodedBinaryPiece(
                     $details['BinaryValue']
                 );
             } else {
                 $attributeValues .= pack('c', 0x01);
-                $attributeValues .= SqsClient::getEncodedStringPiece(
+                $attributeValues .= self::getEncodedStringPiece(
                     $details['StringValue']
                 );
             }
         }
 
         return md5($attributeValues);
+    }
+
+    private static function calculateBodyMd5($message)
+    {
+        return md5($message['body']);
     }
 
     private static function getEncodedStringPiece($piece)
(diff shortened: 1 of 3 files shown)


src/Sqs/SqsClient.php
```diff
@@ -105,7 +105,68 @@ private function queueUrl()
...
+     * @param array $message Message containing attributes for validation.
+     *                       Retrieved when using MessageAttributeNames on
+     *                       ReceiveMessage.
+     *
+     * @return string|null The md5 hash of the message attributes according to
+     *                     the encoding scheme. Returns null when there are no
+     *                     attributes.
+     * @link http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html#message-attributes-items-validation
+     */
+    public static function getExpectedMessageAttributesMd5($message)
```
cjyclaire: Maybe a better naming for this helper method? like `computeMessageAttributesMd5` or perhaps `caculateMessageAttributesMd5` or perhaps `messageAttributesMd5`? `get` feels like fetching value from response instead of constructing it. Perhaps to echo this for `MD5OfBody`, you could have another function `bodyMd5(..)` that simply returns `md5($msg['Body']` when available.

tests/Sqs/SqsClientTest.php
```diff
@@ -51,6 +51,117 @@ public function testValidatesMd5OfBody()
...
+            'region'  => 'us-west-2',
+            'version' => 'latest'
+        ]);
+
+        $mock = new Result(['Messages' => [[
+            'Body' => 'Test',
+            'MD5OfMessageAttributes' => 'foo',
+        ]]]);
+        $this->addMockResults($client, [$mock]);
+        $client->receiveMessage([
```
cjyclaire: From the feature request thread, looks like customer is looking for this functionality for `sendMessage`?

src/Sqs/SqsClient.php
cjyclaire: Any reason for this helper function being public?

tests/Sqs/SqsClientTest.php
kstich: I read the thread as 'Now that `SendMessage` allows you to set `MessageAttributes`, can we validate those on `ReceiveMessage` as well?'. @bakura10, your input here would be beneficial if possible.

Validating on `SendMessage{Batch}` could easily be a separate enhancement request if it is wanted, as the format and closure structure will need to change pretty significantly to capture the necessary data.

src/Sqs/SqsClient.php
kstich: I'm making a small set of changes along these lines for clarity, thanks for the feedback!

src/Sqs/SqsClient.php
```diff
@@ -105,7 +105,73 @@ private function queueUrl()
...
+     *
+     * @return string|null The md5 hash of the message attributes according to
+     *                     the encoding scheme. Returns null when there are no
+     *                     attributes.
+     * @link http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html#message-attributes-items-validation
+     */
+    private static function calculateMessageAttributesMd5($message)
+    {
+        if (empty($message['MessageAttributes'])
+            || !is_array($message['MessageAttributes'])) {
```
imshashank: ) { 
should be on next line

src/Sqs/SqsClient.php
```diff
@@ -105,7 +105,73 @@ private function queueUrl()
...
+                $attributeValues .= self::getEncodedStringPiece(
+                    $details['StringValue']
+                );
+            }
+        }
+
+        return md5($attributeValues);
+    }
+
+    private static function calculateBodyMd5($message)
```
imshashank: Why don't you make this a generic function and pass the "$message['Body']" variable directly to this instead of $message.

kstich: It's consistent with the `calculateMessageAttributesMd5` definition which handles the case of improperly formed 'MessageAttributes'. Since all SQS messages will have a 'Body' field, I'm okay with making this change.

src/Sqs/SqsClient.php
```diff
@@ -124,14 +190,47 @@ private function validateMd5()
                     ->then(
                         function ($result) use ($c, $r) {
                             foreach ((array) $result['Messages'] as $msg) {
+                                $bodyMd5 = self::calculateBodyMd5($msg);
```
cjyclaire: It would be great if we are checking `MD5OfMessageAttributes` and `MD5OfBody` in the same pattern here.
I'd prefer the "non-nesting if" we have for `MD5OfBody` currently:
``` php
if (isset($msg['MD5OfBody'])
    && self::calculateBodyMd5($msg) !== $msg['MD5OfBody']) ...

// Then
if (isset($msg['MD5OfMessageAttributes'])
    && self:: calculateMessageAttributesMd5($msg) !== $msg['MD5OfMessageAttributes']) ...
```

tests/Sqs/SqsClientTest.php
cjyclaire: Sounds cool, thanks for the explain : )

src/Sqs/SqsClient.php
kstich: This was because the implicit else condition of `!isset($msg['MD5OfMessageAttributes'])` is pre-confirmed on the initial if, when we check to see `isset($msg['MessageAttributes'])`.

PR #1287, opened 2017-05-16
merge Update

forkeer:


kstich:
@forkeer It looks like something broke with this PR. We'll be happy to take a look if you have changes to submit, although it might be cleaner to create a new PR.

kstich:
@forkeer We're closing this PR because there hasn't been a response to the concern about it being broken. If you do have changes to submit, please do so in a new PR.

PR #1293, opened 2017-05-23
Log on Broken Doc Shapes

kstich:
Instead of skipping Code Snippets that have issues, track them and add an option to log them out to a file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ api: api-get-apigen
 
 api-models:
 	# Build custom docs
-	php build/docs.php
+	php build/docs.php $(if $(ISSUE-LOGGING-ENABLED),--issue-logging-enabled,)
 
 redirect-map:
 	# Build redirect map
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ help:
 	@echo "  integ          to run integration tests. Provide TEST to perform a specific test."
 	@echo "  guide          to build the user guide documentation"
 	@echo "  guide-show     to view the user guide"
-	@echo "  api            to build the API documentation"
+	@echo "  api            to build the API documentation. Provide ISSUE_LOGGING_ENABLED to save build issues to file."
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -91,7 +91,7 @@ api: api-get-apigen
 
 api-models:
 	# Build custom docs
-	php build/docs.php
+	php build/docs.php $(if $(ISSUE-LOGGING-ENABLED),--issue-logging-enabled,)
```
imshashank: Can you please rename ISSUE-LOGGING-ENABLED to ISSUE_LOGGING_ENABLED

more suited for an env variable format.

build/docs/classes/DocsBuilder.php
```diff
@@ -182,6 +193,24 @@ private function updateQuickLinks(array $services)
...
+                    if (!empty($text)) {
+                        $text .= PHP_EOL;
+                    }
+                    $text .= $serviceName . '-' . $serviceVersion
+                        . ': ' . $message;
+                }
+            }
+        }
+
+        return (bool) file_put_contents("{$this->outputDir}/issues.txt", $text);
```
imshashank: would suggest better renaming like error-issues.log

build/docs/classes/DocsBuilder.php
```diff
@@ -711,17 +740,47 @@ private function createHtmlForOperation(Service $service, $name, Operation $oper
...
+                $generatorIssues
+            );
         }
 
         $html->close(); // operation-container
 
         return $html;
     }
 
+    private function logIssues($serviceName, $serviceVersion, $issuesToLog)
```
imshashank: Suggestion: Use the standard log issue format example:->
`date LEVEL Issue`

imshashank:
Is there a reason this is reducing code coverage?

kstich:
This doesn't touch any of those files, I think this is an issue with Codecov instead.

PR #1294, opened 2017-05-24
fix typo in sqs client description

ebylund:


imshashank:
@ebylund Thanks for opening this PR but the SDK classes use third person derivative language format for English comments.

Example: https://github.com/aws/aws-sdk-php/blob/master/src/S3/S3Client.php#L22

Closing this PR as it does not follow our language style used in the SDK.

Please feel free to update this thread if you think we closed this PR as a mistake.

PR #1299, opened 2017-05-30
S3 Host-Style URIs By Default

kstich:
Change the default URI to use host style over path style, including with dual-stack, with a fallback for incompatibilities. Configuration options have been added and detailed for the S3Client and its sub-command. Tests that relied on path-style but did not alter the URI have been updated to bucket style. A significant amount of tests that relied on it and altered the URI have been duplicated to verify both the host style and path style.

### DIFF
--- a/.changes/nextrelease/host_style_default.json
+++ b/.changes/nextrelease/host_style_default.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "NEW_FEATURE",
+    "category": "S3",
+    "description": "S3 calls are now done with a host style URL by default. Options for path style on the client and command levels are available as `use_path_style_endpoint` and `@use_path_style_endpoint`, respectively. [More details on the differences between the styles can be found here.](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro)"
+  }
+]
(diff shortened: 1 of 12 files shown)


msvrtan:
Hey,

Can changes like this at least get a minor version bump instead of pushing them into patch? Just updated from 3.28.1 to 3.28.6 and had my testing setup broken due to this?

kstich:
@msvrtan Thank you for bringing up the concern with the version bump; we'll discuss how to handle this size of change in the future.

What about the update broke your testing setup? Any details you have may help us to mitigate that style of issue in the future.

msvrtan:
Hi @kstich ,

On our CI we are running mock s3 on http:/localhost:3000/project-name and this change started throwing errors cause http://projectname.localhost:3000/ was not found.

As we had more then few packages updated, it was hard to locate from the error what broke it .I skipped sdk package few times as it had a patch release change and it took me few hours of updating one by one library to locate when it breaks :(

I'm not sure if this change should even trigger the major version update or just a minor one, but I really didnt expect it in patch version (as only bugfixed should be done there by semantic versioning)

kstich:
@msvrtan Appreciate the additional info, thank you! As I said above, we'll discuss how to handle this size of change in the future to hopefully alleviate this kind of issue from happening.

msvrtan:
👍

Issue #1312, opened 2017-06-19
SSEC on multipart uploads not passing parameters or encoding


gabriel403:
When trying to upload large files with SSEC I'm getting an error:

```
 (Aws\\S3\\Exception\\S3MultipartUploadException(code: 0): An exception occurred while uploading parts to a multipart upload. The following parts had errors:
- Part 1: Error executing \"UploadPart\" on \"https://s3-eu-west-1.amazonaws.com/bucket-name/folder-name/filename.csv?partNumber=1&uploadId=someid\"; AWS HTTP error: Client error: `PUT https://s3-eu-west-1.amazonaws.com/bucket-name/folder-name/filename.csv?partNumber=1&uploadId=someid` resulted in a `400 Bad Request` response:
<?xml version=\"1.0\" encoding=\"UTF-8\"?>
<Error><Code>InvalidRequest</Code><Message>The multipart upload initiate requeste (truncated...)
 InvalidRequest (client): The multipart upload initiate requested encryption. Subsequent part requests must include the appropriate encryption parameters.
```

This pull request #1291 is purported to fix this issue and when I try it, the message no longer appears. However when checking the files uploaded they definitely aren't encrypted.
Doing a dump of `$command` at this point https://github.com/aws/aws-sdk-php/blob/master/src/S3/SSECMiddleware.php#L49 shows the expected SSEC keys for the first round, but when the middleware is hit for the subsequent parts they are missing.

If I upload a small file (less than 16mb iirc), it's successfully encrypted

Sample code

```
try {
    $fileKey = 'very-large-file.csv';
    $stream = fopen('/var/tmp/file/directory/very-large-file.csv', 'rb');

    $container['s3Client']->upload(
        getenv('S3_BUCKET_FILES'),
        $fileKey,
        $stream,
        'private',
        ['params' =>
             [
                'SSECustomerAlgorithm' => "AES256",
                'SSECustomerKey'       => base64_encode(getenv('S3_ENC_PHRASE')),
                'SSECustomerKeyMD5'    => md5(base64_encode(getenv('S3_ENC_PHRASE')), true),
            ]
        ]
    );
} finally {
    fclose($stream);
}
```

kstich:
[...]

PR #1291, opened 2017-05-21
Pass params through so multipart uploads work with SSE-C

weshooper:
Fixes following:

```
[2017-05-21 07:19:50] ERROR: Aws\S3\Exception\S3MultipartUploadException: An exception occurred while uploading parts to a multipart upload. The following parts had errors:
- Part 1: Error executing "UploadPart" on "https://s3-eu-west-1.amazonaws.com/bucket-name/foobar/large-test.gz?partNumber=1&uploadId=1234.5678"; AWS HTTP error: Client error: `PUT https://s3-eu-west-1.amazonaws.com/bucket-name/foobar/large-test.gz?partNumber=1&uploadId=1234.5678` resulted in a `400 Bad Request` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRequest</Code><Message>The multipart upload initiate requeste (truncated...)
 InvalidRequest (client): The multipart upload initiate requested encryption. Subsequent part requests must include the appropriate encryption parameters. - <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRequest</Code><Message>The multipart upload initiate requested encryption. Subsequent part requests must include the appropriate encryption parameters.</Message><RequestId>ABC123</RequestId><HostId>BLAH</HostId></Error>
```

PR #1313, opened 2017-06-22
Bug Fix + Functionality Updates for MultipartUploader/Copy and ObjectUploader/Copier

kstich:
Added 'params' passthrough for MultipartUploader/Copy. Increased functionality of ObjectUploader/Copier with regards to 'params'. Added 'before_lookup' option to ObjectCopier to operate on its 'HeadObject' command.

Addresses #1312 / #1291 and #1232 / #1238 in a consistent way that upgrades functionality while maintaining the contract established by each high-level abstraction.

### DIFF
--- a/src/S3/MultipartCopy.php
+++ b/src/S3/MultipartCopy.php
@@ -35,6 +35,11 @@ class MultipartCopy extends AbstractUploadManager
      * - concurrency: (int, default=int(5)) Maximum number of concurrent
      *   `UploadPart` operations allowed during the multipart upload.
      * - key: (string, required) Key to use for the object being uploaded.
+     * - params: (array) An array of key/value parameters that will be applied
+     *   to each of the sub-commands run by the uploader as a base.
+     *   Auto-calculated options will override these parameters. If you need
+     *   more granularity over parameters to each sub-command, use the before_*
+     *   options detailed above to update the commands directly.
      * - part_size: (int, default=int(5242880)) Part size, in bytes, to use when
      *   doing a multipart upload. This must between 5 MB and 5 GB, inclusive.
      * - state: (Aws\Multipart\UploadState) An object that represents the state
@@ -107,19 +112,27 @@ class MultipartCopy extends AbstractUploadManager
 
     private function createPart($partNumber, $partsCount)
     {
+        $data = [];
+
+        // Apply custom params to UploadPartCopy data
+        $config = $this->getConfig();
+        $params = isset($config['params']) ? $config['params'] : [];
+        foreach ($params as $k => $v) {
+            $data[$k] = $v;
+        }
+
+        $data['CopySource'] = $this->source;
+        $data['PartNumber'] = $partNumber;
+
         $defaultPartSize = $this->determinePartSize();
         $startByte = $defaultPartSize * ($partNumber - 1);
-        $partSize = $partNumber < $partsCount
+        $data['ContentLength'] = $partNumber < $partsCount
             ? $defaultPartSize
             : $this->getSourceSize() - ($defaultPartSize * ($partsCount - 1));
-        $endByte = $startByte + $partSize - 1;
+        $endByte = $startByte + $data['ContentLength'] - 1;
+        $data['CopySourceRange'] = "bytes=$startByte-$endByte";
 
-        return [
-            'ContentLength' => $partSize,
-            'CopySource' => $this->source,
-            'CopySourceRange' => "bytes=$startByte-$endByte",
-            'PartNumber' => $partNumber,
-        ];
+        return $data;
     }
 
     protected function extractETag(ResultInterface $result)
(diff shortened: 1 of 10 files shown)


src/S3/MultipartCopy.php
```diff
@@ -107,19 +112,27 @@ protected function getUploadCommands(callable $resultHandler)
 
     private function createPart($partNumber, $partsCount)
     {
+        $data = [];
+
+        // Apply custom params to UploadPartCopy data
+        $config = $this->getConfig();
+        $params = isset($config['params']) ? $config['params'] : [];
+        foreach ($params as $k => $v) {
+            $data[$k] = $v;
```
jeskew: One potential problem with this approach is that it assumes parameters always have the same form and semantic meaning for each constituent command of a multipart operation. I believe that's true now (though I can't say for sure without looking at every parameter used by {Create,Complete,Abort}MultipartUpload/UploadPart), but is that something the can be relied on indefinitely? There's nothing in S3's API structure that guarantees that, which is I believe why multipart uploads must today be customized [by supplying `'before_X'` callbacks](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-multipart-upload.html#customizing-a-multipart-upload).

Take the following scenario: Say S3 adds a `foo` parameter to `CreateMultipartUpload` that takes a structure. Assume that it's a new kind of tagging. A couple months later, they release a feature that lets you apply additional `foo` to the individual parts of a multipart upload. Because `UploadPart` is a streaming operation, `foo` must be sent as a header, which means the SDK must receive it as a serialized string (like how the SDK takes `Tagging` as a string in [`PutObject`](http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3-2006-03-01.html#putobject) (where it's a header) and as a structure in [`PutObjectTagging`](http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3-2006-03-01.html#putobjecttagging) (where it's serialized as part of the request body)). After merging in this PR, that would be a breaking change for the PHP SDK, whereas today it wouldn't be.

tests/S3/ObjectCopierTest.php
```diff
@@ -105,51 +105,64 @@ public function testDoesCorrectOperationAsynchronouslyWithPathStyle(
...
 
-    public function getCopyTestCases()
+    private function getSmallPutObjectMockResult()
     {
         $smallHeadObject = new Result(['ContentLength' => 1024 * 1024 * 6]);
         $putObject = new Result();
+
+        return [$smallHeadObject, $putObject];
+    }
+    private function getMultipartMockResults()
```
jeskew: Not a blocker, but there should be an empty line between this method and the previous one.

tests/S3/ObjectCopierTest.php
```diff
@@ -105,51 +105,64 @@ public function testDoesCorrectOperationAsynchronouslyWithPathStyle(
...
         $putPart = new Result(['ETag' => 'bar']);
         $complete = new Result(['Location' => 'https://bucket.s3.amazonaws.com/key']);
 
+        return array_merge(
+            [$smallHeadObject, $initiate],
+            array_fill(0, $partCount, $putPart),
+            [$complete]
+        );
+    }
+    public function getCopyTestCases()
```
jeskew: Not a blocker, but missing preceding new line (as above).

tests/S3/ObjectCopierTest.php
```diff
@@ -105,51 +105,64 @@ public function testDoesCorrectOperationAsynchronouslyWithPathStyle(
...
         $putPart = new Result(['ETag' => 'bar']);
         $complete = new Result(['Location' => 'https://s3.amazonaws.com/bucket/key']);
 
+        return array_merge(
+            [$smallHeadObject, $initiate],
+            array_fill(0, $partCount, $putPart),
+            [$complete]
+        );
+    }
+    public function getCopyTestCasesWithPathStyle()
```
jeskew: Not a blocker: missing new line

src/S3/ObjectCopier.php
```diff
@@ -74,8 +77,15 @@ public function __construct(
     public function promise()
     {
         return \GuzzleHttp\Promise\coroutine(function () {
+            $headObjectCommand = $this->client->getCommand(
+                'HeadObject',
+                $this->options['params'] + $this->source
+            );
+            if (is_callable($this->options['before_lookup'])) {
```
imshashank: Wouldn't this change behavior or existing clients?

src/S3/MultipartCopy.php
kstich: Any granularity on customizing these operations should still be handled through the 'before_X' callbacks, including the new `before_lookup` in ObjectCopier. The changes around 'params' here are to build out handling the contracts in the ObjectUploader/ObjectCopier and also allow ease of use for the parameters that are consistent across the operations.

src/S3/ObjectCopier.php
kstich: Yes, but only to fulfill what was previously defined and handled in a broken and inconsistent manner. See #1232

imshashank: Fari enough. Please update the changelog message to reflect that this might change the behavior of the client.

imshashank:
LGTM

PR #1314, opened 2017-06-23
SSECMiddleware fix in order to use a custom base64-encoded encryption…

dioogo91:
The previous implementation only accepted printable ASCII characters as the password for the SSE encryption. This reduces the password strength since it is not possible to use the full 256 values per byte, and only around 95 (printable ascii) values, effectively limiting the possible password entropy. This solution accepts both options: a custom base64-encoded encryption key or a 32 characters string.

reference: https://en.wikipedia.org/wiki/Password_strength#Random_passwords

### DIFF
--- a/src/S3/SSECMiddleware.php
+++ b/src/S3/SSECMiddleware.php
@@ -63,7 +63,12 @@ class SSECMiddleware
     {
         // Base64 encode the provided key
         $key = $command[$prefix . 'SSECustomerKey'];
-        $command[$prefix . 'SSECustomerKey'] = base64_encode($key);
+        if(base64_encode(base64_decode($key, true)) === $key) {
+            $command[$prefix . 'SSECustomerKey'] = $key;
+            $key = base64_decode($key, true);
+        } else {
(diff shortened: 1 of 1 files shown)


imshashank:
@dioogo91 Will this change behavior of existing clients?

dioogo91:
@imshashank I closed the pull request because there's no way to **safely guarantee** that some string is base 64 encoded. The code I used only guarantees that it **can** be base 64 encoded.
Nonetheless I still think the SSECMiddleware should also allow the client the power to choose a custom base 64 encoded encryption key because of what I said on my first comment:

> This reduces the password strength since it is not possible to use the full 256 values per byte, and only around 95 (printable ascii) values, effectively limiting the possible password entropy.

Maybe another SSE Parameter that can be used as a flag to choose between the two choices. I'm working on it right now.

dioogo91:
@imshashank to answer your question: it would change if the old SSE keys used would pass through base64_decode PHP function as a valid value. I think the solution I presented to you on my previous comment wouldn't change behaviour and would be painless to use it with old and new clients simultaneously.

PR #1317, opened 2017-06-30
Added Athena recently supported regions

match:
Added eu-west-1 (Ireland), ap-northeast-1 (Tokyo) and ap-southeast-1 (Singapore) to the Athena endpoints definition.


imshashank:
@match Thanks for opening this PR, but I will close this PR as the regions were updated in the endpoints file by the Athena team. 

We apologize for the delay in updating the endpoints file. We will make sure that we update these endpoints as soon as they become GA.

PR #1298, opened 2017-05-27
Fix for upcoming PHP 7.2

remicollet:
Without this:
```
$ php72 vendor/bin/phpunit  -d memory_limit=1G  --testsuite=unit --no-coverage
PHPUnit 5.7.20 by Sebastian Bergmann and contributors.

.............................................................   61 / 2121 (  2%)
.............................................................  122 / 2121 (  5%)
.............................................................  183 / 2121 (  8%)
.............................................................  244 / 2121 ( 11%)
.............................................................  305 / 2121 ( 14%)
.............................................................  366 / 2121 ( 17%)
.................................................SSSSSSSSSSSS  427 / 2121 ( 20%)
SSSSSSSSSSSSSSSSSS..........................................E  488 / 2121 ( 23%)
...............................E.............................  549 / 2121 ( 25%)
......................EEE....................................  610 / 2121 ( 28%)
.............................................................  671 / 2121 ( 31%)
.............................................................  732 / 2121 ( 34%)
.............................................................  793 / 2121 ( 37%)
.............................................................  854 / 2121 ( 40%)
.............................................................  915 / 2121 ( 43%)
.............................................................  976 / 2121 ( 46%)
............................................................. 1037 / 2121 ( 48%)
.........................S..................SSSSSS........... 1098 / 2121 ( 51%)
..........E.................................................. 1159 / 2121 ( 54%)
............................................................. 1220 / 2121 ( 57%)
............................................................. 1281 / 2121 ( 60%)
............................................E................ 1342 / 2121 ( 63%)
............................................................. 1403 / 2121 ( 66%)
............................................................. 1464 / 2121 ( 69%)
............................................................. 1525 / 2121 ( 71%)
............................................................. 1586 / 2121 ( 74%)
............................................................. 1647 / 2121 ( 77%)
............................................................. 1708 / 2121 ( 80%)
............................................................. 1769 / 2121 ( 83%)
............................................................. 1830 / 2121 ( 86%)
............................................................. 1891 / 2121 ( 89%)
............................................................. 1952 / 2121 ( 92%)
...............E............................................. 2013 / 2121 ( 94%)
............................................................. 2074 / 2121 ( 97%)
...............................................               2121 / 2121 (100%)

Time: 3.24 seconds, Memory: 34.01MB

There were 8 errors:

1) Aws\Test\Credentials\CredentialProviderTest::testPersistsToCache
count(): Parameter must be an array or an object that implements Countable

[...]

PR #1318, opened 2017-06-30
Fix for upcoming PHP 7.2 Supplement

kstich:
fix count(): Parameter must be an array or an object that implements Countable

Supplemental PR for #1298


kstich:
PHP 7.2 is forked and feature frozen, and building the SDK on beta1 with these combined fixes is successful.

Issue #1319, opened 2017-07-05
Documentation is out of date


TakesTheBiscuit:
See here:
http://docs.aws.amazon.com/aws-sdk-php/v2/guide/service-s3.html#creating-a-pre-signed-url

`createPresignedUrl` doesn't exist / isn't valid

`Guzzle\Http` is now `GuzzleHttp\`

Probably many other issues, a bit confusing in terms of quick-start access which i was using this document for.

kstich:
Hi @TakesTheBiscuit, it looks like you're viewing the v2 guide. [Here's the `createPresignedUrl` guide for v3.](http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-presigned-url.html)

TakesTheBiscuit:
Appreciate that -- the public links on the front end link there (to v2), i didn't notice anything that said V2 until you just mentioned it, and then it's only in the address bar / URI -- ideally the document would say on it V2? 
<img width="1096" alt="screen shot 2017-07-05 at 16 47 17" src="https://user-images.githubusercontent.com/15122732/27872725-a1c922ae-61a1-11e7-92c1-3b6d0ef5e627.png">

kstich:
Do you have a page that directed you to the v2 docs? We can look in to what's causing that and potentially improving the messaging around v2 -> v3 docs as well.

kstich:
#1320 has been merged in.

PR #1320, opened 2017-07-05
Minor README.md changes

TakesTheBiscuit:
Very minor changes to README.md - see https://github.com/aws/aws-sdk-php/issues/1319#issuecomment-313144883 
Because of the layout and white spacing in this file the prominent links at the bottom of the list were for V2, this should guide people more clearly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,17 +15,14 @@ or by downloading a single zip or phar file from our [latest release][latest-rel
 
 ## Resources
 
-* [User Guide][docs-guide] – For both getting started and in-depth **current** SDK usage information
-* [API Docs][docs-api] – For details about operations, parameters, and responses on **current** SDK version
+* [User Guide][docs-guide] – For both getting started and in-depth SDK usage information
+* [API Docs][docs-api] – For details about operations, parameters, and responses
 * [Blog][sdk-blog] – Tips & tricks, articles, and announcements
 * [Sample Project][sdk-sample] - A quick, sample project to help get you started
 * [Forum][sdk-forum] – Ask questions, get help, and give feedback
 * [Issues][sdk-issues] – Report issues, submit pull requests, and get involved
   (see [Apache 2.0 License][sdk-license])
 * [@awsforphp][sdk-twitter] – Follow us on Twitter
-* For **the previous** version 2 of the SDK:
-  * [User Guide][docs-guide-v2]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,15 +15,15 @@ or by downloading a single zip or phar file from our [latest release][latest-rel
...
+* [User Guide][docs-guide] – For both getting started and in-depth **current** SDK usage information
+* [API Docs][docs-api] – For details about operations, parameters, and responses on **current** SDK version
 * [Blog][sdk-blog] – Tips & tricks, articles, and announcements
 * [Sample Project][sdk-sample] - A quick, sample project to help get you started
 * [Forum][sdk-forum] – Ask questions, get help, and give feedback
 * [Issues][sdk-issues] – Report issues, submit pull requests, and get involved
   (see [Apache 2.0 License][sdk-license])
 * [@awsforphp][sdk-twitter] – Follow us on Twitter
-* For Version 2 of the SDK:
+* For **the previous** version 2 of the SDK:
```
kstich: Since we're so far removed from the transition from V2, these links could be moved all the way down to a subsection of the `### Related AWS Projects` section. This would also mean removing the `**current**` label changes from this changeset, as it's further implied by not being near the V2 information.

TakesTheBiscuit: @kstich makes total sense

kstich: @TakesTheBiscuit Would you like to update this PR to handle the above change? Let us know either way!

TakesTheBiscuit: @kstich sorry, yes, i think that's done now in b174abc

README.md
```diff
@@ -124,6 +121,9 @@ try {
 * [AWS SDK Bundle for Symfony][mod-symfony]
 * [Amazon SNS Message Validator for PHP][sns-validator] - SNS validator without requiring SDK
 * [Guzzle Version 6][guzzle-docs] – PHP HTTP client and framework
+* For **the previous** version 2 of the SDK:
```
kstich: This should either be 'For the previous version of the SDK:' or 'For Version 2 of the SDK:', as the current is grammatically incorrect. I'd prefer the latter in this case.

TakesTheBiscuit: You're totally right, i don't know why i was using such poor _Engrish!_ 
550a325

Issue #1322, opened 2017-07-06
Edge Case Error: Valid 1 week expiration throws error


daum:
On a few projects that we upgraded to the v4 API every once and a while we get the error: "The expiration date of a signature version 4 presigned URL must be less than one week. "

This is when pass through just "1 week" to the presign function.  Looking through code it appears that in certain circumstances when you pass through no start time and try to use exactly one week it can cause an error.  

https://github.com/aws/aws-sdk-php/blob/master/src/Signature/SignatureV4.php#L85 uses `time()` if no start time is passed through.  However it isn't until later https://github.com/aws/aws-sdk-php/blob/master/src/Signature/SignatureV4.php#L97, where the expiration is checked.  The problem arises if the script is on the upper bounds of a second (ie 12:01:01.999999999999999) when it makes a call to the presign function.  In this case it is possible the script would not get to the expiration duration check until after it turns 12:01:02.0000000.  This causes the exception to be thrown as the expiration duration check would at 12:01:02 get the result of `strtotime('1 week')`, which would now have an extra second, triggering the exception to be thrown in the `convertExpires` section.  A simple proof of concept of this occurring is located in this gist: https://gist.github.com/daum/0010c9133a6ba27c373e898e2bc87c43

A somewhat simple fix could be moving the actual conversion of the timestamp of "expires" to be the line before https://github.com/aws/aws-sdk-php/blob/master/src/Signature/SignatureV4.php#L85  .  This would guarantee that the expiration time is ran first, so if you do break over a second the start time would simply be 6 days, 23 hours, 59 minutes, 59 seconds from the start time, rather than what currently can happen (7 days, 0 min, 1 seconds).  

Let me know what you think about this problem and that as a valid solution.  I'd be happy to make a PR if it is agreed this is an edge case that should be fixed.

imshashank:
@daum Thanks for pointing out this issue. Please give us some time to repro this issue. In the meantime, please feel free to open a PR with the above-discussed fix.

daum:
Sure just made a PR here: https://github.com/aws/aws-sdk-php/pull/1325 .  Let me know if you want me to change or add anything.

daum:
@imshashank On the tests the reason they are failing is because they use reflection to allow direct access to the private convertExpires method (https://github.com/aws/aws-sdk-php/blob/master/tests/Signature/SignatureV4Test.php#L98).  Since I've updated the method to take the converted timestamp already they are now failing.  

[...]

PR #1325, opened 2017-07-06
Move expires timestamp in SignatureV4 to be generated before startTimestamp

daum:
As described in issue #1322 an edge case of using exactly 1 week expiration when a second rolls between
the startTimestamp calculation via time() and the expires timestamp can cause incorrect exceptions. This
occurs when the second rolls over (ie 12:01:01.9999999 to 12:02:02.00000) between the startTime (time())
and endTime conversion from strtotime().  This causes the exception of the expiration must be less than a week
to be thrown.  By moving the endTime to be converted before startTime this guarantees that will never occur.


daum:
@imshashank All set just updated those lines.

src/Signature/SignatureV4.php
```diff
@@ -82,6 +82,7 @@ public function presign(
         $expires,
         array $options = []
     ) {
+        $expiresTimestamp = $this->convertToTimestamp($expires);
```
kstich: Isn't this still susceptible to being incorrect? The only way to guarantee the `strtotime` case for `convertExpires()` will be in coordination with the `$startTimestamp` [is to pass in a `$now`](http://php.net/manual/en/function.strtotime.php).

Consider the case:
1:23:45.9999 - `$expiresTimestamp` generated from '+1 week'
1:23:46.0005 - `$startTimestamp` generated via time()
1:23:46:0100 - `convertExpires()` is called, generating 604799

daum: You'd technically get a different duration however it won't throw an error.  Assuming that server times are synced up noone could request it in historical time (requiring the extra second of duration).    The goal of this PR is to fix the opposite case where the second rolls and you fall past the 1 week allow duration and it throws an error, even though the initial request is valid.

.changes/nextrelease/signatureExpiratinoEdgeCaseFix.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "Signature",
+        "description": "Fixed edgecase in expiration duration check on signature when seconds roll between implicit startime and relative end time.."
```
daum: All set, removed.

daum:
@kstich @imshashank  Just wanted to check in on this to see if you think it will be integrated soon or if there are other questions.

src/Signature/SignatureV4.php
kstich: It does fix the issue of throwing an exception on the calculation, but not the correctness of the relative time cases. For instance, [the tests](https://github.com/aws/aws-sdk-php/blob/master/tests/Rds/AuthTokenGeneratorTest.php) for the [`AuthTokenGenerator`](https://github.com/aws/aws-sdk-php/blob/master/src/Rds/AuthTokenGenerator.php#L58) detect an explicit expires length. Actually, it looks like these are [already failing very infrequently](https://travis-ci.org/aws/aws-sdk-php/jobs/257922253).

We could take this PR as is and supply another fix for relative times, though it'd be preferable to fix both at the same time. The change for this would be using `convertToTimestamp` in `convertExpires` for the non-`strtotime` cases and passing the `$startTimestamp` to `strtotime`'s second param. `strtotime` will properly ignore being given a `$now` parameter when not using a relative time. This would be in lieu of doing a `convertToTimestamp` before sending in to `convertExpires`.

daum: @kstich Not sure I follow what the fix for both scenarios would be.  I think part of the issue is that if the SDK allows any relative timestamp to be passed there will always be scenarios when the duration for the expires length could be off.  The reason for that is that the thread could be paused/roll over a second at any given point, so unless the caller of presign includes a "now" parameter (options perhaps?) there is always the possibility of a slight skew.

We could add that as an option to pass, if passed any strtotime in the convertToTimestamp will use the "now" as the second parameter for the time it is based on.  

This would solve any specific duration requests from possibly being off by a tiny amount.  We could maintain BC by making it an optional parameter, so if it isn't passed it will just use the current time that the createSign is invoked, which majority of the time would be the same time.  

Let me know if that makes sense.  I think the change to the existing PR would be fairly minimal.  I'd update convertToTimestamp to take a second parameter, which would be passed to the strtotime conversion in it.  Then in the options array it'd check for the "now" parameter, if non-existent it'd take current time.  What do you think the option should be?  "relativeTimeBase" could be more descriptive than now?

kstich: @daum Thanks for working through this with us. I think we're both on the same page with regards to the problem and that the solution being worked towards both fixes the issues at hand and maintains our contracts.

For a relative expires, which will be the only used case for an optional `$relativeTimeBase ` in `convertToTimestamp`, we'll already have `$startTimestamp` and the intent of the contract is the relative expires is based on that. The default/non-existent value should just be null, as that will have `strtotime` use its own default if we're not overriding. This should probably also warrant a new test on the relative time handling of SignatureV4 of the '+6 days' the suite is set up for; you'll have to pass in the 'start_time' based on the 'aws_time' as an option.

daum: Ok sounds good, will try to update the PR with the additional test + other changes either later today or tomorrow hopefully.

src/Signature/SignatureV4.php
```diff
@@ -82,7 +82,9 @@ public function presign(
         $expires,
         array $options = []
     ) {
-        $startTimestamp = isset($options['start_time']) ? $this->convertToTimestamp($options['start_time']) : time();
+        $relativeTimeBase = isset($options['relativeTimeBase']) ? $options['relativeTimeBase'] : time();
+        $expiresTimestamp = $this->convertToTimestamp($expires,$relativeTimeBase);
```
kstich: Thanks for making these updates @daum!

It seems like I was unclear previously when talking about this section, apologies. I don't think there needs to be a separate `relativeTimeBase` *option*, but instead use the `$startTimestamp` we already have for this line passed to `$relativeTimeBase` of `convertToTimestamp` when generating `$expiresTimestamp`. I can't think of any cases where someone would want `'relativeTimeBase'` to not be the same as `'start_time'`, let me know if I'm missing something.

And a small formatting update: please add a space after the `,` in this function call.

daum: Just updated to use the startTimestamp.  Let me know if you are ok with that ternary in the strtotime function for the convertTimestamp, wasn't sure if you'd prefer that or a nested if/else.

kstich: The [ternary located here](https://github.com/aws/aws-sdk-php/pull/1325/files#diff-50459fdb298219f1d103da728ab4afb2R295) isn't actually needed. `strtotime` will use its own default (`$now = time()`) on being passed null and will correctly ignore the `$now` when passed a non-relative `$time` string.

```php
$time = time();
$date = date(DATE_ATOM, $time);
echo strtotime($date); // 1501704830
echo strtotime($date, null); // 1501704830
echo strtotime($date, $time); // 1501704830
echo strtotime('+15 minutes', $time); // 1501705730
```

daum: @kstich I think the segfault on the builds is coming from the precise to trusty upgrade, I can't replicate locally.  It also appears this is a change Travis made recently: https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming .  There were no segfaults before the upgrade, I can add the precise variable into the travis config if you want.  Let me know how you want to proceed.

daum: @kstich Actually that's a fun PHP gotcha, here is what happens if I send it a null when using a relative timestamp (see last test case):

```php
$time = time();
$date = date(DATE_ATOM, $time);
echo strtotime($date)."\n"; // 1501704830
echo strtotime($date, null)."\n"; // 1501704830
echo strtotime($date, $time)."\n"; // 1501704830
echo strtotime('+15 minutes', $time)."\n"; // 1501705730
echo strtotime('+15 minutes', null)."\n"; // 900
```

kstich: That's lovely, this should stay a ternary then. Sorry for the misdirection!

#1339 will temporarily take care of the Travis issue and we'll rerun these jobs.

daum: Ok - will wait for you guys to merge that, then will merge those updates into this branch.

kstich: #1339 has been merged in.

daum: All set!

src/Signature/SignatureV4.php
```diff
@@ -82,7 +82,9 @@ public function presign(
         $expires,
         array $options = []
     ) {
-        $startTimestamp = isset($options['start_time']) ? $this->convertToTimestamp($options['start_time']) : time();
+
+        $startTimestamp = isset($options['start_time']) ? $this->convertToTimestamp($options['start_time'], null) : time();
```
imshashank: Please break this into multiple lines. We try to keep all line length to be of less than 80 characters in the PHP SDK.

src/Signature/SignatureV4.php
```diff
@@ -285,22 +287,22 @@ private function getCanonicalizedQuery(array $query)
...
     }
 
-    private function convertToTimestamp($dateValue)
+    private function convertToTimestamp($dateValue, $relativeTimeBase = null)
     {
         if ($dateValue instanceof \DateTime) {
             $timestamp = $dateValue->getTimestamp();
         } elseif (!is_numeric($dateValue)) {
-            $timestamp = strtotime($dateValue);
+            $timestamp = strtotime($dateValue, $relativeTimeBase === null ? time() : $relativeTimeBase );
```
imshashank: Please break this into multiple lines. We try to keep all line length to be of less than 80 characters in the PHP SDK.

daum: How do you want to break the ternary statements?  I personally think they are much more readable on one line like this but understand the 80 character comment, was just following other examples in the code like https://github.com/daum/aws-sdk-php/blob/6cd12a2744b9da9813faf1f7d6bf3f52d336c67a/src/Signature/SignatureV4.php#L39

Can you just give me an example of how you'd like the ternary formatted for the multiple lines and I'll get that fixed up asap.

src/Signature/SignatureV4.php
daum: Similar to my other comment - how would you like the ternary broken up.  If you would prefer I could also switch it to a if/else which would have a shorter line length.

kstich: An example of what our 3 line ternary looks like [can be found here.](https://github.com/aws/aws-sdk-php/blob/master/src/S3/BatchDelete.php#L170)

src/Signature/SignatureV4.php
kstich: For this case, the ternary can be one line but the arguments would be separated.
```php
strtotime(
    $dateValue,
    $relativeTimeBase === null ? time() : $relativeTimeBase
);
```

daum: Thanks just updated

kstich:
Thanks again for this work @daum!

PR #1327, opened 2017-07-10
SSECMiddleware fix in order to use a custom base64-encoded encryption key

dioogo91:
The previous implementation only accepted printable ASCII characters as the password for the SSE encryption. This reduces the password strength since it is not possible to use the full 256 values per byte, and only around 95 (printable ascii) values, effectively limiting the possible password entropy. This solution accepts both options: a custom base64-encoded encryption key or a 32 characters string.

reference: https://en.wikipedia.org/wiki/Password_strength#Random_passwords

### DIFF
--- a/src/S3/SSECMiddleware.php
+++ b/src/S3/SSECMiddleware.php
@@ -63,7 +63,7 @@ class SSECMiddleware
     {
         // Base64 encode the provided key
         $key = $command[$prefix . 'SSECustomerKey'];
-        if(base64_encode(base64_decode($key, true)) === $key) {
+        if($command['SSEBase64Key']) {
             $command[$prefix . 'SSECustomerKey'] = $key;
             $key = base64_decode($key, true);
         } else {
--- a/src/S3/SSECMiddleware.php
+++ b/src/S3/SSECMiddleware.php
@@ -63,7 +63,7 @@ class SSECMiddleware
     {
         // Base64 encode the provided key
         $key = $command[$prefix . 'SSECustomerKey'];
-        if($command['SSEBase64Key']) {
+        if($command['SSECustomerKeyIsBase64']) {
             $command[$prefix . 'SSECustomerKey'] = $key;
             $key = base64_decode($key, true);
         } else {
--- a/src/S3/SSECMiddleware.php
+++ b/src/S3/SSECMiddleware.php
@@ -63,8 +63,12 @@ class SSECMiddleware
     {
         // Base64 encode the provided key
         $key = $command[$prefix . 'SSECustomerKey'];
-        $command[$prefix . 'SSECustomerKey'] = base64_encode($key);
-
+        if(base64_encode(base64_decode($key, true)) === $key) {
+            $command[$prefix . 'SSECustomerKey'] = $key;
+            $key = base64_decode($key, true);
+        } else {
+            $command[$prefix . 'SSECustomerKey'] = base64_encode($key);
+        }
         // Base64 the provided MD5 or, generate an MD5 if not provided
         if ($md5 = $command[$prefix . 'SSECustomerKeyMD5']) {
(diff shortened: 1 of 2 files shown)


dioogo91:
@imshashank this version won't change behavior of existing clients: clients that don't want to use a custom base64 encoded key can either leave SSECustomerKeyIsBase64 option value undefined or set it as false or null. You can only use a custom base64 encoded key if you set the value as true.

imshashank:
@dioogo91 The changes LGTM. Have asked @kstich to review and approve.

HugoMestre:
Nice fix @dioogo91. This will allow stronger passwords to be placed in configuration files, e.g. YAML files.

dioogo91:
@kstich any news?

imshashank:
@dioogo91 Please add a changelog blog as described here with this PR -> https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md

imshashank:
@dioogo91 On a second look I have some questions.

Please have a look at the below code snippet which works for me:
```
  $key = openssl_random_pseudo_bytes(256/8);
  $client->putObject([
        'Bucket' => $bucket,
        'Key'    => 'testfile',
        'Body'   => "Test this",
        'SSECustomerAlgorithm' => 'AES256',
        'SSECustomerKey' => $key
    ]);
```

This is using total of 256 bits as the key for the SSE-C. The algorithm will fail if the key size is not correct.

Also customers need to provide a key that is not base64_encoded.  So for `yml` files you can save keys in base64 and then base64 decode them before passing to the s3Client.

Please help us in understanding the scenario if we are missing some usecase which can't be handled without this change.

dioogo91:
@imshashank You're right: we could save it in base64, decode it and then pass it to the s3Client. The thing is every library or adapter that uses this SDK has to change its code in order to do that. Nonetheless and most importantly I think this enhances the flexibility of the SDK because this gives the customer the alternative of using an already base64 encoded key without having to decode it every time the API methods are called. As you said "customers need to provide a key that is not base64_encoded" and that is a limitation to the customer that my pull request tries to mitigate. Surely there are ways to go around this but I think this would be a nice enhancement to this SDK since it doesn't even change its default behaviour in any way.

src/S3/SSECMiddleware.php
```diff
@@ -63,8 +63,12 @@ private function prepareSseParams(CommandInterface $command, $prefix = '')
     {
         // Base64 encode the provided key
         $key = $command[$prefix . 'SSECustomerKey'];
-        $command[$prefix . 'SSECustomerKey'] = base64_encode($key);
-
+        if($command['SSECustomerKeyIsBase64']) {
```
kstich: This check should verify that the field is `!empty()`, not just check if it's truthy, as its current state will give off . It also needs to work for all `$prefix` values.

tests/S3/SSECMiddlewareTest.php
```diff
@@ -71,6 +71,20 @@ public function getListenerTestCases()
                     'SSECustomerKeyMD5' => null,
                 ]
             ],
+            [
+                'PutObject',
```
kstich: This update needs to supply similar behavior for `ListObjects` and `CopyObject` as the tests above it.

src/S3/SSECMiddleware.php
```diff
@@ -63,8 +63,12 @@ private function prepareSseParams(CommandInterface $command, $prefix = '')
...
+            $command[$prefix . 'SSECustomerKey'] = $key;
+            $key = base64_decode($key, true);
+        } else {
+            $command[$prefix . 'SSECustomerKey'] = base64_encode($key);
```
kstich: Should the initial `$key` assignment be at the beginning of this `else` block?

kstich:
Closing due to lack of update for the requested changes.

PR #1330, opened 2017-07-14
Fix for S3MultiRegionClient receiving AuthorizationHeaderMalformed with host-style

### LINKED ISSUE #1324: Recent S3MultiRegionClient changes caused issue
Hello PHP team,

I didn't see any changes to the [`S3MultiRegionClient`](https://github.com/aws/aws-sdk-php/blob/master/src/S3/S3MultiRegionClient.php) called out in any release notes, but it looks like some fairly extensive work was done for 3.25.0. After a recent deployment, we started seeing some issues like:
```
Error executing "PutObject" on "https://{BUCKET}/{OBJECT_KEY}"; AWS HTTP error: Client error: `PUT https://{BUCKET}/{OBJECT_KEY}` resulted in a `400 Bad Request` response:
    AuthorizationHeaderMalformed
    The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-west-2'
    {SIGNATURE_STUFF}
```
But we didn't change any code or config other than the SDK version. So, would any of the recent changes to `S3MultiRegionClient` effect already running code? (e.g., changes to caching, etc.)

kstich:
Properly handle reading mismatched regions from S3's AuthorizationHeaderMalformed exception for S3MultiRegionClient, updated host-style tests to reflect correct behavior.

Fixes #1324

### DIFF
--- a/src/S3/S3ClientTrait.php
+++ b/src/S3/S3ClientTrait.php
@@ -1,6 +1,7 @@
 <?php
 namespace Aws\S3;
 
+use Aws\Api\Parser\PayloadParserTrait;
 use Aws\CommandInterface;
 use Aws\Exception\AwsException;
 use Aws\HandlerList;
@@ -15,6 +16,8 @@ use GuzzleHttp\Promise\RejectedPromise;
  */
 trait S3ClientTrait
 {
+    use PayloadParserTrait;
+
     /**
      * @see S3ClientInterface::upload()
      */
@@ -213,13 +216,36 @@ trait S3ClientTrait
         return $handler($command)
             ->then(static function (ResultInterface $result) {
                 return $result['@metadata']['headers']['x-amz-bucket-region'];
-            }, static function (AwsException $exception) {
-                $response = $exception->getResponse();
+            }, function (AwsException $e) {
+                $response = $e->getResponse();
                 if ($response === null) {
-                    throw $exception;
+                    throw $e;
                 }
+
+                if ($e->getAwsErrorCode() === 'AuthorizationHeaderMalformed') {
+                    $region = $this->determineBucketRegionFromExceptionBody(
+                        $response->getBody()
+                    );
+                    if (!empty($region)) {
+                        return $region;
+                    }
+                }
+
                 return $response->getHeaderLine('x-amz-bucket-region');
             });
+    }
+
+    private function determineBucketRegionFromExceptionBody($responseBody)
+    {
+        try {
+            $element = $this->parseXml($responseBody);
+            if (!empty($element->Region)) {
+                return (string)$element->Region;
+            }
+        } catch (\Exception $e) {
+            return null;
+        }
+        return null;
     }
 
     /**
--- a/src/S3/S3ClientTrait.php
+++ b/src/S3/S3ClientTrait.php
@@ -244,7 +244,7 @@ trait S3ClientTrait
                 return (string)$element->Region;
             }
         } catch (\Exception $e) {
-            return false;
+            // Fallthrough on exceptions from parsing
         }
         return false;
     }
--- a/src/S3/S3ClientTrait.php
+++ b/src/S3/S3ClientTrait.php
@@ -229,6 +229,7 @@ trait S3ClientTrait
                     if (!empty($region)) {
                         return $region;
                     }
+                    throw $e;
                 }
 
                 return $response->getHeaderLine('x-amz-bucket-region');
@@ -243,9 +244,9 @@ trait S3ClientTrait
                 return (string)$element->Region;
             }
         } catch (\Exception $e) {
-            return null;
+            return false;
         }
-        return null;
+        return false;
     }
 
     /**
--- a/src/S3/S3MultiRegionClient.php
+++ b/src/S3/S3MultiRegionClient.php
@@ -243,14 +243,14 @@ class S3MultiRegionClient extends BaseClient implements S3ClientInterface
                             );
                             if (!empty($region)) {
                                 $this->cache->set($cacheKey, $region);
-                            } else {
-                                $region = $this->getDetermineBucketRegionGenerator(
-                                    $command['Bucket']
-                                );
-                            }
 
-                            $command['@region'] = $region;
-                            yield $handler($command);
+                                $command['@region'] = $region;
+                                yield $handler($command);
+                            } else {
+                                throw $e;
+                            }
+                        } else {
+                            throw $e;
                         }
                     }
                 });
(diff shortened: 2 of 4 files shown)


src/S3/S3MultiRegionClient.php
```diff
@@ -224,17 +227,31 @@ private function determineRegionMiddleware() {
...
                                 $command['Bucket']
-                            ));
+                            );
                         }
 
-                        $this->cache->set($cacheKey, $region);
                         $command['@region'] = $region;
                         yield $handler($command);
+                    } catch (AwsException $e) {
+                        if ($e->getAwsErrorCode() === 'AuthorizationHeaderMalformed') {
```
cjyclaire: If the exception caught is not "AuthorizationHeaderMalformed", does this just make the exception "slient"?

src/S3/S3ClientTrait.php
```diff
@@ -213,15 +216,38 @@ public function determineBucketRegionAsync($bucketName)
...
-            }, static function (AwsException $exception) {
-                $response = $exception->getResponse();
+            }, function (AwsException $e) {
+                $response = $e->getResponse();
                 if ($response === null) {
-                    throw $exception;
+                    throw $e;
+                }
+
+                if ($e->getAwsErrorCode() === 'AuthorizationHeaderMalformed') {
```
cjyclaire: Just a bit thought, I'm open on this. Could we make the error code check at `determineBucketRegionFromExceptionBody` function instead perhaps? This helps reduce duplicate code, and easy to extend if there will be another special case in the future. 
Hmm, yet it might be a bit trickier to do this at the Multi region client.

src/S3/S3MultiRegionClient.php
```diff
@@ -224,17 +227,31 @@ private function determineRegionMiddleware() {
...
+                            $region = $this->determineBucketRegionFromExceptionBody(
+                                $e->getResponse()->getBody()
+                            );
+                            if (!empty($region)) {
+                                $this->cache->set($cacheKey, $region);
+                            } else {
+                                $region = $this->getDetermineBucketRegionGenerator(
```
cjyclaire: Looks like calling `getDetermineBucketRegionGenerator` here is checking `AuthorizationHeaderMalformed` twice if cache is not found. This doesn't hurt, though might be worth considering separate the checking logic in my first comment?

src/S3/S3ClientTrait.php
```diff
@@ -213,15 +216,38 @@ public function determineBucketRegionAsync($bucketName)
...
 
+    private function determineBucketRegionFromExceptionBody($responseBody)
+    {
+        try {
+            $element = $this->parseXml($responseBody);
+            if (!empty($element->Region)) {
+                return (string)$element->Region;
+            }
+        } catch (\Exception $e) {
+            return null;
```
jeskew: More of a nit, but I would expect this method to return `false` on error if it's swallowing exceptions.

src/S3/S3MultiRegionClient.php
```diff
@@ -279,11 +301,26 @@ public function determineBucketRegionAsync($bucketName)
...
         return $regionalClient->determineBucketRegionAsync($bucketName)
-            ->then(function ($region) use ($cacheKey) {
-                $this->cache->set($cacheKey, $region);
+            ->then(
+                function ($region) use ($cacheKey) {
+                    $this->cache->set($cacheKey, $region);
+
+                    return $region;
+                },
+                function (AwsException $e) use ($cacheKey) {
```
jeskew: Wouldn't this case already be handled by the `determineBucketRegionAsync` on the regional client?

src/S3/S3MultiRegionClient.php
kstich: Good catch, re-throwing on failure to get a region (other reason for `AuthorizationHeaderMalformed` and on other error codes.

src/S3/S3ClientTrait.php
```diff
@@ -213,15 +216,38 @@ public function determineBucketRegionAsync($bucketName)
...
+                    throw $e;
+                }
+
+                if ($e->getAwsErrorCode() === 'AuthorizationHeaderMalformed') {
+                    $region = $this->determineBucketRegionFromExceptionBody(
+                        $response->getBody()
+                    );
+                    if (!empty($region)) {
+                        return $region;
+                    }
```
kstich: Yep, I caught this when reviewing with @cjyclaire's mention above, re-throwing here as well.

src/S3/S3ClientTrait.php
kstich: Going to keep these separate since the Client handles the cache and the middleware generates into the coroutine instead of returning.

src/S3/S3MultiRegionClient.php
kstich: I thought this was reachable but couldn't generate an exception that escaped to this level that would be caught here; cleaning up.

src/S3/S3MultiRegionClient.php
```diff
@@ -224,17 +227,31 @@ private function determineRegionMiddleware() {
...
+                            );
+                            if (!empty($region)) {
+                                $this->cache->set($cacheKey, $region);
+
+                                $command['@region'] = $region;
+                                yield $handler($command);
+                            } else {
+                                throw $e;
+                            }
+                        } else {
```
cjyclaire: To be consistent with the way you throw exceptions above, perhaps just throw this after the `if` check instead of in a `else` block?

src/S3/S3MultiRegionClient.php
```diff
@@ -224,17 +227,31 @@ private function determineRegionMiddleware() {
...
+                            $region = $this->determineBucketRegionFromExceptionBody(
+                                $e->getResponse()->getBody()
+                            );
+                            if (!empty($region)) {
+                                $this->cache->set($cacheKey, $region);
+
+                                $command['@region'] = $region;
+                                yield $handler($command);
+                            } else {
+                                throw $e;
```
cjyclaire: Same here

src/S3/S3MultiRegionClient.php
kstich: The generator will continue execution when looking for a potential next value to `yield` and would proceed to throw in all states past the `yield` invocation on L248 in doing so. This is different from the re-throw in `S3ClientTrait` because a the `return` blocks further execution for that scope.

src/S3/S3MultiRegionClient.php
```diff
@@ -269,6 +286,11 @@ public function getObjectUrl($bucket, $key)
         return $regionalClient->getObjectUrl($bucket, $key);
     }
 
+    public function getDetermineBucketRegionGenerator($bucketName)
```
jeskew: Why use this method instead of just directly calling `yield $this->determineBucketRegionAsync($bucketName);`? If the intention is just to avoid repeating `yield`, I think this should be private

PR #1352, opened 2017-08-11
Fixed typo

wjgilmore:
Just wading into making my first contribution to the documentation. :-) Fixes typo, timestamp was mistyped as timesamp.


kstich:
Thanks for the fix!

PR #1367, opened 2017-09-06
Set ContentType only when no params ContentType

endlessz:
I have an issue when upload pdf to S3 storage. It always has content type 'application/octet-stream' even though I force it to 'application/pdf' but It's still the same.

I thought it was pdf is too big and It call getSourceMimeType before it generate finish. That makes it wrong ContentType.

If anything go wrong please suggest me. This is my first time. 
Thank you


tiran133:
I have the same problem. All my multipart uploads end up having application/octet-stream as Content-Type.
Regardless if I set the param "ContentType"

Any solution on this?
Help is appreciated.

Alex

PR #1370, opened 2017-09-11
Introduce DBSnapshotAvailable and DBSnapshotDeleted waiters for RDS DBSnapshot

rquadling:
Modified json and compiled php.

Resolves https://github.com/aws/aws-sdk-php/issues/1368

PR #1371, opened 2017-09-11
Compiled json to php

rquadling:
Whilst working on https://github.com/aws/aws-sdk-php/pull/1370, I ran the `make compile-json` command and it resulted in more compilations that expected. This PR covers the additional compilations.


kstich:
These occur due to differences in PHP versions for running compile-json in and would be rewritten by our process. Closing.

rquadling:
So should these generated files be excluded from the repo? Maybe built and included in the versioned archive? If I want to work with the source, it would be my responsibility to build via `make compile-json` (and keep them uptodate if I'm playing around with the contents of `data`, but if I use `composer install` or `composer update` (and using dist rather than source), then I'd get them pre-built?

kstich:
That's correct - if you open up the vendor folder when using composer, you'll only see the .json.php files. We include them here in case you wish to build from source.

rquadling:
I'm not sure I explained things fully.

Ideally locally compiling shouldn't result in a diff if they are only build artefacts.

Only the .json files should be in the repo. The .json.php should only be in the distributions. The .json.php should be .gitignored.

Anyway, thanks for looking at this.

PR #1373, opened 2017-09-13
Travis updates

kstich:
Update Travis to run on defauly (Trusty). Disable xdebug in the container due to segfault issues on 5.5. Add 7.2 (previously nightly) to included list and allow failure on '--prefer-lowest' for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,26 +2,27 @@ language: php
 
 dist: trusty
 
-php:
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - 7.2
-  - hhvm
-  - nightly
-
-sudo: false
-
-env:
-  - COMPOSER_OPTS=""
-  - COMPOSER_OPTS="--prefer-lowest"
-
 matrix:
+  php:
+    - 5.5
+    - 5.6
+    - 7.0
+    - 7.1
+    - 7.2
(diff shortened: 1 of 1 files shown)


kstich:
Will squash before merging - I was unable to recreate the issue in a local Travis instance so had to do a bunch of commit/push sets to trigger builds.

No changelog document as this is not applicable to consumers of the SDK.

.travis.yml
```diff
@@ -1,31 +1,32 @@
...
-
+  - nightly
 env:
   - COMPOSER_OPTS=""
   - COMPOSER_OPTS="--prefer-lowest"
 
 matrix:
   allow_failures:
+    - php: 7.2
+      env: COMPOSER_OPTS="--prefer-lowest"
```
jeskew: Is this scenario only failing because of incompatibilities in PHPUnit? They don't officially support PHP 7.2 yet, so it makes sense not to treat this as a release blocker.

This shouldn't be a blocker for this change, but in a follow-up PR it might be worth upgrading the tests to PHPUnit's namespaced class style so that you can also allow newer environments to run PHP 6.

kstich: There are a few requirements that look like they will need their floors raised. I've added this to the backlog to take care of before 7.2 official release, as well as looking at updating our test suite.

PR #1376, opened 2017-09-14
Optionally preserve iterator keys when using a CommandPool

### LINKED ISSUE #1378: Preserve iterator keys when using Command Pool
An existing PR (https://github.com/aws/aws-sdk-php/pull/1240) seems to have been abandoned.

I created a new PR (https://github.com/aws/aws-sdk-php/pull/1376) with the appropriate requested changes.

rquadling:
Implementation of https://github.com/aws/aws-sdk-php/pull/1240 with suggested configuration option changes.

Resolves #1378


src/CommandPool.php
```diff
@@ -47,9 +49,12 @@ public function __construct(
         if (!isset($config['concurrency'])) {
             $config['concurrency'] = 25;
         }
+        if (!array_key_exists('preserve_iterator_keys', $config)) {
```
kstich: Instead of setting the config value here, the conditional below should just be:
```php
if (!empty($config['preserve_iterator_keys']))
```

rquadling: Sure. I'll use `if (empty())` though. :-D

kstich: >.> Yep!

src/CommandPool.php
```diff
@@ -49,7 +49,7 @@ public function __construct(
         if (!isset($config['concurrency'])) {
             $config['concurrency'] = 25;
         }
-        if (!array_key_exists('preserve_iterator_keys', $config)) {
+        if (empty($config['preserve_iterator_keys'])) {
```
kstich: Apologies for this quick redux, I meant in the other comment to remove this whole conditional setting of `'preserve_iterator_keys'` and replace the conditional in the `$mapFn` loop below with the `empty()` check. Sorry for the confusion!

rquadling: Aha. Yep!

.changes/nextrelease
```diff
@@ -0,0 +1,7 @@
+[
```
kstich: Just noticed this after merging in the RDS waiter PR, the changelog file needs to be in a named file inside a nextrelease folder:

>A changelog document is a small JSON blob placed in the .changes/nextrelease folder. It should be named clearly and uniquely, akin to a branch name.

PR #1385, opened 2017-09-29
S3 Stream Uri Parsing

### LINKED ISSUE #1383: URI parsing for "s3://bucket/key" format
The `Aws\S3\S3UriParser` only supports parsing of URIs [with `http` scheme](https://github.com/aws/aws-sdk-php/blob/master/tests/S3/S3UriParserTest.php).

I have a [stream URI](https://github.com/aws/aws-sdk-php/blob/master/docs/service/s3-stream-wrapper.rst) in the format of `s3://bucket/key` that I was trying to parse with it. Could support for that format be added?

kstich:
Properly parse s3:// uri used with StreamWrapper, with tests.

Resolves #1383

### DIFF
--- a/src/S3/S3UriParser.php
+++ b/src/S3/S3UriParser.php
@@ -10,6 +10,7 @@ use Psr\Http\Message\UriInterface;
 class S3UriParser
 {
     private $pattern = '/^(.+\\.)?s3[.-]([A-Za-z0-9-]+)\\./';
+    private $streamWrapperScheme = 's3';
 
     private static $defaultResult = [
         'path_style' => true,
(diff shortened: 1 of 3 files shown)


src/S3/S3UriParser.php
```diff
@@ -34,6 +35,11 @@ class S3UriParser
     public function parse($uri)
```
cjyclaire: Maybe add a bit document about supporting [stream wrapper](https://github.com/aws/aws-sdk-php/blob/master/docs/service/s3-stream-wrapper.rst) format there

src/S3/S3UriParser.php
```diff
@@ -34,6 +36,11 @@ class S3UriParser
...
     {
         $url = Psr7\uri_for($uri);
+
+        if ($url->getScheme() == $this->streamWrapperScheme) {
```
eliysha: This is not working for `s3://bucket.s3-regionaws.amazonaws.com/filepath`

PR #1387, opened 2017-10-03
Set ContentType only when no params ContentType

### LINKED ISSUE #1386: All multipart uploads end up having Content-Type "application/octet-stream"
Recently I was wondering why my uploaded videos won't play in a browser when accusing the url directly. I discovered that the wrong content-type was set during upload.

I traced the issue back to this peace of code.
Where getInitiateParams() overwrites the previously set content-type in $params.
https://github.com/aws/aws-sdk-php/blob/c0321509cccc20e6f0051ce9e36369e465388029/src/S3/MultipartUploadingTrait.php#L96-L111

It calls the function getSourceMimeType() from
https://github.com/aws/aws-sdk-php/blob/4da9a3b61473b378dea34c2b211cb3654ec3effb/src/S3/MultipartUploader.php#L136-L142

Which tries to figure out the content-type from the source. Since the source is a  "php://temp" stream. It will return "application/octet-stream".
This forces the function getInitiateParams() to overwrite the previously set content-type in $params.

Which is annoying all files larger uploaded via MultipartUpload will have a content-type of  "application/octet-stream".

I'm using Flysysteme AWS-S3_V3 Atapter  to upload files but all its doing its calling the upload function on the S3Client

https://github.com/thephpleague/flysystem-aws-s3-v3/blob/3837b6a354fb2a3cac58ec84bcb135b20a183d31/src/AwsS3Adapter.php#L573-L594

So there shouldn't be anything wrong with that.

Do I miss something or is is intended and if so why?

There was a pull request addressing that issue but there is not much information for it.
https://github.com/aws/aws-sdk-php/pull/1367

Hope somebody can help.

Thank You 
Alex

endlessz:
I have an issue when upload pdf to S3 storage. It always has content type 'application/octet-stream' even though I force it to 'application/pdf' but It's still the same.

I thought it was pdf is too big and It call getSourceMimeType before it generate finish. That makes it wrong ContentType.

Fixes #1386


src/S3/MultipartUploadingTrait.php
```diff
@@ -102,8 +102,8 @@ protected function getInitiateParams()
             $params['ACL'] = $config['acl'];
         }
 
-        // Set the content type
-        if ($type = $this->getSourceMimeType()) {
+        // Set the content type if ContentType is not set
```
kstich: This can just read 'Set the ContentType if not already present' for clarity.

src/S3/MultipartUploadingTrait.php
```diff
@@ -102,8 +102,8 @@ protected function getInitiateParams()
...
+        if (!isset($params['ContentType']) && $type = $this->getSourceMimeType()) {
```
kstich: This `!isset(` should be `empty(`, otherwise we'll allow setting the content type to `''`.

.changes/nextrelease/s3_content_type_params
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "S3",
+        "description": "Set the ContentType if not already present."
```
kstich: This was already the case, it was overriding the supplied ContentType as well. This should read:

> No longer override supplied ContentType parameter when performing a multipart upload.

PR #1395, opened 2017-10-13
S3 Client Side Encryption

### LINKED ISSUE #831: client side encryption
Hi,

Has S3 client side encryption been implemented in this library? Or is there any alternatives maybe?

Thank you!

Best regards,
Lajos

kstich:
Adds the `S3EncryptionClient` and supporting implementations. This implements [envelope encryption](http://docs.aws.amazon.com/kms/latest/developerguide/workflow.html) and utilizes [OpenSSL](https://www.openssl.org/) to provide CBC and GCM (on PHP >= 7.1) support . Uses pluggable strategies (via `MetadataStrategyInterface`) for handling a `MetadataEnvelope` in conjunction with a `MaterialsProvider`. A `KmsMaterialsProvider` is provided for key management . Supports `->putObject[Async]` and `->getObject[Async]` operations. This is [cross compatible with Java, Ruby, Go, and C++ for the provided feature set](http://docs.aws.amazon.com/general/latest/gr/aws_sdk_cryptography.html). A service guide is also included.

/cc @jeskew @mtdowling @cjyclaire 

Resolves #831


kstich:
Codecov report of decreased coverage is incorrect. Details show that 95.99% of the diff is covered which is higher than the current 91.90% coverage.

features/s3Encryption/encryption.feature
```diff
@@ -0,0 +1,42 @@
+@s3Encryption @integ
+Feature: S3 Client Side Encryption
+
+  Scenario: Upload PHP's GCM encrypted fixtures
+    When I get all fixtures for "aes_gcm" from "aws-s3-shared-tests"
```
jeskew: Is this the name of a particular bucket? Including a resource identifier here seems fragile.

src/Crypto/EncryptionTrait.php
```diff
@@ -0,0 +1,177 @@
...
+    protected function getEncryptingStream(
+        Stream $plaintext,
+        $cek,
+        &$cipherOptions
+    ) {
+        switch ($cipherOptions['Cipher']) {
+            case 'gcm':
+                $cipherOptions['TagLength'] = 16;
+
+                $taggedStream = new AesTaggedGcmEncryptingStream(
```
jeskew: Couldn't `AesTaggedGcmEncryptingStream` be removed if you just used an [`AppendStream`](https://github.com/guzzle/psr7#appendstream) here?

src/Crypto/KmsMaterialsProvider.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
```
jeskew: Very nit but passim: I believe SDK style requires no newline between the file opener and the namespace declaration.

src/Crypto/KmsMaterialsProvider.php
```diff
@@ -0,0 +1,109 @@
...
+            true
+        );
+        if (empty($materialsDescription['kms_cmk_id'])) {
+            throw new \RuntimeException('Not able to detect kms_cmk_id from kms'
+                . ' materials description.');
+        }
+
+        return new KmsMaterialsProvider(
+            $this->kmsClient,
+            $materialsDescription['kms_cmk_id']
```
jeskew: What's a `cmk`? Could abbreviations be avoided in this parameter name, or is the expanded version too long to reasonably type?

src/Crypto/KmsMaterialsProvider.php
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * @param string $unencryptedCek Key for use in encrypting other data
+     *                               that itself needs to be encrypted by the
+     *                               Provider.
+     * @param string $materialDescription Material Description for use in
+     *                                    encrypting the $cek.
+     *
+     * @return string
+     */
+    public function encryptCek($unencryptedCek, $materialDescription)
```
jeskew: Same as above comment but for 'cek'

src/S3/Crypto/InstructionFileMetadataStrategy.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param MetadataEnvelope $envelope Encryption data to save according to
+     *                                   the strategy.
+     * @param array $args Starting arguments for PutObject, used for saving
+     *                    extra the instruction file.
+     *
+     * @return array Updated arguments for PutObject.
+     */
+    public function save(MetadataEnvelope $envelope, array $args)
+    {
+        $this->client->putObject([
```
jeskew: What if saving this instruction file fails? Is the ciphertext blob deleted from S3?

src/S3/Crypto/S3EncryptionClient.php
```diff
@@ -0,0 +1,381 @@
...
+
+        $instructionFileSuffix = $this->getInstructionFileSuffix($args);
+        unset($args['@InstructionFileSuffix']);
+
+        $strategy = $this->getMetadataStrategy($args, $instructionFileSuffix);
+        unset($args['@MetadataStrategy']);
+
+        $envelope = new MetadataEnvelope();
+
+        return Promise\promise_for($this->encrypt(
```
jeskew: `encrypt` isn't returning a promise, is it?

tests/Crypto/AesEncryptionStreamTestTrait.php
```diff
@@ -0,0 +1,89 @@
...
+            [128],
+            [192],
+            [256],
+        ];
+    }
+
+    public function plainTextProvider() {
+        return [
+            [Psr7\stream_for('The rain in Spain falls mainly on the plain.')],
+            [Psr7\stream_for('Ø¯Ø³Øªâ€ŒÙ†ÙˆØ´ØªÙ‡â€ŒÙ‡Ø§ Ù†Ù…ÛŒâ€ŒØ³ÙˆØ²Ù†Ø¯')],
```
jeskew: I believe this is a copy-paste error. IIRC, this case (and the following one) were meant to exercise round-tripping of multibyte characters, e.g.:

```php
[Psr7\stream_for('دست‌نوشته‌ها نمی‌سوزند')],
[Psr7\stream_for('Рукописи не горят')],
```

tests/Crypto/KmsMaterialsProviderTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function testProviderBasics()
+    {
+        /** @var KmsClient $client */
+        $client = $this->getTestClient('Kms', []);
+        $keyId = '11111111-2222-3333-4444-555555555555';
+
+        $provider = new KmsMaterialsProvider($client, $keyId);
+
+        $this->assertEquals(
+            ['kms_cmk_id' => $keyId],
```
jeskew: From context, I'm guessing that `CEK` stands for "content encryption key" and `CMK` stands for "content master key"; why not just use `EncryptionKey` and `MasterKey` as the parameter names?

tests/S3/Crypto/HeadersMetadataStrategyTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
```
jeskew: The test classes should probably be in a namespace corresponding to the directory structure.

.travis.yml
```diff
@@ -34,4 +34,4 @@ script:
   - if [ -z "$COMPOSER_OPTS" ]; then make package; fi
 
 after_success:
-  - bash <(curl -s https://codecov.io/bash)
+  - 'if [ $(phpenv version-name) == "7.1" ]; then bash <(curl -s https://codecov.io/bash); fi'
```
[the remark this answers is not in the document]
kstich: Yeah, it looks like CodeCov is taking the first report and using that, which explains the wild shifts as well. I'm going to set this up to use 7.1 with latest dependencies only, putting us on stable grounds that we can update with further PHP releases.

src/S3/Crypto/S3EncryptionClient.php
kstich: It is not returning a promise, but an `AesStreamInterface`. This is why it's the only piece wrapped in the `Promise\promise_for` call.

src/S3/Crypto/InstructionFileMetadataStrategy.php
kstich: The [`$strategy->save()` call](https://github.com/aws/aws-sdk-php/pull/1395/files#diff-a4b2e1f92c8fda8b31a82e07e90466e3R186) is made before the `PutObject` call for the ciphertext. The instruction file will remain on S3 if there is a failure after it has been uploaded, not automatically deleted. I'll add a note to the guide and update the doc block for this file.

src/S3/Crypto/S3EncryptionClient.php
```diff
@@ -0,0 +1,381 @@
...
+     *   to default.
+     * - @InstructionFileSuffix: (string|null) Suffix used when writing to an
+     *   instruction file if using an InstructionFileMetadataHandler.
+     *
+     * @return PromiseInterface
+     *
+     * @throws \InvalidArgumentException Thrown when arguments above are not
+     *                                   passed or are passed incorrectly.
+     */
+    public function putObjectAsync(array $args) {
```
cjyclaire: What's the failed scenario look like? Especially for "getObject"? It will be great if some error scenario (besides invalid arguments) can be tested as well

src/Crypto/KmsMaterialsProvider.php
kstich: `'kms_cmk_id'` is how the key is indexed in the materials description, this is [consistent with the Java (and other) SDK implementations](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/KMSEncryptionMaterials.java#L33). The ['KeyId' naming on `Encrypt`](https://github.com/aws/aws-sdk-php/pull/1395/files#diff-46de444f818130423cc34de9445afd90R83) is used for the `$kmsKeyId` variable.

src/Crypto/KmsMaterialsProvider.php
kstich: CEK is a highly used domain specific noun. I'd hesitate to change this to something longer or changing it to `EncryptionKey` as suggested below. I'll update the documentation for each method that handles a CEK to read:

>content encryption key (CEK)`

features/s3Encryption/encryption.feature
kstich: It is a bucket name and it is fragile. This is how [the AWS SDK for Go handles it](https://github.com/aws/aws-sdk-go/blob/c47b5f826383ef61914397ce2ca736cd92365eee/awstesting/integration/customizations/s3/s3crypto/s3_crypto.feature#L6) and we do something similar for the [other full integration tests as well](https://github.com/aws/aws-sdk-php/blob/master/tests/Integ/S3Context.php#L45). These tests need to talk to other SDKs and don't get a `getResourcePrefix` added. I'm open to suggestions on this.

src/S3/Crypto/S3EncryptionClient.php
```diff
@@ -0,0 +1,381 @@
...
+                    . ' was passed to "MetadataStrategy" was unrecognized.');
+            }
+        } elseif ($instructionFileSuffix) {
+            return new InstructionFileMetadataStrategy(
+                $this->client,
+                $instructionFileSuffix
+            );
+        }
+
+        return null;
```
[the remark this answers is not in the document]
kstich: Yeah, I wanted to be explicit here though since there are other return statements in this function.

src/S3/Crypto/S3EncryptionClient.php
kstich: The thought on async basing the manipulation was if someone invokes `$s3EncryptionClient->putObjectAsync()`, let's defer the heavy lifting that makes up the encryption work and other calls until they are executed. If someone invokes `$s3EncryptionClient->putObject()`, the work is immediately waited for anyways.

jeskew: gotcha that makes sense

PR #1402, opened 2017-10-18
2.8

alshamiri3:


kstich:
These branches are for different major versions and will not be merged.

PR #1414, opened 2017-10-31
HHVM version in User-Agent

kstich:
Add HHVM version to User-Agent if it is defined. Expand ClientResolver test to handle it optionally.


kstich:
A call to `define()` cannot be undone and is, therefore, not safely testable due to side effects. That makes for 2 of the 3 lines affected; this is a hit we have to take on coverage.

PR #1423, opened 2017-11-10
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/.changes/nextrelease/use-phpunit-framework-test
+++ b/.changes/nextrelease/use-phpunit-framework-test
@@ -1,0 +1,7 @@
+[
+   {
+       "type": "enhancement",
+       "category": "Test",
+       "description": "use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase"
+   }
+]
--- a/.changes/nextrelease/use-phpunit-framework-test
+++ b/.changes/nextrelease/use-phpunit-framework-test
@@ -2,6 +2,6 @@
    {
        "type": "enhancement",
        "category": "Test",
-       "description": "use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase"
+       "description": "Use PHPUnit\\Framework\\TestCase instead of PHPUnit_Framework_TestCase"
    }
 ]
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "require-dev": {
         "ext-openssl": "*",
         "ext-dom": "*",
-        "phpunit/phpunit": "^4.8.35|^5.4.0",
+        "phpunit/phpunit": "^4.8.35|^5.4.3",
         "behat/behat": "~3.0",
         "doctrine/cache": "~1.4",
         "aws/aws-php-sns-message-validator": "~1.0",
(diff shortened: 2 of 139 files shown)


kstich:
Thanks for doing these updates!

Can you please point to the documentation that states that a version req of `^5.7` is required for the updates in this PR? Is there something I'm missing that wasn't covered in [`5.4.3`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.4.md#543---2016-06-09)?

gabriel-caruso:
@kstich Actually, I didn’t know that was a previous version of `PHPUnit 5` that support this new `namespace`. Do you want me to down to `5.4.3`?

kstich:
When updating a dependency, it's best to verify that the new minimum you take is exactly the lowest necessary for the updates that are being made. Please go ahead and take this down to `^4.8.35|^5.4.3`.

kstich:
Can you please also add a [Changelog Document to this PR](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md#changelog-documents)?

gabriel-caruso:
Done :+1:

.changes/3.38.2
```diff
@@ -1,5 +1,9 @@
 [
     {
+        "type": "enhancement",
```
kstich: This update needs to go in its own file, not the previous merged change blob.

>A changelog document is a small JSON blob placed in the .changes/nextrelease folder. It should be named a clearly and uniquely, akin to a branch name.

You can find a sample [in this PR](https://github.com/aws/aws-sdk-php/pull/1376/files#diff-058c075a21877f6b3127b93c0b89bf32).

Nit since this has to move: the `"category"` should be the singular `"Test"` to match the namespaces for those files.

gabriel-caruso: I guess now everything is right. Sorry for the inconvenient.

.changes/nextrelease/use-phpunit-framework-test
```diff
@@ -0,0 +1,7 @@
+[
+   {
+       "type": "enhancement",
+       "category": "Test",
+       "description": "use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase"
```
kstich: Some of these characters will be removed due to backslashing. Can you also please capitalize the first letter in the description contents?

carusogabriel: Done

kstich:
Thanks @carusogabriel!

PR #1433, opened 2017-12-01
Add S3EncryptionMultipartUploader

kstich:
Create traits with client-side encryption related parameter and cipher handling functionality (`CryptoParamsTrait`, `CipherBuilderTrait`). Add `prepare_data_source` to `AbstractUploadManager` to execute before the `'initiate'` phase of the upload workflow, document this on the `Aws\Glacier\MultipartUploader` and `Aws\S3\MultipartUploader`. Uses the previously listed functionality to create the `S3EncryptionMultipartUploader`. Updates guides and tests for new functionality.


tests/Integ/S3EncryptionContext.php
```diff
@@ -112,7 +112,7 @@ public function iEncryptEachFixtureWith($wrapAlgorithm, $alias, $region, $cipher
                     $shortCipher = substr($cipher, 4);
                     break;
                 default:
-                    continue;
+                    continue 2;
```
cjyclaire: Does this integration test used to succeed with "continue"?

kstich: Yes, previously it would have just fallen out of the switch and `continue`d the loop via line 119 since `$shortCipher` would be `null` and not supported.

kstich:
There are no comparable tests for the `Glacier\MultipartUploader`. Since the changes exercise the underlying `AbstractUploadManager`, the tests would not be validating any functionality that is different from the `S3MultipartUploader`.

kstich:
Anything beyond our previous conversations, @jeskew?

PR #1463, opened 2018-01-09
Retry Specific cURL Error Codes

### LINKED ISSUE #924: random cURL errors on HTTPS requests to SWF and DynamoDB
Hello!

we recently switched on our EC2 instances from the v2 to v3 (most recent version) of the PHP SDK. We use SWF, DynamoDB and S3. Since we switched to the v3, we are facing cURL errors that appear completely randomly it seems:

The application has thrown an exception!
 Aws\Swf\Exception\SwfException
 Error executing "PollForActivityTask" on "https://swf.us-east-1.amazonaws.com"; AWS HTTP error: cURL error 56: SSL read: error:00000000:lib(0):func(0):reason(0), errno 104 (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)

Sometimes the exception appears twice per hour, sometimes once in three hours... there's no scheme.

We already switched back to an older AMI and updated cURL from 7.35.0 (there is a problem with chunked upload in the 7.35) to 7.36.0 but nothing helps. We are polling SWF for activity tasks with long polling enabled, so there shouldn't be too many requests.

I googled this of course before opening an issue, but the only topics I found either date from 2012-2013 or were related to a broken load balancer on AWS. And that would probably be too much of a coincidence in our case...

Does anyone know about this problem..? I can't even say if it's a problem in the SDK or something else. So I'm grateful for every hint you can give me!

Thanks a lot!
Pascal

kstich:
Implements a method to retry `CURLE_*_ERROR` based failures for `AwsExceptions` preceeded by `ConnectExceptions` that do not return true from `->isConnectionError()`.

Resolves #924 

/cc @jeskew

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -2,7 +2,9 @@
 namespace Aws;
 
 use Aws\Exception\AwsException;
+use GuzzleHttp\Exception\ConnectException;
 use Psr\Http\Message\RequestInterface;
+use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Promise;
 
@@ -28,6 +30,10 @@ class RetryMiddleware
         'RequestThrottled'                       => true,
         'BandwidthLimitExceeded'                 => true,
         'RequestThrottledException'              => true,
+    ];
+
+    private static $retryCurlErrors = [
+        CURLE_RECV_ERROR => true,
     ];
 
     private $decider;
@@ -56,13 +62,14 @@ class RetryMiddleware
      */
     public static function createDefaultDecider($maxRetries = 3)
     {
+        $guzzleMajorVersion = (string) ClientInterface::VERSION[0];
         return function (
             $retries,
             CommandInterface $command,
             RequestInterface $request,
             ResultInterface $result = null,
             $error = null
-        ) use ($maxRetries) {
+        ) use ($maxRetries, $guzzleMajorVersion) {
             // Allow command-level options to override this value
             $maxRetries = null !== $command['@retries'] ?
                 $command['@retries']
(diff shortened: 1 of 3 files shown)


tests/RetryMiddlewareTest.php
```diff
@@ -81,6 +83,33 @@ public function testDeciderIgnoresPHPError()
...
+                new \GuzzleHttp\Message\Request('GET', 'http://www.example.com')
+            );
+        }
+        $err = new AwsException(
+            'e',
+            $command,
+            ['connection_error' => false],
+            $previous
+        );
+        $this->assertTrue($decider(0, $command, $request, null, $err));
```
jeskew: My impression from https://github.com/aws/aws-sdk-php/issues/924#issuecomment-355646600 was that `CURLE_RECV_ERROR` would not be retried by default. Did you mean it wouldn't be retried by Guzzle but would be retried by the SDK?

kstich: I've left a [new comment](https://github.com/aws/aws-sdk-php/issues/924#issuecomment-356445082) on that thread (reproduced below) that states the change:

> After some further discussion based around the operations being handled, it is believed we can retry this at the SDK level since it's not safe to do at the Guzzle level. Support is being added for automatic retry of `CURLE_RECV_ERROR` via #1463.

src/RetryMiddleware.php
```diff
@@ -80,6 +87,18 @@ public static function createDefaultDecider($maxRetries = 3)
                 return true;
             } elseif (isset(self::$retryStatusCodes[$error->getStatusCode()])) {
                 return true;
+            } elseif (($previous = $error->getPrevious())
+                && $previous instanceof ConnectException) {
+                if ($guzzleMajorVersion === '6') {
+                    return isset(self::$retryCurlErrors[$previous->getHandlerContext()['errno']]);
```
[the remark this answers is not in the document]
kstich: Guzzle5's [`ConnectException`](https://github.com/guzzle/guzzle/blob/5.3/src/Exception/ConnectException.php) is just a sub-classed [`RequestException`](https://github.com/guzzle/guzzle/blob/5.3/src/Exception/RequestException.php#L13). The `RequestException` doesn't surface any of the information in `getHandlerContext()` (which doesn't exist in Guzzle5) from what I can tell.

src/RetryMiddleware.php
```diff
@@ -80,6 +87,18 @@ public static function createDefaultDecider($maxRetries = 3)
                 return true;
             } elseif (isset(self::$retryStatusCodes[$error->getStatusCode()])) {
                 return true;
+            } elseif (($previous = $error->getPrevious())
+                && $previous instanceof ConnectException) {
+                if ($version[0] === '6') {
```
jeskew: Rather than checking the version, couldn't you check if the `getHandlerContext` method exists?

PR #1465, opened 2018-01-10
Update `Aws\PhpHash::$context` docblock with `HashContext` type.

qmcree:
With PHP 7.2, the `hash` extension was updated to use objects instead of resources. See https://secure.php.net/manual/en/migration72.incompatible.php#migration72.incompatible.hash-ext-to-objects .

### DIFF
--- a/src/PhpHash.php
+++ b/src/PhpHash.php
@@ -1,14 +1,12 @@
 <?php
 namespace Aws;
 
-use HashContext;
-
 /**
  * Incremental hashing using PHP's hash functions.
  */
 class PhpHash implements HashInterface
 {
-    /** @var HashContext */
+    /** @var resource|\HashContext */
     private $context;
 
     /** @var string */
@@ -65,7 +63,7 @@ class PhpHash implements HashInterface
     /**
      * Get a hash context or create one if needed
      *
-     * @return resource
+     * @return resource|\HashContext 
      */
     private function getContext()
     {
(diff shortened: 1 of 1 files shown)


src/PhpHash.php
```diff
@@ -1,12 +1,14 @@
 <?php
 namespace Aws;
 
+use HashContext;
```
kstich: I don't believe doc blocks will require having the `use` statement for the class.

src/PhpHash.php
```diff
@@ -1,12 +1,14 @@
...
+
 /**
  * Incremental hashing using PHP's hash functions.
  */
 class PhpHash implements HashInterface
 {
-    /** @var resource */
+    /** @var HashContext */
```
kstich: This should be `resource|HashContext` since it may also be a `resource` type for older versions of PHP. The same change should apply to the `@return` annotation on [Line 68](https://github.com/aws/aws-sdk-php/pull/1465/files#diff-9870abb8d621b0ea2887e4ba5cc617b7R68)

src/PhpHash.php
kstich: That's an acceptable change, it's done [in other places](https://github.com/aws/aws-sdk-php/blob/88171efe2ad8b048cd0459957ae69eaf656dbf80/src/S3/StreamWrapper.php#L79).

PR #1466, opened 2018-01-11
S3 "Size" Update for ListObjects, ListObjectsV2, ListObjectVersions, and ListParts

### LINKED ISSUE #1458: S3Client listObjectsV2 Size maxes out at 2147483647
I am working with the AWS SDK to pull information from our S3 buckets, and I am running into some issues with the Size property. It seems that for any file larger than 2 GB, the result of listObjectsV2 is returning up to 2147483647 for the Size. This is the max value for a signed integer.

We need to be able to get the actual size of the file as it exists on S3, and I am hoping this is a simple enough fix to get implemented. If not, maybe I am going about this the wrong way? Or maybe some other option is available to get the real size of the file.

Any response is appreciated.

Thanks
Brian

kstich:
Updates the type of the S3 `'Size'` object to `long` (from `integer`) to properly reflect objects over `PHP_INT_MAX` in size. This will affect `ListObjects`, `ListObjectsV2`, `ListObjectVersions`, and `ListParts`. This bug fix may be a breaking change for customers who relied on the previously `PHP_INT_MAX` capped behavior or on the type of the field. You may see objects at their full size, as a string, if they are over `PHP_INT_MAX` in size.

The `"feature"` designation is to ensure this is marked as a minor version due to the potentially breaking change.

Fixes #1458

/cc @jeskew


xthiago:
@kstich is correct. This PR is a breaking change for those that uses `strict_types=1` and rely on the type of fields. 😞 

I would like to suggest the release of changes like that as major in the future.

PR #1471, opened 2018-01-17
RDS PreSignedUrl Missing DestinationRegion Fix

### LINKED ISSUE #1462: RDS CopyDBClusterSnapshot error: PreSignedurl could not be authenticated
I’m trying to make a CopyDBClusterSnapshot request to copy a cluster snapshot from us-east-1 to us-west-2 from the AWS SDK for PHP v3.48.7 installed via composer. The following is my code:

```
$copyResult = $clients['RDSDestination']->copyDBClusterSnapshot(['CopyTags' => true,
                     'SourceDBClusterSnapshotIdentifier' => $sourceSnapshotArn,
                     'SourceRegion' => $sourceRegion,
                     'TargetDBClusterSnapshotIdentifier' => $destinationSnapshotId,
                     
![rolepermissions](https://user-images.githubusercontent.com/22547771/34685984-29ebddfc-f478-11e7-8c1f-6f5926ee7183.PNG)
'KmsKeyId' => $kmsKeyIdWest]);
```

$clients[‘RDSDestination’] is a RDS client in us-west-2 authenticated with RDS access to the attached snapshot of permissions (“RolePermissions.PNG”). $sourceSnapshotArn is the same ARN as used below in the CLI call, $sourceRegion is ‘us-east-1’, $destinationSnapshotId is the same as used below in the CLI call (except with a 2 instead of a 1 so that it's unique) and $kmsKeyIdWest is the default RDS KMS key for us-west-2. The following is the error message:

```
Error executing "CopyDBClusterSnapshot" on "https://rds.us-west-2.amazonaws.com"; AWS HTTP error: Client error: `POST https://rds.us-west-2.amazonaws.com` resulted in a `400 Bad Request` response:
<ErrorResponse xmlns="http://rds.amazonaws.com/doc/2014-10-31/">
  <Error>
    <Type>Sender</Type>
    <Code>InvalidPara (truncated...)
InvalidParameterValue (client): PreSignedurl could not be authenticated - <ErrorResponse xmlns="http://rds.amazonaws.com/doc/2014-10-31/">
  <Error>
    <Type>Sender</Type>
    <Code>InvalidParameterValue</Code>
    <Message>PreSignedurl could not be authenticated</Message>
  </Error>
  <RequestId>35daae55-e1fa-49d1-bd96-cef6513eb0a9</RequestId>
</ErrorResponse>
```

It’s also worth noting that if I do the following command with the same credentials and in us-west-2, the snapshot successfully copies:

`aws rds copy-db-cluster-snapshot --source-region us-east-1 --source-db-cluster-snapshot-identifier cluster-snapshot-arn --target-db-cluster-snapshot-identifier copy-snapshot-test-1 --kms-key-id default-rds-kms-key`

Obviously, the identifiers and kms-key-id are different, but I just put in placeholders.

I’ve spent a while looking through documentation and other user’s issues trying to figure this out. I’ve found two different places that appear to say two different things:

1. The API reference (https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-rds-2014-10-31.html#copydbclustersnapshot) says that PreSignedUrl (seen in the error message, and also a parameter for the command) is computed by the SDK (“The SDK will compute this value for you on your behalf.”).

2. In the cross-region replication user guide (https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CopyDBClusterSnapshot.CrossRegion.html), under “Copying an Encrypted DB Cluster Snapshot by Using the AWS CLI or Amazon RDS API” -> “API”, is says that PreSignedUrl must be specified in the call, and points me to other pages to construct such a request (which I’ve tried, but get the same error).

I don’t know if I’m doing or reading something wrong, but I’m out of ideas.

kstich:
Fixes a bug where `'DestinationRegion'` was not being added to automatically generated `'PreSignedUrl'` parameters for RDS operations (`CopyDBSnapshot`, `CreateDBInstanceReadReplica`, `CopyDBClusterSnapshot`, and `CreateDBCluster`.)

Fixes #1462 
Rebuild of #1468 

/cc @jeskew


src/data/rds/2014-10-31/api-2.json
```diff
@@ -1844,6 +1844,7 @@
         "TargetDBClusterSnapshotIdentifier":{"shape":"String"},
         "KmsKeyId":{"shape":"String"},
         "PreSignedUrl":{"shape":"String"},
+        "DestinationRegion":{"shape":"String"},
```
jeskew: Does the model need to be modified if the member is being added in `Aws\Rds\RdsClient::applyDocFilters`?

kstich: Yes, `applyDocFilters` functions are only called at time of doc building, not at time of json model compiling.

PR #1474, opened 2018-01-19
Simplify return

carusogabriel:


.changes/nextrelease/simplify_return.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "Aws",
```
kstich: This should be "Api\\\\Parser" to reflect the namespace.

PR #1487, opened 2018-02-01
Removed useless else after return

alexmanno:
As a title, I removed the useless else after returns.

(Without changing the logic)


alexmanno:
@kstich I fixed the style problems as you said

alexmanno:
Ping @kstich

PR #1496, opened 2018-02-14
Add missing S3 url format not supported

eliysha:
Related to https://github.com/aws/aws-sdk-php/pull/1385

### DIFF
--- a/.changes/nextrelease/s3_uri_parser
+++ b/.changes/nextrelease/s3_uri_parser
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "S3",
+        "description": "Add S3 url format not supported"
+    }
+]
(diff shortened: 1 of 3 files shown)


kstich:
Hi @eliysha, the [`S3\StreamWrapper`](https://github.com/aws/aws-sdk-php/blob/master/src/S3/StreamWrapper.php) only supports the `://bucket/key` format after the passed `$protocol` target (default `'s3'`). It relies on the passed `$client` for handling the region of the bucket. This URI would successfully parse out if the scheme was `https` instead. Are you able to successfully use the URI you're testing for in any case? I don't see where the SDK would provide this URI, how is it generated?

Adding this parsing to the `S3UriParser` when it is not supported by the `S3StreamWrapper` would create a false expectation. I don't think adding the related capabilities to the `S3StreamWrapper` makes sense either, given the above information about `$client`.

PR #1497, opened 2018-02-14
S3\Transfer::construct(): correct docs

danielbeardsley:
The docs didn't match the implementation, adjust the docs to be
correct.

It's been this way since 2015 -- 74f8c50c0bde2ef8d8ec57b195c4a20857056f9f


jschwarzwalder:
I will verify that the new instructions are correct.

Thank you for submitting the change.

PR #1506, opened 2018-03-05
CURLE_RECV_ERROR  error retry fixed

Halama:
Hello,
we have upgraded to the latest AWS SDK because the retry of `CURLE_RECV_ERROR` was fixed in https://github.com/aws/aws-sdk-php/releases/tag/3.52.0

Unfortunately after deploy of the new application version with updated AWS SDK we are still receiving occasional `CURLE_RECV_ERROR` errors.

I've tried to debug a code and I've noticed that `GuzzleHttp\Exception\ConnectException` exception is expected in AWS SDK https://github.com/aws/aws-sdk-php/blob/299ac47ff5171d6a25f788dc333c0a3c8db40f69/src/RetryMiddleware.php#L104 but `GuzzleHttp\Exception\RequestException` is  thrown in Guzzle https://github.com/guzzle/guzzle/blob/748d67e23a984e7a2562847994fd133fd4e629a6/src/Handler/CurlFactory.php#L154 . 
So I think that `CURLE_RECV_ERROR` is never retried. 
This PR should fix it.

thanks

### DIFF
--- a/.changes/nextrelease/fix_curl_retry.json
+++ b/.changes/nextrelease/fix_curl_retry.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "bugfix",
+    "category": "RetryMiddleware",
+    "description": "Retry the CURLE_RECV_ERROR error."
+  }
+]
--- a/.changes/nextrelease/fix_curl_retry.json
+++ b/.changes/nextrelease/fix_curl_retry.json
@@ -2,6 +2,6 @@
   {
     "type": "bugfix",
     "category": "RetryMiddleware",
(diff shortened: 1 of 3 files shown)


.changes/nextrelease/fix_curl_retry.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "bugfix",
+    "category": "RetryMiddleware",
+    "description": "Retry the CURLE_RECV_ERROR error."
```
kstich: Can you please update this to read:

>Retries CURLE_RECV_ERROR on all RequestException, not just ConnectException.

PR #1514, opened 2018-03-20
Signatures to be generated from any \DateTimeInterface

### LINKED ISSUE #1513: Widen SignatureV4 timing support to DateTimeInterface
SignatureV4::convertToTimestamp checks against \DateTime instead of \DateTimeInterface. Pushing a PR to fix.

theMadness:
Added the ability for signatures to be generated off of `\DateTimeInterface` objects that are not based on `\DateTime`.

Resolves #1513

### DIFF
--- a/src/Signature/SignatureV4.php
+++ b/src/Signature/SignatureV4.php
@@ -292,7 +292,7 @@ class SignatureV4 implements SignatureInterface
 
     private function convertToTimestamp($dateValue, $relativeTimeBase = null)
     {
-        if ($dateValue instanceof \DateTime) {
+        if ($dateValue instanceof \DateTimeInterface) {
             $timestamp = $dateValue->getTimestamp();
         } elseif (!is_numeric($dateValue)) {
             $timestamp = strtotime($dateValue,


src/Signature/SignatureV4.php
```diff
@@ -292,7 +292,7 @@ private function getCanonicalizedQuery(array $query)
 
     private function convertToTimestamp($dateValue, $relativeTimeBase = null)
     {
-        if ($dateValue instanceof \DateTime) {
+        if ($dateValue instanceof \DateTimeInterface) {
```
kstich: Can you add test(s) that pass in objects that aren't `\DateTime` but extend `\DateTimeInterface`? Looks like that's only [`\DateTimeImmutable`](http://php.net/manual/en/class.datetimeimmutable.php).

theMadness: Well, any object that implements that interface from scratch, or extend \DateTimeImmutable would also match. Personally I'm not entirely sure what we should be testing here. \DateTime implements \DateTimeInterface and already passes the tests.

kstich: The thought is that the current tests already passed with this line as `\DateTime`, the addition of a test with `\DateTimeImmutable` expresses that the change is working as intended.

PR #1517, opened 2018-03-28
Updated gitattributes

KasperFranz:
this includes codecov.yml which shouldn't be needed in the export


KasperFranz:
A discussion about this is also .changes is also worth having as that folder is distrubted currently

KasperFranz:
Any eta on this?

kstich:
I think a discussion around the `.changes` folder could be warranted, given we currently exclude `CHANGELOG.md`. How about an issue (and potentially PR) about it so we can have some time for community comment?

PR #1519, opened 2018-04-01
use namespaced phpunit

DQNEO:
additional fix to https://github.com/aws/aws-sdk-php/pull/1423

### DIFF
--- a/tests/S3/Crypto/S3EncryptionMultipartUploaderTest.php
+++ b/tests/S3/Crypto/S3EncryptionMultipartUploaderTest.php
@@ -10,7 +10,7 @@ use Aws\Test\UsesServiceTrait;
 use Aws\Test\Crypto\UsesMetadataEnvelopeTrait;
 use GuzzleHttp\Psr7;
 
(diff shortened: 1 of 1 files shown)


tests/S3/Crypto/S3EncryptionMultipartUploaderTest.php
```diff
@@ -10,7 +10,7 @@
 use Aws\Test\Crypto\UsesMetadataEnvelopeTrait;
 use GuzzleHttp\Psr7;
 
-class S3EncryptionMultipartUploaderTest extends \PHPUnit_Framework_TestCase
```
kstich: Instead of specifying the whole namespace in the `extends`, can you port this to a `use` statement and just reference the class name?

```php
use PHPUnit\Framework\TestCase;

class S3EncryptionMultipartUploaderTest extends TestCase
```

DQNEO: Sure, I did it d14c8e3

PR #1571, opened 2018-06-19
Service id fallback

howardlopez:
Added fallback generation of Service ID property.


tests/Api/ServiceTest.php
```diff
@@ -80,6 +80,39 @@ function () { return []; }
         $this->assertArrayHasKey('foo', $s->getOperations());
     }
 
+    public function serviceIdDataProvider()
```
kstich: Would be nice to have a set of cases in here with trailing numbers as well.

howardlopez: Added a couple of cases with trailing numbers.

PR #1604, opened 2018-08-18
compatible to namespaced phpunit classes

DQNEO:
As of PHPUnit 6, Long name classes (`\PHPUnit_Framework_*`)  don't exist any more.
So we should prepare for it.


DQNEO:
Sure! did it. 1e6b06a9e5e8dfcf993d800decb84d810b821634

.changes/3.64.16
```diff
@@ -0,0 +1,7 @@
+[
```
kstich: Missed this in the other PR so I'm submitting an update, but these should go in a unique file in the `.changes/nextrelease/` folder.

DQNEO:
OK, both PRs are made for the same purpose, I merged the messages into one line.

kstich:
Still needs to move to the proper location like `.changes/nextrelease/namespaced_phpunit_classes.json`.

DQNEO:
Oh, sorry I completely misunderstood the rule.
I renamed it.  73d72101

PR #1605, opened 2018-08-18
Compat expect exception

DQNEO:
As of PHPUnit6, `setExpectedException()` does not exist.
So we should prepare for it.


DQNEO:
Sure! I did it 5ee9421

PR #1636, opened 2018-09-27
Lambda TCP Keep-Alive

howardlopez:
Added default TCP Keep-Alive Curl setting for the Lambda client.

### DIFF
--- a/.changes/nextrelease/lambda_tcp_keepalive.json
+++ b/.changes/nextrelease/lambda_tcp_keepalive.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "Lambda",
+        "description": "Added default TCP Keep-Alive Curl setting for the Lambda client."
+    }
+]
--- a/.changes/nextrelease/lambda_tcp_keepalive.json
+++ b/.changes/nextrelease/lambda_tcp_keepalive.json
@@ -1,6 +1,6 @@
 [
     {
-        "type": "enhancement",
+        "type": "feature",
         "category": "Lambda",
         "description": "Added default TCP Keep-Alive Curl setting for the Lambda client."
     }
--- a/src/Lambda/LambdaClient.php
+++ b/src/Lambda/LambdaClient.php
@@ -72,14 +72,6 @@ use Aws\Middleware;
 class LambdaClient extends AwsClient
 {
     /**
-     * Default Curl options for Lambda
-     * @var array
-     */
-    static $defaultCurlOptions = [
-        CURLOPT_TCP_KEEPALIVE => 1,
-    ];
-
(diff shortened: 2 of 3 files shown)


tests/Lambda/LambdaClientTest.php
```diff
@@ -0,0 +1,32 @@
...
+            'region' => 'us-east-1',
+            'version' => 'latest'
+        ]);
+
+        $list = $client->getHandlerList();
+        $list->setHandler(function ($command, $request) {
+            $this->assertArraySubset(
+                [
+                    'curl' => LambdaClient::$defaultCurlOptions,
+
```
kstich: Extra whitespace.

src/Lambda/LambdaClient.php
```diff
@@ -67,4 +69,40 @@
...
  */
-class LambdaClient extends AwsClient {}
+class LambdaClient extends AwsClient
+{
+    /**
+     * Default Curl options for Lambda
+     * @var array
+     */
+    static $defaultCurlOptions = [
+        CURLOPT_TCP_KEEPALIVE => 1,
```
kstich: This configuration should be local to the middleware generation below and behind a `curl` extension check since we don't require cURL. [See this PR and it's related bug](https://github.com/aws/aws-sdk-php/pull/1481/files#diff-85d927b5b18f9813bfd8535665c1e98dR62).

.changes/nextrelease/lambda_tcp_keepalive.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
```
kstich: Since this is changing a transport level option, I'd recommend adding setting this as a `"feature"` level to get a minor version bump.

src/Lambda/LambdaClient.php
```diff
@@ -67,4 +69,40 @@
...
+
+    /**
+     * Provides a middleware that sets default Curl options for the command
+     *
+     * @return callable
+     */
+    public function getDefaultCurlOptionsMiddleware()
+    {
+        return Middleware::mapCommand(function (CommandInterface $cmd) {
+            if (!isset($cmd['@http']['curl'])) {
```
kstich: Is this wanted for all operations or just a subset?

src/Lambda/LambdaClient.php
microrhythms: Agreed about the `curl` extension check, but keeping this as a class variable allows re-use for testing, minimizing future code changes should options change.

src/Lambda/LambdaClient.php
howardlopez: Current scope is for all operations.

PR #1644, opened 2018-10-11
Add MaxRetriesExceeded entry to ApiCall events

howardlopez:


kstich:
Looks like this could use a tiny update for test coverage.

PR #1647, opened 2018-10-15
Bump patch version of Guzzle for PHP 7.3 compatibility

carusogabriel:
See https://github.com/guzzle/guzzle/pull/2112 and https://github.com/laravel/framework/pull/26125

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ php:
   - 7.0
   - 7.1
   - 7.2
+  - 7.3
   - hhvm
   - nightly
 env:
@@ -14,6 +15,7 @@ env:
 
 matrix:
   allow_failures:
+    - php: 7.3
     - php: hhvm
     - php: nightly
   fast_finish: true
--- a/.changes/nextrelease/bump-guzzlehttp
+++ b/.changes/nextrelease/bump-guzzlehttp
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -6,16 +6,16 @@ php:
...
+  - 7.3
   - hhvm
-  - nightly
 env:
   - COMPOSER_OPTS=""
   - COMPOSER_OPTS="--prefer-lowest"
 
 matrix:
   allow_failures:
+    - php: 7.3
```
carusogabriel: As PHP 7.3 has reached "Release Candidates" versions, I believe we can drop this. But, I'll leave to the maintainers.

.travis.yml
```diff
@@ -6,16 +6,16 @@ php:
   - 7.0
   - 7.1
   - 7.2
+  - 7.3
   - hhvm
-  - nightly
```
howardlopez: Let's leave `nightly` in here. It may be redundant with 7.3 for some period of time, but going forward we'll always want to test against the latest build.

.travis.yml
```diff
@@ -6,16 +6,16 @@ php:
...
 env:
   - COMPOSER_OPTS=""
   - COMPOSER_OPTS="--prefer-lowest"
 
 matrix:
   allow_failures:
+    - php: 7.3
     - php: hhvm
-    - php: nightly
```
howardlopez: As above, let's leave `nightly` in here.

Issue #1655, opened 2018-10-23
Cannot redeclare Aws\manifest()


tbarry-eminence:
## Issue Overview
`Cannot redeclare Aws\manifest() (previously declared in /vendor/aws/aws-sdk-php/src/functions.php:341)`

## Expected Behavior
`src/functions.php` included once.

## Current Behavior
`src/functions.php` included twice.

This is possibly related to the fact that the composer.json file defines an autoload for both the `Aws` classmap at `src/` and the files at `src/functions.php`.

```json
"autoload": {
    "psr-4": {
        "Aws\\": "src/"
    },
    "files": ["src/functions.php"]
}
```

## Possible Solution
I'm not entirely certain if composer will autoload namespaced functions if they are in the namespaced directory, but perhaps this would resolve it? (I really only have a fundamental understanding of composer package management).

Alternatively, functions inside of `src/functions.php` could be wrapped in a `function_exists()` if check eg.

```php
if (!function_exists('manifest'))
{
    function manifest($service = null)
    {
       //  ...
    }
}
```

## Steps to Reproduce
1. `composer require aws/aws-sdk-php`
2. `require_once 'vendor/autoload.php';`

## Context (Environment)
OS: Ubuntu 16.04
PHP: 7.1.22-1
Composer: 1.7.2

## Detailed Description
I've seen a couple variants of this posted either as issues here or on stackoverflow, but most seem to be caused be including composers `vendor/autoload.php` multiple times. I've been quite thorough within my companies web project to confirm that this is not the case here. I am certain we are *only including composers vendor/autoload.php once*.

I've attempted to recreate this issue as it's own project, but admit that it does seem to work fine outside of my companies environment. Unfortunately I cannot share my business code, and while I fully admit this then has to be an issue on my end I am at a loss for how to resolve it without altering the `aws/aws-sdk-php` package.

I do not believe this to be a big request, but it would certainly resolve my issue of implementing the AWS PHP SDK into my companies application.

Thanks,

Tony B.

howardlopez:
First of all, thank you for your well-structured issue report. 

A couple of things to try:
- Delete your `composer.lock` file and your `vendor` directory, and then run `composer update` again. This will basically reinstall all your dependencies from scratch and could help clear up any funkiness in your composer install.
[...]

PR #1657, opened 2018-10-23
Fix an issue where functions inside of src/functions.php were declared multiple times.

tbarry-eminence:
This pull request applies to a fix to issue #1655 by wrapping all functions inside of functions.php with an function_exists() check.

Example:

```php
if (!function_exists(__NAMESPACE__ . '\\function_name')) {
    function function_name()
    {
        // ...
    }
}
```


howardlopez:
Closing this PR as discussed in #1655 .

PR #1673, opened 2018-11-18
Use in_array function to simplify foreach

carusogabriel:


.changes/nextrelease/simplify-foreach-with-in-array
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "Aws",
```
howardlopez: The Aws namespace is added by default, so the category here should just be "Waiter".

howardlopez:
Thanks for making the change, will be merged after Reinvent.

Issue #1694, opened 2018-12-18
S3 PutObject is missing a Content-MD5 header when writing to a bucket with a default object lock configuration


huubvdw:
When writing a new object to a bucket with a default object lock configuration, the request will result is a 400 bad request response. ` InvalidRequest: Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters`

I noticed in [ApplyChecksumMiddleware](https://github.com/aws/aws-sdk-php/blob/master/src/S3/ApplyChecksumMiddleware.php) the PutObject operation only sets a sha256 content hash header. 

As you don't know upfront if a bucket has a object lock configuration, I'm thinking of setting both the sha256 content hash header and the content MD5 header for the PutObject operation. 

Any other suggestions?

howardlopez:
@huubvdw, your investigation and PR are appreciated. Unfortunately, `PutObject` only requires the `Content-MD5` header when there are object lock parameters, so we won't be able to accept your PR to change the general `PutObject` behavior in this way. 

S3 and the SDK teams are aware of the issue and have discussed options internally, but for now, you'll have to supply your own `ContentMD5` parameter to object lock requests. For example:

```
$client->putObject([
    'Bucket' => 'your_bucket',
    'Key' => 'your_key',
    'Body' => 'your_body',
    'ObjectLockMode' => 'GOVERNANCE',
    'ObjectLockRetainUntilDate' => $someDate
    'ContentMD5' => base64_encode(md5('your_body', true))
]);
```

mmomasters:
Hi everyone, I am having the same issue and was wondering where can I supply my own ContentMD5 parameter for s3fs to work correctly while using guidance ?

I tried s3fs with -o enable_content_md5 with no success

Thank you!

harshadparekh:
This suggestion of passing contentMD5 with PutObject() request assumes that the data we want to write as an object in AWS is available as a file locally so that we can pre-compute the MD5 on the content. For cases where we're generating data on the fly and providing it as an input stream to PutObject() request, how do we get the MD5 to pass alongwith? Also, how does this work when using MultiPartUpload for large objects?

TheJosh:
It appears that the `ContentMD5` field is not listed in the SDK Docs:
https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3-2006-03-01.html#putobject

Does the field still exist, even though it's not in the docs?

dean-huczok:
[...]

PR #1695, opened 2018-12-19
Set Content-MD5 header for PutObject operation

huubvdw:
Fix issue for missing Content-MD5 header on PutObject operation for object lock configured S3 buckets.

Addresses issue #1694


howardlopez:
Thank you for the PR, but it can't be accepted per [the discussion on the associated issue](https://github.com/aws/aws-sdk-php/issues/1694#issuecomment-449115740).

PR #1709, opened 2019-01-08
Avoid repeatedly include same files for variables

hezachary:
PHP include (require) only release file handles after script stop, use include for variables, may cause "failed to open stream too many open files" error.
I found this problem when I run coverage test, also happen to long running tasks (such as use loop to load sqs without quit script 24/7).

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -141,9 +141,18 @@ function or_chain()
  */
 function load_compiled_json($path)
 {
+    static $compiledList = [];
+
     $compiledFilepath = "{$path}.php";
-    if (is_readable($compiledFilepath)) {
-        return include($compiledFilepath);
+
+    if(!isset($compiledList[$compiledFilepath])) {
+        if (is_readable($compiledFilepath)) {
+            $compiledList[$compiledFilepath] = include($compiledFilepath);
+        }
+    }
+
+    if (isset($compiledList[$compiledFilepath])) {
+        return $compiledList[$compiledFilepath];
     }
 
     if (!file_exists($path)) {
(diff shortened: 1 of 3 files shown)


tests/FunctionsTest.php
```diff
@@ -93,6 +99,38 @@ public function testUsesPhpCompilationOfJsonIfPossible()
...
+        file_put_contents(
+            $compiledPath,
+            '<?php return ' . var_export($soughtData, true) . ';',
+            LOCK_EX
+        );
+
+        $this->assertSame($soughtData, Aws\load_compiled_json($jsonPath));
+        $compiledAtime = fileatime($compiledPath);
+
+        sleep(1);
```
howardlopez: Maybe use [usleep](https://www.php.net/manual/en/function.usleep.php) to take less time?

srchase: fileatime() returns seconds, so we need a full second to guarantee there would be difference in the access times on the test.

src/functions.php
```diff
@@ -141,9 +141,18 @@ function or_chain()
  */
 function load_compiled_json($path)
 {
+    static $compiledList = [];
+
     $compiledFilepath = "{$path}.php";
-    if (is_readable($compiledFilepath)) {
-        return include($compiledFilepath);
+
+    if(!isset($compiledList[$compiledFilepath])) {
```
howardlopez: Style: spacing after `if`

srchase: added space.

srchase:
@hezachary 

Thanks for opening this PR!

Issue #1722, opened 2019-02-04
Api\Validator incorrectly erring on assoc. arrays that contain a `0` index


bwg:
The heuristic for determining whether an array is assoc `Api\Validator::checkAssociativeArray` fails when an array contains a `0` index, regardless of the other keys.  

This is not correct.  The simple presence of a `0` index does not indicate a non-assoc array.  Both `json_encode` and `Aws\DynamoDb\Marshaler::marshalValue()` consider an array non-assoc only when the keys are 0 based sequential integers.

given the array (actual data I'm working with):
```
$a = [
    0   => 4,
    1   => 6,
    "b" => 2,
    "d" => 1,
    "e" => 29,
]
```

json_encode will encode it as a map
```
json_encode($a);
{"0":4,"1":6,"b":2,"d":1,"e":29}
```

more importantly, so does `DynamoDB\Marshaler`

```
$m = new Aws\DynamoDb\Marshaler();
$m->marshalValue($a);
array(1) {
  'M' =>
  array(5) {
    [0] =>
    array(1) {
      'N' =>
      string(1) "4"
    }
    [1] =>
    array(1) {
      'N' =>
      string(1) "6"
    }
    'b' =>
    array(1) {
      'N' =>
      string(1) "2"
    }
    'd' =>
    array(1) {
      'N' =>
      string(1) "1"
    }
    'e' =>
    array(1) {
      'N' =>
      string(2) "29"
    }
  }
}
```

This manifests itself when validating a DynamoDB request, since DynamoDb has marshaled the data as a Map, but the validator treats it as an array.  Disabling validation in the SDK allows the request to successfully complete.

This issue was previously reported as a PR here: https://github.com/aws/aws-sdk-php/pull/930 but the proposed solution was rejected; for reasons I'm not entirely clear on.  

I'm submitting this as an issue instead of a PR because of the maintainers comments in #930.  As stated there are many ways to check whether an array is acting as a map or list, but I'm not sure what would be an acceptable solution.  Perhaps the way that `DynamoDb\Marshaler` does it?

At the end of the day validation has to be correct, or at the very least consistent. I'm happy to submit this as a PR with some guidance on an acceptable implementation.

Cheers

howardlopez:
[...]

PR #1731, opened 2019-02-20
Fixes validation on assoc arrays having a 0 index

bwg:
https://github.com/aws/aws-sdk-php/issues/1722

The existing heuristic for determining an assoc array looks only
for the presence of a `0` index.  However to correctly determine
whether a php array will be serialized (json) or marshaled (DynamoDB)
into a list (array) or map (assoc), the hueristic must be to check
for sequential integer keys starting at 0.

This change will walk the array checking the key against an incrementing
integer.  at any point where the key does not meet the expected integer
value, the iteration will stop and the array will be considered
associative. (this implementation is similar to the DynamoDB marshaler)

For any array not having a 0 index, this method will return after
the first iteration, so for the shape of the assoc arrays that this
method successfully validated before this change, there will be little
to no performance impact.

the only difference will be for arrays that have a 0 index, which this
method did NOT properly validate previously so there is no baseline
for a performance comparison.

### DIFF
--- a/.changes/nextrelease/api_validate_assoc_array
+++ b/.changes/nextrelease/api_validate_assoc_array
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "Api",
+        "description": "Fixes validation on assoc arrays having a 0 index."
+    }
+]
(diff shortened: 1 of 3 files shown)


bwg:
well I screwed this up bad.  put the changefile in the wrong dir, didn't have a local branch.  closing to save myself some dignity.

PR #1732, opened 2019-02-20
Fixes validation on assoc arrays having a 0 index

bwg:
https://github.com/aws/aws-sdk-php/issues/1722

The existing heuristic for determining an assoc array looks only
for the presence of a `0` index.  However to correctly determine
whether a php array will be serialized (json) or marshaled (DynamoDB)
into a list (array) or map (assoc), the hueristic must be to check
for sequential integer keys starting at 0.

This change will walk the array checking the key against an incrementing
integer.  at any point where the key does not meet the expected integer
value, the iteration will stop and the array will be considered
associative. (this implementation is similar to the DynamoDB marshaler)

For any array not having a 0 index, this method will return after
the first iteration, so for the shape of the assoc arrays that this
method successfully validated before this change, there will be little
to no performance impact.

the only difference will be for arrays that have a 0 index, which this
method did NOT properly validate previously so there is no baseline
for a performance comparison.

### DIFF
--- a/src/Api/Validator.php
+++ b/src/Api/Validator.php
@@ -249,20 +249,20 @@ class Validator
 
     private function checkAssociativeArray($value)
     {
-        $assoc = false;
+        $isAssociative = false;
 
         if (is_array($value)) {
-            $i   = 0;
+            $expectedIndex = 0;
             $key = key($value);
 
             do {
-                $assoc = $key !== $i++;
+                $isAssociative = $key !== $expectedIndex++;
                 next($value);
                 $key = key($value);
-            } while (!$assoc && null !== $key);
+            } while (!$isAssociative && null !== $key);
         }
 
-        if (!$assoc) {
+        if (!$isAssociative) {
             $this->addError('must be an associative array. Found '
                 . Aws\describe_type($value));
             return false;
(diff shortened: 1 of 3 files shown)


src/Api/Validator.php
```diff
@@ -249,7 +249,20 @@ private function checkCanString($value)
 
     private function checkAssociativeArray($value)
     {
-        if (!is_array($value) || isset($value[0])) {
+        $assoc = false;
```
howardlopez: Style nitpick: can the variable name be more descriptive, like `$isAssociative`?

src/Api/Validator.php
```diff
@@ -249,7 +249,20 @@ private function checkCanString($value)
...
+
+        if (is_array($value)) {
+            $i   = 0;
```
howardlopez: More style nitpicks:
* Can the variable be more descriptive, like `$expectedIndex`? 
* Can you remove the whitepsace after the variable?

tests/Api/ValidatorTest.php
```diff
@@ -160,6 +160,33 @@ public function validationProvider()
                 ['foo' => [1, 3]],
                 "Found 1 error while validating the input provided for the Foo operation:\n[foo] must be an associative array. Found array(2)"
             ],
+            // empty array must validate as an assoc array
```
howardlopez: Can you add in another test to check for the case of an array with the proper indexed array keys, but out of order? For example:
```
[2 => 'foo', 0 => 'bar', 1 => 'baz']
```

PR #1734, opened 2019-02-26
Clone sdk instance

martinssipenko:
*Description of changes:*
Adds ability to clone existing Sdk instance with extra arguments.

This is useful when Sdk is injected, but user needs to override some default arguments that were used when creating the Sdk object.


tests/SdkTest.php
```diff
@@ -63,4 +63,26 @@ public function testCreatesClientsWithAlias()
...
+            'version' => 'latest',
+            'region'  => 'us-east-1',
+        ]);
+
+        $this->assertSame(
+            'us-east-1',
+            $sdk->createDynamoDb()->getRegion()
+        );
+
+        $copy = $sdk->copy([
```
howardlopez: Can you add in a test to check that the copy keeps config options from the original (for options that were not overwritten)?

martinssipenko:
@howardlopez I've added an extra assertion, I hope it does what you meant.

PR #1752, opened 2019-03-18
CloudFront Signer: parse private key once

danielbeardsley:
openss_sign() is abysmally slow. On a fast CPU with a small amount of
data it takes 1.5ms of CPU time.

Turns out, half of that was parsing the private key. If we have
openssl parse the key ahead of time, `openssl_sign()` takes half the
time.

This change brought it from 1.5ms to 0.7ms per call, still really
slow but much better.

Note: We started running into performance issues with *lots* of image
urls being rendered on long pages.

---

### DIFF
--- a/tests/CloudFront/SignerTest.php
+++ b/tests/CloudFront/SignerTest.php
@@ -26,6 +26,21 @@ class SignerTest extends TestCase
     }
 
     /**
+     * Assert that the key is parsed during construction
+     *
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessageRegExp /PEM .*no start line/
+     */
+    public function testBadPrivateKey() {
+        $filename = tempnam(sys_get_temp_dir(), 'cloudfront-fake-key');
+        file_put_contents($filename, "Not a real private key");
+        $s = new Signer(
+            "not a real keypair id",
+            $filename
+        );
+    }
+
+    /**
      * @expectedException \InvalidArgumentException
      * @expectedExceptionMessage PK file not found
      */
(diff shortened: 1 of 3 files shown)


danielbeardsley:
> that verifies the key is only parsed once for multiple signings?

I've added a test that asserts parsing happens in the constructor, but there's no way of enforcing it doesn't also happen in the `sign()` call without further exposing the innards of the class (protected methods we could override).

Given that the output is identical before and after this change, I'd consider this an internal optimization and not a contract that should be enforced with a test.

We *could* find a way to test it (a protected one-line `parseKey()` function that we could ensure is only called once), but the test would still be lacking cause it'd be trivial to pass all tests by just only using the function in the constructor.

> Also, can you add a changelog entry, as specified in the [Contributing](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md) document?

Yes! Done!

tests/CloudFront/SignerTest.php
```diff
@@ -25,6 +25,21 @@ public function setUp()
...
 
+    /**
+     * Assert that the key is parsed during construction
+     *
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessageRegExp /PEM .*no start line/
+     */
+    public function testBadPrivateKey() {
+        $filename = tempnam(sys_get_temp_dir(), 'cloudfront-fake-key');
+        file_put_contents($filename, "Not a real private key");
```
howardlopez: Can you clean this up with an `unlink` after use? You may need to catch and rethrow the exception.

danielbeardsley: Done!  Went with a `finally` cause it's available in php 5.5 and it looks like this repo targets php 5.5 and above.

PR #1756, opened 2019-03-25
Make Multipart\AbstractUploadManager accept PHP 7+ Throwable as a failure type

### LINKED ISSUE #1755: TypeError: Argument 1 of Aws\Multipart\AbstractUploadManager::Aws\Multipart\{closure}() must be an instance of Exception, instance of Error given
The `AbstractUploadManager` is catching `Exception` types, however, if a PHP 7.0+ `Error` is thrown, the actual error that occurred is suppressed by the registered error handler of the promise. 

https://github.com/aws/aws-sdk-php/blob/bccb5b3509ed7517e50a75bba6b19ac4506aae70/src/Multipart/AbstractUploadManager.php#L140-L146

bradynpoulsen:
Closes #1755
#  ------------------------ >8 ------------------------
# Do not modify or remove the line above.
# Everything below it will be ignored.

Requesting a pull to aws:master from bradynpoulsen:bug/1755_php_7_throwable_support_by_multipart_abstractuploadmanager

Write a message for this pull request. The first block
of text is the title and the rest is the description.

### DIFF
--- a/src/Multipart/AbstractUploadManager.php
+++ b/src/Multipart/AbstractUploadManager.php
@@ -137,13 +137,29 @@ abstract class AbstractUploadManager implements Promise\PromisorInterface
             // Complete the multipart upload.
             yield $this->execCommand('complete', $this->getCompleteParams());
             $this->state->setStatus(UploadState::COMPLETED);
-        })->otherwise(function (\Exception $e) {
-            // Throw errors from the operations as a specific Multipart error.
-            if ($e instanceof AwsException) {
-                $e = new $this->config['exception_class']($this->state, $e);
-            }
-            throw $e;
-        });
+        })->otherwise($this->buildFailureCatch());
+    }
+
+    private function transformException($e)
+    {
+        // Throw errors from the operations as a specific Multipart error.
+        if ($e instanceof AwsException) {
+            $e = new $this->config['exception_class']($this->state, $e);
+        }
+        throw $e;
+    }
+
+    private function buildFailureCatch()
+    {
+        if (interface_exists("Throwable")) {
+            return function (\Throwable $e) {
+                return $this->transformException($e);
+            };
+        } else {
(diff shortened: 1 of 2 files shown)


src/Multipart/AbstractUploadManager.php
```diff
@@ -137,13 +137,29 @@ public function promise()
...
+        // Throw errors from the operations as a specific Multipart error.
+        if ($e instanceof AwsException) {
+            $e = new $this->config['exception_class']($this->state, $e);
+        }
+        throw $e;
+    }
+
+    private function buildFailureCatch()
+    {
+        if (PHP_MAJOR_VERSION >= 7) {
```
bradynpoulsen: This may be overkill, but I did want to propose a solution that kept the type-checking instead of just dropping it in favor of a naked `otherwise(function ($e)`

srchase: @bradynpoulsen 

This solution makes sense to keep type-checking here.  Could you add test coverage for this change?  

Also, instead of checking ```PHP_MAJOR_VERSION``` we could check Throwable directly: 
```
if (interface_exists('Throwable')) {
```

srchase: checking in again.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.92.2 - TBD
```
srchase: Refer to the instructions for adding a changelog here: [CONTRIBUTING - changelog](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md#changelog-documents).

This will get added to CHANGELOG.md automatically in the right release.

bradynpoulsen: That makes sense! Fixed, thanks :)

PR #1763, opened 2019-04-11
Fix code

tolbon:
*Issue #, if available:*

*Description of changes:*

- Replace literal array declaration with modern [] array declaration.
- Fix in S3\StreamWrapper a function say return true but never return true
- Fix in LambdaClient a double method have same name.

### DIFF
--- a/src/Lambda/LambdaClient.php
+++ b/src/Lambda/LambdaClient.php
@@ -46,6 +46,8 @@ use Aws\Middleware;
  * @method \GuzzleHttp\Promise\Promise getPolicyAsync(array $args = [])
  * @method \Aws\Result invoke(array $args = [])
  * @method \GuzzleHttp\Promise\Promise invokeAsync(array $args = [])
+ * @method \Aws\Result invokeAsync(array $args = [])
+ * @method \GuzzleHttp\Promise\Promise invokeAsyncAsync(array $args = [])
  * @method \Aws\Result listAliases(array $args = [])
  * @method \GuzzleHttp\Promise\Promise listAliasesAsync(array $args = [])
  * @method \Aws\Result listEventSourceMappings(array $args = [])
--- a/src/Lambda/LambdaClient.php
+++ b/src/Lambda/LambdaClient.php
@@ -46,8 +46,6 @@ use Aws\Middleware;
  * @method \GuzzleHttp\Promise\Promise getPolicyAsync(array $args = [])
  * @method \Aws\Result invoke(array $args = [])
  * @method \GuzzleHttp\Promise\Promise invokeAsync(array $args = [])
- * @method \Aws\Result invokeAsync(array $args = [])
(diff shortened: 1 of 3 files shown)


src/S3/StreamWrapper.php
```diff
@@ -459,6 +459,8 @@ public function dir_rewinddir()
             $this->dir_opendir($this->openedPath, null);
             return true;
         });
+
+        return true;
```
howardlopez: Since `boolCall` may return true or false depending on success or failure, this should be:
```
return $this->boolCall(function() {
...
}
```

src/Lambda/LambdaClient.php
```diff
@@ -46,8 +46,6 @@
  * @method \GuzzleHttp\Promise\Promise getPolicyAsync(array $args = [])
  * @method \Aws\Result invoke(array $args = [])
  * @method \GuzzleHttp\Promise\Promise invokeAsync(array $args = [])
- * @method \Aws\Result invokeAsync(array $args = [])
- * @method \GuzzleHttp\Promise\Promise invokeAsyncAsync(array $args = [])
```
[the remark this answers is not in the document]
tolbon: Oh I not see invokeAsync in api-2.json sorry -_-'
Yeah ok no problem.

ps: This is not strange to Async an Async function ?

src/S3/StreamWrapper.php
tolbon: Yeah I'm agree with you but
if you see boolCall always return true cause this function always return true
And dir_opendir seems always return true too.

Same effect ? or I need to fix with 

```php
/**
     * This method is called in response to rewinddir()
     *
     * @return boolean true on success
     */
    public function dir_rewinddir()
    {
        return $this->boolCall(function() {
            $this->objectIterator = null;
            return $this->dir_opendir($this->openedPath, null);
        });
    }
```

src/Lambda/LambdaClient.php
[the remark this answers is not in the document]
howardlopez: `invokeAsync` is a Lambda service operation. The PHP SDK also uses `Async` as a suffix to indicate that a promise will be returned (instead of a result object). So, `invokeAsyncAsync` is indeed valid for the SDK, strangely enough! The issue we will have to address internally is how to define the non `Async` version of functions that actually end in the word `Async`.

src/S3/StreamWrapper.php
howardlopez: `boolCall` may return false if the `catch` block is hit, so please correct with the return statement as shown.

src/Lambda/LambdaClient.php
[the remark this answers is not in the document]
tolbon: Now fix normally

src/S3/StreamWrapper.php
tolbon: Now fix normally

PR #1769, opened 2019-04-18
Fix test invokes fulfilled ignore keys

rquadling:
Fix the unit test that demonstrates the iterator keys are actually ignored in the CommandPool, rather than ignoring them in the test.


.changes/nextrelease/process_credential_provider.json
```diff
@@ -2,6 +2,6 @@
     {
         "type": "feature",
         "category": "Credentials",
-        "description": "This update adds a process credential provider. Credetentials can be sourced from a command returning json on stdout."
+        "description": "This update adds a process credential provider. Credentials can be sourced from a command returning json on stdout."
```
srchase: Thanks for addressing this typo, but could you drop this commit?  It's conflicting due to nextrelease having already been pulled in 3.92.0.  I'll followup to look at getting this corrected separately.

rquadling: Sure.

srchase:
Thanks for fixing this test.

PR #1786, opened 2019-05-09
Changed type hints from Stream to StreamInterface in DecryptionTrait

ahaaje:
*Issue #1787 *

When using the _S3EncryptionClient::getObject()_ to fetch previously client side encrypted document, as described [here](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/s3-encryption-client.html), errors like this was encountered.

`
Argument 1 passed to Aws\S3\Crypto\S3EncryptionClient::getTagFromCiphertextStream() must be an instance of GuzzleHttp\Psr7\Stream, instance of GuzzleHttp\Psr7\LazyOpenStream given, called in .../vendor/aws/aws-sdk-php/src/Crypto/DecryptionTrait.php on line 149

Argument 1 passed to Aws\S3\Crypto\S3EncryptionClient::getStrippedCiphertextStream() must be an instance of GuzzleHttp\Psr7\Stream, instance of GuzzleHttp\Psr7\LazyOpenStream given, called in .../vendor/aws/aws-sdk-php/src/Crypto/DecryptionTrait.php on line 156
`

I fixed this by changing the type hint to Psr\Http\Message\StreamInterfacem which is implemented by the different stream types like LazyOpenStream.

### DIFF
--- a/.changes/nextrelease/changelog_decryption_trait.json
+++ b/.changes/nextrelease/changelog_decryption_trait.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "Crypto",
+        "description": "Changed type hint to StreamInterface"
+    }
+]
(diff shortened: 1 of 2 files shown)


srchase:
@ahaaje 

Thanks for the contribution!

PR #1805, opened 2019-05-29
Fixes double annontation invokeAsync

dada-amater:
*Issue #, if available:*
There are 2 annotations for function `invokeAsync`. It makes phpstan to fail.
*Description of changes:*
Removed wrong annotation.


srchase:
@dada-amater 

Thanks for opening this PR! Appreciate you taking the time to submit.

Unfortunately, the annotations are generated and this double annotation cannot be removed in order to maintain backwards compatibility.

```invokeAsync``` is a valid Lambda operation and to use that operation with the PHP SDK asynchronously, it's necessary to use the admittedly awkward ```invokeAsyncAsync```.

In the next version of the SDK, we'll take a look at aliasing this operation to avoid this issue.

PR #1817, opened 2019-06-11
added pagination support for cognito-idp/ListUsers

vjirovsky:
*Issue #, if available:*

*Description of changes:*

There is missing pagination support for method ListUsers in cognito-idp service. 

In this PR there is no change to paginators-1.json.php (because it should be autogenerated from the source).

Result changes to paginators-1.json.php:
```
'ListUsers' => [ 'input_token' => 'PaginationToken', 
'limit_key' => 'Limit', 'output_token' => 'PaginationToken', 'result_key' => 'Users', ], 
```

### DIFF
--- a/src/data/cognito-idp/2016-04-18/paginators-1.json
+++ b/src/data/cognito-idp/2016-04-18/paginators-1.json
@@ -30,6 +30,12 @@
       "output_token": "NextToken",
       "result_key": "ResourceServers"
     },
+    "ListUsers": {
+      "input_token": "PaginationToken",
+      "limit_key": "Limit",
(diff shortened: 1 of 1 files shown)


diehlaws:
Thanks for submitting this PR @vjirovsky. Since waiters, paginators, and API models for each service are provided by the corresponding service team, this change would unfortunately be overwritten on the next release so we cannot accept and merge this PR. 

That being said, I've reached out to the Cognito Identity Provider team internally about adding a paginator for `ListUsers` with the changes present in your PR so this should be added relatively quickly in a way that would propagate to all AWS SDKs and not just the AWS SDK for PHP.

PR #1824, opened 2019-06-18
assume role using profile(s) in ~/.aws/config

srchase:
Enables profiles in ~/.aws/config to be used in auto-assuming role.

When environment variable ```AWS_SDK_LOAD_NONDEFAULT_CONFIG``` is set, profiles will be read from ```~/.aws/config```.  Any profiles returned from ```~/.aws/config``` can be used as a ```source_profile``` or the profile of the role to be assume.

Given this credential file at ~/.aws/credentials:
```
[profile1]
output = json
region = us-east-1
aws_access_key_id = A**************
aws_secret_access_key = *****************
```

And this config file at ~/.aws/config:
```
[profile profile2]
output = json
region = us-east-1
role_arn = arn:aws:iam::123456789012:role/Developer
source_profile = profile1
```

profile2 will be assumed using the credentials from profile1.  The reverse also works, as well as using one profile in ```~/.aws/config``` to load another profile also in ```~/.aws/config```.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -591,12 +591,10 @@ class CredentialProvider
         // If loading .aws/credentials, also load .aws/config when AWS_SDK_LOAD_NONDEFAULT_CONFIG is set
         if ($filename === self::getHomeDir() . '/.aws/credentials'
             && getenv('AWS_SDK_LOAD_NONDEFAULT_CONFIG')
-        )
-        {
+        ) {
             $configFilename = self::getHomeDir() . '/.aws/config';
             $configProfileData = \Aws\parse_ini_file($configFilename, true, INI_SCANNER_RAW);
-            foreach ($configProfileData as $name => $profile)
-            {
+            foreach ($configProfileData as $name => $profile) {
                 // standardize config profile names
                 $name = str_replace('profile ', '', $name);
                 if (!isset($profileData[$name])) {
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -313,7 +313,7 @@ class CredentialProvider
             if (!is_readable($filename)) {
                 return self::reject("Cannot read credentials from $filename");
             }
-            $data = \Aws\parse_ini_file($filename, true, INI_SCANNER_RAW);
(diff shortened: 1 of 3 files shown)


src/Credentials/CredentialProvider.php
```diff
@@ -581,6 +581,33 @@ private static function getHomeDir()
...
+     * Gets profiles from specified $filename, or default ini files.
+     */
+    private static function loadProfiles($filename)
+    {
+        $profileData = \Aws\parse_ini_file($filename, true, INI_SCANNER_RAW);
+
+        // If loading .aws/credentials, also load .aws/config when AWS_SDK_LOAD_NONDEFAULT_CONFIG is set
+        if ($filename === self::getHomeDir() . '/.aws/credentials'
+            && getenv('AWS_SDK_LOAD_NONDEFAULT_CONFIG')
+        )
```
howardlopez: Style: `{` should be on same line as `)`

src/Credentials/CredentialProvider.php
```diff
@@ -581,6 +581,33 @@ private static function getHomeDir()
...
+        {
+            $configFilename = self::getHomeDir() . '/.aws/config';
+            $configProfileData = \Aws\parse_ini_file($configFilename, true, INI_SCANNER_RAW);
+            foreach ($configProfileData as $name => $profile)
```
howardlopez: Style: `{` should be on same line as the `foreach`

src/Credentials/CredentialProvider.php
```diff
@@ -581,6 +581,33 @@ private static function getHomeDir()
...
+            {
+                // standardize config profile names
```
howardlopez: Does this behavior match other SDKs/CLI?

srchase: When loading from config, JavaScript strips the 'profile ' prefix:
https://github.com/aws/aws-sdk-js/blob/9b64887b9f6d12b8f87389724aa31bb90651573f/lib/shared-ini/ini-loader.js#L10

And then merges all of the credential and config profiles into one object.  config profiles get overwritten by credential profiles if they have the same name.

nvcken:
@srchase 
Should this work if MFA enabled ?

PR #1836, opened 2019-07-01
Assume role with web identity

srchase:
Implements credential provider that allows role to be assumed using web_identity_token_file, specified as an environmental variable, ```AWS_WEB_IDENTITY_TOKEN_FILE``` or as ```web_identity_token_file``` option in shared credentials/config.

### DIFF
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -93,7 +93,7 @@ class AssumeRoleWithWebIdentityCredentialProvider
 
         return $client->assumeRoleWithWebIdentityAsync($assumeParams)
             ->then(function (Result $result) {
-                    return $this->client->createCredentials($result);
+                return $this->client->createCredentials($result);
             })->otherwise(function (\RuntimeException $exception) {
                 throw new CredentialsException(
                     "Error assuming role from web identity credentials",
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -13,9 +13,6 @@ use GuzzleHttp\Promise;
 class AssumeRoleWithWebIdentityCredentialProvider
 {
     const ERROR_MSG = "Missing required 'AssumeRoleWithWebIdentityCredentialProvider' configuration option: ";
-    const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
-    const ENV_ARN = 'AWS_ROLE_ARN';
-    const ENV_SESSION = 'AWS_ROLE_SESSION';
 
     /** @var string */
     private $tokenFile;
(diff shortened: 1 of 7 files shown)


src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -0,0 +1,105 @@
...
+use GuzzleHttp\Promise;
+
+/**
+ * Credential provider that provides credentials via assuming a role with a web identity
+ * More Information, see: https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sts-2011-06-15.html#assumerolewithwebidentity
+ */
+class AssumeRoleWithWebIdentityCredentialProvider
+{
+    const ERROR_MSG = "Missing required 'AssumeRoleWithWebIdentityCredentialProvider' configuration option: ";
+    const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
```
howardlopez: Is this ever used? Seems redundant with the class constant from `CredentialProvider`. Ditto for `ENV_ARN`.

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -0,0 +1,105 @@
...
+    const ENV_ARN = 'AWS_ROLE_ARN';
+    const ENV_SESSION = 'AWS_ROLE_SESSION';
```
howardlopez: This doesn't overlap with the same class constant from `CredentialProvider`, but is it ever used?

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        if (!isset($config['RoleArn'])) {
+            throw new \InvalidArgumentException(self::ERROR_MSG . "'RoleArn'.");
+        }
+        $this->arn = $config['RoleArn'];
+
+        if (!isset($config['WebIdentityTokenFile'])) {
+            throw new \InvalidArgumentException(self::ERROR_MSG . "'WebIdentityTokenFile'.");
+        }
+        $this->tokenFile = $config['WebIdentityTokenFile'];
```
howardlopez: The specification calls for an error to be thrown if the token is not an absolute path - perhaps that should go in this constructor?

src/Credentials/CredentialProvider.php
```diff
@@ -69,13 +72,18 @@ class CredentialProvider
...
     public static function defaultProvider(array $config = [])
     {
-        $localCredentialProviders = self::localCredentialProviders();
+        $defaultChain = [
+            self::env(),
+            self::assumeRoleWithWebIdentityCredentialProvider(),
+            self::ini(),
+            self::ini('profile default', self::getHomeDir() . '/.aws/config'),
+        ];
         $remoteCredentialProviders = self::remoteCredentialProviders($config);
```
howardlopez: Since the remote and local credential providers are now all interspersed with each other, does this distinction still make sense? Should this just be a single default chain of providers?

src/Credentials/CredentialProvider.php
```diff
@@ -278,6 +286,64 @@ public static function assumeRole(array $config=[])
...
+                    'client' => $stsClient
+                ]);
+
+                return $provider();
+            }
+
+            $profileName = getenv(self::ENV_PROFILE) ?: 'default';
+            $filename = isset($config['filename'])
+                ? $config['filename']
+                : self::getHomeDir() . '/.aws/credentials';
```
howardlopez: This doesn't seem specifically addressed by the SEP, but should `/.aws/config` also be looked at?

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -0,0 +1,106 @@
...
+
+        if (!preg_match("/^\w\:|^\/|^\\\/", $this->tokenFile)) {
+            throw new \InvalidArgumentException("'WebIdentityTokenFile' must be absolute path.");
```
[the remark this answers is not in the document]
srchase: Added assertions to cover the session name getting set in config via an environmental variable or via a profile.

Issue #1664, opened 2018-11-02
InstanceProfileProvider does not handle when the response contains an invalid JSON


alquerci:
Hello,

We have a dynamically modified infrastructure managed by auto-scaling rules. EC2 instances are launched and terminated upon load.

Issue
------

Some times the instance metadata server respond with an invalid JSON and we got this exception.

The response body here is partial.

```
Unexpected instance profile response code:  
Stack trace:
#0 /project/vendor/aws/aws-sdk-php/src/Credentials/InstanceProfileProvider.php(53): Aws\Credentials\InstanceProfileProvider->decodeResult('{\n  "Code" : "S...')
```

See https://github.com/aws/aws-sdk-php/blob/3.70.2/src/Credentials/InstanceProfileProvider.php#L102

Note
-------

* When occurred the common retry behaviour is skipped.
* I saw boto provides a fix for such issue https://github.com/boto/botocore/issues/1049

diehlaws:
Hi @alquerci, thanks for reaching out about this. Unfortunately the Instance Profile Credential Provider does not use the same retry middleware that is used when issuing service API calls, and its default Guzzle handler does not automatically retry the Instance Profile retrieval when an unsuccessful response is returned by the Instance Metadata service. 

That being said, it is possible to pass your own handler to the Credential Provider so that retries can be implemented to fit your use case. Below is an example of implementing a custom handler on an Instance Profile Credential Provider for an S3 client based on our [Handlers and Middleware](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_handlers-and-middleware.html) documentation page.

```php
$myHandler = function (CommandInterface $cmd, RequestInterface $request) {
    $result = new Result(['foo' => 'bar']);
    // Implement retry logic here
    return Promise\promise_for($result);
};

$provider = CredentialProvider::instanceProfile([
    'handler' => $myHandler
]);

$creds = $provider()->wait();

$client = new S3Client([
    'region' => 'us-west-2',
    'version' => 'latest',
    'credentials' => $creds
]);
```

alquerci:
Hello @diehlaws, thanks to take care about this.

For sure extends the library is the solution to provide the expected behaviour. I create this issue because such feature/patch may have the place on `aws-php-sdk` and not only library user side. Because this issue is hard to detect and need to use a dedicated retry bahaviour because it is bound to AWS infrastructure.

[...]

PR #1837, opened 2019-07-02
Instance profile provider retries

srchase:
resolves https://github.com/aws/aws-sdk-php/issues/1664

Implements exponential backoff when invalid JSON is returned to InstanceProfileProvider.

By default, the provider will retry three times.

Retries can be adjusted on the provider with the ```retries``` config option:
```
$provider = new InstanceProfileProvider([
    'retries' => 5
]);
```
And can be disabled by setting to ```0```:
```
$provider = new InstanceProfileProvider([
    'retries' => 0
]);
```

### DIFF
--- a/src/Credentials/InstanceProfileProvider.php
+++ b/src/Credentials/InstanceProfileProvider.php
@@ -5,6 +5,7 @@ use Aws\Exception\CredentialsException;
 use Aws\Exception\InvalidJsonException;
 use Aws\Sdk;
 use GuzzleHttp\Promise;
+use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Http\Message\ResponseInterface;
@@ -75,6 +76,14 @@ class InstanceProfileProvider
                             'Invalid JSON Response, retries exhausted.'
                         );
                     }
+                } catch (RequestException $e) {
+                    if ($this->attempts < $this->retries) {
+                        sleep(pow(1.2, $this->attempts));
+                    } else {
+                        throw new CredentialsException(
+                            'Networking error, retries exhausted.'
+                        );
+                    }
                 }
                 $this->attempts++;
             }
@@ -115,6 +124,9 @@ class InstanceProfileProvider
                 return (string) $response->getBody();
             })->otherwise(function (array $reason) {
                 $reason = $reason['exception'];
+                if ($reason instanceof \GuzzleHttp\Exception\RequestException) {
+                    throw $reason;
+                }
                 $msg = $reason->getMessage();
                 throw new CredentialsException(
                     $this->createErrorMessage($msg)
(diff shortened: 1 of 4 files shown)


src/Credentials/InstanceProfileProvider.php
```diff
@@ -24,18 +25,27 @@ class InstanceProfileProvider
...
      * - profile: Optional EC2 profile name, if known.
+     * - retries: Optional number of retries to be attempted.
      *
      * @param array $config Configuration options.
      */
     public function __construct(array $config = [])
     {
         $this->timeout = isset($config['timeout']) ? $config['timeout'] : 1.0;
         $this->profile = isset($config['profile']) ? $config['profile'] : null;
+        $this->retries = isset($config['retries']) ? $config['retries'] : 0;
```
howardlopez: Should this retry by default?

srchase: Once?  Or preserve the existing behavior of this not getting retried by default.

srchase: 3 sounds like a good compromise.

src/Credentials/InstanceProfileProvider.php
```diff
@@ -52,8 +62,22 @@ public function __invoke()
...
                 $this->profile = (yield $this->request(self::CRED_PATH));
             }
-            $json = (yield $this->request(self::CRED_PATH . $this->profile));
-            $result = $this->decodeResult($json);
+            $result = null;
+            while ($result == null) {
+                try {
+                    $json = (yield $this->request(self::CRED_PATH . $this->profile));
+                    $result = $this->decodeResult($json);
+                } catch (InvalidJsonException $e) {
```
howardlopez: Should there be a retry for things beyond invalid JSON?

srchase: I think for now, we should limit this to the JSON error.  The original issue that precipitated needing to retry was for invalid JSON.  We may want to extend this to throttling errors in the future, but will want to be careful to not exacerbate that issue.

tests/Credentials/InstanceProfileProviderTest.php
```diff
@@ -200,4 +202,63 @@ public function testThrowsExceptionOnInvalidJsonResponse()
...
+    public function testNetworkingErrorsAreRetried()
+    {
+        $retries = 1;
+
+        $t = time() + 1000;
+        $result = json_encode($this->getCredentialArray('foo', 'baz', null, "@{$t}"));
+        $responses = [new Response(200, [], Psr7\stream_for($result))];
+
+        $client = function () use (&$retries, $responses) {
+            if (0 === --$retries) {
```
howardlopez: With the `--` before `$retries`, won't this only ever execute once without ever hitting the rejection below?

srchase: Yes, should be a post-decrement so the rejection is thrown on the first request.

PR #1850, opened 2019-07-15
Indirect deprecation notices fixed

bvanderstee:
Added docblock from "Aws\S3\S3ClientInterface::getObjectUrl()" to
"Aws\S3\S3Client::getObjectUrl()"

*Description of changes:*
Recent update got me a Indirect deprecation notice on my own phpunit tests inside my project that is using this library

>   1x: The "Aws\S3\S3Client::getObjectUrl()" method will require a new "array $options" argument in the next major version of its parent class "Aws\S3\S3ClientInterface", not defining it is deprecated.
>     1x in S3ManagerTest::testPutFileOnS3 from App\Manager\Tests

Resolved this notice by adding the docblock from the interface method to the object method itself.

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -361,7 +361,7 @@ class S3Client extends AwsClient implements S3ClientInterface
     /**
      * Returns the URL to an object identified by its bucket and key.
      *
-     * The URL returned by this method is not signed nor does it ensure the the
+     * The URL returned by this method is not signed nor does it ensure that the
      * bucket and key given to the method exist. If you need a signed URL, then
      * use the {@see \Aws\S3\S3Client::createPresignedRequest} method and get
      * the URI of the signed request.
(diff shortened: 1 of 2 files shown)


src/S3/S3Client.php
```diff
@@ -358,6 +358,20 @@ public function createPresignedRequest(CommandInterface $command, $expires, arra
...
+     * Returns the URL to an object identified by its bucket and key.
+     *
+     * The URL returned by this method is not signed nor does it ensure the the
+     * bucket and key given to the method exist. If you need a signed URL, then
+     * use the {@see \Aws\S3\S3Client::createPresignedRequest} method and get
+     * the URI of the signed request.
+     *
+     * @param string $bucket  The name of the bucket where the object is located
+     * @param string $key     The key of the object
+     * @param array  $options Aws\Signature\SignatureInterface::presign options
```
srchase: $options was added to the doc block in Aws\S3\S3ClientInterface by mistake.

Instead of adding it here, would you remove it there?  That should fix the deprecation notice.

bvanderstee: $options removed on both S3ClientInterface and S3Client itself

srchase:
@bvanderstee 

Thanks for the PR and making this change!

PR #1853, opened 2019-07-24
STS regional endpoints

howardlopez:
* Adds support for STS regional endpoints and corresponding configuration.


src/Sts/RegionalEndpoints/ConfigurationProvider.php
```diff
@@ -0,0 +1,314 @@
...
+            return Promise\promise_for(
+                new Configuration($data[$profile][self::INI_ENDPOINTS_TYPE])
+            );
+        };
+    }
+
+    /**
+     * Wraps a config provider and caches previously provided configuration.
+     *
+     * Ensures that cached configuration is refreshed when it expires.
```
srchase: Do these cached configs actually expire? If not, this is a bit confusing.

src/Sts/RegionalEndpoints/ConfigurationProvider.php
```diff
@@ -0,0 +1,314 @@
...
+            static $result;
+            static $isConstant;
+
+            // Constant config will be returned constantly.
+            if ($isConstant) {
+                return $result;
+            }
+
+            // Create the initial promise that will be used as the cached value
+            // until it expires.
```
srchase: Also, expiration mentioned here.  Is this needed?

PR #1858, opened 2019-07-30
retry InvalidIdentityToken errors

srchase:
Retry InvalidIdentityToken errors.  Used by AssumeRoleWithWebIdentityCredentialProvider when StsClient makes call to asssume role.

### DIFF
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -26,6 +26,11 @@ class AssumeRoleWithWebIdentityCredentialProvider
     /** @var StsClient */
     private $client;
 
+    /** @var integer */
+    private $retries;
+
+    /** @var integer */
+    private $attempts;
 
     /**
      * The constructor attempts to load config from environment variables.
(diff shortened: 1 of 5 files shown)


src/RetryMiddleware.php
```diff
@@ -31,6 +31,7 @@ class RetryMiddleware
         'RequestThrottledException'              => true,
         'TooManyRequestsException'               => true,
         'IDPCommunicationError'                  => true,
+        'InvalidIdentityToken'                   => true,
```
howardlopez: Do we want this error code to be retried for all services?

srchase: [AssumeRoleWithSAML](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithSAML.html) also returns ```InvalidIdentityToken```.  Other services shouldn't return this error.

howardlopez: Ideally retrying `InvalidIdentityToken` would only be limited in scope to STS (as it's a special use case), as it's not inherently a throttling or retryable error, and we should allow for the possibility of other services using that exception name at some point.

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -75,32 +83,49 @@ public function __construct(array $config = [])
...
+                    'RoleSessionName' => $this->session,
+                    'WebIdentityToken' => $token
+                ];
+
+                try {
+                    $result = $client->assumeRoleWithWebIdentity($assumeParams);
+                } catch (\Exception $e) {
+                    if ($e->getAwsErrorCode() == 'InvalidIdentityToken') {
+                        if ($this->attempts < $this->retries) {
+                            sleep(pow(1.2, $this->attempts));
```
howardlopez: Is this retry strategy enough for the cold start scenario we're covering for?

srchase: A single retry may not succeed when encountering cold starts.  Bumping the default retries to 3 should be sufficient for that scenario.

PR #1614, opened 2018-08-29
[WIP] Basic static analysis with PHPStan

lookyman:

PR #1861, opened 2019-07-30
Static analysis & Travis updates

howardlopez:
* Added static code analysis using [PHPStan](https://github.com/phpstan/phpstan) to Travis CI configuration.
* Added PHP 7.4 to build job matrix
* Added comments and readability tweaks to Travis file
* Thanks to @lookyman for initial [PR](https://github.com/aws/aws-sdk-php/pull/1614) in 2018


.travis.yml
```diff
@@ -19,23 +20,56 @@ matrix:
...
+  # Resolve dependencies
   - composer --version
   - travis_retry composer update $COMPOSER_OPTS --no-interaction --prefer-source
 
 script:
+  # Unit tests
   - make test
+
+  # Static analysis
+  - >
```
srchase: Should this get moved into the Makefile so it can be run offline?  And update CONTRIBUTING.md too?

howardlopez: I hesitate to put this in the Makefile since it requires installing composer packages that are not part of require-dev, and which will only work with PHP 7.1+. 

Will update `CONTRIBUTING.md` for new guidelines.

PR #1881, opened 2019-10-03
Change cache clearing in StreamWrapper to account for custom protocol use.

omissis:
Replace hardcoded `s3` protocol with the configured one in `S3`'s `StreamWrapper` class in `stream_flush` method, to avoid leaving wrong items in the cache should a custom protocol name be used.


srchase:
@omissis 

Thanks for opening this PR!

The new test is failing on PHP 5.5 and PHP 5.6.  Would you mind addressing that and adding a changelog entry?

omissis:
Hi! Sure, I will look into it as soon as I can :)

srchase:
@omissis,

Could you also add a changelog when you update this?

Thanks

PR #1892, opened 2019-11-07
Allow setting credential instance metadata timeouts and retries from environment variables

callum-p:
*Issue #, if available:*
N/A

*Description of changes:*
botocore allows setting instance metadata timeout and retries via environment variables. This update brings the PHP SDK in line with that functionality. See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L62 and https://github.com/boto/botocore/blob/9f5afe26cc4a2695dcb62db453f9db7c299f3ffc/botocore/configprovider.py#L63 for botocore allowing environment variables.

Use cases for this include things like when using a framework (e.g. Laravel), the SES mail transport abstracts away the SES client constructor from the user, so it's hard to set the timeout or retries without a lot of code changes. It's easier to simply set an environment variable.


howardlopez:
Thanks @callum-p for the PR! Merged in #1933 after a couple changes.

PR #1898, opened 2019-11-12
S3 regional endpoint

howardlopez:
* Added support for S3 us-east-1 regional endpoint setting and corresponding configuration.


tests/S3/RegionalEndpoint/ConfigurationProviderTest.php
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * @covers \Aws\S3\RegionalEndpoint\ConfigurationProvider
+ */
+class ConfigurationProviderTest extends TestCase
+{
+    private static $originalEnv;
+
+    private $iniFile = <<<EOT
+[custom]
+s3_us_east_1_regional_endpoint = regional
```
srchase: Should there be a case here for different casings (title, upper) of these options?  Same for ENV.

howardlopez: Added commit with tests in the provider and configuration class for a non-lowercase value.

PR #1912, opened 2019-11-30
Fixed the `Trying to access array offset on value of type null` error on php7.4

jefersonralmeida:
Issue https://github.com/aws/aws-sdk-php/issues/1889

I just included a simple check to ensure that we're not using a null variable as an array.

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -143,7 +143,7 @@ class RetryMiddleware
         }
 
         if (!$error) {
-            return is_array($result['@metadata']) ? isset($statusCodes[$result['@metadata']['statusCode']]) : false;
+            return isset($statusCodes[$result['@metadata']['statusCode'] ?? null]);
         }
 
         if (!($error instanceof AwsException)) {
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -143,7 +143,7 @@ class RetryMiddleware
         }
 
         if (!$error) {
-            return isset($statusCodes[$result['@metadata']['statusCode']]);
+            return is_array($result['@metadata']) ? isset($statusCodes[$result['@metadata']['statusCode']]) : false;
         }
 
         if (!($error instanceof AwsException)) {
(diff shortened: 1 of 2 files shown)


howardlopez:
@jefersonralmeida Thanks for the PR! It looks like the syntax used `??` does not work with PHP 5.x and fails those Travis CI tests. Can you refactor to use 5.x-friendly code, such as: 

```
            if (!isset($result['@metadata']['statusCode'])) {
                return false;
            }
            return isset($statusCodes[$result['@metadata']['statusCode']]);
```

Also, can you add a [changelog](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md)?

src/RetryMiddleware.php
```diff
@@ -143,7 +143,7 @@ private static function isRetryable(
         }
 
         if (!$error) {
-            return isset($statusCodes[$result['@metadata']['statusCode']]);
+            return isset($statusCodes[$result['@metadata']['statusCode'] ?? null]);
```
georgeboot:
```suggestion
            if (!isset($result['@metadata']['statusCode'])) {
                return false;
            }
            
            return isset($statusCodes[$result['@metadata']['statusCode']]);
```

src/S3/PutObjectUrlMiddleware.php
```diff
@@ -44,7 +44,7 @@ function (ResultInterface $result) use ($command) {
                 switch ($name) {
                     case 'PutObject':
                     case 'CopyObject':
-                        $result['ObjectURL'] = $result['@metadata']['effectiveUri'];
+                        $result['ObjectURL'] = $result['@metadata']['effectiveUri'] ?? null;
```
georgeboot:
```suggestion
                        if (isset($result['@metadata']['effectiveUri']) {
                            $result['ObjectURL'] = $result['@metadata']['effectiveUri'];
                        } else {
                            $result['ObjectURL'] = null;
                        }
```

howardlopez:
@jefersonralmeida We decided to go ahead and ship some fixes in #1916 due to this currently causing user pain and wanting to mitigate this as soon as possible. Will close this PR as it now duplicates that work, but thanks so much for the submission!

PR #1916, opened 2019-12-03
PHP 7.4 fixes

### LINKED ISSUE #1915: Trying to access array offset on value of type null
Please fill out the sections below to help us address your issue.

### Version of AWS SDK for PHP?
3.125

### Version of PHP (`php -v`)?
7.4

### What issue did you see?
Trying to access array offset on value of type null in /srv/www/vendor/aws/aws-sdk-php/src/RetryMiddleware.php:146

### Steps to reproduce
I don't currently have a good way to reproduce this.  

But, looking at the RetryMiddleware class, there are many things to note.  Firstly, it's not typed at all, so `$result` could literally be anything and it's not even verified to be a legitimate value before attempting to access array keys on it.

howardlopez:
* Fixes array offset null issue and test stream class for PHP 7.4.
* Adds php `7.4` travis build (alongside `7.4snapshot`) as an allowed-to-fail build, for monitoring.
* Thanks to @jefersonralmeida for his initial [PR](https://github.com/aws/aws-sdk-php/pull/1912)

Resolves #1915


.travis.yml
```diff
@@ -9,6 +9,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
+  - 7.4
   - 7.4snapshot
```
srchase: Any reason to keep 7.4snapshot now that 7.4 is available?

howardlopez: 7.4 is not available (i.e. Travis is still failing to build it), but figured we should be monitoring it when it does become so.

srchase: 👍

PR #1921, opened 2019-12-09
S3 control account id validation

srchase:
Validate S3Control input paramater for host name compatibility.

### DIFF
--- a/tests/S3Control/S3ControlEndpointMiddlewareTest.php
+++ b/tests/S3Control/S3ControlEndpointMiddlewareTest.php
@@ -69,6 +69,7 @@ class S3ControlEndpointMiddlewareTest extends TestCase
 
     /**
      * @expectedException \InvalidArgumentException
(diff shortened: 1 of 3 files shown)


tests/S3Control/S3ControlEndpointMiddlewareTest.php
```diff
@@ -67,6 +67,21 @@ public function testDoesNothingWhenDualStackDisabledOnOperationLevel()
         $middleware($command, $this->getRequest($command));
     }
 
+    /**
+     * @expectedException \InvalidArgumentException
+     */
```
howardlopez: Should there be an expected message, to make sure we're failing on the right thing?

srchase: Fixed in b456071cb501d1be6262151fb885118c4c7c82ab

PR #1933, opened 2019-12-17
Allow setting credential instance metadata timeouts and retries from environment variables

howardlopez:
* Brings PHP SDK inline with Python's boto3, allowing setting instance metadata timeout and retries from environment variables
* Original [PR](https://github.com/aws/aws-sdk-php/pull/1892) by @callum-p. It was updated to accommodate changes to the Instance Profile Provider since the PR was created.


tdmalone:
FYI, documentation at https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-metadata_service_num_attempts.html and https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-metadata_service_timeout.html says that the PHP SDK doesn't support these env vars; I have just submitted feedback to ask for those pages to be updated given this PR.

PR #1934, opened 2019-12-18
HHVM 4 doesn't support PHP anymore, as such

GrahamCampbell:
HHVM 3.18 LTS can run both composer and PHPUnit, so it should be used to test against.


GrahamCampbell:
Ping @howardlopez. ;)

GrahamCampbell:
Ping.

PR #1937, opened 2019-12-21
Allow a Sleep Between Pages when Using a Paginator

ssigwart:
*Description of changes:*
This is useful for APIs such as the Route53 `ListResourceRecordSets`
operation, which has a rate limit of 5 requests/sec (see
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests).

It allows you to add a sleep between pages to avoid going over rate limits on APIs with low limits.

### DIFF
--- a/.changes/nextrelease/paginatorSleep.json
+++ b/.changes/nextrelease/paginatorSleep.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "ResultPaginator",
+    "description": "Allow paginator to be configured with a sleep between pages."
+  }
+]
--- a/docs/guide/paginators.rst
+++ b/docs/guide/paginators.rst
@@ -49,6 +49,11 @@ the paginator would need to do 10 requests total. When you iterate through the
 results, the first request is executed when you start iterating, the second in
 the second iteration of the loop, and so forth.
 
+Because some APIs are rate limited (e.g. Route53 ``ListResourceRecordSets``), you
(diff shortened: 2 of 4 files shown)


srchase:
@ssigwart 

Thanks for submitting this contribution.

Does inserting the sleep into the paginator result work for your use case?

```
$client = new Route53Client([
    'profile' => 'default',
    'version' => 'latest',
    'region' => 'us-east-1',
]);

$results = $client->getPaginator('ListResourceRecordSets', [
    'HostedZoneId' => 'myZoneId'
]);

foreach ($results as $result) {
    foreach ($result['ResourceRecordSets'] as $object) {
        //...
    }
    sleep(1);
}
```

ssigwart:
For the most part, yes.  The one issue is that it adds an extra sleep after the last page.  I could add code to check for that too, but I thought it might be nice to put it in the paginator so it's a one line call to set it up.

srchase:
@ssigwart 

Thanks for the quick response.

At this time, we've decided not to accept this PR.

For maximum flexibility, it makes sense to have delays to be added when using paginators, rather than by adding a single type of delay as a configuration option within the paginator itself.  While this configuration option wouldn't preclude adding delays outside the paginator, we'd prefer to not add this capability since there is an existing way to implement the same thing in a straight forward way.

ssigwart:
Ok.  Thanks for looking at it.

srchase:
Thanks for opening this!

PR #1938, opened 2019-12-28
Bumped minimum jmespath.php

GrahamCampbell:
`jmespath.php` causes deprecation exceptions to be raised on PHP 7.2 or higher, and totally crashes on PHP 7.4 due to the usage of the `fn` reserved word. This is fixed in the 1.5.0 release.

---

NB There is one outstanding jmespath spec compliance issue, affecting (all versions of) the PHP library: https://github.com/jmespath/jmespath.php/issues/56.


howardlopez:
Thanks for the PR - this will go out in the next release.

PR #1940, opened 2019-12-30
Allow Guzzle 7

Nyholm:
We just released a beta of Guzzle 7 today. The [BC breaking changes](https://github.com/guzzle/guzzle/blob/master/UPGRADING.md#60-to-70) have been kept to a minimum. 

This PR allows users to use AWS SDK with Guzzle 7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,12 @@ matrix:
     - php: hhvm
     - php: nightly
   fast_finish: true
+  include:
+    - name: "Guzzle 7 beta"
+      php: 7.3
+      env: COMPOSER_OPTS="--no-interaction"
+      before_install:
+        - composer require "guzzlehttp/guzzle:^7.0@beta"
 
 sudo: false
 
@@ -49,6 +55,7 @@ install:
   # Resolve dependencies
   - composer --version
   - travis_retry composer update $COMPOSER_OPTS --no-interaction --prefer-source
+  - composer show guzzlehttp/guzzle
 
 script:
   # Unit tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
     - php: nightly
   fast_finish: true
   include:
-    - name: "Guzzle 7 beta"
+    - name: "PHP: 7.3 - Guzzle 7 beta"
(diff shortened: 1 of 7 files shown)


src/functions.php
```diff
@@ -291,18 +293,44 @@ function default_http_handler()
...
+{
+    if (defined('\GuzzleHttp\ClientInterface::VERSION')) {
+        $version = (string) ClientInterface::VERSION;
+        if ($version[0] === '6') {
+            return 6;
+        }
+        if ($version[0] === '5') {
+            return 5;
+        }
+    } elseif (method_exists('\GuzzleHttp\Client', 'sendRequest')) {
```
GrahamCampbell: Can use `::class` notation here.

src/functions.php
```diff
@@ -291,18 +293,44 @@ function default_http_handler()
...
 
+/**
+ * Get the major version of guzzle that is installed.
+ *
+ * @internal This function is internal and should not be used outside aws/aws-sdk-php.
+ * @return int
+ * @throws \RuntimeException
+ */
+function guzzle_major_version()
+{
```
GrahamCampbell: Maybe keep a static variable cache of the result, within this function body?

howardlopez: I think this makes sense, esp considering how often this method will be called.

.travis.yml
```diff
@@ -21,6 +21,12 @@ matrix:
     - php: hhvm
     - php: nightly
   fast_finish: true
+  include:
+    - name: "Guzzle 7 beta"
```
howardlopez: For consistency and clarity's sake, could this be named something like `PHP: 7.3 - Guzzle 7 beta`

Nyholm:
I updated the PR with Travis config, changelog and a cache. 

Im hesitant the cache is needed to be honest. I ran a small test: 

```
Without cache: 0,018428ms
With cache:    0,006540ms
```

With cache is indeed much faster. It runs in 35% of the time. But this is running 10 million times! So I think it is fair to go without a cache. =)

Especially considering that we are often doing a HTTP call after calling this function.

Nyholm:
Thank you for merging.

PR #1946, opened 2020-01-08
Load private key for cloudfront signer from variable

### LINKED ISSUE #1928: File Path for PEM file vs. Loading into PHP Variable
Until recently, I was able to store the contents of the .pem file into the database, and then load it in real-time into a PHP variable, and then use that to create a CloudFront signed URL.

```php
//$cfprivkey is contents of the RSA private key stored in db
$privateKey = openssl_get_privatekey($cfprivkey);

$signedURL = $cloudFront->getSignedUrl([
    'url'         => $filename,
    'expires'     => $expires,
    'key_pair_id' => $cfpubkey,
   'private_key' => $privateKey
]);
```

But of late, I'm getting an error:

`Warning: file_exists() expects parameter 1 to be a valid path, resource given in /path/to/vendor/aws/aws-sdk-php/src/CloudFront/Signer.php on line 33
PK file not found: Resource id #262`

Did something change in a recent version of the CloudFront API's in AWS PHP SDK? How was this possible before and why is this breaking now? And more importantly, how can I create a signed URL without having to store the .pem contents into an actual file, and instead load it from the database and pass it as a variable?

All AWS docs are suggesting that I store it as a file and give the path to the file when creating the signed URL, but that's not what I want. Is there no choice any more but to do that?

Thanks!

diehlaws:
*Issue #, if available:* 
Resolves #1928 

*Description of changes:*
Added logic to CF signer's construct function to allow for private keys stored as variables in addition to those stored on disk.

### DIFF
--- a/tests/CloudFront/SignerTest.php
+++ b/tests/CloudFront/SignerTest.php
@@ -27,18 +27,32 @@ class SignerTest extends TestCase
      * Assert that the key is parsed during construction
      *
      * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessageRegExp /PK .*Not a real private key/
+     */
+    public function testBadPrivateKeyContents() {
+        $privateKey = "Not a real private key";
+        $s = new Signer(
+            "not a real keypair id",
+            $privateKey
+        );
+    }
+
+    /**
+     * Assert that the key is parsed during construction
+     *
+     * @expectedException \InvalidArgumentException
      * @expectedExceptionMessageRegExp /PEM .*no start line/
      */
-    public function testBadPrivateKey() {
+    public function testBadPrivateKeyPath() {
         $filename = tempnam(sys_get_temp_dir(), 'cloudfront-fake-key');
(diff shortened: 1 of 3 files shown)


tests/CloudFront/SignerTest.php
```diff
@@ -23,22 +23,36 @@ public function setUp()
...
+        $privateKey = "Not a real private key";
+        $s = new Signer(
+            "not a real keypair id",
+            $privateKey
+        );
+    }
+
     /**
      * Assert that the key is parsed during construction
      *
```
howardlopez: Should this doc string description be differentiated from the previous test (i.e. variable vs file)?

tests/CloudFront/SignerTest.php
```diff
@@ -23,22 +23,36 @@ public function setUp()
         );
     }
 
+    /**
+     * Assert that the key is parsed during construction
```
howardlopez: Should this doc string description be differentiated from the following test (i.e. variable vs file)?

Issue #1953, opened 2020-01-15
Incorrect detection of host-style endpoint pattern while using IP address 


sigalx:
Please fill out the sections below to help us address your issue.

### Version of AWS SDK for PHP?

Latest (3.132.5). Error has began to reproduce since updating from version 3.18.28 to 3.132.5

### Version of PHP (`php -v`)?

Various (reproduced on 7.1 and 7.4 branches)

### What issue did you see?

> exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "PutObject" on "http://test.192.168.74.111:9000/files/76/762b7dc27c00d9243e60361b6b29e865b036d2ec"; AWS HTTP error: cURL error 6: Could not resolve host: test.192.168.74.111 (see https://curl.haxx.se/libcurl/c/libcurl-errors.html)'

> GuzzleHttp\Exception\ConnectException: cURL error 6: Could not resolve host: test.192.168.74.111 (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) in /var/www/html/timebook-dev/protected/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:200
Stack trace:
. . .

### Steps to reproduce

Try to upload any object to the storage with this config to bucket named «test»:
>                            'endpoint' => 'http://192.168.74.111:9000',
>                            'credentials' => [
>                                'key' => '***',
>                                'secret' => '***',
>                            ],
>                            'region' => 'us-east-1',
>                            'version' => 'latest',

### Additional context

Possible error in S3EndpointMiddleware.php on line 144 (109) — we only check if bucket name is DNS-compatible but not host too (IP address case)

diehlaws:
Hi @sigalx, thanks for reaching out to us. Starting with [version 3.28.5](https://github.com/aws/aws-sdk-php/blob/master/CHANGELOG.md#3285---2017-05-30) of the AWS SDK for PHP, S3 calls are performed with host-style URLs by default. You can change this behavior by setting `use_path_style_endpoint` to `true` in the service client options as shown below.

```php
$client = new S3Client([
    'region' => 'us-east-1',
    'version' => 'latest',
    'endpoint' => 'http://192.168.74.111:9000',
    'use_path_style_endpoint' => true,
    ]);
```

sigalx:
[...]

PR #1960, opened 2020-01-27
#1953 Host validation for IP address while trying to use host-style requests

sigalx:
Resolves issue #1953 

We validate for host while trying to use host-style requests (we have to make sure that it's not a valid IP address)

### DIFF
--- a/.changes/nextrelease/1953-host-style-ip-address-check.json
+++ b/.changes/nextrelease/1953-host-style-ip-address-check.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "bugfix",
+    "category": "S3",
+    "description": "Fixing incorrect detection of host-style endpoint pattern while using IP address"
+  }
+]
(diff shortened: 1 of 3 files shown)


sigalx:
> Thanks for the PR. Change makes sense, but can you also add a test for the new functionality as well as a [changelog](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md#changelog-documents)?

I have added a test for the code in my PR, but I don't think so it adds something new just fixes the existing bug, so it doesn't require a changelog. What do you think about?

howardlopez:
> I have added a test for the code in my PR, but I don't think so it adds something new just fixes the existing bug, so it doesn't require a changelog. What do you think about?

Thanks for adding the test. We normally add changelogs for bugfixes as well, as they do change behavior, even if it is a fix. You can mark the `type` as `bugfix`.

sigalx:
Added a changelog

howardlopez:
Thanks for adding the changelog - this will go out in the next release.

PR #1964, opened 2020-02-07
Fix error thrown when invalid debug option is passed

romainneutron:
Hello, 
When a S3 transfer is initiated with the `debug` option set to false, an error occurs, preventing us using the library like this:
```php
$transfer = new Transfer($client, $folder, $target, [
    'debug' => $output->isVeryVerbose(),
]);
```

This patch fixes this, asserting the output is a resource before writing

### DIFF
--- a/.changes/nextrelease/fix_invalid_debug_option
+++ b/.changes/nextrelease/fix_invalid_debug_option
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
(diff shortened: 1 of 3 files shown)


romainneutron:
Hello @howardlopez 
Thanks for the review, it's done !

howardlopez:
Thanks @romainneutron ! This will go out in the next release.

PR #1974, opened 2020-02-28
Correct name of instance metadata service

john-aws:
*Description of changes:*

Correcting the name of the instance metadata service (not server).


howardlopez:
Thanks for the PR - good catch on the inconsistent naming.

PR #2002, opened 2020-04-23
S3 ambiguous success tweak

howardlopez:
* Adds change to S3 parsing logic to correctly interpret certain 200 responses with a functionally empty body as connection errors.


somnaderi:
Hello, I'm currently seeing an issue in 3.133.0 where my MultiPartCopies are resulting in corrupted files that are sometimes double the original file size.

The copies that fail like this will also have an exception ("cURL error 52: Empty reply from server") that we catch and retry the way the recommended sample code suggests (getting the copy state and retrying the copy in a do/while loop). The MultiPartCopy then "succeeds" despite the destination file being corrupted.

It appears that this PR address the empty reply issue. I'm curious if this will address my specific problem...? Or can someone point me to where this problem was discussed?

PR #2018, opened 2020-05-19
Allow changing RDS AuthToken lifetime

philrennie:
All the docs I can find for IAM based RDS access say tokens have a 15 minute lifetime, but I can't find a documented reason why this value couldn't be configurable, it appears to be signed using standard AWS signature version 4.

Having generated tokens manually they seem to work fine with much shorter lifetimes, which is what I would like to happen in a particular project of mine. Shorter token lifetime == a bit more secure.

Seems to be a fairly small edit to make this user adjustable.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.changes/nextrelease/allow_changing_RDS_authtoken_lifetime.json
+++ b/.changes/nextrelease/allow_changing_RDS_authtoken_lifetime.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "Rds",
+    "description": "Added optional configurable lifetime value to Rds AuthTokenGenerator"
+  }
+]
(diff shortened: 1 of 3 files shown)


SamRemis:
@philrennie I apologize for the long delay- I contacted the RDS team to see if this was allowed, and it's not a currently supported feature.  What you're doing should not throw any sort of error, but the lifetime won't be changed.  They are currently looking into whether or not they would like to implement this feature.  I will give you an update once they get back to me.

SamRemis:
@philrennie Upon looking into it a bit more, I realized this works with the presign URL, not with the RDS service, so it's okay to merge.  This does need a few validations that the lifetime is <16 and >0 as well as a type check.  I can add those and the test that go along with it if you give me write access to your branch

philrennie:
@SamRemis great news :) I would have tried to add the validations myself, but I couldn't figure out from the existing code what the house style is for errors and exceptions, so I'll take your kind offer and have invited you to collaborate on the repo.

Cheers

SamRemis:
@philrennie thank you for your contribution :) I have to wait at least one day to merge it, but will definitely do so by the end of the week

ajredniwja (approved):

PR #2026, opened 2020-05-28
feat: Polyfill for AES-GCM on PHP < 7.1

scottarc:
*Issue #, if available:* N/A

*Description of changes:*

If you are using PHP 7.1 or newer, this will not affect you at all, because AES-GCM is already available for you.

If you're still on an older version of PHP, this change will allow you to interoperate with other SDK implementations and/or PHP versions that use AES-GCM to encrypt/decrypt customer data.

**Technical Details**:

This pull request implements a pure-PHP implementation of GMAC; which, when combined with OpenSSL's AES implementations (both AES-ECB and AES-CTR) allows us to safely provide AES-GCM on PHP versions older than 7.1. (We specifically target 5.5 at minimum.)

These changes were developed by [AWS Crypto Tools](https://github.com/aws/crypto-tools) with the following goals:

1. Correctness to the AES-GCM specification
2. Algorithmic constant-time
3. Low memory overhead

Unfortunately, it is impossible to meet those three goals with PHP code and still have performance competitive with OpenSSL's AES-GCM implementation. Consequently, if you're on an older version of PHP and must use this implementation, it will be slower than what PHP 7.1 provides.

The provided AES-GCM implementation contained in this pull request only supports 96-bit nonces and 128-bit authentication tags.

In addition to implementing AES-GCM, the following classes have been updated to use the new code on PHP versions older than 7.1:

* `Aws\Crypto\AesGcmDecryptingStream`
* `Aws\Crypto\AesGcmEncryptingStream`

The unit tests include the relevant NIST test vectors for AES-GCM.


tests/Crypto/AesGcmEncryptingStreamTest.php
```diff
@@ -30,6 +30,11 @@ public function testStreamOutputSameAsOpenSSL(
         StreamInterface $plainText,
         $keySize
     ) {
+        if (version_compare(PHP_VERSION, '7.1', '<')) {
```
howardlopez, line 33: Should there be a separate test for `AesGcmDecryptingStreamTest` and `AesGcmEncryptingStreamTest` that covers the < PHP 7.1 case (which is now supported with the `AesGcm` class implementation)?

src/Crypto/Polyfill/ByteArray.php
```diff
@@ -0,0 +1,265 @@
...
+     * @param int $index
+     * @param int $newval
+     * @return void
+     */
+    public function offsetSet($index, $newval)
+    {
+        try {
+            parent::offsetSet($index, $newval & 0xff);
+        } catch (\Exception $ex) {
+            var_dump([$index, $newval]);
```
howardlopez, line 147: Should there be a test to cover this exception in `ByteArrayTest`?

src/Crypto/Polyfill/ByteArray.php
```diff
@@ -0,0 +1,265 @@
...
+ * @package Aws\Crypto\Polyfill
+ */
+class ByteArray extends \SplFixedArray
+{
+    use NeedsTrait;
+
+    /**
+     * ByteArray constructor.
+     *
+     * @param int|string|int[] $size
```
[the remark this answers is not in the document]
scottarc: Good catch!

tests/Crypto/AesGcmEncryptingStreamTest.php
scottarc: There is. This is just an OpenSSL compatibility test with random inputs, which isn't possible to run on PHP <7.1.

src/Crypto/Polyfill/ByteArray.php
scottarc: It should be removed. I'll fix that in the next commit.

src/Crypto/Polyfill/AesGcm.php
```diff
@@ -0,0 +1,217 @@
+<?php
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
scottarc: That's standard procedure for AWS Crypto Tools. If it needs to be removed, I can remove it.

PR #2033, opened 2020-06-05
Fixed guzzle v6 compatibility

michalcharvat:
This fix resolve compatibility problem with Guzzle v6 closure. After search trought internet there is probably same error in v5 handler too. 

Major problem in this case is incompatibility with PHP 5.5+ - throwable is available since PHP 7.0. However PHP < 7.2 is dead - currently only 7.2, 7.3. and 7.4 is supported. 

`Uncaught Exception TypeError: "Argument 1 passed to Aws\Handler\GuzzleV6\GuzzleHandler::Aws\Handler\GuzzleV6\{closure}() must be an instance of Exception, instance of Error given, called in /application/vendor/guzzlehttp/promises/src/Promise.php on line 203" at /application/vendor/aws/aws-sdk-php/src/Handler/GuzzleV6/GuzzleHandler.php line 45`

### DIFF
--- a/src/Exception/AwsException.php
+++ b/src/Exception/AwsException.php
@@ -48,7 +48,7 @@ class AwsException extends \RuntimeException implements
         $message,
         CommandInterface $command,
         array $context = [],
-        \Exception $previous = null
+        \Throwable $previous = null
     ) {
         $this->data = isset($context['body']) ? $context['body'] : [];
         $this->command = $command;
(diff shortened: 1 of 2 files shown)


howardlopez:
Can you provide an example of the issue that's prompting this PR, with a code sample and error message? 

Note that any PR will have to make sure it works for PHP 5.5 and later, as the SDK still supports those versions.

SamRemis:
Since this PR is incompatible with php 5.x, we are closing this for now.

Issue #1753, opened 2019-03-19
DateTime::__construct(): Failed to parse time string (XXXX) at position: Unexpected character


liubo1985:
Recently, we migrated to AWS SDK V3, but got this error sometimes. Seems like the error is from parsing s3 response. Could someone give some help? Thanks in advance. 

DateTime::__construct(): Failed to parse time string (511301473873) at position 11 (3): Unexpected character

The stack trace:
../vendor/aws/aws-sdk-php/src/Api/Parser/
in DateTime::__construct called at ../vendor/aws/aws-sdk-php/src/Api/Parser/AbstractRestParser.php
AbstractRestParser.php (123)
…/vendor/aws/aws-sdk-php/src/Api/Parser/
AbstractRestParser.php (46)
…/vendor/aws/aws-sdk-php/src/S3/
RetryableMalformedResponseParser.php (37)
…/vendor/aws/aws-sdk-php/src/S3/
AmbiguousSuccessParser.php (58)
…at /vendor/aws/aws-sdk-php/src/S3/
GetBucketLocationParser.php (29)
… called at ../vendor/aws/aws-sdk-php/src/
WrappedHttpHandler.php (125)
… called at ../vendor/aws/aws-sdk-php/src/
WrappedHttpHandler.php (92)
…losure} called at ../vendor/guzzlehttp/promises/src/
Promise.php (203)
…Handler called at /vendor/guzzlehttp/promises/src/
Promise.php (156)
…sure} called at /vendor/guzzlehttp/promises/src/
TaskQueue.php (47)
…at /vendor/guzzlehttp/guzzle/src/Handler/
CurlMultiHandler.php (98)
…at /vendor/guzzlehttp/guzzle/src/Handler/
CurlMultiHandler.php (125)
…execute called at /vendor/guzzlehttp/promises/src/
Promise.php (246)
…eWaitFn called at /vendor/guzzlehttp/promises/src/
Promise.php (223)
…Pending called at /vendor/guzzlehttp/promises/src/
Promise.php (267)
…aitList called at /vendor/guzzlehttp/promises/src/
Promise.php (225)
…Pending called at /vendor/guzzlehttp/promises/src/
Promise.php (267)
…aitList called at /vendor/guzzlehttp/promises/src/
Promise.php (225)
…Pending called at /vendor/guzzlehttp/promises/src/
Promise.php (62)
…wait called at /vendor/aws/aws-sdk-php/src/
AwsClientTrait.php (58)

[...]

PR #2038, opened 2020-06-10
Feature: defensive timestamp parsing

SamRemis:
Issue #, if available:
#1753
Description of changes:
Added defensive timestamp parsing


tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+
+use Aws\Api\Parser\Exception\ParserException;
+use Aws\Api\Service;
+use Aws\AwsClient;
+use GuzzleHttp\Promise\FulfilledPromise;
+use GuzzleHttp\Psr7\Response;
+use PHPUnit\Framework\TestCase;
+
+/*
+ * This class tests the PHP specific functionality of the JsonParser;
```
howardlopez, line 11: This doc bloc should be folded into the doc block for the class itself (i.e. above the @covers statements). [PHPDoc reference](https://docs.phpdoc.org/latest/getting-started/your-first-set-of-documentation.html)

Also, rather than mentioning PHP vs cross-SDK, it'd probably be more accurate to say something like generic parsing functionality vs. custom parsing functionality (since this defensive timestamp parsing is essentially custom handling for the benefit of the user).

tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+            [
+                'metadata' => [
+                    "protocol" => "json",
+                    "apiVersion" => "2014-01-01"
+                ],
+                'shapes' => [
+                    "ParseJsonRequest" => [
+                        "type" => "structure",
+                        "members" => [
+                            "timestamp" =>[
```
howardlopez, line 174: This is a super nitpick that doesn't really matter in a test model, but shapes and members are supposed to be capitalized (a few services deviate, but those are legacy models).

tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+ * cross-SDK testing is done in ComplianceTest.php
+ */
+
+/**
+ * @covers \Aws\Api\Parser\JsonRpcParser
+ * @covers \Aws\Api\Parser\JsonParser
+ */
+class JsonParserTest extends TestCase
+{
+    public function testUnixTimestampReturnedFromService()
```
howardlopez, line 21: Should the name be more reflective of the actual test case, like `testUnixTimestampWhenModeledAsIso8601`?

tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+        $service = $this->generateTestService();
+        $client = $this->generateTestClient($service, ['timestamp' => 932169600]);
+        $command = $client->getCommand('ParseJson');
+        $list = $client->getHandlerList();
+        $handler = $list->resolve();
+        $result = $handler($command)->wait()['timestamp']->__toString();
+
+        self::assertEquals("1999-07-17T00:00:00+00:00", $result);
+    }
+    public function testTextDateReturnedFromService()
```
howardlopez, line 32: Perhaps `testTextDateWhenModeledAsIso8601`? And so on, for other tests. Although, if you end up using the data provider pattern, they'd end up under an umbrella test anyway.

tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+/*
+ * This class tests the PHP specific functionality of the JsonParser;
+ * cross-SDK testing is done in ComplianceTest.php
+ */
+
+/**
+ * @covers \Aws\Api\Parser\JsonRpcParser
+ * @covers \Aws\Api\Parser\JsonParser
+ */
+class JsonParserTest extends TestCase
```
howardlopez, line 19: Should there also be tests for an operation which is modeled as a unix timestamp, and returns the same kinds of things tested for in these cases (i.e. returning a unix timestamp, text date, an iso8601 date, etc)?

Since many of the tests are essentially repeated code with different inputs and expectations, the PHPUnit [@dataProvider pattern ](https://phpunit.readthedocs.io/en/9.1/writing-tests-for-phpunit.html#data-providers)(used throughout the SDK) seems like it would be a good fit here.

tests/Api/Parser/JsonParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    public function testStringReturnedFromService()
+    {
+        $service = $this->generateTestService();
+        $client = $this->generateTestClient($service, ['timestamp' => "this text is not a date"]);
+        $command = $client->getCommand('ParseJson');
+        $list = $client->getHandlerList();
+        $handler = $list->resolve();
+        $this->setExpectedException(ParserException::class);
```
howardlopez, line 72: PHPUnit has @expectedException and @expectedExceptionMessage that can be used to annotate tests. However, if you use a data provider, `setExpectedException` is likely the right approach, as that can take an input. Note that it's probably also wise to test for the exception message.

tests/Api/Parser/XmlParserTest.php
```diff
@@ -0,0 +1,206 @@
+<?php
```
howardlopez, line 1: The comments from the `JsonParserTest` should also apply to this test class.

src/Api/Parser/XmlParser.php
```diff
@@ -138,11 +139,10 @@ private function parse_boolean(Shape $shape, $value)
 
     private function parse_timestamp(Shape $shape, $value)
     {
-        if (!empty($shape['timestampFormat'])
-            && $shape['timestampFormat'] === 'unixTimestamp') {
-            return DateTimeResult::fromEpoch((string) $value);
+        if(is_string($value) || is_int($value) || (is_object($value) && method_exists($value, '__toString'))){
```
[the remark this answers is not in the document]
SamRemis: Yes objects do.  I'm not sure when, but it happens in the compliance test

tests/Api/Parser/JsonParserTest.php
```diff
@@ -159,7 +80,7 @@ private function generateTestClient(Service $service, $args = [])
         );
     }
 
-    private function generateTestService()
+    private function generateTestService($expectedFormat)
```
howardlopez, line 83: Instead of having a dynamic test service, it may be more idiomatic of the way models are constructed to just have 2 operations that have distinct response shapes, one that uses a unix timestamp member, and one that uses an iso8601 timestamp member. In the data provider above, you can then specify which operation to invoke in the test for each test case. Same would go for the XmlParserTest.

src/Api/DateTimeResult.php
```diff
@@ -76,7 +76,7 @@ public static function fromTimestamp($timestamp, $expectedFormat = null)
                 return self::fromISO8601($timestamp);
 
         } catch (Exception $exception) {
-            throw new ParserException('Invalid timestamp value passed to DateTimeResult::fromTimestamp');
+                throw new ParserException('Invalid timestamp value passed to DateTimeResult::fromTimestamp');
```
howardlopez, line 79: Style nitpick: Looks like this is double-indented (i.e. 8 spaces) now?

src/Api/DateTimeResult.php
```diff
@@ -13,19 +13,32 @@ class DateTimeResult extends \DateTime implements \JsonSerializable
...
     /**
      * Create a new DateTimeResult from a unix timestamp.
-     *
+     * The Unix epoch (or Unix time or POSIX time or Unix
+     * timestamp) is the number of seconds that have elapsed since
+     * January 1, 1970 (midnight UTC/GMT).
      * @param $unixTimestamp
      *
      * @return DateTimeResult
+     * @throws Exception
```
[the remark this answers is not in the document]
SamRemis: No, it does not.  It's a generic exception.  If I try to specify it, "PHPDoc comment doesn't contain all the necessary @throws tags"

src/Api/DateTimeResult.php
```diff
@@ -53,30 +53,29 @@ public static function fromTimestamp($timestamp, $expectedFormat = null)
...
+                } else if ($expectedFormat == 'unixTimestamp') {
+                    try {
+                        return self::fromEpoch($timestamp);
+                    }
+                    catch (Exception $exception) {
+                        return self::fromISO8601($timestamp);
                     }
                 }
+
                 else if (\Aws\is_valid_epoch($timestamp)) {
```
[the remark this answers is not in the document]
SamRemis: No, you were clear, it was just late when I did this and I was very tired.  It seems I read it too quickly.  Sorry about that

tests/Api/Parser/JsonParserTest.php
SamRemis: No you were clear, I just was very tired and read it too quickly, my apologies.  I think I got what you asked for this time.

tests/Api/Parser/JsonParserTest.php
```diff
@@ -15,116 +15,96 @@
...
-        $service = $this->generateTestService();
-        $client = $this->generateTestClient($service, ['UnknownTimestamp' => $timestamp]);
-        $command = $client->getCommand('ParseJson');
-        $list = $client->getHandlerList();
-        $handler = $list->resolve();
-        $this->setExpectedException($expectedException, $expectedMessage);
-        $handler($command)->wait();
-    }
 
     private function generateTestClient(Service $service, $args = [])
```
howardlopez, line 109: To avoid repetition, could the test client and test service code be shared as a trait across the `JsonParserTest` and `XmlParserTest`?

If the only difference is the metadata `protocol` value, could that be set from a method parameter if it's shared across the two test classes?

SamRemis: It's not the only difference.  They have some different input tests because of how the two protocols store values differently.  We could potentially combine them and have some tests that are shared between the two and some that are protocol specific using three data providers.  Would you rather that?

SamRemis: Yes; what would be a good name for the trait class?  ParserTestTrait?

howardlopez: `ParserTestServiceTrait` perhaps?

SamRemis: Great, I just renamed it and pushed it.  Thank you for the suggestion

src/Api/DateTimeResult.php
```diff
@@ -9,16 +13,69 @@ class DateTimeResult extends \DateTime implements \JsonSerializable
...
+    public static function fromTimestamp($timestamp, $expectedFormat = null)
+    {
+            if (empty($timestamp) || !(is_string($timestamp) || is_int($timestamp))) {
+                throw new ParserException('Invalid timestamp value passed to DateTimeResult::fromTimestamp');
+            }
+            try {
+                if ($expectedFormat == 'iso8601') {
+                    try {
+                        return self::fromISO8601($timestamp);
+                    }
```
howardlopez, line 59: Style nitpick: `catch` should be on the same line as the closing bracket, as per PSR-2 for [try-catch](https://www.php-fig.org/psr/psr-2/#56-try-catch).

src/Api/DateTimeResult.php
```diff
@@ -9,16 +13,69 @@ class DateTimeResult extends \DateTime implements \JsonSerializable
...
+                    catch (Exception $exception) {
+                        return self::fromEpoch($timestamp);
+                    }
+                } else if ($expectedFormat == 'unixTimestamp') {
+                    try {
+                        return self::fromEpoch($timestamp);
+                    }
```
howardlopez, line 66: See above comment for try-catch.

tests/Api/Parser/ParserTestServiceTrait.php
```diff
@@ -0,0 +1,135 @@
...
+                    ],
+                ],
+            ],
+            function () {
+                return [];
+            }
+        );
+    }
+
+}
```
howardlopez, line 135: Nitpick: files should have newline ending

.changes/nextrelease/feature-defensive-timestamp-parsing.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "feature",
+    "category": "Api",
+    "description": "Added defensive parsing of timestamps so it can take epoch or ISO8601 without knowing the type."
+  }
+]
```
howardlopez, line 7: Nitpick: files should have newline ending

PR #2039, opened 2020-06-11
Update major version of guzzle calculation

gmponos:


gmponos:
cc @Nyholm maybe you would also like to have a look here?

Nyholm:
This will it work with beta1. Is that a problem? Should we add a conflict in composer.json? If so, when do we remove that?

Or many we don’t care since we define this to work with guzzle7. It will work with guzzle7.

gmponos:
hm... maybe this should wait until the RC. 

IIRC versions with RC do not allow BC so things can not break after..

gmponos:
not sure why the CI is not running.. but I think that now that guzzle has RC tagged this can be merged.

gmponos:
Unfortunately phpstan build is failing. There is a PHPStan known issue with constant and `defined`. Should I add an `ignoreError` in the phpstan.neon ?

Nyholm:
Yes please. It is a priority that this is fixed.

Nyholm:
I made a [small PR](https://github.com/aws/aws-sdk-php/pull/2050) to show that master is red.

gmponos:
The failing build is unrelated.

Nyholm:
Thank you for merging.

PR #2044, opened 2020-06-17
Fix broken link for Guzzle mock plugin

### LINKED ISSUE #2029: 404 not found link in Static Client Facades
**Describe the issue with documentation**
[Feature facades docs](https://docs.aws.amazon.com/aws-sdk-php/v2/guide/feature-facades.html) ends with: 

> you might consider using the [Guzzle Mock Plugin](http://guzzlephp.org/plugins/mock-plugin.html). 

However this links does not exist anymore. I found the proper one for [Guzzle 3](https://guzzle3.readthedocs.io/plugins/mock-plugin.html) though. Maybe [that](http://docs.guzzlephp.org/en/stable/testing.html?highlight=mock) for Guzzle6?

diehlaws:
*Issue #, if available:*
Fix #2029 

*Description of changes:*
Update link for Guzzle mock plugin to working URL.

### DIFF
--- a/docs/feature-facades.rst
+++ b/docs/feature-facades.rst
@@ -167,4 +167,4 @@ showing how this is done:
     }
 
 Alternatively, if you are specifically only mocking responses from clients, you might consider using the `Guzzle Mock
-Plugin <http://guzzlephp.org/plugins/mock-plugin.html>`_.
+Plugin <https://guzzle3.readthedocs.io/plugins/mock-plugin.html>`_.
--- a/docs/feature-facades.rst
+++ b/docs/feature-facades.rst
@@ -166,5 +166,5 @@ showing how this is done:
         }
     }
 
-Alternatively, if you are specifically only mocking responses from clients, you might consider using the `Guzzle Mock
-Plugin <https://guzzle3.readthedocs.io/plugins/mock-plugin.html>`_.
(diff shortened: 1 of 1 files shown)


docs/feature-facades.rst
```diff
@@ -167,4 +167,4 @@ showing how this is done:
     }
 
 Alternatively, if you are specifically only mocking responses from clients, you might consider using the `Guzzle Mock
-Plugin <http://guzzlephp.org/plugins/mock-plugin.html>`_.
+Plugin <https://guzzle3.readthedocs.io/plugins/mock-plugin.html>`_.
```
gmponos, line 170: This is not the correct link.. this is from Guzzle v3 which is not supported anymore...

This is the link for G6-7: http://docs.guzzlephp.org/en/stable/testing.html#mock-handler
And this is the link fro G5: http://docs.guzzlephp.org/en/5.3/testing.html#mock-subscriber

Unless if the doc here is intent is to point to Guzzle 3

diehlaws: This change is for the documentation for version 2 of the AWS SDK for PHP, which depends on Guzzle 3. That said, I suppose there's no reason not to include links for mocking HTTP responses using newer versions of Guzzle.

PR #2047, opened 2020-06-25
Feature client constructor no config option

SamRemis:
*Issue #, if available:*

*Description of changes:*
Added feature to use no ini file in configuration provider using client constructor parameter


tests/Credentials/CredentialProviderTest.php
```diff
@@ -224,6 +224,54 @@ public function iniFileProvider()
         ];
     }
 
+    private static $standardIni = <<<EOT
```
howardlopez, line 227: Class members should all be declared before any methods.

src/AwsClient.php
```diff
@@ -171,6 +171,8 @@ public static function getArguments()
      *   signature version to use with a service (e.g., v4). Note that
      *   per/operation signature version MAY override this requested signature
      *   version.
+     * - use_aws_config_file: (bool, default=bool(true)) Set to false to
```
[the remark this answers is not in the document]
SamRemis: Changed to use_aws_shared_files

src/ClientSideMonitoring/ConfigurationProvider.php
```diff
@@ -77,11 +77,15 @@ class ConfigurationProvider extends AbstractConfigurationProvider
...
-            self::ini(),
-            self::fallback()
-        ];
+        $configProviders = [];
+        array_push($configProviders, self::env());
+        if (
+            !isset($config['use_aws_config_file'])
+            || $config['use_aws_config_file'] != false
+        ) {
+            array_push($configProviders, self::ini());
```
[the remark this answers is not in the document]
SamRemis: Since they're effectively the same, I checked and found one usage of array_push in the src folder and modified that as well; it was in src/S3/PostObjectV4.  I figured since there was only a minor change it wouldn't need its own PR, but let me know if you want me to revert that or anything

tests/S3/UseArnRegion/ConfigurationProviderTest.php
```diff
@@ -99,6 +99,37 @@ public function testCreatesDefaultFromFallback()
         $this->assertSame($expected->toArray(), $result->toArray());
     }
 
+    public function testUsesIniWithUseAwsConfigFileTrue()
+    {
+        $dir = $this->clearEnv();
+        putenv(ConfigurationProvider::ENV_CONFIG_FILE . '=' . $dir . "/alt_config");
+        $expected = new Configuration(true);
+        file_put_contents($dir . '/config', $this->iniFile);
+        file_put_contents($dir . '/alt_config', $this->altIniFile);
```
howardlopez, line 108: Is there any reason this test needs an alt_config?

SamRemis: Yes, I removed the /config to clear that up.  The default values come in with no change if I just use the normal config file, so the test doesn't know for sure whether or not it actually used the file. 

 The other options are adding a new redundant configuration with a new name, or I could putting the altIniFile's contents in as the normal config file.  Let me know if you'd prefer I did it another way, I kept in the /alt_config file to be clear that we were using the input of the alt ini file instead of the standard one

howardlopez: The way you changed it seems reasonable.

PR #2050, opened 2020-06-30
Guzzle 7 is released, no beta test needed

Nyholm:
We test with Guzzle 5 on "prefer-lowest".
We test with Guzzle 6 on PHP < 7.2
We test with Guzzle 7 on PHP ≥ 7.2


Nyholm:
The CI failure is unrelated to this PR. It seams like they are related because of the release to Guzzle 7. I dont know the build sequence, but it seams like it is just CI that fails, but the library is working.

PR #2061, opened 2020-07-16
Region validation

howardlopez:
* Adds region validation as a valid host label when region is being used to construct an endpoint. Note this does not take effect when a custom endpoint is supplied.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -440,8 +440,7 @@ function is_valid_hostname($hostname)
  */
 function is_valid_hostlabel($label)
 {
-    return preg_match("/^([a-z\d](-*[a-z\d])*)$/", $label)
-        && preg_match("/^.{1,63}$/", $label);
+    return preg_match("/^(?!-)[a-zA-Z0-9-]{1,63}(?<!-)$/", $label);
 }
 
 /**
--- a/src/functions.php
+++ b/src/functions.php
@@ -433,6 +433,18 @@ function is_valid_hostname($hostname)
 }
 
 /**
+ * Checks if supplied parameter is a valid host label
+ *
+ * @param $label
+ * @return bool
+ */
+function is_valid_hostlabel($label)
+{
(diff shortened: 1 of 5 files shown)


src/functions.php
```diff
@@ -432,6 +432,18 @@ function is_valid_hostname($hostname)
...
+/**
+ * Checks if supplied parameter is a valid host label
+ *
+ * @param $label
+ * @return bool
+ */
+function is_valid_hostlabel($label)
+{
+    return preg_match("/^([a-z\d](-*[a-z\d])*)$/", $label)
+        && preg_match("/^.{1,63}$/", $label);
```
SamRemis, line 444: Go is using a different regex: "[a-zA-Z0-9-]{1,63}"
Is there anywhere we could find agreed upon rules for the regex?

SamRemis, line 444: Genuine question that I'm not sure how to find the answer to: how does preg_match with that specific input vary in computational time from just checking the length of the string and doing a less than or greater than?

SamRemis: ruby is using this: /^(?![0-9]+$)(?!-)[a-zA-Z0-9-]{,63}(?<!-)$/

SamRemis: Follow up: after asking everyone, it seems like there's no standard agreed upon option, so this works great :) hopefully in the future there will be one regex that every team implements

skmcgrail: Chiming in here @SamRemis since I know you brought it up earlier, https://github.com/aws/aws-sdk-go/blob/master/aws/endpoints/v3model.go#L10 is the specific regex that we are using to validate this functionality.

howardlopez: Updated to use a newly proposed regex. Will wait until cross-sdk test cases are established and update with those as well.

howardlopez: New regex includes the length validation within itself, so need to worry about this case.

PR #2069, opened 2020-07-28
Client-side encryption client user agent string

howardlopez:
* Add crypto-specific user-agent string to encryption clients.


src/S3/Crypto/S3EncryptionMultipartUploaderV2.php
```diff
@@ -19,7 +19,10 @@
  */
 class S3EncryptionMultipartUploaderV2 extends MultipartUploader
 {
-    use EncryptionTraitV2, CipherBuilderTrait, CryptoParamsTraitV2;
+    use CipherBuilderTrait;
```
SamRemis, line 22: Thank you for alphabetizing these

src/S3/Crypto/S3EncryptionMultipartUploaderV2.php
```diff
@@ -105,6 +108,7 @@ public function __construct(
         $source,
         array $config = []
     ) {
+        $this->appendUserAgent($client, 'S3CryptoV2.0');
```
SamRemis, line 111: At some point, have we considered a global file for these magic strings that are repeated across files like "S3CryptoV2.0".  There's so many different ones that I feel like implementing that would be messier and may be more trouble than it's worth, I'm just curious if the idea has been floated before

src/S3/Crypto/S3EncryptionMultipartUploader.php
```diff
@@ -19,7 +19,12 @@
  */
 class S3EncryptionMultipartUploader extends MultipartUploader
 {
-    use EncryptionTrait, CipherBuilderTrait, CryptoParamsTrait;
+    use CipherBuilderTrait;
+    use CryptoParamsTrait;
+    use EncryptionTrait;
+    use UserAgentTrait;
+
+    const VERSION = '1';
```
SamRemis, line 27: I'd rather see this be named something like "CRYPTOGRAPHY_VERSION" so we know what it's referring to without having to find where it's used later in the code.

PR #2083, opened 2020-08-18
added credential source

SamRemis:
*Issue #, if available:*

*Description of changes:*


.changes/nextrelease/credential_source.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "feature",
+    "category": "Credentials",
+    "description": "This change adds support for the credential_source option in the credential file."
+  }
+]
```
howardlopez, line 7: Nitpick: missing newline

src/Credentials/CredentialProvider.php
```diff
@@ -584,43 +584,57 @@ private static function loadRoleProfile($profiles, $profileName, $filename, $sts
...
-            ];
-            $sourceCredentials = call_user_func(
-                CredentialProvider::ini($sourceProfileName, $filename, $config)
-            )->wait();
-            $stsClient = new StsClient([
-                'credentials' => $sourceCredentials,
-                'region' => $sourceRegion,
-                'version' => '2011-06-15',
-            ]);
+        if (
```
howardlopez, line 596: Is this conditional necessary given the previous conditional with the rejection?

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +706,35 @@ private static function loadDefaultProfiles() {
...
+    public static function getAssumeRoleCredentials(
+        $sourceProfileName = '',
+        $profileName = '',
+        $filename = '',
+        $config = []
+    )
+    {
+        $data = self::loadProfiles($filename);
+        $credentialSource = !empty($data[$profileName]['credential_source']) ? $data[$profileName]['credential_source'] : null;
+        if (isset($credentialSource) && is_string($credentialSource)) {
```
howardlopez, line 718: Do you need the `is_string`, considering you have a default exception throw in the switch below?

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +706,35 @@ private static function loadDefaultProfiles() {
...
+            switch ($credentialSource) {
+                case 'Environment':
+                    return self::env();
```
howardlopez, line 721: For each of the cases, if, say in this case, the env() doesn't contain proper credentials, does it make sense to instead throw a new intermediate exception explicitly specifying that the env credentials were specified by the credential source and not found? I'm thinking of the scenario in which a customer may see an env exception, but not realize why it's failing there (instead of going through the full default chain).

src/Credentials/CredentialProvider.php
SamRemis: Good catch, I think I initially had the logic working differently so I had to do that, but changed it to an xor in the form of an == and forgot to change it

src/Credentials/CredentialProvider.php
SamRemis: Whats the most idiomatic way of throwing this exception?  Since self::env returns a callback, i can't put a try catch around this specifically.  The only thing I could think off off the top of my head is to add an optional parameter like so:

    public static function env($isCredentialsSource = false){
//happy path here, removed for brevity
            $errorMessage = 'Could not find environment variable '
                . 'credentials in ' . self::ENV_KEY . '/' . self::ENV_SECRET);
            if ($isCredentialsSource){
                $errorMessage = "Provided credential_source 'Environment' in credentials failed to retrieve credentials" .
                                              "with exception: " . $errorMessage
             }
            return self::reject();
        };
}

//please ignore the awful spacing I wrote this on github

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +702,52 @@ private static function loadDefaultProfiles() {
         return $profiles;
     }
 
+    public static function getAssumeRoleCredentials(
+        $sourceProfileName = '',
+        $profileName = '',
+        $filename = '',
+        $config = []
+    )
```
howardlopez, line 710: Style nitpick: bracket should be on same line as parentheses

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +702,52 @@ private static function loadDefaultProfiles() {
...
+                    $credentialsPromise = self::ecsCredentials();
+                    break;
+                default:
+                    throw new CredentialsException(
+                        "Invalid credential_source found in config file: {$credentialSource}. Valid inputs "
+                        . "include Environment, Ec2InstanceMetadata, and EcsContainer."
+                    );
+            }
+            $credentialsResult = null;
+            try{
```
howardlopez, line 733: Style nitpick: need space before `{`

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +702,52 @@ private static function loadDefaultProfiles() {
         return $profiles;
     }
 
+    public static function getAssumeRoleCredentials(
```
howardlopez, line 705: Should this function be separated out into 2 functions, one that handles source_profile and one that handles credential source? They don't seem to share much in the way of logic or parameters with each other, so it could be cleaner to have one function for each.

src/Credentials/CredentialProvider.php
```diff
@@ -692,6 +702,52 @@ private static function loadDefaultProfiles() {
...
+        $sourceProfileName = '',
+        $profileName = '',
+        $filename = '',
+        $config = []
+    )
+    {
+        $data = self::loadProfiles($filename);
+        $credentialSource = !empty($data[$profileName]['credential_source']) ? $data[$profileName]['credential_source'] : null;
+        $credentialsPromise = null;
```
howardlopez, line 714: Should this statement be inside the `if` block, since it's not used outside of it?

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+        putenv('HOME=' . dirname($dir));
+
+        try {
+            $result = CredentialProvider::getAssumeRoleCredentials(
+                'assume',
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
```
howardlopez, line 688: Is there any way to test this rejected promise is specifically the env rejection we're looking for?

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+        putenv('HOME=' . dirname($dir));
+        try {
+            $result = CredentialProvider::getAssumeRoleCredentials(
+                'assume',
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
+            try{
```
howardlopez, line 745: Style nitpick: spacing

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+                $result->wait();
+            }
+            catch (\Exception $exception){
+                self::assertContains(
```
howardlopez, line 749: Style nitpick:

```
self::assertContains(
    'Error retrieving credentials from the instance profile metadata service',
    $exception->getMessage()
);
```

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+            $result = CredentialProvider::getAssumeRoleCredentials(
+                'assume',
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
+            try{
+                $result->wait();
+            }
```
howardlopez, line 747: Style: `catch` should be on same line as close bracket

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+            catch (\Exception $exception){
```
howardlopez, line 748: Style: spacing before bracket

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+        putenv('HOME=' . dirname($dir));
+        try {
+            $result = CredentialProvider::getAssumeRoleCredentials(
+                'assume',
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
+            try{
```
howardlopez, line 777: Style: spacing

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+                $result->wait();
+            }
+            catch (\Exception $exception){
```
howardlopez, line 780: Style: `catch` should be on same line as bracket

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+        putenv('HOME=' . dirname($dir));
+        try {
+            $result = CredentialProvider::getAssumeRoleCredentials(
+                'assume',
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
+            try{
```
howardlopez, line 813: Style: spacing

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+                $result->wait();
+            }
+            catch (\Exception $exception){
```
howardlopez, line 816: Style: `catch` should be on same line as bracket.

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
...
+                self::assertContains(
```
howardlopez, line 817: Style: see previous test

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,167 @@ public function testEnsuresSourceProfileIsSpecified()
         }
     }
 
+    public function testAssumeRoleInConfigFromCredentialsSourceFails()
```
howardlopez, line 667: * Can you add a case where the user sets both credential_source and source_profile?
* Is there a way to mock the success case for Ec2InstanceMetadata and EcsContainer? Ideally there'd be a fail and success case for each type.

src/Credentials/CredentialProvider.php
SamRemis: Sure, but source_profile doesn't have much logic so I will just add that up above instead if that is alright

src/Credentials/CredentialProvider.php
SamRemis: Removed the if block since it was no longer necessary

tests/Credentials/CredentialProviderTest.php
SamRemis: No unfortunately not, I tried for a while to get it to work.  The reason it won't work is because of the fact that you can't inject a mock into this function.  Take the ecs for instance.  Given that it is a static method that only takes a config, you can't inject a mock in any way:
    public static function ecsCredentials(array $config = [])
    {
        return new EcsCredentialProvider($config);
    }

src/Credentials/CredentialProvider.php
```diff
@@ -604,9 +610,18 @@ private static function loadRoleProfile($profiles, $profileName, $filename, $sts
...
             ];
-            $sourceCredentials = call_user_func(
-                CredentialProvider::ini($sourceProfileName, $filename, $config)
-            )->wait();
+            $sourceCredentials = null;
+            if ($profiles[$profileName]['source_profile']){
+                $sourceCredentials = call_user_func(
+                    CredentialProvider::ini($sourceProfileName, $filename, $config)
+                )->wait();
+            }
```
howardlopez, line 618: Style nitpick: `else` should be on same line as bracket

tests/Credentials/CredentialProviderTest.php
```diff
@@ -664,6 +664,188 @@ public function testEnsuresSourceProfileIsSpecified()
...
+            $result = CredentialProvider::getCredentialsFromSource(
+                'assume',
+                $dir . '/credentials',
+                []
+            );
+            self::assertInstanceOf('GuzzleHttp\Promise\RejectedPromise', $result);
+            $result->wait();
+        } catch (\Exception $exception) {
+            throw $exception;
+        }finally {
```
howardlopez, line 695: Style: needs space before `finally`

tests/Credentials/CredentialProviderTest.php
howardlopez: > Is there any way to test this rejected promise is specifically the env rejection we're looking for?

Bump

tests/Credentials/CredentialProviderTest.php
howardlopez: Makes sense. More invasive measures could be taken by adding mock functions to replace `instanceProfile()` and the like, but at that point you'd only be testing the method is called, which is already being tested via the expected exception message here.

howardlopez (changes requested):
Should a note be added in the code doc description for `ini` that `credential_source` is an option? The current description doesn't necessarily indicate that the `ini` will be able to redirect to another credential source. Also, should there be a corresponding PR for the dev guide for credentials?

_____

(Sam): I can't reply to these top level comments, so I'm just going to reply here:
https://github.com/awsdocs/aws-php-developers-guide/pull/88
Above is the PR for the dev guide, and I added the note to the ini

howardlopez (approved):

PR #2101, opened 2020-09-16
feature-cached-cli-sso-credential-provider

SamRemis:
*Issue #, if available:*

*Description of changes:*
Added support for using cached SSO credentials from the CLI


src/Credentials/CredentialProvider.php
```diff
@@ -296,6 +302,82 @@ public static function instanceProfile(array $config = [])
...
+                . '/.aws/sso/cache/'
+                . utf8_encode(sha1($ssoProfile['sso_start_url']))
+                . ".json";
+
+            if (!is_readable($tokenLocation)) {
+                return self::reject("Unable to read token file at $tokenLocation");
+            }
+
+            $tokenData = json_decode(file_get_contents($tokenLocation), true);
+            if (!empty($tokenData['accessToken']) && !empty($tokenData['expiresAt'])) {
```
howardlopez, line 341: Should this throw an exception if either `accessToken` or `expiresAt` is empty?

src/Credentials/CredentialProvider.php
```diff
@@ -296,6 +302,82 @@ public static function instanceProfile(array $config = [])
...
+                    return self::reject("Cached SSO credentials returned an invalid expiration");
+                }
+                $now = new DateTimeResult();
+                if ($expiration < $now) {
+                    return self::reject("Cached SSO credentials returned expired credentials");
+                }
+            }
+
+            $ssoClient = null;
+            if(empty($config['ssoClient'])) {
```
howardlopez, line 354: Style: spacing on if

tests/Credentials/CredentialProviderTest.php
```diff
@@ -897,6 +897,108 @@ public function testEnsuresSourceProfileHasCredentials()
...
+    public function testSsoProfileProviderBadFile()
+    {
+        $dir = $this->clearEnv();
+
+        $filename = $dir . '/config';
+        putenv('HOME=' . dirname($dir));
+
+        try {
+            call_user_func(CredentialProvider::sso('default', $filename))->wait();
+        } catch (\Exception $e) {
```
howardlopez, line 973: Is there any reason to wrap this in a try-catch and rethrow, if the test has an expected exception?

.changes/nextrelease/feature-sso-credential-provider.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "feature",
+    "category": "Credentials",
+    "description": "Added credential provider which retrieves cached SSO credentials from the CLI"
+  }
+]
```
howardlopez, line 7: Nitpick: needs newline at end of file.

tests/Credentials/CredentialProviderTest.php
```diff
@@ -897,6 +897,108 @@ public function testEnsuresSourceProfileHasCredentials()
         }
     }
```
howardlopez, line 899: Should there be a test that checks to make sure the `sso` provider is part of the default provicer chain?

howardlopez (changes requested):

howardlopez:
Forgot to add - can you modify the title and description of the PR to match its contents (i.e. to be useful for future reference)?

tests/Credentials/CredentialProviderTest.php
```diff
@@ -897,6 +897,108 @@ public function testEnsuresSourceProfileHasCredentials()
...
+        try {
+            call_user_func(CredentialProvider::sso('default', $filename))->wait();
+        } catch (\Exception $e) {
+            throw $e;
+        }
+    }
+
+    /**
+     * @expectedException \Aws\Exception\CredentialsException
+     * @expectedExceptionMessage  must contain the following keys: sso_start_url, sso_region, sso_account_id, and sso_role_name
```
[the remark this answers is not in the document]
SamRemis: Yes, that was intentional.  Since 'expectedExceptionMessage' just does a contains, and the message is variable with the filename as shown below.  This takes the most significant part of the message and checks for it.  The other option was using @expectedExceptionMessageRegExp which would add a bit of overhead that doesn't seem necessary.   It's just a small amount and only happens when tests are run, so I can add it if you'd prefer

                   ```
 "Profile {$ssoProfileName} in {$filename} must contain the following keys: "
                    . 'sso_start_url, sso_region, sso_account_id, and sso_role_name'
```

File: src/Credentials/CredentialProvider.php [hunk 7]
```diff
@@ -478,8 +560,8 @@ public static function ini($profile = null, $filename = null, array $config = []
             if (empty($data[$profile]['aws_session_token'])) {
                 $data[$profile]['aws_session_token']
                     = isset($data[$profile]['aws_security_token'])
-                        ? $data[$profile]['aws_security_token']
-                        : null;
+                    ? $data[$profile]['aws_security_token']
```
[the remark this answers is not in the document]
SamRemis: Yeah not sure why that happens.  I've seen it in a few commits now and usually catch it before you review- it seems to be something my IDE does automatically sometimes, even when I don't remember running any sort of autoformat.  I'll have to look into that.  Thank you

tests/Credentials/CredentialProviderTest.php
SamRemis: No particular reason other than consistency within the test file with some older tests.  I've removed it now.

howardlopez: Looks like the try-catch is still here?

howardlopez (changes requested):
Looking good, just missing a change in the test file.

tests/Credentials/CredentialProviderTest.php
howardlopez: Think you may have targeted a different test in your latest commit.

howardlopez (approved):

src/Credentials/CredentialProvider.php
```diff
@@ -88,6 +89,11 @@ public static function defaultProvider(array $config = [])
             !isset($config['use_aws_shared_config_files'])
             || $config['use_aws_shared_config_files'] !== false
         ) {
+            $defaultChain['sso'] = self::sso(
```
jkenn99, line 92: This introduces a new PHP notice when there is no `[profile default]` section in the ~/.aws./config file, which should be always, since the default section is supposed to be `[default]`:

src/Credentials/CredentialProvider.php
```diff
@@ -296,6 +302,85 @@ public static function instanceProfile(array $config = [])
...
+    public static function sso($ssoProfileName, $filename = null, $config = [])
+    {
+        $filename = $filename ?: (self::getHomeDir() . '/.aws/config');
+
+        return function () use ($ssoProfileName, $filename, $config) {
+            if (!is_readable($filename)) {
+                return self::reject("Cannot read credentials from $filename");
+            }
+            $data = self::loadProfiles($filename);
+            $ssoProfile = $data[$ssoProfileName];
```
jkenn99, line 319: This should check for existence before assignment.
```
$ssoProfile = !empty($data[$ssoProfileName]) ? $data[$ssoProfileName] : null;
```

jkenn99 (commented):
This change has introduced this PHP notice whenever trying to lookup credentials using the default provider chain:
```
PHP Notice:  Undefined index: profile default in somepath/vendor/aws/aws-sdk-php/src/Credentials/CredentialProvider.php on line 319
```

Issue #2097, opened 2020-09-09
AssumeRoleWithWebIdentityCredentialProvider: failed to open stream, permission denied


nesar77:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Trying to delete or putobject on S3 storage using PODs credential: it works but I'm getting a warning 
<b>Warning</b>:  file_get_contents(/var/run/secrets/eks.amazonaws.com/serviceaccount/token): failed to open stream: Permission denied in <b>/usr/local/backend/vendor/aws/aws-sdk-php/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php</b> on line <b>97</b><br />
<br />

**Version of AWS SDK for PHP?**
v3.151.6

aws/aws-sdk-php                    3.151.6 AWS SDK for PHP - Use Amazon Web Services in your PHP project

**Version of PHP (`php -v`)?**

PHP 7.4.10 (cli)

**To Reproduce (observed behavior)**

`<?php

namespace BackEnd\Services;

define('PROJECT_ROOT', realpath(__DIR__ . '/..'));
require PROJECT_ROOT . '/../../vendor/autoload.php';

use Aws\S3\S3Client;
use Aws\Exception\AwsException;
use Aws\Credentials\CredentialProvider;
use Aws\S3\Exception\S3Exception;

class AWSS3
{

    // ACL flags
    const ACL_PRIVATE = 'private';
    const ACL_PUBLIC_READ = 'public-read';
    const ACL_PUBLIC_READ_WRITE = 'public-read-write';
    const ACL_AUTHENTICATED_READ = 'authenticated-read';

    const STORAGE_CLASS_STANDARD = 'STANDARD';
    const STORAGE_CLASS_RRS = 'REDUCED_REDUNDANCY';
    const STORAGE_CLASS_STANDARD_IA = 'STANDARD_IA';

    const SSE_NONE = '';
[...]

PR #2103, opened 2020-09-17
suppress file warning

SamRemis:
*Issue #, if available:*
#2097

*Description of changes:*
Upon trying to make a file_get_contents call, the following PHP error was thrown:
`failed to open stream: Permission denied `

This fix attempts to correct that by making a is_readable call first


src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -94,12 +94,13 @@ public function __invoke()
             $result = null;
             while ($result == null) {
                 try {
-                    $token = file_get_contents($this->tokenFile);
-                    if (false === $token) {
+                    if (is_readable($this->tokenFile)) {
```
howardlopez, line 97: Is the main purpose to throw a `CredentialsException` instead of having the file warning? If so, if the `is_readable` call fails, won't it reach the same warning within the `else` block?

src/.main/main.php
```diff
@@ -0,0 +1,87 @@
+<?php
```
howardlopez, line 1: Looks like this file got committed by mistake?

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -94,12 +94,13 @@ public function __invoke()
...
+                        $token = file_get_contents($this->tokenFile);
+                    } else {
                         clearstatcache(true, dirname($this->tokenFile) . "/" . readlink($this->tokenFile));
                         clearstatcache(true, dirname($this->tokenFile) . "/" . dirname(readlink($this->tokenFile)));
                         clearstatcache(true, $this->tokenFile);
-                        $token = file_get_contents($this->tokenFile);   
+                        $token = file_get_contents($this->tokenFile);
                     }
                 } catch (\Exception $exception) {
```
howardlopez, line 105: Can you add a test case for the logic?

howardlopez (changes requested):

src/.main/main.php
SamRemis: yep, sorry about that.  I'll get rid of it

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
SamRemis: The logic is a bit different now from when you wrote this; a test case was still added to ensure that the error was not thrown

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -94,12 +94,19 @@ public function __invoke()
...
-                    $token = file_get_contents($this->tokenFile);
+                    $token = is_readable($this->tokenFile)
+                        ? file_get_contents($this->tokenFile)
+                        : false;
                     if (false === $token) {
                         clearstatcache(true, dirname($this->tokenFile) . "/" . readlink($this->tokenFile));
                         clearstatcache(true, dirname($this->tokenFile) . "/" . dirname(readlink($this->tokenFile)));
                         clearstatcache(true, $this->tokenFile);
-                        $token = file_get_contents($this->tokenFile);   
+                        if (!is_readable($this->tokenFile)){
```
howardlopez, line 104: Style nitpick: space before bracket

howardlopez (commented):
Looks good, just had a style nitpick

howardlopez (approved):

PR #2104, opened 2020-09-21
added circular source_profile check

SamRemis:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ clover.xml
 /.project
 /.buildpath
 /.settings
+src/.main/Main.php
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ clover.xml
 /.project
 /.buildpath
 /.settings
-src/.main/Main.php
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -817,7 +817,7 @@ class CredentialProvider
         $data = self::loadProfiles($filename);
         $credentialSource = !empty($data[$profileName]['credential_source']) ? $data[$profileName]['credential_source'] : null;
         $credentialsPromise = null;
-        
+
         switch ($credentialSource) {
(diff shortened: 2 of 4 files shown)


src/Credentials/CredentialProvider.php
```diff
@@ -801,6 +811,13 @@ public static function getCredentialsFromSource(
         $credentialSource = !empty($data[$profileName]['credential_source']) ? $data[$profileName]['credential_source'] : null;
         $credentialsPromise = null;
 
+        if ($credentialSource == null && empty($data[$profileName]['role_arn'])) {
```
howardlopez, line 814: Is this block necessary? If `$credentialSource` is null, it will be caught in the switch block below, and not sure `$data[$profileName]['role_arn']` is ever used by anything past this point?

SamRemis: I agree but it was in the SEP: 
"If a role arn is not provided along with credential_source then the SDK must fail / return an error."

As far as placement goes, it has to be in this function since it's public.  Through the default chain, it will never enter this section without a role_arn

SamRemis: I did rewrite it to invert the first condition

.gitignore
```diff
@@ -24,3 +24,4 @@ clover.xml
 /.project
 /.buildpath
 /.settings
+src/.main/Main.php
```
howardlopez, line 27: This should probably not be checked in to the repo and instead remain for your local .gitignore only.

src/Credentials/CredentialProvider.php
howardlopez: It doesn't really make sense to be checked if it's not actually used by anything in this section. Is there another section for which this check and throw makes sense, where the `role_arn` is actually being used, that might be more what the SEP is referring to?

howardlopez (approved):

PR #2127, opened 2020-10-21
validate non empty bucket

SamRemis:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/InputValidationMiddleware.php
+++ b/src/InputValidationMiddleware.php
@@ -57,17 +57,15 @@ class InputValidationMiddleware
             if (!empty($input = $service['shapes'][$op['input']['shape']])) {
                 if (!empty($input['required'])) {
                     foreach ($input['required'] as $key => $member) {
-                        if (!in_array($member, $this->mandatoryAttributeList))
-                            continue;
-                        if (isset($cmd[$member])) {
+                        if (in_array($member, $this->mandatoryAttributeList)) {
                             $argument = is_string($cmd[$member]) ?  trim($cmd[$member]) : $cmd[$member];
-                            if ($argument !== '')
-                                continue;
+                            if ($argument === '' || $argument === null) {
+                                $commandName = $cmd->getName();
+                                throw new \InvalidArgumentException(
+                                    "The {$commandName} operation requires non-empty parameter: {$member}"
+                                );
+                            }
                         }
-                        $commandName = $cmd->getName();
-                        throw new \InvalidArgumentException(
-                        "The {$commandName} operation requires non-empty parameter: {$member}"
-                        );
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


src/InputValidationMiddleware.php
```diff
@@ -0,0 +1,59 @@
+<?php
+namespace Aws;
+
+use Aws\Api\Service;
+
+/**
+ * Validates the required input parameters of commands are non empty
```
howardlopez, line 7: After some discussion, an approach to limit the scope to specific members may be the better approach, in case new operations are released with required members that may not necessarily conform to this validation. The middleware should be able to take in an argument containing relevant members.

src/InputValidationMiddleware.php
```diff
@@ -0,0 +1,78 @@
...
+    public function __invoke(CommandInterface $cmd) {
+        $nextHandler = $this->nextHandler;
+        $op = $this->service->getOperation($cmd->getName())->toArray();
+        if (!empty($op['input']['shape'])) {
+            $service = $this->service->toArray();
+            if (!empty($input = $service['shapes'][$op['input']['shape']])) {
+                if (!empty($input['required'])) {
+                    foreach ($input['required'] as $key => $member) {
+                        if (!in_array($member, $this->mandatoryAttributeList))
+                            continue;
```
howardlopez, line 61: For this pattern, should the throwing of the exception be inside the if blocks, and if it doesn't need to throw the exception, it just passes through to the next element without needing the continues?

tests/InputValidationMiddlewareTest.php
```diff
@@ -0,0 +1,189 @@
...
+            );
+        }
+    }
+
+    /**
+     * @dataProvider getInvalidEndpointExceptions
+     *
+     * @param $input
+     */
+    public function testNoValidationWithoutInputLIst($input)
```
howardlopez, line 88: Typo: testNoValidationWithoutInputLIst -> testNoValidationWithoutInputList

tests/InputValidationMiddlewareTest.php
```diff
@@ -0,0 +1,189 @@
...
+        $list->setHandler(function ($command) {
+            return;
+        });
+        $list->appendValidate(
+            InputValidationMiddleware::wrap($service, $mandatoryInputList)
+        );
+        $handler = $list->resolve();
+
+        $handler($command, new Request('POST', 'https://foo.com'));
+        self::assertTrue(true);
```
howardlopez, line 104: Why is this trivial assert here? Perhaps a slightly more meaningful assert would be capturing the result of the $handler() and checking that it's going through the stack as expected? You may have to modify the defined handler to return something useful.

howardlopez (changes requested):

howardlopez (approved):
Looks good pending Travis tests

PR #2135, opened 2020-11-02
Use aws shared credential file

### LINKED ISSUE #1518: ini provider doesn't respect AWS_SHARED_CREDENTIALS_FILE env variable
## Issue
When using the `ini` credential provider I realized that `AWS_SHARED_CREDENTIALS_FILE` is not used to try and resolve the shared file (like how it does in the awscli) if no filename was provided. Code below:
https://github.com/aws/aws-sdk-php/blob/c414752f2f88a036958438a3d973475c6786a387/src/Credentials/CredentialProvider.php#L283-L286

## Expected
I expected the `AWS_SHARED_CREDENTIALS_FILE` would be used to resolve the correct credential filename as it does for awscli. Example implementation below:

```php
const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';


public static function ini($profile = null, $filename = null)
{
    if (!$filename) {
        $filename = getenv(self::ENV_SHARED_CREDENTIALS_FILE) ?: (self::getHomeDir() . '/.aws/credentials');
    }
}
```

I think that the `localCredentialProvider()` function should also try and use `AWS_SHARED_CREDENTIALS_FILE` env variable for consistency.

## Possible issues
Technically this could be breaking if people have come to rely on the sdk ignoring this environment variable but it seems unlikely to me since it'd be opposite of the awscli. 

Curious what you guys think. Thanks!

SamRemis:
*Issue #, if available:*
fix #1518

*Description of changes:*

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -51,7 +51,7 @@ class CredentialProvider
     const ENV_SECRET = 'AWS_SECRET_ACCESS_KEY';
     const ENV_SESSION = 'AWS_SESSION_TOKEN';
     const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
-
+    const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
     /**
      * Create a default credential provider that first checks for environment
      * variables, then checks for the "default" profile in ~/.aws/credentials,
@@ -503,7 +503,7 @@ class CredentialProvider
      */
     public static function ini($profile = null, $filename = null, array $config = [])
     {
-        $filename = $filename ?: (self::getHomeDir() . '/.aws/credentials');
+        $filename = self::getFileName($filename);
         $profile = $profile ?: (getenv(self::ENV_PROFILE) ?: 'default');
 
         return function () use ($profile, $filename, $config) {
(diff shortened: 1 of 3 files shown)


tests/Credentials/CredentialProviderTest.php
```diff
@@ -369,6 +370,68 @@ public function testCreatesFromProcessCredentialWithFilename()
...
+
+        $creds = call_user_func(
+            CredentialProvider::process('baz', $dir . '/mycreds')
+        )->wait();
+        unlink($dir . '/mycreds');
+        $this->assertEquals('foo', $creds->getAccessKeyId());
+        $this->assertEquals('bar', $creds->getSecretKey());
+    }
+
+    public function testCreatesFromProcessCredentialWithSharedFilename()
```
howardlopez, line 393: You've covered calling the providers directly, but should there also be a test that the default provider ends up using the new ENV variable as well?

src/Credentials/CredentialProvider.php
```diff
@@ -51,7 +51,7 @@ class CredentialProvider
     const ENV_SECRET = 'AWS_SECRET_ACCESS_KEY';
     const ENV_SESSION = 'AWS_SESSION_TOKEN';
     const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
-
+    const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
     /**
```
howardlopez, line 55: Nitpick: there should be a line break between const declarations and this doc block.

howardlopez (changes requested):

howardlopez (approved):

mullermp (approved):

Issue #2143, opened 2020-11-16
Test error handling of S3Client uploadAsync method


matjazmav:
Confirm by changing [ ] to [x] below:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers

**Version of AWS SDK for PHP?**
`3.80.3`

**Version of PHP (`php -v`)?**
```
PHP 7.4.11 (cli) (built: Oct  8 2020 17:31:46) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.11, Copyright (c), by Zend Technologies
```

**Describe the question**

Method `uploadAsync` automatically decides and uploads object in single or multiple chunks. And we can expect 2 different kind of expectations `S3Exception` and `S3MultipartUploadException` in our `$result['reason']`. I'm trying to mock part of the S3 client to throw the exception, I have done this by MockHandler:

```
$s3->getHandlerList()->setHandler(new MockHandler([$result], null, null)); 
```

And the `$result` is:
```
new S3Exception("", new Command("mockCommand"), [
                    'code' => 'mockCode',
                    'response' => new Response(401)
                ])
```

Or either this:
```
new S3MultipartUploadException(new UploadState(['testid']))
```

Since the `S3MultipartUploadException` is not instance of `AwsException` I got this exception:
`InvalidArgumentException: Expected an Aws\ResultInterface or Aws\Exception\AwsException.`

How can I handle such scenario?

ajredniwja:
Hey @matjazmav thanks for opening this issue, created a PR for this which should fix this issue.

matjazmav:
Great news, thank you :)

PR #2150, opened 2020-11-25
MockHandler exceptions fix

ajredniwja:
*Issue #2143, 

*Description of changes:*

Code changes to mock exceptions, for example `S3MultipartUploadException` which extends from Exception and not `AwsException`, using the previous code will throw an exception like: `InvalidArgumentException: Expected an Aws\ResultInterface or Aws\Exception\AwsException.`

### DIFF
--- a/.changes/nextrelease/mockhandlerExcpetionFix.json
+++ b/.changes/nextrelease/mockhandlerExcpetionFix.json
@@ -1,0 +1,5 @@
+{
+    "type": "enhancement",
+    "category": "",
+    "description": "Code changes to mock exceptions correctly in mockhandler"
+  }
--- a/.changes/nextrelease/mockhandlerExcpetionFix.json
+++ b/.changes/nextrelease/mockhandlerExcpetionFix.json
@@ -1,5 +1,7 @@
-{
+[
+  {
     "type": "enhancement",
     "category": "",
     "description": "Code changes to mock exceptions correctly in mockhandler"
   }
+]
--- a/src/MockHandler.php
+++ b/src/MockHandler.php
@@ -5,6 +5,7 @@ use Aws\Exception\AwsException;
 use GuzzleHttp\Promise;
 use GuzzleHttp\Promise\RejectedPromise;
 use Psr\Http\Message\RequestInterface;
+use Exception;
 
 /**
  * Returns promises that are rejected or fulfilled using a queue of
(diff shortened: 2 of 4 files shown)


src/MockHandler.php
```diff
@@ -48,12 +48,12 @@ public function append()
...
         foreach (func_get_args() as $value) {
             if ($value instanceof ResultInterface
-                || $value instanceof AwsException
+                || $value instanceof Exception
                 || is_callable($value)
             ) {
                 $this->queue[] = $value;
             } else {
-                throw new \InvalidArgumentException('Expected an Aws\ResultInterface or Aws\Exception\AwsException.');
+                throw new \InvalidArgumentException('Expected an Aws\ResultInterface or Aws\Exception.');
```
SamRemis, line 56: This should be the PHP core exception, not aws\exception (we don't have a class named exception).  It will also need the following use statement added at the top: 
use Exception;

tests/MockHandlerTest.php
```diff
@@ -18,7 +18,7 @@ class MockHandlerTest extends TestCase
 {
     /**
      * @expectedException \InvalidArgumentException
-     * @expectedExceptionMessage Expected an Aws\ResultInterface or Aws\Exception\AwsException
+     * @expectedExceptionMessage Expected an Aws\ResultInterface or Aws\Exception
```
SamRemis, line 21: Change this accordingly with the last comment

src/MockHandler.php
```diff
@@ -48,12 +48,12 @@ public function append()
     {
         foreach (func_get_args() as $value) {
```
SamRemis, line 49: please add a changelog too (nothing to do with this line, just needed to comment somewhere)

.changes/nextrelease/mockhandlerExcpetionFix.json
```diff
@@ -0,0 +1,5 @@
+{
```
SamRemis, line 1: The whole thing should be in square brackets, that's why this is failing travis- the formatting is very finicky because it is used to make our versioning files

.changes/nextrelease/mockhandlerExcpetionFix.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "",
+    "description": "Code changes to mock exceptions correctly in mockhandler"
+  }
+]
```
SamRemis, line 7: Not gonna block approval of this, but worth noting: see the circle with the red line at the end of this file?  That's because PSR-2 guidelines recommending files ending with a blank line (https://www.php-fig.org/psr/psr-2/#22-files).  This file is going to get automatically deleted and the contents moved to a changelog file anyway though, so being nit-picky about this one isn't necessary

SamRemis (approved):

SamRemis:
@ajredniwja feel free to merge whenever :)

Issue #2154, opened 2020-11-27
Env vars not accessible via getenv()


mheki:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Hello,
we have an issue with `EcsCredentialProvider` getting env variables by a non-thread save `getenv()`.
The same code works fine when reached by a browser, but doesn't work when triggered by a cron.
Our use case was to push CSV file to S3 bucket in another AWS account.
The error from a cronjob was 
```
Aws\Exception\CredentialsException Error retrieving credentials from the instance profile metadata service. (cURL error 7:  (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://169.254.169.254/latest/meta-data/iam/security-credentials/
```
This meant `EcsCredentialProvider` wasn't used to provide credentials. When I investigated the issue I realised `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`  env var returned by `getenv()` was false, but was available in both `$_SERVER` and `$_ENV` variables.
I'll create a PR with the code change which helped us fix the issue.

**Version of AWS SDK for PHP?**
3.164.0
**Version of PHP (`php -v`)?**
7.4.9

**To Reproduce (observed behavior)**
- Create an S3 bucket on `A` AWS account.
- Add relevant permissions to the `A` account bucket allowing adding files by `B` AWS role/account.
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::xxxxxxxxx:role/foo"
            },
            "Action": [
                "s3:PutObject",
                "s3:PutObjectAcl"
            ],
            "Resource": "arn:aws:s3:::test-bucket-upload/*",
            "Condition": {
                "StringEquals": {
                    "s3:x-amz-acl": "bucket-owner-full-control"
                }
            }
        }
    ]
}
```
- send CSV file from PHP code triggered by cron (I use FlySystem lib abstraction for that)

[...]

PR #2155, opened 2020-11-27
EcsCredentialProvider - Replace getenv() with $_SERVER

mheki:
*Issue #2154 

*Description of changes:*
Changes how env variable is taken for `EcsCredentialProvider` (non-thread safe `getenv()` to `$_SERVER`)
This fixed the issue we had on PHP script triggered by a cronjob on ECS task where `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` was available only in `$_SERVER` and `$_ENV` but not by `getenv()`.
This resulted with a different CredentialsProvider being used and failing with an error.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -105,7 +105,9 @@ class CredentialProvider
         $shouldUseEcsCredentialsProvider = getenv(EcsCredentialProvider::ENV_URI);
         // getenv() is not thread safe - fall back to $_SERVER
         if ($shouldUseEcsCredentialsProvider === false) {
-            $shouldUseEcsCredentialsProvider = isset($_SERVER[EcsCredentialProvider::ENV_URI]) ? $_SERVER[EcsCredentialProvider::ENV_URI] : false;
+            $shouldUseEcsCredentialsProvider = isset($_SERVER[EcsCredentialProvider::ENV_URI])
+                ? $_SERVER[EcsCredentialProvider::ENV_URI]
+                : false;
         }
 
         if (!empty($shouldUseEcsCredentialsProvider)) {
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -102,7 +102,13 @@ class CredentialProvider
             );
         }
 
-        if (!empty($_SERVER[EcsCredentialProvider::ENV_URI])) {
+        $shouldUseEcsCredentialsProvider = getenv(EcsCredentialProvider::ENV_URI);
+        // getenv() is not thread safe - fall back to $_SERVER
+        if ($shouldUseEcsCredentialsProvider === false) {
+            $shouldUseEcsCredentialsProvider = isset($_SERVER[EcsCredentialProvider::ENV_URI]) ? $_SERVER[EcsCredentialProvider::ENV_URI] : false;
+        }
+
+        if (!empty($shouldUseEcsCredentialsProvider)) {
             $defaultChain['ecs'] = self::ecsCredentials($config);
         }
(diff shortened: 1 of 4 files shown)


anamariacla (approved):

SamRemis:
Hi @mheki ,
Thank you so much for your pull request.  I am worried that if we merge, it may pose a breaking change for some customers who rely on the case-insensitivity of getenv().  I would suggest changing the code to try getenv and call $_SERVER in a catch block.
Thanks,
-Sam

mheki:
You're absolutely right @SamRemis 
I have changed proposed fix to take `getenv()` into account first.
That should not introduce BC breaks.
I tried to implement a new test for `EcsCredentialProvider` but regardless of passed lowercase or uppercase ECS_URI test results were passing. http://169.254.170.2 returned expected credentials with or without `/latest/credentials?id=7e9114eb-6b2f-426e-908a-7f0a318e1786`

src/Credentials/CredentialProvider.php
```diff
@@ -102,7 +102,13 @@ public static function defaultProvider(array $config = [])
             );
```
SamRemis, line 102: Please add a changelog file in the .changes/nextrelease folder as described at the bottom of [this page](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md)

src/Credentials/CredentialProvider.php
```diff
@@ -102,7 +102,13 @@ public static function defaultProvider(array $config = [])
...
         }
 
-        if (!empty(getenv(EcsCredentialProvider::ENV_URI))) {
+        $shouldUseEcsCredentialsProvider = getenv(EcsCredentialProvider::ENV_URI);
+        // getenv() is not thread safe - fall back to $_SERVER
+        if ($shouldUseEcsCredentialsProvider === false) {
+            $shouldUseEcsCredentialsProvider = isset($_SERVER[EcsCredentialProvider::ENV_URI]) ? $_SERVER[EcsCredentialProvider::ENV_URI] : false;
```
SamRemis, line 108: Depending on who is running this and on what platform $_ENV may be worth checking too- that can be a later addition though

SamRemis (changes requested):

src/Credentials/CredentialProvider.php
mheki: > In both the CGI and FastCGI SAPIs, $_SERVER is also populated by values from the environment;
https://www.php.net/manual/en/ini.core.php#ini.variables-order

I think using `$_SERVER` should be fine as a fallback. 
Is it possible that `getenv()` and `$_SERVER` return nothing but `$_ENV`contains values? 
The only example I can think of is  explicitly set with `$_ENV['foo'] = 'bar'`, but that's a rear edge case.

src/Credentials/CredentialProvider.php
mheki: done, thx

src/Credentials/CredentialProvider.php
SamRemis: agreed, and thank you for getting those fixes in so quickly as well as for your contribution :) I will merge soon

SamRemis:
Hi @mheki, 
I apologize, we had to revert the changes temporarily because they were causing some internal issues with tests running on EC2 instances.  I will work on redoing the changes tonight to get them to work with the tests on EC2 and will remerge as soon as I can get them running and approved.  The new PR will reference this one.
-Sam

SamRemis:
Just to update: I found out what was going on, the $_SERVER variables needed to be cleared out before certain tests ran on the credential provider's default chain.  Will work on a fix ASAP

SamRemis:
Redone in #2176

PR #2166, opened 2020-12-11
bugfix empty token file

### LINKED ISSUE #2164: AssumeRoleWithWebIdentityCredentialProvider: Missing or locked AWS_WEB_IDENTITY_TOKEN_FILE

**Describe the bug**
We get sporadically the following failure on assumeRoleWithWebIdentity():

    [WebIdentityToken] must be a string or an object that implements __toString(). Found bool(false) (0)

*Method: Aws\Credentials\AssumeRoleWithWebIdentityCredentialProvider->__invoke()*

```file_get_contents()``` return's ```bool(false)```.

It seems that the AWS_WEB_IDENTITY_TOKEN_FILE is locked or missing during loading of the file content. I don't know exactly what happens, when the token-file is updated with a new token but there seems to be a kind of race condition.

v3.166.0

PHP 7.3.25

I was able to reproduce it with an empty token, which leads to a similar problem:

First create an empty token file and set needed ENV's:

    touch /tmp/token.file
    export AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/token.file
    export AWS_ROLE_ARN=arn:aws:iam::123456789123:policy/AnyPolicy

Code:
```php
use Aws\Credentials\CredentialProvider;
use GuzzleHttp\Promise\Coroutine;

$provider = CredentialProvider::assumeRoleWithWebIdentityCredentialProvider([]);
/** @var Coroutine $coroutine */
$coroutine = $provider();
$coroutine->then(function () {
    echo 'success';
})->otherwise(function (\RuntimeException $e) {
    echo sprintf('%s: %s', $e->getCode(), $e->getMessage());
});
```
Failure:

    [WebIdentityToken] expected string length to be >= 4, but found string length of 1 (0)

**Expected behavior**
My guess is that the token file is sporadically locked or missing when it is in exchange.
I think this should lead to an AwsException with code "InvalidIdentityToken" to trigger the retry mechanism.

SamRemis:
*Issue #, if available:*
fix #2164
*Description of changes:*

### DIFF
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -108,6 +108,9 @@ class AssumeRoleWithWebIdentityCredentialProvider
                         }
                         $token = file_get_contents($this->tokenFile);
                     }
+                    if (empty($token)) {
+                        throw new CredentialsException("InvalidIdentityToken from file: {$this->tokenFile}");
+                    }
                 } catch (\Exception $exception) {
                     throw new CredentialsException(
                         "Error reading WebIdentityTokenFile from " . $this->tokenFile,
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -32,7 +32,10 @@ class AssumeRoleWithWebIdentityCredentialProvider
     private $retries;
 
     /** @var integer */
-    private $attempts;
+    private $authenticationAttempts;
+
+    /** @var integer */
+    private $tokenFileReadAttempts;
 
     /**
      * The constructor attempts to load config from environment variables.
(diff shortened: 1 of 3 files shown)


src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -108,6 +108,9 @@ public function __invoke()
                         }
                         $token = file_get_contents($this->tokenFile);
                     }
+                    if (empty($token)) {
+                        throw new CredentialsException("InvalidIdentityToken from file: {$this->tokenFile}");
+                    }
```
tobiashaase, line 113: Hello @SamRemis,
thank you very much for your quick response.

I think this change just leads to another failure message?
In addition this should trigger a retry in my opinion.

I would prefer something like this at line 129 before calling assumeRoleWithWebIdentity():

```php
if (empty($assumeParams['WebIdentityToken'])) {
    throw new AwsException(
        "Empty InvalidIdentityToken",
        null,
        [
            'code' => 'InvalidIdentityToken',
        ]
    );
}
```

Best regards
Tobias

SamRemis: Hi @tobiashaase,

I definitely agree that is the proper behavior; the reason I implemented it this way was just to get a quick fix in while I can do a bit more investigating.  I will need to look into how other SDKs are handling retries to make sure our behavior lines up and is appropriate.  I hesitate to copy the logic from below since the retries are meant for actual guzzle calls made by the client, so it's difficult to say how retries on reading the file should behave.  If there isn't a standard already established, I can make one :)

-Sam

tobiashaase: Ok - I understand. Thank very you.

ajredniwja (approved):

PR #2172, opened 2020-12-14
Resolves #2171: Honor 'use_aws_shared_config_files' for process…

### LINKED ISSUE #2171: Process Provider does not honor 'use_aws_shared_config_files' boolean similar to ini provider
It appears there is a conditional against the iniProvider here:

https://github.com/aws/aws-sdk-php/blob/5b59db8d5d3d5ce1e93b8c4de5908437cc2eb40f/src/Credentials/CredentialProvider.php#L89-L103

And it is not honored in the same way when evaluating for processes:

https://github.com/aws/aws-sdk-php/blob/5b59db8d5d3d5ce1e93b8c4de5908437cc2eb40f/src/Credentials/CredentialProvider.php#L108-L112

It is not clear why the ecsProvider is nestled between the two.

I'll file a PR to add the conditional to both stanzas. I can amend it if somebody feels the process method should be adjacent to the iniProvider method.

cuppett:
…-based CredentialProvider

*Issue #, if available:*

Resolves #2171 

For the credential resolver whereby the default ~/.aws/config and ~/.aws/credentials are evaluated for access credentials, the config variable 'use_aws_shared_config_files' is considered. However, for the process-based credential resolver, checking the same $HOME directory shared config files, it is not. This adds the same conditional check to ensure those files are not considered for either method where configuration is explicitly supplied indicating they should not be.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -56,6 +56,7 @@ class CredentialProvider
     /**
      * Create a default credential provider that
      * first checks for environment variables,
+     * then checks for assumed role via web identity,
      * then checks for cached SSO credentials from the CLI,
      * then check for credential_process in the "default" profile in ~/.aws/credentials,
      * then checks for the "default" profile in ~/.aws/credentials,
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -109,9 +109,9 @@ class CredentialProvider
 
         if (!empty(getenv(EcsCredentialProvider::ENV_URI))) {
             $defaultChain['ecs'] = self::ecsCredentials($config);
+        } else {
+            $defaultChain['instance'] = self::instanceProfile($config);
         }
-
-        $defaultChain['instance'] = self::instanceProfile($config);
 
         if (isset($config['credentials'])
             && $config['credentials'] instanceof CacheInterface
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -76,9 +76,9 @@ class CredentialProvider
         $cacheable = [
             'web_identity',
             'sso',
-            'ecs',
             'process_credentials',
             'process_config',
+            'ecs',
             'instance'
         ];
 
@@ -100,23 +100,17 @@ class CredentialProvider
                 'profile default',
                 self::getHomeDir() . '/.aws/config'
             );
-        }
-
-        if (!empty(getenv(EcsCredentialProvider::ENV_URI))) {
-            $defaultChain['ecs'] = self::ecsCredentials($config);
-        }
-        
-        if (
-            !isset($config['use_aws_shared_config_files'])
-            || $config['use_aws_shared_config_files'] !== false
-        ) {
             $defaultChain['process_credentials'] = self::process();
             $defaultChain['process_config'] = self::process(
                 'profile default',
                 self::getHomeDir() . '/.aws/config'
             );
         }
-        
+
+        if (!empty(getenv(EcsCredentialProvider::ENV_URI))) {
+            $defaultChain['ecs'] = self::ecsCredentials($config);
+        }
+
         $defaultChain['instance'] = self::instanceProfile($config);
 
         if (isset($config['credentials'])
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -54,13 +54,14 @@ class CredentialProvider
     const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
 
     /**
-     * Create a default credential provider that first checks for
-     * environment variables,
-     * then checks for the "default" profile in ~/.aws/credentials,
+     * Create a default credential provider that
+     * first checks for environment variables,
+     * then check for role-based auth in ~/.aws/config,
      * then check for credential_process in the "default" profile in ~/.aws/credentials,
+     * then checks for the "default" profile in ~/.aws/credentials,
+     * then for credential_process in the "default profile" profile in ~/.aws/config,
      * then checks for "profile default" profile in ~/.aws/config (which is
      * the default profile of AWS CLI),
-     * then for credential_process in the "default profile" profile in ~/.aws/config,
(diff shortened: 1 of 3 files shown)


cuppett:
I've pushed a couple more commits to order this according to the documentation:

[defaultProvider provider](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_provider.html#defaultprovider-provider)

1. environment variables
  a) Original environment variables
  b) Web Identity environment variables
2. AWS config/credentials files
  a) SSO/role based
  b) process based
  c) static credentials
3) ECS/EC2 roles
  a) ECS if environment variable present
  b) instance role

The ordering in the existing file is blended a bit. I've made the comment block and the code match the documentation. In addition, this ordering now matches the Java SDK for comparison

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.java#L46-L52

The profile types are all lumped together:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/ProfilesConfigFile.java#L207-L216

And then, there's a toggle between ECS and instance profile "either/or" versus "both":

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/EC2ContainerCredentialsProviderWrapper.java#L58-L64

SamRemis:
@cuppett Thanks for your PR, I agree this is a better order than we had before- could you create a changelog file?  I also will close/re-open this to retrigger the github actions since this was pushed while we were in the process of migrating from Travis.

cuppett:
> @cuppett Thanks for your PR, I agree this is a better order than we had before- could you create a changelog file? I also will close/re-open this to retrigger the github actions since this was pushed while we were in the process of migrating from Travis.

Done. Thanks!

SamRemis (approved):

src/Credentials/CredentialProvider.php
```diff
@@ -54,14 +54,16 @@ class CredentialProvider
...
 
     /**
-     * Create a default credential provider that first checks for environment
-     * variables, then checks for the "default" profile in ~/.aws/credentials,
+     * Create a default credential provider that
+     * first checks for environment variables,
+     * then check for role-based auth in ~/.aws/config,
+     * then check for credential_process in the "default" profile in ~/.aws/credentials,
+     * then checks for the "default" profile in ~/.aws/credentials,
+     * then for credential_process in the "default profile" profile in ~/.aws/config,
```
SamRemis, line 62: I believe this is referring to the SSO sign on, which should be before the first credential_process

SamRemis (changes requested):
Looks good, just one of the docs is out of order if I'm reading it correctly.  Please let me know if I misunderstood

src/Credentials/CredentialProvider.php
cuppett: Yup, it does and it is first in there. All good.

cuppett:
> Looks good, just one of the docs is out of order if I'm reading it correctly. Please let me know if I misunderstood

Double checked and raw file shows them all in good order. The markup view with line breaks a little confusing how it interrupts some of the sentences. 

I've got each then, then, then on separate lines to let future adds and changes be easier to see and review; hope that helps!

SamRemis:
@cuppett I apologize for my confusion- I think I misread and the documentation is missing one step: the SSO provider.  I saw the note about the CLI's default profile and my mind jumped to SSO.

cuppett:
> @cuppett I apologize for my confusion- I think I misread and the documentation is missing one step: the SSO provider. I saw the note about the CLI's default profile and my mind jumped to SSO.

No worries, getting these right super important.

cuppett:
Does this PR need flipped back to approved?

src/Credentials/CredentialProvider.php
```diff
@@ -54,14 +54,16 @@ class CredentialProvider
     const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
 
     /**
-     * Create a default credential provider that first checks for environment
-     * variables, then checks for the "default" profile in ~/.aws/credentials,
+     * Create a default credential provider that
+     * first checks for environment variables,
+     * then check for role-based auth in ~/.aws/config,
+     * then check for credential_process in the "default" profile in ~/.aws/credentials,
```
SamRemis, line 60: Above this line (after role-based, before credential_process), please add one about the SSO- for example:

      * then checks for cached SSO credentials from the CLI

SamRemis (changes requested):
Please add the a line about the SSO- for example, add before line 60

src/Credentials/CredentialProvider.php
cuppett: Done. Actually, replaces existing line 59 to make the terminology line up. Thanks!

SamRemis: There are 9 entries in the default provider, and only 8 listed in the documentation- now the assume role one is missing

SamRemis: the default provider has the following entries in order after your changes:
1) env
2) web identity (assume role)
3) sso
4) credential_process from credential file
5) credentials from credentials ini file
6) credential_process from config file
7) credentials from config ini file
8) ECS
9) EC2

Your documentation lists them as:
1) checks for environment variables
2) checks for cached SSO credentials from the CLI
3) check for credential_process in the "default" profile in ~/.aws/credentials
4)  checks for the "default" profile in ~/.aws/credentials
5) then for credential_process in the "default profile" profile in ~/.aws/config
6) checks for "profile default" profile in ~/.aws/config (which is the default profile of AWS CLI)
7) tries to make a GET Request to fetch credentials if ECS environment variable is presented
8) finally checks for EC2 instance profile credentials.

cuppett: Oh, I see what you are saying. Yeah, the environment variable is both the original stuff, ACCESS_KEY and secret + the web identity env vars + mounted volumes. I can add that. Np.

cuppett: Done, great catch!

SamRemis:
Looks good!  Going to wait for all the github actions to pass and will merge ASAP.  Thank you for your contribution

SamRemis (approved):

PR #2184, opened 2021-01-13
feature copy file path

### LINKED ISSUE #2130: [BUG] [MultipartCopy] Can't copy file with「?」mark
`src/S3/MultipartCopy.php` → `fetchSourceMetadata()`

```
private function fetchSourceMetadata()
{
    if ($this->config['source_metadata'] instanceof ResultInterface) {
        return $this->config['source_metadata'];
    }

    list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
    $headParams = [
        'Bucket' => $bucket,
        'Key' => $key,
    ];
    if (strpos($key, '?')) {
        list($key, $query) = explode('?', $key, 2);
        $headParams['Key'] = $key;
        $query = Psr7\parse_query($query, false);
        if (isset($query['versionId'])) {
            $headParams['VersionId'] = $query['versionId'];
        }
    }
    return $this->client->headObject($headParams);
}
```

If `?` in `$key`, it's determined as query separator.
And `$key` is splitted to `$key` and `$query`.
But how can I copy file with source key is `path/ololo?imnotaquery.jpg` (Yes, it is possible to create file with ? mark in S3)

So characters after `?` can be both file name and query parameters

Why here `Key` is reassigned? Is there any reasons?
```
$headParams['Key'] = $key;
```

AWS working fine even if pass `headParams['Key']` with query string as is.

Looks like it will be better to remove this line.
In other case you need to separate `$this->source` and `$this->query`.

Here is my logs when I try to copy `/BUCKET/somepath/coc?py.mpg` (source) file
```
Error executing "HeadObject" on "https://BUCKET.s3.ap-northeast-2.amazonaws.com/somepath/coc"; AWS HTTP error: Client error: `HEAD https://BUCKET.s3.ap-northeast-2.amazonaws.com/somepath/coc` resulted in a `404 Not Found` response NotFound (client): 404 Not Found (Request-ID: D3B81271BF4D2891) - 
```

SamRemis:
*Issue #, if available:*
fix #2130 
*Description of changes:*
Adds the option to overload the source parameter string in multipart copy with an array with required keys 'source_key' and 'source_bucket' and optional 'source_version_id'.  This would be only be necessary if the source key contains a '?'.


ajredniwja (approved):

ajredniwja (approved):

src/S3/MultipartCopy.php
```diff
@@ -128,15 +129,17 @@ private function createPart($partNumber, $partsCount)
             $data[$k] = $v;
         }
 
-        list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
-        $data['CopySource'] = '/' . $bucket . '/' . implode(
```
kstich, line 132: This implode sequence is lost in the new `else`.

src/S3/MultipartCopy.php
```diff
@@ -128,15 +129,17 @@ private function createPart($partNumber, $partsCount)
...
-        );
+        if (is_array($this->source)) {
+            $key = str_replace('?', '%3F', $this->source['source_key']);
+            $data['CopySource'] = '/' . $this->source['source_bucket'] . '/' . $key;
+        } else {
+            list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
+            $data['CopySource'] = '/' . $bucket . '/' . $key;
+        }
         $data['PartNumber'] = $partNumber;
+        if (!empty($this->sourceVersionId)) {
```
kstich, line 140: Doesn't look like `sourceVersionId` can be set at all. I'd recommend removing this for now and putting it in to a separate PR so the changes can be reviewed in isolation.

src/S3/MultipartCopy.php
```diff
@@ -178,20 +181,45 @@ private function fetchSourceMetadata()
...
+        } else {
+            list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
+            $headParams = [
+                'Bucket' => $bucket,
+                'Key' => $key,
+            ];
+            if (strpos($this->source, '?')) {
+                list($key, $query) = explode('?', $key, 2);
+                $headParams['Key'] = $key;
+                $query = Psr7\parse_query($query, true);
```
kstich, line 200: The urlencode boolean changing to `true` here is potentially breaking.

src/S3/MultipartCopy.php
```diff
@@ -128,15 +129,17 @@ private function createPart($partNumber, $partsCount)
...
 
-        list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
-        $data['CopySource'] = '/' . $bucket . '/' . implode(
-            '/',
-            array_map(
-                'urlencode',
-                explode('/', rawurldecode($key))
-            )
-        );
+        if (is_array($this->source)) {
```
kstich, line 132: These if/else pieces need comments to explain what behavior change is happening.

kstich (changes requested):
The PR description needs to call out what behavior is being supported by these changes and how. This is helpful when reviewing the approach taken to enable the changes.

src/S3/MultipartCopy.php
SamRemis: What do you mean?  sourceVersionId is set and used directly below, and is set on line 190- it's only needed on the headobject and used here

src/S3/MultipartCopy.php
kstich: It looks like this has been added back. However, I'm worried this might have slipped through - are there tests (better yet, integration tests) for this to update or could some be added to cover this class and it's edges?

src/S3/MultipartCopy.php
```diff
@@ -178,20 +187,45 @@ private function fetchSourceMetadata()
...
-            'Key' => $key,
-        ];
-        if (strpos($key, '?')) {
-            list($key, $query) = explode('?', $key, 2);
-            $headParams['Key'] = $key;
-            $query = Psr7\parse_query($query, false);
-            if (isset($query['versionId'])) {
-                $headParams['VersionId'] = $query['versionId'];
+        if (is_array($this->source)) {
+            $headParams['Key'] = $this->source['source_key'];
```
kstich, line 191: `$headParams` is undefined at the point when this key is set, which I think generates warnings in some versions.

```php
$headParams = [
    'Bucket' => $this->source['source_bucket'],
    'Key' => $this->source['source_key'],
];
```

src/S3/MultipartCopy.php
```diff
@@ -178,20 +187,45 @@ private function fetchSourceMetadata()
...
-            'Bucket' => $bucket,
-            'Key' => $key,
-        ];
-        if (strpos($key, '?')) {
-            list($key, $query) = explode('?', $key, 2);
-            $headParams['Key'] = $key;
-            $query = Psr7\parse_query($query, false);
-            if (isset($query['versionId'])) {
-                $headParams['VersionId'] = $query['versionId'];
+        if (is_array($this->source)) {
```
kstich, line 190: This if/else should probably also have a comment explaining the differences.

src/S3/MultipartCopy.php
```diff
@@ -178,20 +187,45 @@ private function fetchSourceMetadata()
...
+                    $this->sourceVersionId = $query['versionId'];
+                    $headParams['VersionId'] = $query['versionId'];
+                }
             }
         }
         return $this->client->headObject($headParams);
     }
+
+    /**
+     * @param $inputSource
```
kstich, line 217: These should be filled out.

src/S3/MultipartCopy.php
```diff
@@ -127,16 +128,24 @@ private function createPart($partNumber, $partsCount)
...
-            '/',
-            array_map(
-                'urlencode',
-                explode('/', rawurldecode($key))
-            )
-        );
+        // The source parameter here is usually a string, but can be overloaded as an array
+        // if the key contains a '?' character to specify where the query parameters start
+        if (is_array($this->source)) {
+            $key = str_replace('?', '%3F', $this->source['source_key']);
```
kstich, line 134: Does this replacement of the `?` need to be composed with the `rawurlencode` and `urlencode` calls?

src/S3/MultipartCopy.php
kstich: nit: Any reason this is defined over 3 statements still?

src/S3/MultipartCopy.php
```diff
@@ -61,12 +63,11 @@ public function __construct(
         $source,
         array $config = []
     ) {
-        if (ArnParser::isArn($source)) {
-            $this->source = '';
+        if (is_array($source)) {
```
kstich, line 66: The constructor documentation for `$source` needs to be updated to define the new behavior.

src/S3/MultipartCopy.php
SamRemis: Good call.  Now it's only one statement :)

src/S3/MultipartCopy.php
SamRemis: I fixed it to call rawurlencode instead

src/S3/MultipartCopy.php
SamRemis: Documentation added (pasted here so you don't have to go searching): 
```
     * @param string $source Location of the data to be copied (in the form
     *                       /<bucket>/<key>).  If the key contains a '?'
     *                       character, instead pass an array of source_key,
     *                       source_buckey, and source_version_id.
```

src/S3/MultipartCopy.php
```diff
@@ -127,16 +130,24 @@ private function createPart($partNumber, $partsCount)
...
+        // if the key contains a '?' character to specify where the query parameters start
+        if (is_array($this->source)) {
+            $key = str_replace('%2F', '/', rawurlencode($this->source['source_key']));
+            $data['CopySource'] = '/' . $this->source['source_bucket'] . '/' . $key;
+        } else {
+            list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
+            $data['CopySource'] = '/' . $bucket . '/' . rawurlencode(
+                    implode(
+                        '/',
+                        array_map('urlencode', explode('/', $key))
```
kstich, line 143: The old code for this was calling `rawurldecode` on the `$key`, is the difference intentional?

src/S3/MultipartCopy.php
```diff
@@ -52,21 +54,22 @@ class MultipartCopy extends AbstractUploadManager
      *   result of executing a HeadObject command on the copy source.
      *
      * @param S3ClientInterface $client Client used for the upload.
-     * @param string $source Location of the data to be copied
-     *                                  (in the form /<bucket>/<key>).
+     * @param string $source Location of the data to be copied (in the form
+     *                       /<bucket>/<key>).  If the key contains a '?'
+     *                       character, instead pass an array of source_key,
+     *                       source_buckey, and source_version_id.
```
kstich, line 60: "source_buckey" should be "source_bucket"

Shouldn't the type for this `@param` be updated to reflect `string|array`?

SamRemis: good catches :) thank you

src/S3/MultipartCopy.php
SamRemis: Just changed that code entirely.  The idea behind it is to turn go through the string encoding everything but slashes.  Since there's a much cleaner way to do that (which is also used elsewhere in the SDK), I moved around the logic for consistency and code cleanup.    

Tests are all still passing and all of the manual testing I did showed no change.  rawurldecode is now just called in the getInputSource method instead

src/S3/MultipartCopy.php
```diff
@@ -127,16 +130,20 @@ private function createPart($partNumber, $partsCount)
...
-                explode('/', rawurldecode($key))
-            )
-        );
+        // The source parameter here is usually a string, but can be overloaded as an array
+        // if the key contains a '?' character to specify where the query parameters start
+        if (is_array($this->source)) {
+            $key = str_replace('%2F', '/', rawurlencode($this->source['source_key']));
+            $data['CopySource'] = '/' . $this->source['source_bucket'] . '/' . $key;
+        } else {
+            list($bucket, $key) = explode('/', ltrim($this->source, '/'), 2);
```
kstich, line 139: Does this `else` block provide the exact same values as the previous code in all situations? I don't think the integration tests are validating that this is actually the same result, more that it doesn't break the process overall.

PR #2186, opened 2021-01-28
bugfix to update content type in lexmodelsv2

SamRemis:
*Issue #, if available:*

*Description of changes:*
Added handler to update the content type of the LexModelsV2 client to 'application/x-amz-json-1.1'

### DIFF
--- a/src/LexModelsV2/LexModelsV2Client.php
+++ b/src/LexModelsV2/LexModelsV2Client.php
@@ -99,7 +99,7 @@ class LexModelsV2Client extends AwsClient {
     }
 
     /**
-     * Creates a middleware that fixes the Content-Type header when it is present
+     * Creates a middleware that updates the Content-Type header when it is present
      *
      * @return callable
      */
--- a/src/LexModelsV2/LexModelsV2Client.php
+++ b/src/LexModelsV2/LexModelsV2Client.php
@@ -99,7 +99,10 @@ class LexModelsV2Client extends AwsClient {
     }
 
     /**
-     * Creates a middleware that updates the Content-Type header when it is present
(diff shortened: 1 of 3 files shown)


tests/LexModelsV2/LexModelsV2ClientTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $command = $client->getCommand('DeleteBot', [
+            'botId' => '0123456789'
+        ]);
+        $request = \Aws\serialize($command);
+
+        // Corrected the header in a post request
+        $contentType = $request->getHeader('Content-Type');
+        $this->assertEmpty($contentType);
+    }
+}
```
skotambkar, line 58: Have to done an integ test with the change?

 May or may not include the integ test in the SDK, but should test it.

src/LexModelsV2/LexModelsV2Client.php
```diff
@@ -86,4 +88,37 @@
...
+    {
+        parent::__construct($args);
+
+        // Setup middleware.
+        $stack = $this->getHandlerList();
+        $stack->appendBuild($this->updateContentType(), 'models.lex.v2.updateContentType');
+    }
+
+    /**
+     * Creates a middleware that updates the Content-Type header when it is present
```
alextwoods, line 102: May make sense to add a little more detail to justify why this is needed (just that service has a protocol of rest-json but the operations require it be set differently)

alextwoods (approved):

tests/LexModelsV2/LexModelsV2ClientTest.php
SamRemis: Yes, it did work

Issue #2189, opened 2021-02-03
Duplicated scheme "http" in Uri by the S3EndpointMiddleware


aldenw:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
When the endpoint option is set and path style endpoints are set to true, the S3EndpointMiddleware produces a Uri with both the scheme and the host in the host property of the Uri.  This causes issues when the Uri get's to Guzzle as Guzzle replaces the blank scheme with 'http', duplicating 'http' in the final Uri.

**Version of AWS SDK for PHP?**
v3.173.0

**Version of PHP (`php -v`)?**
na

**To Reproduce (observed behavior)**
We encountered this bug while using the sdk configured via the Symfony aws bundle, with the options for path style endpoints set to true and custom endpoints for localstack.

```yaml
aws:
  S3:
    use_path_style_endpoint: true
    endpoint: http://my-endpoint:123
```

**Expected behavior**
Here is a failing unit test showing what I thought the expected behavior should be.

```php
    public function testApplyingEndpointWhenEndpointOptionAndPathStyleAreSet()
    {
        $nextHandler = function ($command, Request $request){
            $uri = $request->getUri();
            $this->assertEquals($expectedHost = 'my-endpoint', $uri->getHost());
            $this->assertEquals($expectedScheme = 'http', $uri->getScheme());
        };
        $command = new Command('CreateBucket', ['Bucket' => 'abc']);
        $middleware = new S3EndpointMiddleware( $nextHandler, 'us-west-2', [
            'endpoint' => 'http://my-endpoint:123',
            'path_style' => true
        ]);
        $requestUri = new Uri('http://my-endpoint:123');
        $request = new Request('PUT',$requestUri);
        $middleware($command, $request);
    }
```

**Additional context**
[...]

PR #2191, opened 2021-02-03
Fix issue with s3 request uri generation

aweddleton:
When the endpoint option is set the way the Uri was being generated resulted in a Uri with a blank scheme, and the scheme included in the host value.

*Issue #, if available: 2189
https://github.com/aws/aws-sdk-php/issues/2189

*Description of changes: Changed to pass host to Uri via constructor rather than generating new Uri using the withHost method.

### DIFF
--- a/src/S3/BucketEndpointArnMiddleware.php
+++ b/src/S3/BucketEndpointArnMiddleware.php
@@ -13,6 +13,7 @@ use Aws\Endpoint\PartitionEndpointProvider;
 use Aws\Exception\InvalidRegionException;
 use Aws\Exception\UnresolvedEndpointException;
 use Aws\S3\Exception\S3Exception;
+use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
 
 /**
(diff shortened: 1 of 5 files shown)


src/S3/BucketEndpointArnMiddleware.php
```diff
@@ -13,6 +13,7 @@
 use Aws\Exception\InvalidRegionException;
 use Aws\Exception\UnresolvedEndpointException;
 use Aws\S3\Exception\S3Exception;
+use InvalidArgumentException;
```
aldenw, line 16: Added to resolve phpstan failure.

src/S3Control/S3ControlEndpointMiddleware.php
```diff
@@ -22,6 +22,8 @@ class S3ControlEndpointMiddleware
     private $region;
     /** @var callable */
     private $nextHandler;
+    /** @var string|null */
+    private $endpoint;
```
aldenw, line 26: Added to resolve phpstan failure.

src/S3/S3EndpointMiddleware.php
```diff
@@ -46,6 +46,8 @@ class S3EndpointMiddleware
     private $endpointProvider;
     /** @var callable */
     private $nextHandler;
+    /** @var string */
+    private $endpoint;
```
aldenw, line 50: Added to resolve phpstan failure.

PR #2196, opened 2021-02-17
PHPDoc update on signature

maxhelias:
Updated PHPDoc about the signatures.
The change concerns more particularly the "presign" method which can take a larger object than a simple DateTime.

Edit : I am using a DateTimeImmutable and my PHPStan detect an error on this


SamRemis:
@maxhelias  Thank you for catching that and contributing, you're right this accepts any DateTimeInterface.  I will merge shortly

Issue #1959, opened 2020-01-24
$client->uploadDirectory() always uploads every file, even when no changes


curtiscarlson:
Please fill out the sections below to help us address your issue.

### Version of AWS SDK for PHP?
3.133.5

### Version of PHP (`php -v`)?
PHP 7.2.21

### What issue did you see?
when running uploadDirectory(), it creates a new version of every file in the folder

### Steps to reproduce
If you have a runnable example, please include it as a snippet or link to a repository/gist for larger code examples.

No, just the basic function with a local source folder and a bucket destination, no other options or arguments

### Additional context
Any additional information relevant to the issue. Examples include any framework you may be using (e.g. Laravel, Wordpress) in conjunction with the AWS SDK for PHP, or PHP/environment config settings if the issue is related to memory or performance.

windows / xampp

using slim/pimple and env, my setup looks like this

```
<?php

namespace App\Providers;

use Pimple\Container;
use Pimple\ServiceProviderInterface;
use Aws\S3\S3Client;

class S3ServiceProvider implements ServiceProviderInterface
{
     public function register(Container $container) {

        // set up s3
        $container['s3client'] = function() {
            $client = new S3Client([
                'credentials' => [
                    'key'    => $_ENV['AWS_ACCESS_KEY_ID'],
                    'secret' => $_ENV['AWS_SECRET_ACCESS_KEY'],
                ],
                'region' => $_ENV['AWS_DEFAULT_REGION'],
                'version' => 'latest',
            ]);

            return $client;
        };

   }
}
```

Symfony console command: 
```
<?php

namespace App\Commands;

use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Console\Question\Question;
use curtiscarlson\phplibs\ServiceLocator;

Class BackupStorage extends Command {

  protected function configure() {
    $this->setName('storage:backup')
      ->setDescription('Syncs the /storage folder to S3')
      ->setHelp('Syncs the /storage folder to S3');
  }

  public function execute(InputInterface $input, OutputInterface $output) {


      $client = ServiceLocator::get('s3client');

      $client->uploadDirectory(ROOT . '/storage', 'mybucketname');

  }

}
```

curtiscarlson:
could it be a permissions issue?

curtiscarlson:
[...]

PR #2224, opened 2021-04-02
Add "overwrite" option to Transfer

natitech:
Allow upload to be done once based on the file names

Issue #1959

Add "overwrite" option (bool, default=true) to Transfer. If disabled, the files will only be uploaded if the key is not found on the bucket (tested through HeadObject method). 

It is probably not the exact feature that is suggested in the issue #1959. It would be better if the decision to upload would be based on the ETag instead. But it will hopefully be already useful for people looking for a quick way to backup files on S3 with the PHP SDK (as I do).


SamRemis:
Hi @natitech,
Thank you for your PR- sorry about these github actions failures, they just started showing up a few days ago without any code changes.  I'm working on the fix now, and will restart your build after it's working again
-Sam

src/S3/Transfer.php
```diff
@@ -340,7 +346,37 @@ private function getDownloadsIterator()
         return $this->source;
     }
 
-    private function upload($filename)
+    private function uploadAsNeeded($file)
+    {
+        if (!$this->overwrite) {
+            return $this->client
+                ->executeAsync($this->hasBeenUploadedCommand($file))
```
SamRemis, line 353: There is already a doesObjectExist function built into the SDK as part of the S3ClientTrait which would be useful here to reduce code duplication.  This could be fewer lines of code by having it just a small addition to the upload function instead of it's own function:
if (!this->overwrite)  { 
    if ($this->client ->doesObjectExist($file)) {
        return
    }
}
//normal flow of logic continues

src/S3/Transfer.php
```diff
@@ -340,7 +346,37 @@ private function getDownloadsIterator()
...
+                ->otherwise(function(Aws\S3\Exception\S3Exception $e) use ($file) {
+                    if ($e->getAwsErrorCode() === 'AccessDenied') {
+                        // The file does exist
+                        return;
```
SamRemis, line 357: Should we be returning a value here to let a user know that the file already existed?

SamRemis (changes requested):
Added a few comments, but I like the idea of having this as a feature.  I do need to check with other SDKs and see if any of them have a similar option and if they do, then we would want consistent behavior.

src/S3/Transfer.php
natitech: It does indeed work better.

natitech:
> Added a few comments, but I like the idea of having this as a feature. I do need to check with other SDKs and see if any of them have a similar option and if they do, then we would want consistent behavior.

Hi @SamRemis, thank you for your comments. I had a quick look in the java SDK before implementing this solution and I wasn't able to find any similar option.

.changes/nextrelease/feature-upload-once.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "feature",
+    "category": "S3",
+    "description": "Add Transfer option to upload only non existing files"
```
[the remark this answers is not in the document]
natitech: Indeed, I'm not sure about the point to mix versioning and no overwrite. That said, I think this option should take precedence. As the customer needs to change the default overwrite option (from true to false), there won't be any unexpected behavior. If you agree, I can clarify the documentation (for instance, adding "(...) even when versionning enabled").

src/S3/Transfer.php
natitech: I initially thought about using the "doesObjectExist" function. I think it depends on how we want to achieve concurrency :
* one option is to first filter files that need to be uploaded and then create upload promises only for these files. In that case, using the doesObjectExist would be the best solution
* the other option is to test existence concurently : for each files, we create a promise that will test AND upload as needed. I think this is faster, that's why I prefered this implementation.

src/S3/Transfer.php
natitech: We could but what do you think should be returned ? 

That makes me think about a different possibility : adding an explicit message to the debug output. And to do this, I think I can do the same as what is done in the "upload" functions : call the before option callback but this time with the HeadObject command. And then change a little bit the debugging message (for instance changing "Transferring" by "Ignoring" when HeadObject). What do you think ?

.changes/nextrelease/feature-upload-once.json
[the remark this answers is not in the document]
SamRemis: I agree, and I don't think anyone would likely intentionally mix versioning and overwrite, I'd more see it as a potential point of confusion for customers who enable one, forget about it, and later enable the other.  So I wanted your thoughts on how to best clarify that; putting it in the documentation is probably the best option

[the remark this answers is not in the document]
natitech: I have updated the option documentation in the code. If this gets merged, it could be good to update the developers guide as well (https://github.com/awsdocs/aws-php-developers-guide/blob/master/doc_source/s3-transfer.rst).

natitech:
Hi there, any chance to get some feedback (https://github.com/aws/aws-sdk-php/pull/2224#discussion_r610918308 and https://github.com/aws/aws-sdk-php/pull/2224#discussion_r610918565) ?

PR #2235, opened 2021-04-21
trim time to micro-seconds when the time get from server shows in nano-seconds.

keijack:
Some of the third party implementation like Ceph will return timestamp in nano-seconds with the ISO8061 timestamp format like `2021-04-20T14:47:30.234566239Z`, which is not supported in PHP (7.4.2 in my version), so this branch add code to trim the timestamp to micro-seconds like `2021-04-20T14:47:30.234566Z` to avoid this exception.

### DIFF
--- a/src/Api/DateTimeResult.php
+++ b/src/Api/DateTimeResult.php
@@ -9,8 +9,8 @@ use Exception;
  * DateTime overrides that make DateTime work more seamlessly as a string,
  * with JSON documents, and with JMESPath.
  */
-class DateTimeResult extends \DateTime implements \JsonSerializable
-{
+class DateTimeResult extends \DateTime implements \JsonSerializable {
+
     /**
      * Create a new DateTimeResult from a unix timestamp.
      * The Unix epoch (or Unix time or POSIX time or Unix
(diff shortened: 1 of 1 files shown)


src/Api/DateTimeResult.php
```diff
@@ -31,12 +30,39 @@ public static function fromEpoch($unixTimestamp)
...
+     * 
+     * @return string
+     */
+    private static function toPrecisionMs($iso8601Timestamp) {
+        $timeIso8601 = strtoupper($iso8601Timestamp);
+        $startPos = strpos($timeIso8601, '.');
+        if ($startPos != false) {
+            $startPos += 1;
+            $endPos = strpos($timeIso8601, '+');
+            if ($endPos == false) {
```
SamRemis, line 53: This part could be fewer lines:
if ($endPos = strpos($timeIso8601, '+') == false) {
  if ($endPos = strpos($timeIso8601, "Z") == false) {
  $endPos = strlen($timeIso8601);
  }
}

src/Api/DateTimeResult.php
```diff
@@ -31,12 +30,39 @@ public static function fromEpoch($unixTimestamp)
...
+                $endPos = strpos($timeIso8601, "Z");
+            }
+            if ($endPos == false) {
+                $endPos = strlen($timeIso8601);
+            }
+            $ns = substr($timeIso8601, $startPos, $endPos - $startPos);
```
SamRemis, line 59: This should only happen if endpos > startpos.  That would only happen if it was malformed, but still worth a check

SamRemis (changes requested):
This also needs a changelog file

PR #2236, opened 2021-04-21
fix: don't attempt flushHttpDebug when disabled

boite:
When TraceMiddleware is configured with http=false, createHttpDebug does not create a debug resource.  This fix ensures that flushHttpDebug does not try to use the non-existent debug resource.

Fixes:

```
Notice: Undefined index: debug

  at vendor/aws/aws-sdk-php/src/TraceMiddleware.php:288
  at Aws\TraceMiddleware->flushHttpDebug(object(Command))
     (vendor/aws/aws-sdk-php/src/TraceMiddleware.php:111)
```

### DIFF
--- a/src/TraceMiddleware.php
+++ b/src/TraceMiddleware.php
@@ -285,7 +285,7 @@ class TraceMiddleware
 
     private function flushHttpDebug(CommandInterface $command)
     {
-        if ($res = $command['@http']['debug']) {
+        if ($this->config['http'] && $res = $command['@http']['debug']) {
             rewind($res);
             $this->write(stream_get_contents($res));
             fclose($res);


amaitu (approved):

src/TraceMiddleware.php
```diff
@@ -285,7 +285,7 @@ private function createHttpDebug(CommandInterface $command)
 
     private function flushHttpDebug(CommandInterface $command)
     {
-        if ($res = $command['@http']['debug']) {
+        if ($this->config['http'] && $res = $command['@http']['debug']) {
```
SamRemis, line 288: In case the underlying logic changes later, wouldn't it be safer to change it to if (isset($command['@http']['debug']) && $res = $command['@http']['debug'])?

SamRemis (changes requested):
Please add a changelog file as well :)

stobrien89:
Resolved by https://github.com/aws/aws-sdk-php/pull/2435

PR #2245, opened 2021-05-11
Suppress error for potential empty token

DimitrijeManic:
*Description of changes:*
An empty token is expected and handled so there is no need to log the warning errors the we get from `file_get_contents`

Currently, every time the token is refreshed the following gets logged
```
ERROR - 2021-05-11 14:48:06 --> {"description":"Severity: Warning --> file_get_contents(\/var\/run\/secrets\/eks.amazonaws.com\/serviceaccount\/token): failed to open stream: No such file or directory \/var\/www\/test\/vendor\/aws\/aws-sdk-php\/src\/Credentials\/AssumeRoleWithWebIdentityCredentialProvider.php 102"}
```

### DIFF
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -99,7 +99,7 @@ class AssumeRoleWithWebIdentityCredentialProvider
             while ($result == null) {
                 try {
                     $token = is_readable($this->tokenFile)
-                        ? file_get_contents($this->tokenFile)
+                        ? @file_get_contents($this->tokenFile)
                         : false;
                     if (false === $token) {
                         clearstatcache(true, dirname($this->tokenFile) . "/" . readlink($this->tokenFile));
@@ -110,7 +110,7 @@ class AssumeRoleWithWebIdentityCredentialProvider
                                 "Unreadable tokenfile at location {$this->tokenFile}"
                             );
                         }
-                        $token = file_get_contents($this->tokenFile);
+                        $token = @file_get_contents($this->tokenFile);
                     }
                     if (empty($token)) {
                         if ($this->tokenFileReadAttempts < $this->retries) {


polesen:
:+1: 

I actually want/need this too. And to us, it is not only a warning being logged. We are running with the laravel framework, and laravel sets its own PHP error_handler, which raises errors as exceptions. And when this happens, the AWS SDK actually never gets to the lines of clearstatcache calls.

I know this is laravel specific, but the first call to `file_get_contents` actually don't need the error, so it should have an `@` in front, I agree.

I am not sure the second call after clearstatcache calls should have it though.

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -99,7 +99,7 @@ public function __invoke()
             while ($result == null) {
                 try {
                     $token = is_readable($this->tokenFile)
-                        ? file_get_contents($this->tokenFile)
+                        ? @file_get_contents($this->tokenFile)
```
polesen, line 102: :+1:  I agree that this calls should have a `@` in front, because it really doesn't need the error if it fails, and it does an explicit check itself afterwards.
Also see: https://github.com/aws/aws-sdk-php/pull/2245#issuecomment-875464278

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -110,7 +110,7 @@ public function __invoke()
                                 "Unreadable tokenfile at location {$this->tokenFile}"
                             );
                         }
-                        $token = file_get_contents($this->tokenFile);
+                        $token = @file_get_contents($this->tokenFile);
```
polesen, line 113: I don't think it as such hurts to have an `@` here also, but I think it is not needed. When we get to this place, it would actually *be* and error, if there still is no token file, after the statcache has been cleared.
Unless of course you rely on fallbacks to other kind of auth, if this one fails. Then there could/should be an `@` in front also here.

polesen (approved):

jycr753 (approved):
Good change, but I agree @pelo

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
DimitrijeManic: @polesen My thinking here was because we are retrying with a backoff we should silence these and allow the timeout to throw and log if we aren't able to read the credentials.

ie. We are expecting an empty token token and have a way of handling it

DimitrijeManic:
@polesen 
@jycr753 
Looks like you've both approved, could we get this merged?

polesen:
@DimitrijeManic I would love to see it merged, but I am just a client, so I at least can't do it :D ... I know my employer has contacted AWS support to get them to look at it, but nothing seems to have happended.

SamRemis:
@polesen, I haven't heard from AWS support yet, sorry to hear about that.   Contacting me or @ajredniwja directly on here will probably be faster.  If I don't respond, feel free to ping me again.  I'm not sure what the AWS stance on suppressing warnings is; it may be an issue if anyone relies on it, which is my biggest hesitation on merging this.  The other minor thing is that it's missing a changelog file.  Instructions can be found [here](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md#code-contributions) for how to make one. 

In the meantime, I'll do some research on how to suppress error warnings

Issue #2251, opened 2021-05-30
Uploading to S3 with BucketKeyEnabled throws an error


humancopy:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Trying to upload to `S3` with `"BucketKeyEnabled" => true`  throws an error.

**Version of AWS SDK for PHP?**
v3.183.7

**Version of PHP (`php -v`)?**
v8.0.2

**To Reproduce (observed behavior)**
```php
$uploader = new Aws\S3\ObjectUploader(
    $client,
    $bucket,
    $key,
    $source,
    'private',
    [
        'params' => [
            'ServerSideEncryption' => 'aws:kms',
            'BucketKeyEnabled'     => true,
        ]
    ]
);
```

**Expected behavior**
The file would get uploaded and the `BucketKeyEnabled` should be enabled on it.

**Screenshots**
Throws an error:
```
An uncaught Exception was encountered
Type: Aws\S3\Exception\S3Exception

Message: Error executing "PutObject" on "..."; AWS HTTP error: Client error: `PUT ...` resulted in a `400 Bad Request` response: InvalidArgumentValue should be either 'true' or 'fal (truncated...) InvalidArgument (client): Value should be either 'true' or 'false' - InvalidArgumentValue should be either 'true' or 'false'x-amz-server-side-encryption-bucket-key-enabled4K7BDPJMK4Z88YP9JBIz94Iy3WjUsNGJpE53SPl8JScPVJW6VzOQYy04oMymLEvWYxpnFT2zdgg+s2nTmujPb4OeWtY=

Filename: /var/www/vendor/aws/aws-sdk-php/src/WrappedHttpHandler.php

Line Number: 195

Backtrace:

/var/www/vendor/aws/aws-sdk-php/src/WrappedHttpHandler.php
Line: 97
Function: parseError

/var/www/vendor/guzzlehttp/promises/src/Promise.php
[...]

PR #2252, opened 2021-06-01
Boolean in header should be 'true' or 'false'

humancopy:
*Issue #2251*

*Description of changes:*

This pull request makes boolean headers sent as string (`true` or `false`) instead of integer (`1` or `0`)


SamRemis:
Thank you so much for the PR!  I'm hesitant to merge since the boolean value should be sent over the wire as a boolean, not a string.  This could have unintended consequences on other REST services.  I agree that there's an issue with the headers, but I think it's on the service side, not the SDK side since '0' and '1' are valid boolean values.

humancopy:
@SamRemis it's been a while :)

I don't think this will be fixed by the service as other libs and code rely already on it being like this so I can't see how they will make this change now. Would love to hear your opinion...

stobrien89:
Hi @humancopy,

Thanks for your patience.  After doing a bit of research, I think this is something we can accept.  I added some tests and will need to do a bit more review to see if we should change this for other protocols, but it's not looking like it at this time.  May need to add some more tests as well.  I'll let you know as soon as I have an update.

PR #2268, opened 2021-07-08
Test on PHP 8.0

driesvints:
This also allows the library to be built on PHP 8.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       #for each of the following versions of PHP, with and without --prefer-lowest
       matrix:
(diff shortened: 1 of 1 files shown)


driesvints:
I can't seem to start the tests to run unless I get approval because of GitHub's new policies. Therefor I have no idea if the test suite will pass or not. After approval has been given I can look into making the tests pass. 

I also can't run them locally because an incompatible PHPUnit version is loaded because of its very loose constraints: https://github.com/sebastianbergmann/phpunit/blob/4.8.36/composer.json#L23

driesvints:
Going to close this one as there's much work to be done to migrate the entire test suite to a newer PHPUnit version. I unfortunately don't have the time for this.

PR #2269, opened 2021-07-09
Add ReturnTypeWillChange attribute

driesvints:
This PR adds the new `ReturnTypeWillChange` attribute from PHP 8.1 to several internal method overrides so these are compatible with PHP 8.1. This is similarly to how we've done it in the Laravel framework: https://github.com/laravel/framework/pull/37838

At the moment the PHP 8.1 build for the Laravel framework is failing because AWS is not yet compatible with it (understandable because it's still a while out). By merging this PR, the AWS PHP SDK will be one step closer to PHP 8.1 compatibility and we can run our own build to prepare Laravel. Now with the recent PHP 8.1 Alpha 3 being out it's a good time to start testing: https://www.php.net/archive/2021.php#2021-07-08-1

See the failing build on `laravel/framework` here: https://github.com/laravel/framework/runs/3022524466#step:7:61

PS.: [I tried to add a PHP 8.0 build](https://github.com/aws/aws-sdk-php/pull/2268) (and was planning on a 8.1 after it) but there seems to be much more work involved as the current test suite has greatly aged. I suggest doing a new 4.0 major release and dropping a few older PHP versions and maintain parallel 3.x and 4.x branches so you can at least modernize the codebase and upgrade to a newer PHPUnit version.

### DIFF
--- a/src/Api/DateTimeResult.php
+++ b/src/Api/DateTimeResult.php
@@ -4,6 +4,7 @@ namespace Aws\Api;
 
 use Aws\Api\Parser\Exception\ParserException;
 use Exception;
+use \ReturnTypeWillChange;
 
 /**
  * DateTime overrides that make DateTime work more seamlessly as a string,
@@ -94,6 +95,7 @@ class DateTimeResult extends \DateTime implements \JsonSerializable
      *
      * @return mixed|string
      */
+    #[ReturnTypeWillChange]
     public function jsonSerialize()
     {
         return (string) $this;
--- a/src/Crypto/MetadataEnvelope.php
+++ b/src/Crypto/MetadataEnvelope.php
@@ -6,6 +6,7 @@ use \ArrayAccess;
 use \IteratorAggregate;
 use \InvalidArgumentException;
 use \JsonSerializable;
+use \ReturnTypeWillChange;
 
 /**
  * Stores encryption metadata for reading and writing.
@@ -49,6 +50,7 @@ class MetadataEnvelope implements ArrayAccess, IteratorAggregate, JsonSerializab
         $this->data[$name] = $value;
     }
 
+    #[ReturnTypeWillChange]
     public function jsonSerialize()
     {
         return $this->data;
--- a/src/DynamoDb/BinaryValue.php
+++ b/src/DynamoDb/BinaryValue.php
@@ -2,6 +2,7 @@
 namespace Aws\DynamoDb;
 
 use GuzzleHttp\Psr7;
+use \ReturnTypeWillChange;
 
 /**
  * Special object to represent a DynamoDB binary (B) value.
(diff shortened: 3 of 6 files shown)


driesvints:
@SamRemis can you maybe approve the workflow so I can spot any build failures? Then I can fix them before you have a change to review this.

SamRemis:
@driesvints,
Thanks for the follow up :) - I just did.  Sorry about the delay, I didn't see it still needed to be approved

driesvints:
Cool, thanks @SamRemis! No worries at all. Seems like all tests pass 👍

driesvints:
@SamRemis Heya, I definitely don't want to pressure you but is there any chance we can get this PR merged and tagged any time soon? Atm this is preventing us from going ahead with a passing PHP 8.1 build for Laravel.

SamRemis:
@driesvints No pressure at all!  I apologize for the delay- I was looking into this and got swept up in other things.  Looking at it again, it's definitely backwards compatible.  I'm not quite ready to get to work on supporting PHP 8, but am happy to merge this if it helps you guys.  I'll do my best to start work on getting it merged today.  

My only question is on the use statements; why are they necessary?  They generate a warning: `The use statement with non-compound name 'ReturnTypeWillChange' has no effect in...`.  I also don't see them used at all [here in this example](https://wiki.php.net/rfc/internal_method_return_types).  Is there something I'm missing?

driesvints:
Hey @SamRemis. Thanks a lot!

> My only question is on the use statements; why are they necessary? They generate a warning: 

Hmm, where do you see that? We didn't see any problems with that in Laravel when merging https://github.com/laravel/framework/pull/37838. I can easily undo that if need be.

SamRemis:
@driesvints I got that warning by using this SDK to make a very basic script to fetch an object from S3.  Excuse the bad cose because this was just a test file:
```
<?php
require '../../vendor/autoload.php';

use \ReturnTypeWillChange;

    'version' => 'latest',
    'region'  => 'us-east-2',
]);
$client->headBucket([
    'Bucket' => 'bucketName',
]);
```

When I ran the above code on PHP 7.3, that warning was shown in the output

driesvints:
@SamRemis that's because you're using it directly in that script. If you use any of the classes modified in this PR there shouldn't be an error normally.

SamRemis:
@driesvints You're right, tested it out that way and no warning came up :) 

Just need to test it down to 5.5, and no problems came.  Thanks for your patience, going to merge it soon

driesvints:
Thanks a bunch @SamRemis 🎉

driesvints:
Btw, just FYI: we ended up changing the syntax ourselves to `#[\ReturnTypeWillChange]` to avoid the use statement but purely out of semantic reasoning and ease of use to later on remove those again when the time is right.

PR #2295, opened 2021-08-17
Actually set the key formInput for PostObjectV4

QartemisT:
*Description of changes:* The variable syntax will not work inside of singular quote marks, so just direct set the variable.

### DIFF
--- a/src/S3/PostObjectV4.php
+++ b/src/S3/PostObjectV4.php
@@ -67,7 +67,7 @@ class PostObjectV4
         ];
 
         // setup basic formInputs
-        $this->formInputs = $formInputs + ['key' => '${filename}'];
+        $this->formInputs = $formInputs + ['key' => $filename];
 
         // finalize policy and signature
 


SamRemis:
$filename actually doesn't exist in this context either, so this line should probably be deleted if it's working properly.  I will look into this shortly

SamRemis:
Looks like this is a kind of weird quirk of S3, but the original code is correct.  In [these docs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html) in the "Form Fields" section, the 'key' part of the table explains why we are setting it to `'${filename}'`:

> To use the file name provided by the user, use the ${filename} variable. For example, if a user named Mary uploads the file example.jpg and you specify /user/mary/${filename}, the key name is /user/mary/example.jpg.

This is just telling the PostObjectV4 instance to use whatever filename is uploaded.  This used to be a configurable option, but it was removed in [this commit](https://github.com/aws/aws-sdk-php/commit/050b186cecfc1541b53e1d36d0367655ae1ef34d) in order to support more features.

Issue #2300, opened 2021-08-28
Allow guzzle/psr-7 2.0 to be installed to fully support Guzzle 7


antonioribeiro:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Composer is having trouble resolving package dependencies for Guzzle 7.0, as it's not allowing guzzlehttp/psr7 2.0 to be installed. This package is locked to [guzzlehttp/psr7:^1.7.0](https://github.com/aws/aws-sdk-php/blob/master/composer.json#L21) while Guzzle requirements are ["guzzlehttp/psr7": "^1.7 || ^2.0"](https://github.com/guzzle/guzzle/blob/master/composer.json#L34).

**Solution** 

- Remove `guzzlehttp/psr7:^1.7.0` from composer.json, as it will be required by Guzzle anyway; or
- Allow `guzzlehttp/psr7:^1.7.0 || ^2.0`

SamRemis:
closing as a duplicate of #2264

PR #2276, opened 2021-07-21
added support for document types

SamRemis:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2301, opened 2021-08-28
Allow guzzlehttp/psr7 2.0

antonioribeiro:
Issue https://github.com/aws/aws-sdk-php/issues/2300

# Changes
Allow guzzlehttp/psr7 2.0 to be installed if needed.

Closes https://github.com/aws/aws-sdk-php/issues/2300


SamRemis:
There's already another PR that does this same thing, but this is actually a breaking change and would cause errors in more than a few services and with some cryptography related features.  I'm currently working on this in PR #2276, but it's not quite ready for release yet

PR #2320, opened 2021-09-24
Respect the AWS_PROFILE env variable within the default credentials provider

### LINKED ISSUE #2278: Unable to use non-default profile for SSO credentials
Confirm by changing [x] to [x] below to ensure that it's a bug:
  
**Describe the bug**
I am unable to use PHP to work with AWS using a non-default SSO profile. We use AWS SSO (Single-Sign On) for authenticating to AWS. However, most people have multiple permissions sets (roles, if you will). They are stored in `~/.aws/config` like this:

```
[default]
output=json
region=us-east-1
sso_account_id=<account_number>
sso_region=<region>
sso_role_name=Developer
sso_start_url=<login_url>

[profile admin]
output=json
region=us-east-1
sso_account_id=<account_number>
sso_region=<region>
sso_role_name=Admin
sso_start_url=<login_url>
```

PHP works (for example, calling to S3) with the first set of credentials. However, if I set `export AWS_PROFILE=admin`, which works with AWS CLI (and other language SDKs), it no longer works with PHP.

Initially `3.158`, but tested against `3.186.1` with the same issue.

7.4.21

1. Set up a PHP project and install `aws-php-sdk` via composer.
2. Write PHP to use AWS (e.g. calling to S3).
3. Run `export AWS_PROFILE=myprofile`.
4. Run it via the same terminal session, and when it calls to get credentials, the following happens:
	- In `ClientResolver.php` (line 626), it detects the profile:
	- `$args['credentials'] = CredentialProvider::ini($args['profile']);`
    - In the `ini()` function, in `CredentialsProvider.php` (line 517), it looks to require either a `role_arn` or `aws_access_key` field (neither of which exist in SSO).
    - Finding neither, it fails with `No credentials present in INI profile 'admin' (/Users/<user>/.aws/credentials)`.

If you unset `AWS_PROFILE` and use your default profile, it detects SSO properly and sets up credentials as expected.

**Expected behavior**
When using other SDKs with AWS, I can set a profile (`export AWS_PROFILE=admin`) and it uses it automatically when authenticating to AWS. This should do the same (or at least provide an option or flag to use it).

**Screenshots**
Can provide if needed, but everything is related to the files (and lines) listed above.

**Additional context**
Worth noting (though I suspect it is ultimately unrelated), the error cites `.aws/credentials`, though all SSO credentials are in `.aws/config`. I am guessing it's just because it sees the profile and assume it is a normal profile, and thus isn't processing it as if it was an SSO credential.

cdloh:
Fixes #2278

*Description of changes:*

Rather then always using the default profile in the default credential provider instead respect the AWS_PROFILE env variable if set.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -85,6 +85,8 @@ class CredentialProvider
             'instance'
         ];
 
+        $profileName = getenv(self::ENV_PROFILE) ?: 'default';
+
         $defaultChain = [
             'env' => self::env(),
             'web_identity' => self::assumeRoleWithWebIdentityCredentialProvider($config),
@@ -94,18 +96,18 @@ class CredentialProvider
             || $config['use_aws_shared_config_files'] !== false
         ) {
             $defaultChain['sso'] = self::sso(
-                'profile default',
+                'profile '. $profileName,
                 self::getHomeDir() . '/.aws/config',
                 $config
             );
             $defaultChain['process_credentials'] = self::process();
             $defaultChain['ini'] = self::ini();
             $defaultChain['process_config'] = self::process(
-                'profile default',
+                'profile ' . $profileName,
                 self::getHomeDir() . '/.aws/config'
             );
             $defaultChain['ini_config'] = self::ini(
-                'profile default',
+                'profile '. $profileName,
                 self::getHomeDir() . '/.aws/config'
             );
         }


SamRemis:
I agree this should be fixed; the only reason I hesitate here is that people may now be relying on the broken behavior and have profiles named "profile default".  I will discuss with the team and see what they say

SamRemis:
Great :) it looks good to me, and I've started an internal audit to make sure other SDKs aren't making the same mistake.  Before I merge this I need to look into our metrics to see who this may be a breaking change for and contact them if I'm able to so there may be a bit of a delay, but this will be merged ASAP

PR #2331, opened 2021-10-09
Fix issue where a float could be passed to gmdate, throwing an exception in PHP 8.1

27pchrisl:
Fixes #2327

Adds a coercion to int, which would have been the internal behaviour of the function pre-8.1. The test suite expects the code should fail when passed a string like "this text is not a date", therefore the `is_numeric` test and exception.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2332, opened 2021-10-11
PHP 8.0, 8.1, PHPUnit 9 support

27pchrisl:
*Description of changes:*

This PR adds support for running tests against PHP 8.0 and 8.1, and upgrades to PHPUnit 9 (for supported PHP versions).

The changes implement polyfills for older PHPUnit versions to add missing 9.x methods, and restores older methods that newer PHPUnit versions removed. This is implemented through the `PHPUnitCompatTrait`, dynamically selected in `bootstrap.php`.

As PHPUnit 9 added the `void` return type to `setUp`, `tearDown`, `setUpBeforeClass` and `tearDownAfterClass`, the polyfills provide `_`-prefixed versions of these methods to support all PHP versions (eg `_setUp`).

The only actual failing `src` code in 8.0/8.1 was the same as fixed in #2331. All other changes are only to the test suite. I'm commenting below on notable changes.


stobrien89:
Hi @27pchrisl,

Going to close this in favor of #2403.  Want to try to have that merged relatively soon.

PR #2333, opened 2021-10-13
Fix random test failures in Aws\Test\Crypto\Polyfill\AesGcmTest::testCompat

27pchrisl:
This PR removes the randomness from the test that can cause test failures, and uses only valid tag lengths according to the spec, and specific plaintext length values.

As seen in: https://github.com/aws/aws-sdk-php/runs/3875170019?check_suite_focus=true

### DIFF
--- a/tests/Crypto/Polyfill/AesGcmTest.php
+++ b/tests/Crypto/Polyfill/AesGcmTest.php
@@ -2616,36 +2616,42 @@ class AesGcmTest extends TestCase
     {
         if (PHP_VERSION_ID < 70100) {
             $this->markTestSkipped('This test can only run on PHP 7.1');
+
             return;
         }
-        $ptLen = \random_int(0, 1024);
-        $aadLen = \random_int(0, 1024);
-        $i = 0;
-        $tag1 = $tag2 = '';
-        for ($i = 0; $i < 16; ++$i) {
-            $plaintext = \random_bytes($ptLen + $i);
-            $aad = \random_bytes($aadLen + $i);
-            $key = \random_bytes(32);
-            $nonce = \random_bytes(12);
 
-            $exp = \openssl_encrypt(
-                $plaintext,
-                'aes-256-gcm',
-                $key,
-                OPENSSL_RAW_DATA | OPENSSL_NO_PADDING,
-                $nonce,
-                $tag1,
-                $aad
-            );
-            $got = AesGcm::encrypt(
-                $plaintext,
-                $nonce,
-                new Key($key),
-                $aad,
-                $tag2
-            );
-            $this->assertSame(bin2hex($exp), bin2hex($got));
-            $this->assertSame(bin2hex($tag1), bin2hex($tag2));
+        $tag1 = $tag2 = '';
+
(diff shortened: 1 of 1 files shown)


scottarc:
> I'm not a crypto expert,

No worries. Most people aren't crypto experts, but that doesn't make your contribution less valuable. :)

> so I wasn't sure if there was any implication of the way I rewrote the test that made it less safe or secure as a test.

Strictly speaking, the rewritten test *does* reduce confidence of the implementation. Someone in the future might look at this and think, "Wait, this could have false positives, since they're not testing random inputs!"

However, a flaky test that fails unexpectedly is annoying. 

If possible, I'd like to keep the code largely as written before this PR, but with the underlying cause of the flaky test fixed. If that turns out to not be possible, I think the PR you submitted here is an improvement on the reliability axis.

> Based on reading the spec linked from the code in the PR, it definitely sounded like there were only very specific valid tag lengths that could be used.

Only 96-bit nonces are supported. Any attempt to use a different nonce size results in an exception being thrown.

https://github.com/aws/aws-sdk-php/blob/master/src/Crypto/Polyfill/AesGcm.php#L87-L91

However, we didn't restrict tag lengths in this code, because that can be accomplished in the encrypt path by simple truncation (but this will cause the decrypt path to fail, since that expects 128-bit tags).

https://github.com/aws/aws-sdk-php/blob/f9bf4c2e0f939d39b2dbbd9d891df5b5774d2e36/src/Crypto/Polyfill/AesGcm.php#L156-L158

In the future, we *could* throw a further `needs()` call that asserts a 128-bit tag on decryption, should anyone want to use this code outside of the S3 Encryption Client.

If I identify the root cause, I'll submit a superseding PR and link it to this one.

stobrien89:
Hey @scottarc,

Were you able to make any headway on identifying the root cause?

Zombaya:
I found a probable root-cause while creating the MR about PHP8.0,8.1.

https://github.com/aws/aws-sdk-php/blob/0f5d86c0dbed9b413cae6b77c4b5811bb857c9cb/tests/Crypto/Polyfill/AesGcmTest.php#L2621-L2622
https://github.com/aws/aws-sdk-php/blob/0f5d86c0dbed9b413cae6b77c4b5811bb857c9cb/tests/Crypto/Polyfill/AesGcmTest.php#L2626-L2627

In case `$ptLen` or `$aadLen` is 0, this will break since `random_bytes()` expects a value >= 1.

I believe this can be fixed by doing either `$plaintext = ($ptLen + $i) > 0 ? \random_bytes($ptLen + $i) : '';` or by using a dataProvider and test the cases where `$plaintext` and/or `$aad` are empty.

I made an attempt in https://github.com/Zombaya/aws-sdk-php/pull/5/commits/b3a22877110ec724850caeb2542419b2dc07f188.  The test for this is failing though because I upped the loop-counter from 16 to 3200, causing the test to run out of memory.  It can serve as an example though.
I upped the loop-counter to try and force the test to fail, as I still sometimes saw the test fail.
I added a failure-message to be able to log which randomly generated values caused the test to fail, to be able to reproduce the error.

PR #2334, opened 2021-10-13
Fixes usage of `http_build_query` on PHP 8.1

nunomaduro:
This pull request addresses the error `Passing null to parameter #2 ($numeric_prefix) of type string is deprecated` that gets thrown when using this library with PHP 8.1.


plugowski:
Could be nice to merge it soon :)

PR #2354, opened 2021-12-01
Fixes for PHP 8.1 compatibility

karyna-tsymbal-atwix:
*Issue #, if available:*

*Description of changes:*
This PR adds several fixes for PHP 8.1. compatibility.

1.
In PHP 8.1, >> non-final internal methods now require overriding methods to declare a compatible return type, otherwise a deprecated notice is emitted during inheritance validation.
RFC: https://wiki.php.net/rfc/internal_method_return_types

Because of this, we receive multiple deprecation notices when running our code with PHP8.1.

This PR adds #[ReturnTypeWillChange] attribute where it is needed, in order to avoid the described problem and keep backward compatibility.

2.
In 8.1 it's [deprecated](https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg) to pass `null` to non-nullable arguments of internal functions.
We're receiving such a warning in our logs:
`aws/aws-sdk-php/src/Signature/SignatureV4.php:327 rawurlencode(): Passing null to parameter #1 ($string) of type string is deprecated`

This PR adds an additional checking for null in a problematic place.

### DIFF
--- a/src/Signature/SignatureV4.php
+++ b/src/Signature/SignatureV4.php
@@ -324,11 +324,11 @@ class SignatureV4 implements SignatureInterface
         ksort($query);
         foreach ($query as $k => $v) {
             if (!is_array($v)) {
-                $qs .= rawurlencode($k) . '=' . rawurlencode($v) . '&';
+                $qs .= rawurlencode($k ?? '') . '=' . rawurlencode($v ?? '') . '&';
             } else {
                 sort($v);
                 foreach ($v as $value) {
-                    $qs .= rawurlencode($k) . '=' . rawurlencode($value) . '&';
+                    $qs .= rawurlencode($k ?? '') . '=' . rawurlencode($value ?? '') . '&';
                 }
             }
         }
--- a/build/docs/classes/ShapeIterator.php
+++ b/build/docs/classes/ShapeIterator.php
@@ -49,26 +49,31 @@ class ShapeIterator implements \Iterator
         $this->rewind();
     }
 
+    #[\ReturnTypeWillChange]
     public function current()
     {
         return $this->shapes[$this->index];
     }
 
+    #[\ReturnTypeWillChange]
     public function next()
     {
         $this->index++;
     }
 
+    #[\ReturnTypeWillChange]
     public function key()
     {
         return $this->index;
     }
 
+    #[\ReturnTypeWillChange]
     public function valid()
     {
         return isset($this->shapes[$this->index]);
     }
 
+    #[\ReturnTypeWillChange]
     public function rewind()
     {
         $this->index = 0;
--- a/src/Api/Parser/DecodingEventStreamIterator.php
+++ b/src/Api/Parser/DecodingEventStreamIterator.php
@@ -172,6 +172,7 @@ class DecodingEventStreamIterator implements Iterator
     /**
      * @return array
      */
+    #[\ReturnTypeWillChange]
     public function current()
     {
         return $this->currentEvent;
@@ -180,11 +181,13 @@ class DecodingEventStreamIterator implements Iterator
     /**
      * @return int
      */
+    #[\ReturnTypeWillChange]
     public function key()
     {
         return $this->key;
     }
 
+    #[\ReturnTypeWillChange]
     public function next()
     {
         $this->currentPosition = $this->stream->tell();
@@ -194,6 +197,7 @@ class DecodingEventStreamIterator implements Iterator
         }
     }
 
+    #[\ReturnTypeWillChange]
     public function rewind()
     {
         $this->stream->rewind();
@@ -205,6 +209,7 @@ class DecodingEventStreamIterator implements Iterator
     /**
      * @return bool
      */
+    #[\ReturnTypeWillChange]
     public function valid()
     {
         return $this->currentPosition < $this->stream->getSize();
--- a/src/Api/Parser/EventParsingIterator.php
+++ b/src/Api/Parser/EventParsingIterator.php
@@ -33,26 +33,31 @@ class EventParsingIterator implements Iterator
         $this->parser = $parser;
     }
 
+    #[\ReturnTypeWillChange]
     public function current()
     {
         return $this->parseEvent($this->decodingIterator->current());
     }
 
+    #[\ReturnTypeWillChange]
     public function key()
     {
         return $this->decodingIterator->key();
     }
 
+    #[\ReturnTypeWillChange]
     public function next()
     {
         $this->decodingIterator->next();
     }
 
+    #[\ReturnTypeWillChange]
     public function rewind()
     {
         $this->decodingIterator->rewind();
     }
 
+    #[\ReturnTypeWillChange]
     public function valid()
     {
         return $this->decodingIterator->valid();
--- a/src/CloudTrail/LogFileIterator.php
+++ b/src/CloudTrail/LogFileIterator.php
@@ -127,6 +127,7 @@ class LogFileIterator extends \IteratorIterator
      *
      * @return array|bool
      */
+    #[\ReturnTypeWillChange]
     public function current()
     {
         if ($object = parent::current()) {
--- a/src/ResultPaginator.php
+++ b/src/ResultPaginator.php
@@ -101,21 +101,25 @@ class ResultPaginator implements \Iterator
     /**
      * @return Result
      */
+    #[\ReturnTypeWillChange]
     public function current()
     {
         return $this->valid() ? $this->result : false;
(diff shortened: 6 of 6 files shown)


xmav:
@SamRemis Could you please check this PR ?

src/Signature/SignatureV4.php
```diff
@@ -324,11 +324,11 @@ private function getCanonicalizedQuery(array $query)
         ksort($query);
         foreach ($query as $k => $v) {
             if (!is_array($v)) {
-                $qs .= rawurlencode($k) . '=' . rawurlencode($v) . '&';
+                $qs .= rawurlencode($k ?? '') . '=' . rawurlencode($v ?? '') . '&';
```
SamRemis, line 327: The ?? operator was added in 7.0, but the SDK supports back to 5.5, so this will probably make the tests that are running fail.  Even if it doesn't, it's a non backward-compatible change.  What circumstance is this needed in though? 
I need to look into this snippet of code more to make sure, but it looks like it would never be null since it's iterating through an array and getting the key/value pairs, where $k is the key.  When would a key in a query be null?

SamRemis:
@xmav Of course, I took a look and ran the tests; it's failing because of the '??' operator.  Could this be fixed and I'll take another look?

src/Signature/SignatureV4.php
xmav: the ?? operator could be fixed in a backward compatible way for sure. Unfortunately don't have live debug data for this case now but trace looks like:
```
/var/www/html/vendor/aws/aws-sdk-php/src/Signature/SignatureV4.php:327 rawurlencode(): Passing null to parameter #1 ($string) of type string is deprecated
/var/www/html/vendor/aws/aws-sdk-php/src/Signature/SignatureV4.php:327
/var/www/html/vendor/aws/aws-sdk-php/src/Signature/SignatureV4.php:285
/var/www/html/vendor/aws/aws-sdk-php/src/Signature/SignatureV4.php:103
/var/www/html/vendor/aws/aws-sdk-php/src/Signature/S3SignatureV4.php:46
/var/www/html/vendor/aws/aws-sdk-php/src/Middleware.php:126
/var/www/html/vendor/guzzlehttp/promises/src/Promise.php:204
/var/www/html/vendor/guzzlehttp/promises/src/Promise.php:153
/var/www/html/vendor/guzzlehttp/promises/src/TaskQueue.php:48
/var/www/html/vendor/guzzlehttp/promises/src/Promise.php:248
```

that's with PHP8.1 - https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal

xmav: agree that $k could not be null. But $v could be
https://github.com/guzzle/psr7/blob/c55d23ab371b472342bee943a8bdb9c3cf963a18/src/Query.php#L45
most likely that "Passing null to parameter" recorded for second occurrence of `rawurlencode` in that line

xmav: https://github.com/aws/aws-sdk-php/blob/8c4863a369fd3aab630270ea09b576de81887a3b/src/Signature/SignatureV4.php#L402

xmav: Hi @karyna-tsymbal-atwix !
Could you please do corresponding changes ? (I don't think I can commit\push to PR branch)

karyna-tsymbal-atwix: Hello @xmav  @SamRemis  , I've made changes according to the discussion above

andrewbess:
Hello @SamRemis 
We fixed PR according to your recommendations
Could you please check it?
Thank you in advance.

SamRemis:
I need to look into this a bit more just to ensure backward compatibility on the rawurlencode function, but other than that LGTM.

Please add a change log file as shown here before I merge:
https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md#changelog-documents


Just as a note, this repo needs to maintain backward compatibility to 5.5, so adding these serves to suppress the message, and isn't a guarantee the return types will be added in 9.0; at that point, I'll likely need to make a new repo to make the breaking changes

BenMorel:
> at that point, I'll likely need to make a new repo to make the breaking changes

You just need a new major version. It's OK to stop supporting older PHP versions at some point. PHP 5.5 support has ended more than 5 years ago, and the only supported PHP versions are now [7.4, 8.0 and 8.1](https://www.php.net/supported-versions.php).

Most libraries nowadays require at the bare minimum PHP 7.1, with more and more libraries going more agressive over the years (Symfony 6 now requires PHP 8).

Just like Symfony, you can also decide to support 2 versions: a LTS version that supports older PHP versions and only gets security updates (with a clear end of support date), and a newer version that gets the new features and only supports recent versions of PHP.

I personally think that library maintainers going the extra mile to support very old PHP versions are actually doing a disservice to the community, by postponing the necessary regular maintenance any codebase requires, and keeping old, unpatched PHP versions live.

src/Signature/SignatureV4.php
```diff
@@ -324,11 +324,11 @@ private function getCanonicalizedQuery(array $query)
         ksort($query);
         foreach ($query as $k => $v) {
             if (!is_array($v)) {
-                $qs .= rawurlencode($k) . '=' . rawurlencode($v) . '&';
+                $qs .= rawurlencode($k) . '=' . rawurlencode($v !== null ? $v : '') . '&';
```
SamRemis, line 327: While this isn't a breaking change, why is it necessary?  `rawurlencode(null)` returns an empty string , same as `rawurlencode('')`.  Did this method change somewhere that I didn't see?  I couldn't find it in the PHP changelogs when I looked, but it's possible I'm not considering something or I missed it.

xmav: that has difference on PHP8.1 - https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal
https://www.php.net/manual/en/function.rawurlencode.php
So this generates "Deprecation notice" with PHP8.1 and depending on how php&application is configured may lead to unexpected behavior. (i.e application could be configured to halt in case of deprecation notice))

SamRemis: Thank you :) that makes sense

SamRemis:
Did the testing I was talking about and got my questions answered, so once that changelog is in, I'm happy to merge after the next successful release; after that outage in us-east-1 yesterday, I'd prefer to wait to see a successful release before merging anything else.

SamRemis:
@BenMorel Thanks for the feedback, I'll definitely discuss it with the team.  I'd personally love to be able to use the new syntax and features of more recent versions of PHP, but it's difficult since a non-insignificant portion of our customers still use 5.5 and 5.6.   One of the most important things for us is to never intentionally introduce a breaking change.  

That also makes a repo that rolls supported versions and features along with PHP difficult to imagine, though I like the idea of it.  I'm happy to discuss this on gitter if there's a good way to do it while guaranteeing no breaking changes without forcing users to pin their versions.

.changes/nextrelease/bugfix-php-v81-compatibility.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "type": "bugfix",
+        "category": "",
+        "description": "This release adds #[ReturnTypeWillChange] attribute where it is needed for PHP 8.1 compatibility"
+    },
+    {
```
SamRemis, line 7: While I personally think this should be valid syntax, it's not in the current state of the SDK.  Please condense the two entries into one.  Something along the lines of 

```
[
    {
        "type": "bugfix",
        "category": "",
        "description": "This release adds #[ReturnTypeWillChange] attribute where it is needed and fixes a deprecation notice when pass null to rawurlencode() for PHP 8.1 compatibility"
    }
 ]
 
```

SamRemis: Sorry to be a pain about this

karyna-tsymbal-atwix: hi @SamRemis , no problem :)

andrewbess:
Hello @SamRemis 
Fixed, please check.
Thank you in advance.

SamRemis:
Of course- I'm going to wait for the actions to finish running, and will merge once they have passed :) thanks again

SamRemis:
Just as an FYI, this is merged, but won't be released via composer until the next tagged release.  One is scheduled for today; they generally happen around 2:30pm EST, but they are occasionally later.  

Thanks to everyone who contributed to this and to the discussion :)

PR #2365, opened 2021-12-20
fixed "hash_init(): Passing null to parameter #3 ($key) of type string is deprecated

marc-mabe:
… in PhpHash::getContext()

*Description of changes:*

The signature of `hash_init` is defined as `hash_init(string $algo, int $flags = 0, string $key = "", array $options = []): HashContext`

So if no key is needed the argument should either be skipped or an empty string but not `NULL`.
Passing `NULL` is deprecated since PHP 8.1.

### DIFF
--- a/src/PhpHash.php
+++ b/src/PhpHash.php
@@ -68,7 +68,7 @@ class PhpHash implements HashInterface
     private function getContext()
     {
         if (!$this->context) {
-            $key = isset($this->options['key']) ? $this->options['key'] : null;
+            $key = isset($this->options['key']) ? $this->options['key'] : '';
             $this->context = hash_init(
                 $this->algo,
                 $key ? HASH_HMAC : 0,


ruudk:
Is there something holding back this from being merged?

@marc-mabe Could you maybe rebase this PR to see if that solves the failing test?

ruudk:
@SamRemis Is there something holding back this from being merged?

ruudk:
Created a new PR that includes this + a changelog document, https://github.com/aws/aws-sdk-php/pull/2394  lets see if the tests pass.

SamRemis:
@PrinsFrank Yep :) thanks for the head up.  

Looks like this is the same logic that's in the main repo and live right now: https://github.com/aws/aws-sdk-php/blob/master/src/PhpHash.php

PR #2379, opened 2022-02-24
Update RestSerializer.php

NairanOmura:
Hello,

When I use the class "ApiGatewayManagementApiClient" with the URL generated by api gateway (ex.: https://xxxxxxxxx.execute-api.us-east-2.amazonaws.com/dev) the response is "(client) Forbidden",

I'm using the postToConnection() method and I noticed that the url generated by the aws sdk was stripping the "stage" from the like "https://ixe7e2mjfl.execute-api.us-east-2.amazonaws.com/@connections/OBZeGcyKCYcCGng=" where the correct would be  "https://ixe7e2mjfl.execute-api.us-east-2.amazonaws.com/dev/@connections/OBZeGcyKCYcCGng=".. debugging the aws sdk i find the solution proposed in Pull-Request.

### DIFF
--- a/src/Api/Serializer/RestSerializer.php
+++ b/src/Api/Serializer/RestSerializer.php
@@ -217,7 +217,7 @@ abstract class RestSerializer
         // If endpoint has path, remove leading '/' to preserve URI resolution.
         $path = $this->endpoint->getPath();
         if ($path && $relative[0] === '/') {
-            $relative = substr($relative, 1);
+            $relative = substr($path.$relative, 1);
         }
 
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @NairanOmura,

Thanks for the PR!  And sorry for the delayed response.  We won't be able to accept this change because it appears to contain breaking changes, but if you could [open an issue](https://github.com/aws/aws-sdk-php/issues/new/choose) with us and provide some additional context like sample client code and debug logs, we'd be happy to investigate this further.

PR #2388, opened 2022-02-26
Bugfix: endpoint variant tag resolution

### LINKED ISSUE #2386: Bug: Endpoint Variant Resolution
👋  PHP SDK team,

I was doing some investigatory work on endpoints and noticed a potential issues with the resolution of variants that I wanted to raise to you.

The code in question is here: https://github.com/aws/aws-sdk-php/blob/b98b82205f521c84ebd2d50ca0d6936acd9b5131/src/Endpoint/Partition.php#L286-L315

This implementation appears to be dependent on the order on which the endpoint variants tags are appended to the `$variantTags` array. Thus when the model has the tags in an alternative order (as it doesn't guarantee in the specification the order tags will be presented in) the array comparisons will not be successful.

As an example if I have configured FIPS and DualStack to be enabled then the variant tags would be defined as `$variantTags = ["fips", "dualstack"]`, and when iterating the model it would fail to match variants tags like the one here: https://github.com/aws/aws-sdk-php/blob/master/src/data/endpoints.json#L14 as `["fips", "dualstack"] != ["dualstack", "fips"]`.

This model iteration should be changed to not be dependent on the tag ordering.

stobrien89:
*Issue #, if available:* Closes #2386

*Description of changes:*
Fixes an issue where endpoint variant tags were being evaluated in an order-dependent manner.  Although there aren't any models with alternative ordering currently, it is not guaranteed future models won't be published with an alternative ordering of tags, as ordering is not guaranteed in the specification. 


SamRemis (approved):

stobrien89:
Manually tested today with an S3 client and the `use_fips_endpoint`/`use_dual_stack_endpoint` config options and all looks good.  Made sure to swap the tag order in the `fips.dualstack` S3 endpoint.

PR #2394, opened 2022-03-03
[PHP 8.1] fixed "hash_init(): Passing null to parameter #3 ($key) of type string is deprecated

ruudk:
Retry of #2365 that includes required changelog document

### DIFF
--- a/.changes/nextrelease/bugfix-php81-hash_init
+++ b/.changes/nextrelease/bugfix-php81-hash_init
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "",
+        "description": "Passing null to hash_init is deprecated on PHP 8.1."
+    }
+]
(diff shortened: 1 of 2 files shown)


ruudk:
@SamRemis is now green + contains required changelog. Would be great to have this merged and tagged so that the deprecation disappears on PHP 8.1.

ruudk:
I don't have more information. I just saw the deprecation and checked the php.net/hash_init to confirm.

SamRemis:
@ruudk, I was looking in the wrong place.  It states pretty clearly [here](https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation) that replacing these null values with their defaults is how you upgrade this.  Since you set it to the default value, this should be good to merge.  Thank you for your contribution :)

SamRemis (approved):

Issue #2323, opened 2021-09-29
AWS CloudShell `Error retrieving credentials from the instance profile metadata service.`


kuettai:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Background: The cloudshell environment does not have "aws configure" setup.

I created a python file p.py with the following

```python
import boto3
s3 = boto3.client('s3')
resp = s3.list_buckets()
print(resp)
```

Python3 successfully returns me the list of buckets.

However in PHP SDK, (the scripts work in EC2, with InstanceProfile) but in CloudShell, i get this error: "Error retrieving credentials from the instance profile metadata service"

This is the sample code:

```php
use Aws\Rds\RdsClient;
$o = new RdsClient([
    'signature_version' => 'v4',
    'region' => 'ap-southeast-1',
    'version' => '2014-10-31'
]);
$o->describeDBInstances();
```

What is the recommended workaround for PHP SDK to follow CloudShell IAM-User permission? The expectation on permission granting in PHP SDK should behave the same as Python SDK.

**Version of AWS SDK for PHP?**
v3.192.1

**Version of PHP (`php -v`)?**
PHP 7.2.34 (cli)

**To Reproduce (observed behavior)**
IAM User with readOnlyPermission in RDS, login to AWS Console. Launch cloudshell (in any available region). Install php.
```bash
if(sudo yum list installed  | grep php-cli > /dev/null) then echo 'PHP installed ,skipped'; else sudo amazon-linux-extras install -y php7.2; fi
##aws-sdk requires mbstring and xml
if(sudo yum list installed | grep php-mbstring > /dev/null) then echo 'php-mbstring installed, skipped'; else sudo yum install php-mbstring -y; fi
if(sudo yum list installed | grep php-xml > /dev/null) then echo 'php-xml installed, skipped'; else sudo yum install php-xml -y; fi
```

[...]

PR #2324, opened 2021-09-30
Allows Cloudshell to acquire perm from AWS_CONTAINER_CREDENTIALS_FULL…

kuettai:
…_URI

*Issue #, if available:*
https://github.com/aws/aws-sdk-php/issues/2323

*Description of changes:*
Added supports for AWS_CONTAINER_CREDENTIALS_FULL_URI in EcsCredentialProvider.php, this will allows aws-php-sdk to works with CloudShell correctly. Permission in Cloudshell to follow IAM User.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2249, opened 2021-05-19
Add AWS_CONTAINER_CREDENTIALS_FULL_URI support to EcsCredentialProvider

BenHarris:
With reference to using [amazon/amazon-ecs-local-container-endpoints](https://github.com/awslabs/amazon-ecs-local-container-endpoints) for local development and continuing the discussion within [PR #48 over there](https://github.com/awslabs/amazon-ecs-local-container-endpoints/pull/48)...

When used with docker, the special/static 169.254.170.2 IP requires that a custom network is created and each container is attached to this network and given an IP within this space. This creates significant bloat in docker-compose files which is essentially unnecessary. If we were able to pass in a hostname, it could then be resolved by docker irrespective of any other network configuration. The only configuration then required would be that of the ecs-local-endpoints container.

*Description of changes:*
By implementing a `AWS_CONTAINER_CREDENTIALS_SERVER_URI` environmental variable, we are able to pass in an internal hostname to better support the use of Amazon ECS Local Container Endpoints.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2401, opened 2022-03-14
Allows Cloudshell to acquire perm from AWS_CONTAINER_CREDENTIALS_FULL v2

kuettai:
*Issue #2323 , if available:*

*Description of changes:*
Added supports for AWS_CONTAINER_CREDENTIALS_FULL_URI in EcsCredentialProvider.php, this will allows aws-php-sdk to works with CloudShell correctly. Permission in Cloudshell to follow IAM User.

### DIFF
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -111,6 +111,10 @@ class EcsCredentialProvider
 
         if(empty($credsUri)){
             $credFullUri = getenv(self::ENV_FULL_URI);
+            if($credFullUri === false){
+                $credFullUri = isset($_SERVER[self::ENV_FULL_URI]) ? $_SERVER[self::ENV_FULL_URI] : '';
+            }
+
             if(!empty($credFullUri))
                 return $credFullUri;
         }
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -57,11 +57,7 @@ class EcsCredentialProvider
         $client = $this->client;
         $request = new Request('GET', self::getEcsUri());
         
-        $authToken = self::getEcsAuthToken();
-        $headers = [];
-        if(!empty($authToken))
-            $headers = ['Authorization' => $authToken];
-        
+        $headers = $this->setHeaderForAuthToken();
         return $client(
             $request,
             [
@@ -89,6 +85,15 @@ class EcsCredentialProvider
     private function getEcsAuthToken()
     {
         return getenv(self::ENV_AUTH_TOKEN);
+    }
+
+    public function setHeaderForAuthToken(){
+        $authToken = self::getEcsAuthToken();
+        $headers = [];
+        if(!empty($authToken))
+            $headers = ['Authorization' => $authToken];
+
+        return $headers;
     }
 
     /**
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -103,14 +103,16 @@ class EcsCredentialProvider
      */
     private function getEcsUri()
     {
-        $credFullUri = getenv(self::ENV_FULL_URI);
-        if(!empty($credFullUri))
-            return $credFullUri;
-        
         $credsUri = getenv(self::ENV_URI);
 
         if ($credsUri === false) {
             $credsUri = isset($_SERVER[self::ENV_URI]) ? $_SERVER[self::ENV_URI] : '';
+        }
+
+        if(empty($credsUri)){
+            $credFullUri = getenv(self::ENV_FULL_URI);
+            if(!empty($credFullUri))
+                return $credFullUri;
         }
         
         return self::SERVER_URI . $credsUri;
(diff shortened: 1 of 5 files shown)


kuettai:
ATTN: @stobrien89 

Previous PR #2324 , changes made to the comment.
Would also solve PR #2249 as per highlighted in PR #2324

stobrien89:
Hi @kuettai,

Thanks for taking the time to submit the changes.  I'll discuss with the team and get back to you as soon as I can.  If it's alright with you, I'll close the older PR and focus on this one for now.

src/Credentials/EcsCredentialProvider.php
```diff
@@ -54,11 +56,18 @@ public function __invoke()
...
+        $headers = [];
+        if(!empty($authToken))
+            $headers = ['Authorization' => $authToken];
+        
         return $client(
             $request,
             [
                 'timeout' => $this->timeout,
                 'proxy' => '',
+                'headers' => $headers
```
stobrien89, line 70: We want some component of a new test for the `EcsCredentialProvider` to ensure the Authorization header is set and the value matches the token

tests/Credentials/CredentialProviderTest.php
```diff
@@ -1813,19 +1817,27 @@ public function testCachesCacheableInDefaultChain()
             'process_credentials',
             'process_config',
             'ecs',
+            'ecs_cloudshell',
```
stobrien89, line 1820: We should remove this entry, as `$cacheable` here represents the default chain.  `ecs_cloudshell` credentials should still be categorized under `ecs` since it's using the same provider

stobrien89 (changes requested):
Implementation looks good- the testing just needs to be refined.  Ideally we'd want to have a separate test in `EcsCredentialProviderTest` that simulates a scenario where we'd be using these two env vars and checks that things like the `Authorization` header are being set correctly.

tests/Credentials/CredentialProviderTest.php
kuettai: ### ok, how do you suggest we test the needs of 2 different scenario whereas
Scenario 1: AWS_CONTAINER_CREDENTIALS_RELATIVE_URI
Scenario 2: Without the above ENV variable, but with AWS_CONTAINER_CREDENTIALS_FULL_URI & AWS_CONTAINER_AUTHORIZATION_TOKEN

src/Credentials/EcsCredentialProvider.php
kuettai: Separated the $headers generate logic to separate public function in the same class to make it unit-testable. Added the code to the token. Will push the fix once we align.

tests/Credentials/CredentialProviderTest.php
kuettai: Removed ecs_cloudshell unit-testing from defaultChain

src/Credentials/EcsCredentialProvider.php
kuettai: Code pushed.

stobrien89 (approved):
Conditional approval to run checks

src/Credentials/EcsCredentialProvider.php
```diff
@@ -84,6 +103,10 @@ public function __invoke()
      */
     private function getEcsUri()
     {
+        $credFullUri = getenv(self::ENV_FULL_URI);
```
stobrien89, line 106: Dug into it a bit further and it looks like this will need to be refactored.  Other SDKs evaluate the relative uri before the full uri and that's what we'll need to do for sake of consistency.  The [Java docs](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/EC2ContainerCredentialsProviderWrapper.html) give a good outline of the desired logic here.  More than willing to chip in on this if you need any help.

stobrien89 (changes requested):
Left a comment regarding the uri retrieval logic.  We'll need to re-order some things for sake of consistency with other SDKs

src/Credentials/EcsCredentialProvider.php
kuettai: done

stobrien89 (changes requested):
Left one more comment. Appreciate your patience here.  I think we'll be ready for a final review after it's addressed.

src/Credentials/EcsCredentialProvider.php
```diff
@@ -89,6 +108,12 @@ private function getEcsUri()
         if ($credsUri === false) {
             $credsUri = isset($_SERVER[self::ENV_URI]) ? $_SERVER[self::ENV_URI] : '';
         }
+
+        if(empty($credsUri)){
+            $credFullUri = getenv(self::ENV_FULL_URI);
```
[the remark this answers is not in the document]
kuettai: Restructure and added the fallback accordingly. UnitTest for EcsCredentialProvider and "make test" has been executed and pass locally.

stobrien89 (approved):
Pending team review

stobrien89:
Hi @kuettai and @BenHarris, 

Thanks again for your work on this.  Tested manually using docker as well as cloudshell and everything works as expected.

PR #2405, opened 2022-03-21
Fix: warning when passing null as parameter to ucfirst() in Ec2ParamB…

yenfryherrerafeliz:
…uilder.php

From src/Api\Serializer/Ec2ParamBuilder.php ucfirst() function is being call with the posibility of passing null as parameter, which could cause a warning if using PHP>=8.1 due that passing null to a function parameter not declared nullable throws a deprecation warning in this php version. To prevent this warning from happening, what I did was to make sure that when calling ucfirst() the variable has value in it.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/Api/Serializer/Ec2ParamBuilder.php
+++ b/src/Api/Serializer/Ec2ParamBuilder.php
@@ -12,7 +12,7 @@ class Ec2ParamBuilder extends QueryParamBuilder
     protected function queryName(Shape $shape, $default = null)
     {
         return ($shape['queryName']
-            ?: ucfirst($shape['locationName']))
+                    ?: ( isset($shape['locationName']) ? ucfirst($shape['locationName']) : null) )
                 ?: $default;
     }
 


src/Api/Serializer/Ec2ParamBuilder.php
```diff
@@ -12,7 +12,7 @@ class Ec2ParamBuilder extends QueryParamBuilder
     protected function queryName(Shape $shape, $default = null)
     {
         return ($shape['queryName']
-            ?: ucfirst($shape['locationName']))
```
SamRemis, line 15: I probably gave wrong directions here initially; the default should just be changed to `''` if this is only to fix the warning.  If it's to fix an additional warning around $shape['locationName'] not being set, then it should look more like 
```
        return ($shape['queryName']
            ?: isset ($shape['locationName'])) 
                ? ucfirst($shape['locationName'])
                : "";
```

Issue #2378, opened 2022-02-23
PHP Deprecated:  ucfirst(): Passing null to parameter #1 ($string) of type string is deprecated


utdrmac:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x ] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [ x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [ x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
```
PHP Deprecated:  ucfirst(): Passing null to parameter #1 ($string) of type string is deprecated in phar:///Users/foo/aws/aws.phar/Aws/Api/Serializer/Ec2ParamBuilder.php on line 15
```

**Version of AWS SDK for PHP?**
3.209.28
**Version of PHP (`php -v`)?**
8.1.2

**To Reproduce (observed behavior)**
Use any API in Aws\Ec2\Ec2Client

**Expected behavior**
I expect AWS SDK to work with the latest PHP version and produce no deprecation warnings/errors

utdrmac:
Also,
```
PHP Deprecated:  Implicit conversion from float 4497975.890990347 to int loses precision in phar:///Users/foo/aws/aws.phar/GuzzleHttp/Handler/CurlMultiHandler.php on line 143
```

sapphirecat:
I am also seeing this issue.  I am attaching an anonymized stack trace generated by XDebug.  (We are still running 8.0 in production, but I've updated my laptop to Homebrew's current 8.1 version to start looking for related issues like this.)

```
PHP 8.1.3 (cli) (built: Feb 18 2022 09:24:27) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.3, Copyright (c) Zend Technologies
    with Xdebug v3.1.3, Copyright (c) 2002-2022, by Derick Rethans
    with Zend OPcache v8.1.3, Copyright (c), by Zend Technologies
```

[error-log.txt](https://github.com/aws/aws-sdk-php/files/8133724/error-log.txt)

SamRemis:
[...]

PR #2407, opened 2022-03-22
Fix: warning when passing null as parameter to ucfirst()

yenfryherrerafeliz:
From src\Api\Serializer\Ec2ParamBuilder.php ucfirst() function is being call with the possibility of passing null  as parameter, which could cause a warning if using PHP>=8.1 due that passing null to a function parameter not declared nullable throws a deprecation warning in this php version. To prevent this warning from happening, what I did was to make sure that when calling ucfirst the variable has a value in it.

*Issue #2378, if available:*

*Description of changes:*
I updated the logic of the line in question to make sure that when calling ucfirst a value is passed.


src/Api/Serializer/Ec2ParamBuilder.php
```diff
@@ -12,7 +12,7 @@ class Ec2ParamBuilder extends QueryParamBuilder
     protected function queryName(Shape $shape, $default = null)
     {
         return ($shape['queryName']
-            ?: ucfirst($shape['locationName']))
+            ?: (isset($shape['locationName']) ? ucfirst($shape['locationName']) : ""))
```
SamRemis, line 15: Sorry for giving mixed feedback here a few times; this could be shortenened to 
`ucfirst($shape['locationName'] ?: "")`

SamRemis (approved):

PR #2417, opened 2022-04-06
Chore: Use new syntax for version constraint, bump PHPUnit5 and guzzlehttp/psr7 2

jonnitto:
Use the double pipe instead of the single pipe for version constraints, as the single pipe is obsolete:
https://getcomposer.org/doc/articles/versions.md#version-range


stobrien89:
Hi @jonnitto,

Thanks for the PR.  I just opened one which bumps the min version for PHPUnit 5 and guzzleHttp/psr7— I think it would be best to do both in one go.  I'll push my changes to this PR.

stobrien89 (approved):

Issue #1561, opened 2018-05-28
Object existence methods and delete markers


PabloWestphalen:
I think it would be nice if the [checkExistenceWithCommand ](https://github.com/aws/aws-sdk-php/blob/2d0508ddeac51bbb0e0f86465db6b7ac50a597fb/src/S3/S3ClientTrait.php#L283) method, in addition to returning true or false, threw a special exception or at least a specialized message in S3Exception for the uncovered, grey "object does exist but its latest version is a delete marker" area.

Maybe something in the lines of

        try {
            $this->execute($command);
            return true;
        } catch (S3Exception $e) {
            if ($e->getAwsErrorCode() == 'AccessDenied') {
                return true;
            }
            if ($e->getStatusCode() >= 500) {
                throw $e;
            }

    +       if (!empty($e->getResponse()->getHeaders()["x-amz-delete-marker"])) {
    +           throw new S3Exception('File exists, but its current version is a delete marker', $command, null, $e);
    +       }

            return false;
        }

kstich:
Adding an exception here would be a behavior change and muddy the object being deleted. The object, having been previously deleted, does not exist in the current version of the bucket.

We could add a behavioral flag for `$includeDeleteMarkers = false` to `doesObjectExist` (and subsequently `checkExistenceWithCommand`) to return true if the object being checked is a delete marker. Would that fit what you're looking for?

PabloWestphalen:
Oh absolutely. I'm new to the SDK, and my proposal was the first yet probably naive solution i could think of (which is why i created the issue instead of going straight for a PR). Sure a flag that like that would work too.

stobrien89:
@PabloWestphalen,

Are you still interested in this feature? Currently working on a second version of the existence helper methods.

stobrien89:
@PabloWestphalen,

This has been implemented in #2424 and will be available in tomorrow's release.

PR #2424, opened 2022-04-11
feat: doesObjectExist and doesBucketExist v2

### LINKED ISSUE #2342: doesObjectExist should not return true or false based on assumptions
  
**Describe the bug**
If the request made in `doesObjectExist` fails for any reason, it should throw an exception instead of returning `false` based on any assumptions (e.g. bucket does not exist if endpoint can not be reached). Otherwise, the result may be wrong, i.e. `false` is returned, even though the object does actually exist.

I guess it is a [similar issue](https://github.com/aws/aws-sdk-java/issues/1256) as in the Java SDK. You may consider to introduce `doesBucketExistV2` [as well](https://github.com/aws/aws-sdk-java/issues/1256#issuecomment-331214701)

> Due to some concerns about internal service teams being tightly coupled to the underlying API call we decided to rollback this change so that doesBucketExist continues to use headBucket. We've deprecated this method and introduced a doesBucketExistV2 which uses the new getBucketAcl approach.

Example: v3.191.6

7.4

To use an endpoint URL that can not be resolved, is IMO the easiest way to get an error in the request. Otherwise, it depends on your endpoint configuration. You can run the following example without any changes to the configuration:

```php
  'version' => 'latest',
  'region' => 'eu-central-1',
  'endpoint' => 'http://endpoint-that-can-not-be-resolved',
  'use_path_style_endpoint' => true,
  'credentials' => [
    'key' => 's3CredentialsKey',
    'secret' => 's3CredentialsSecret',
  ],
]);

$bucketName = 'bucketName';
$doesExist = $client->doesObjectExist($bucketName, 'bucketKey');
// should not be reached, since exception should be thrown by doesObjectExist:
var_dump($doesExist); // false
// the following throws an exception as expected
$client->createBucket(['Bucket' => $bucketName]);
```

You can also reproduce this issue with your own endpoint if you use invalid credentials to trigger a 403. In that case, `doesObjectExist` should throw an exception, too.

**Expected behavior**
`doesObjectExist` should throw an exception.

**Additional context**
In my private project, I had an issue, when I uploaded a lot of files. I used `doesObjectExist` to check if I already uploaded the file. This case was actually meant to be used, if I have to run the script again. However, in the first run, `doesObjectExist` returned `true` in some cases (0.1%), even though the file did actually not exist and hence, wasn't uploaded. Are there any assumptions, where `true` is returned? I'm not sure how to reproduce this consistently or how to log relevant information if the result was wrong, but I guess that requests have failed with another error as I described in the reproduce section of this issue.

stobrien89:
*Issue #, if available:*
Implements #1561, closes #2342 

*Description of changes:*

New object/bucket existence helper methods with changed exception handling behavior.  Previously, `false` was returned any time an exception was thrown by the underlying S3 APIs (`headObject` and `headBucket`) unless the exception's status code was >= `500` (then exception was thrown) or the exception's error code was `AccessDenied` (`true` was returned).  This was problematic because `false` could be returned in the cases of an unresolvable endpoint, expired credentials, invalid credentials, or if bucket permissions were not granted to the calling entity and resources did in fact exist.

This implementation makes fewer assumptions in cases where status codes are ambiguous, namely `403` status codes. a `403` status code can signify bad credentials or incorrect bucket-level permissions— this causes issues with `headObject`/`doesObjectExist` calls in particular because a 403 will be returned whether or not an object exists. In this implementation, exceptions will be thrown in all cases when an exception is thrown by the underlying API, unless the exception has a  `404` status code (returns false) or in the case of `DoesBucketExist`, a user can specify if they want to accept `403` status codes, as a 403 will be thrown if a bucket exists, but an entity doesn't have correct bucket-level permissions.

Other changes include accepting bucket redirects (incorrectly specified region) as a `true` case for `DoesBucketExist`, the `$accept403` flag for `doesBucketExist` and the `$includeDeleteMarkers` flag for `doesObjectExist`, which counts a `404` response with the presence of a delete marker header as a `true` case.

### DIFF
--- a/src/S3/S3ClientInterface.php
+++ b/src/S3/S3ClientInterface.php
@@ -41,6 +41,8 @@ interface S3ClientInterface extends AwsClientInterface
     public function getObjectUrl($bucket, $key);
 
     /**
+     * @deprecated Use doesBucketExistV2() instead
+     *
      * Determines whether or not a bucket exists by name.
      *
      * @param string $bucket  The name of the bucket
@@ -50,6 +52,8 @@ interface S3ClientInterface extends AwsClientInterface
     public function doesBucketExist($bucket);
 
     /**
+     * @deprecated
+     *
      * Determines whether or not a bucket exists by name. This method uses S3's
      * HeadBucket operation and requires the relevant bucket permissions in the
      * default case to avoid inaccuracies.
@@ -66,6 +70,8 @@ interface S3ClientInterface extends AwsClientInterface
     public function doesBucketExistV2($bucket, $accept403);
 
     /**
+     * @deprecated Use doesObjectExistV2() instead
+     *
      * Determines whether or not an object exists by name.
      *
      * @param string $bucket  The name of the bucket
--- a/src/S3/S3ClientInterface.php
+++ b/src/S3/S3ClientInterface.php
@@ -52,8 +52,6 @@ interface S3ClientInterface extends AwsClientInterface
     public function doesBucketExist($bucket);
 
     /**
-     * @deprecated
-     *
      * Determines whether or not a bucket exists by name. This method uses S3's
      * HeadBucket operation and requires the relevant bucket permissions in the
      * default case to avoid inaccuracies.
--- a/src/S3/S3ClientInterface.php
+++ b/src/S3/S3ClientInterface.php
@@ -52,8 +52,8 @@ interface S3ClientInterface extends AwsClientInterface
      * Determines whether or not a bucket exists by name.
      *
      * @param string $bucket  The name of the bucket
-     * @param bool $accept403 Changes behavior to return true in the case of a 403. Bucket
-     *                        permissions and credentials MUST be correct to avoid inaccuracies.
+     * @param bool $accept403 Changes behavior to return true in the case of a 403.
+     *                        credentials MUST be correct to avoid inaccuracies.
      *
      * @return bool
      */
--- a/src/S3/S3ClientInterface.php
+++ b/src/S3/S3ClientInterface.php
@@ -4,6 +4,7 @@ namespace Aws\S3;
 use Aws\AwsClientInterface;
 use Aws\CommandInterface;
 use Aws\ResultInterface;
+use Aws\S3\Exception\S3Exception;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Http\Message\RequestInterface;
 
@@ -49,13 +50,18 @@ interface S3ClientInterface extends AwsClientInterface
     public function doesBucketExist($bucket);
 
     /**
-     * Determines whether or not a bucket exists by name.
+     * Determines whether or not a bucket exists by name. This method uses S3's
+     * HeadBucket operation and requires the relevant bucket permissions in the
+     * default case to avoid inaccuracies.
      *
      * @param string $bucket  The name of the bucket
-     * @param bool $accept403 Changes behavior to return true in the case of a 403.
-     *                        credentials MUST be correct to avoid inaccuracies.
+     * @param bool $accept403 Set to true for this method to return true in the case of
+     *                        invalid bucket-level permissions. Credentials MUST be valid
+     *                        to avoid inaccuracies. Using the default value of false will
+     *                        cause an exception to be thrown instead.
      *
      * @return bool
+     * @throws S3Exception|Exception if there is an unhandled exception
      */
     public function doesBucketExistV2($bucket, $accept403);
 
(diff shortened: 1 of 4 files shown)


src/S3/S3ClientTrait.php
```diff
@@ -297,6 +330,42 @@ private function checkExistenceWithCommand(CommandInterface $command)
...
+        CommandInterface $command,
+        $accept403,
+        $includeDeleteMarkers = false
+    )
+    {
+        try {
+            $this->execute($command);
+            return true;
+        } catch (S3Exception $e) {
+            if (($accept403 && $e->getStatusCode() === 403)
```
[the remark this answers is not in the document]
SamRemis: Another option (that I like less well cause of performance) would be to add some variables above this like so:

```
$statusCode = $e->getStatusCode();
$exceptionHasDeleteMarker = !empty($e->getResponse()->getHeader("x-amz-delete-marker"))
if (
       ($accept403 && $statusCode == 403)
       || ($includeDeleteMarkers && $exceptionHasDeleteMarker)
 ) {
       return true;
 }
```

src/S3/S3ClientInterface.php
```diff
@@ -48,6 +48,17 @@ public function getObjectUrl($bucket, $key);
      */
     public function doesBucketExist($bucket);
 
+    /**
+     * Determines whether or not a bucket exists by name.
+     *
+     * @param string $bucket  The name of the bucket
+     * @param bool $accept403 Changes behavior to return true in the case of a 403.
```
SamRemis, line 55: How about "set to true to have this method return true in the case of a 403 credential error; using the default value of false will cause the error to be thrown instead"

src/S3/S3ClientTrait.php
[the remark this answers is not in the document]
stobrien89: I wasn't sure how we felt about separate if statements that result in the same return value, but I like that approach.  Better for readability and less cumbersome.

src/S3/S3ClientTrait.php
```diff
@@ -297,6 +330,42 @@ private function checkExistenceWithCommand(CommandInterface $command)
...
+                || (!empty($e->getResponse()->getHeaders()["x-amz-delete-marker"])
```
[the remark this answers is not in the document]
stobrien89: Agreed. I'll get that updated

src/S3/S3ClientTrait.php
```diff
@@ -297,6 +330,42 @@ private function checkExistenceWithCommand(CommandInterface $command)
...
+                    && $includeDeleteMarkers)
+                || ($command->getName() === 'HeadBucket'
```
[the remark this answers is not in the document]
stobrien89: This is a condition I'll need to test again— the error code _should_ be a 301, but I'm pretty sure it wasn't exposed on the exception itself and it was shorter to write an `instanceof` than opening up the response for it.

src/S3/S3ClientTrait.php
```diff
@@ -260,6 +261,30 @@ public function doesBucketExist($bucket)
...
+     */
+    public function doesBucketExistV2($bucket, $accept403 = false)
+    {
+        $command = $this->getCommand('HeadBucket', ['Bucket' => $bucket]);
+
+        try {
+            $this->execute($command);
+            return true;
+        } catch (S3Exception $e) {
+            if (($accept403 && $e->getStatusCode() === 403)
```
SamRemis, line 275: there should be a newline after the first parenthesis

src/S3/S3ClientTrait.php
```diff
@@ -260,6 +261,30 @@ public function doesBucketExist($bucket)
...
+     * @see S3ClientInterface::doesBucketExistV2()
+     */
+    public function doesBucketExistV2($bucket, $accept403 = false)
+    {
+        $command = $this->getCommand('HeadBucket', ['Bucket' => $bucket]);
+
+        try {
+            $this->execute($command);
+            return true;
+        } catch (S3Exception $e) {
```
SamRemis, line 274: if it's a different kind of exception, it'll just get thrown back to the user, right?

src/S3/S3ClientTrait.php
```diff
@@ -273,6 +298,42 @@ public function doesObjectExist($bucket, $key, array $options = [])
...
+            $response = $e->getResponse();
+
+            if ($includeDeleteMarkers
+                && (!empty($response)
+                    && $response->getHeader('x-amz-delete-marker'))
+            )
+            {
```
SamRemis, line 327: this should be on the line above: `) {`

src/S3/S3ClientTrait.php
```diff
@@ -273,6 +298,42 @@ public function doesObjectExist($bucket, $key, array $options = [])
...
+        );
+
+        try {
+            $this->execute($command);
+            return true;
+        } catch (S3Exception $e) {
+            $response = $e->getResponse();
+
+            if ($includeDeleteMarkers
+                && (!empty($response)
```
SamRemis, line 324: Maybe extract this into a hasDeleteMarker method that checks if a response exists and returns whether or not that header is there?  Condense a a bit of code.

src/S3/S3ClientTrait.php
```diff
@@ -273,6 +298,42 @@ public function doesObjectExist($bucket, $key, array $options = [])
...
+    /**
+     * @see S3ClientInterface::doesObjectExistV2()
+     */
+    public function doesObjectExistV2(
+        $bucket,
+        $key,
+        $includeDeleteMarkers = false,
+        array $options = []
+    )
+    {
```
SamRemis, line 310: should be on the same line as the parenthesis

src/S3/S3ClientTrait.php
stobrien89: yeah, we're only interested in inspecting S3exceptions for true/false conditions and the rest can be thrown back to the user.

src/S3/S3ClientTrait.php
stobrien89: totally agree. just took a second glance and that looks pretty ugly

SamRemis (approved):

SimonMacIntyre:
I'm confused by this changelog, it says `doesBucketExistV2` was deprecated and to use `doesObjectExistV2` instead, is that a mistake?

`doesBucketExistV2` is a brand new method in this PR, so I assume it was not intended to implement a brand new method and deprecate it in the same merge?

image

Issue #2353, opened 2021-11-30
Lots of deprecation messages in symfony 5.4


jokaorgua:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
Lot of deprecation messages when using this SDK with Symfony 5.4

Example
```
Method "ArrayAccess::offsetExists()" might add "bool" as a native return type declaration in the future. Do the same in implementation "Aws\Api\AbstractModel" now to avoid errors or add an explicit @return annotation to suppress this message.

Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "Aws\Endpoint\Partition" now to avoid errors or add an explicit @return annotation to suppress this message.
```

**Version of AWS SDK for PHP?**
v3.206.0

**Version of PHP (`php -v`)?**
8.0.13

**To Reproduce (observed behavior)**
Just run any tests with Symfony 5.4

**Expected behavior**
No deprecation message

XWB:
Same here, PHP 8 and Symfony 5.4:

> Method "ArrayAccess::offsetExists()" might add "bool" as a native return type declaration in the future. Do the same in implementation "Google\Model" now to avoid errors or add an explicit @return annotation to suppress this message.

> Method "Iterator::current()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Google\Collection" now to avoid errors or add an explicit @return annotation to suppress this message.

etc

Not sure where this Google model comes from, is it shipped with Symfony?

jokaorgua:
@XWB I think your errors are from google apis lib, not from the was one
But they are the same. Lack of return types

jokaorgua:
any news here?

ES-Six:
Also tested with php 8.1.1, same reported behaviour, lot of depreciations.

aws/aws-sdk-php : 3.209.1

Here is a sample of the depreciations :

[...]

PR #2361, opened 2021-12-14
Add missing PHPDoc annotations for `ArrayAccess` and `Countable` implementations

ruudk:
While upgrading to Symfony 5.4 I noticed the follow reported deprecations:

* Method `ArrayAccess::offsetExists()` might add `bool` as a native return type declaration in the future. Do the same in implementation `Aws\Api\AbstractModel` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetExists()` might add `bool` as a native return type declaration in the future. Do the same in implementation `Aws\Endpoint\Partition` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetGet()` might add `mixed` as a native return type declaration in the future. Do the same in implementation `Aws\Api\AbstractModel` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetSet()` might add `void` as a native return type declaration in the future. Do the same in implementation `Aws\Api\AbstractModel` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetSet()` might add `void` as a native return type declaration in the future. Do the same in implementation `Aws\Endpoint\Partition` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetUnset()` might add `void` as a native return type declaration in the future. Do the same in implementation `Aws\Api\AbstractModel` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `ArrayAccess::offsetUnset()` might add `void` as a native return type declaration in the future. Do the same in implementation `Aws\Endpoint\Partition` now to avoid errors or add an explicit `@return` annotation to suppress this message.
* Method `Countable::count()` might add `int` as a native return type declaration in the future. Do the same in implementation `Aws\HandlerList` now to avoid errors or add an explicit `@return` annotation to suppress this message.

PR #2434, opened 2022-04-20
Add missing PHPDoc annotation for `PromiseInterface` implementations

ruudk:
*Issue #2353*

*Description of changes:*
This fixes the following deprecation produced by the Symfony PHPUnit Bridge:

```
Method "GuzzleHttp\Promise\PromisorInterface::promise()" might add "PromiseInterface" as a
native return type declaration in the future. Do the same in implementation "Aws\S3\ObjectUploader"
now to avoid errors or add an explicit @return annotation to suppress this message.
```

### DIFF
--- a/src/S3/ObjectCopier.php
+++ b/src/S3/ObjectCopier.php
@@ -6,6 +6,7 @@ use Aws\Arn\S3\AccessPointArn;
 use Aws\Exception\MultipartUploadException;
 use Aws\Result;
 use Aws\S3\Exception\S3Exception;
+use GuzzleHttp\Promise\Coroutine;
 use GuzzleHttp\Promise\PromisorInterface;
 use InvalidArgumentException;
 
@@ -78,7 +79,7 @@ class ObjectCopier implements PromisorInterface
      */
     public function promise()
     {
-        return \GuzzleHttp\Promise\Coroutine::of(function () {
+        return Coroutine::of(function () {
             $headObjectCommand = $this->client->getCommand(
                 'HeadObject',
                 $this->options['params'] + $this->source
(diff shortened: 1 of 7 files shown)


ruudk:
👋  @SamRemis Same kind of PR as last time #2361. Hope you can have a look 😊

ruudk:
@stobrien89 @SamRemis What's holding back merging this PR? Should I make any changes? Please let me know and I'll fix it right away.

src/S3/ObjectCopier.php
```diff
@@ -75,10 +77,12 @@ public function __construct(
      * Perform the configured copy asynchronously. Returns a promise that is
      * fulfilled with the result of the CompleteMultipartUpload or CopyObject
      * operation or rejected with an exception.
+     *
+     * @return Coroutine|PromiseInterface
      */
     public function promise()
     {
-        return \GuzzleHttp\Promise\Coroutine::of(function () {
```
SamRemis, line 81: Is there a reason you removed the first bit of the namespace from the PHPDoc?  I'm not a PHPDoc expert, so maybe I'm missing something, but I'd think that the way they're displayed to a developer varies by IDE.  I'd think it would be best to keep as much info as possible.  

Is it a safe assumption the full namespace always get picked up from the use statement above?

src/Waiter.php
```diff
@@ -86,9 +88,12 @@ public function __construct(
         }
     }
 
+    /**
+     * @return Coroutine|PromiseInterface
```
SamRemis, line 92: Shouldn't this always return a Coroutine?

ruudk: Yes, fixed it.

src/S3/ObjectCopier.php
ruudk: I rebased the PR and did it now in 3 separate commits to better explain why. 

The reason why I cleaned this up is because the FQCN looks bad and are probably added by mistake. 

The rest of the code base imports everything. Let's also do it here.

And yes, it's safe to put short class names in PHPDoc when they are imported at the top.

ruudk:
@SamRemis I rebased the PR and re-arranged it in separate commits to make it more clear. Also answered your question. Please have a look again. 💙

SamRemis:
@ruudk It will be tagged along with our daily release today which should be around 2:00pm EST :) Thank you for your contribution

PR #2435, opened 2022-04-22
Fixed #2412 by making sure that `fclose` in TraceMiddleware is only u…

### LINKED ISSUE #2412: Type Error while using http.debug with a bool value: rewind(): Argument #1 ($stream) must be of type resource, bool given
  
**Describe the bug**
If I'm using the `http.debug` option with the value `true` like this:
```php
$sdk->createFirehose(
    [
        'http' => [
            'debug' => true
        ]
    ]
);
```

I get the error:
> rewind(): Argument #1 ($stream) must be of type resource, bool given

AWS SDK is 3.218.0 and PHP is 8.1.4

I used to send the request with the `http.debug = true` to Firehose

**Expected behavior**
No error

**Screenshots**
image

Dgame:
…sed if there is a valid ressource

*Issue #, if available:*
#2412

*Description of changes:*

### DIFF
--- a/src/TraceMiddleware.php
+++ b/src/TraceMiddleware.php
@@ -286,7 +286,7 @@ class TraceMiddleware
     private function flushHttpDebug(CommandInterface $command)
     {
         if ($res = $command['@http']['debug']) {
-            if (is_ressource($res)) {
+            if (is_resource($res)) {
                 rewind($res);
                 $this->write(stream_get_contents($res));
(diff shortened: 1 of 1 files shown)


stobrien89:
Thanks for the PR @Dgame.  We might want a test on this, but chances are we'll be able to merge this next week without one.

stobrien89 (approved):

Issue #2299, opened 2021-08-26
Copy file with an accesspoint arn does not work


sascha-mummert:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution

**Describe the bug**

I used an ARN to an access point to copy a file within an S3 bucket and this did not work.
It rises the error message "Invalid copy source bucket name".

AWS SDK version: 3.178.1 and 3.191.4
PHP version: 7.4.19

**To Reproduce (observed behavior)**
The following code rises the following exception:
```
Aws\S3\Exception\S3Exception - Exception: Error executing "CopyObject" on "https://YOUR_ACCESS_POINT_NAME-YOUR_ACCOUNT_ID.s3-accesspoint.eu-central-1.amazonaws.com/testDestination/testFile.txt"; AWS HTTP error: Client error: `PUT https://YOUR_ACCESS_POINT_NAME-YOUR_ACCOUNT_ID.s3-accesspoint.eu-central-1.amazonaws.com/testDestination/testFile.txt` resulted in a `400 Bad Request` response:
<Error><Code>InvalidArgument</Code><Message>Invalid copy source bucket name</Message><ArgumentName>x-amz-copy-source</Ar (truncated...)
 InvalidArgument (client): Invalid copy source bucket name - <Error><Code>InvalidArgument</Code><Message>Invalid copy source bucket name</Message><ArgumentName>x-amz-copy-source</ArgumentName><ArgumentValue>/arn:aws:s3:eu-central-1:YOUR_ACCOUNT_ID:accesspoint/YOUR_ACCESS_POINT_NAME/testSource%2FtestFile.txt</ArgumentValue><RequestId>AET833EM26XD5Y3D</RequestId><HostId>Es0IO2drWlYr6Q4qp9GvCviSX2zgwga71qWha3wBV/F7k7G9vBeKGRMzWzUcROYgPfmWEJbaEeU=</HostId></Error> 
[...]

PR #2448, opened 2022-05-09
fix: ObjectCopier does not work when using access-point ARN

yenfryherrerafeliz:
*Issue #2299

*Description of changes:*
When an ObjectCopier instance is used to copy s3 objects, and the object's source is specified using an access-point arn, it fails due that the object's location needs to be prefixed with "/object/", so that the whole path should be something as following "Access-Point-ARN/object/S3-Object-Location". What I did to fix this issue was to add the missing part "/object/" when building the source path in ObjectCopier.php

### DIFF
--- a/src/S3/MultipartCopy.php
+++ b/src/S3/MultipartCopy.php
@@ -147,11 +147,7 @@ class MultipartCopy extends AbstractUploadManager
             );
         }
 
-        $uri = '/';
-        if (ArnParser::isArn($bucket)) {
-            $uri = '';
-        }
-
+        $uri = ArnParser::isArn($bucket) ? '' : '/';
         $uri .= $bucket . '/' . $key;
         $data['CopySource'] = $uri;
         $data['PartNumber'] = $partNumber;
@@ -216,7 +212,6 @@ class MultipartCopy extends AbstractUploadManager
                 'Bucket' => $bucket,
                 'Key' => $key,
             ];
-            var_dump($headParams);
             if (strpos($key, '?')) {
                 list($key, $query) = explode('?', $key, 2);
                 $headParams['Key'] = $key;
@@ -239,11 +234,7 @@ class MultipartCopy extends AbstractUploadManager
      */
     private function getInputSource($inputSource)
     {
-        if (ArnParser::isArn($inputSource)) {
-            $sourceBuilder = '';
-        } else {
-            $sourceBuilder = "/";
-        }
+        $sourceBuilder = ArnParser::isArn($inputSource) ? '' : '/';
         $sourceBuilder .= ltrim(rawurldecode($inputSource), '/');
         return $sourceBuilder;
     }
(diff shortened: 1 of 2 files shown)


src/S3/MultipartCopy.php
```diff
@@ -132,20 +133,28 @@ private function createPart($partNumber, $partsCount)
...
             '/',
             array_map(
                 'urlencode',
                 explode('/', rawurldecode($key))
             )
         );
         }
+
+        $data['CopySource'] = '/';
+        if (ArnParser::isArn($bucket)){
```
SamRemis, line 153: there should be a space after the closing parenthesis before the opening bracket
         if (ArnParser::isArn($bucket)) {

src/S3/MultipartCopy.php
```diff
@@ -132,20 +133,28 @@ private function createPart($partNumber, $partsCount)
...
+            $key = implode(
             '/',
             array_map(
                 'urlencode',
                 explode('/', rawurldecode($key))
             )
         );
         }
+
+        $data['CopySource'] = '/';
```
SamRemis, line 152: We can set this to something else for now- maybe $uri or something like that?  I'm wondering if that would look cleaner than starting the copy source up here.

src/S3/MultipartCopy.php
```diff
@@ -132,20 +133,28 @@ private function createPart($partNumber, $partsCount)
         }
         // The source parameter here is usually a string, but can be overloaded as an array
         // if the key contains a '?' character to specify where the query parameters start
+        $bucket = '';
```
SamRemis, line 136: why did you add these?

tests/S3/ObjectCopierTest.php
```diff
@@ -111,26 +112,6 @@ function (CommandInterface $cmd, RequestInterface $req) {
                         );
                         break;
 
-                    case 'UploadPartCopy':
```
SamRemis, line 114: Why did we delete an entire test case?

src/S3/MultipartCopy.php
```diff
@@ -134,18 +135,25 @@ private function createPart($partNumber, $partsCount)
...
+                '/',
+                array_map(
+                    'urlencode',
+                    explode('/', rawurldecode($key))
+                )
+            );
         }
+
+        $uri = '/';
+        if (ArnParser::isArn($bucket)) {
```
SamRemis, line 151: this could be condensed into one line:
 $uri = ArnParser::isArn($bucket) ? '' : "/";

src/S3/MultipartCopy.php
```diff
@@ -208,6 +216,7 @@ private function fetchSourceMetadata()
                 'Bucket' => $bucket,
                 'Key' => $key,
             ];
+            var_dump($headParams);
```
SamRemis, line 219: This should be deleted

Issue #2237, opened 2021-04-23
Infinite loop in GetLogEvents paginator


pagelypete:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
When using the `GetLogEvents` paginator in the `CloudWatchLogsClient`, the end of the log stream is not properly detected and there is an infinite loop requesting the last page of the log stream.

It seems like a similar bug existed in the Java client - https://github.com/aws/aws-sdk-java-v2/issues/1045

**Version of AWS SDK for PHP?**
v3.178.3

**Version of PHP (`php -v`)?**
7.4.9

**To Reproduce (observed behavior)**

```php
$cwClient = new CloudWatchLogsClient();
$logPaginator = $cwClient->getPaginator("GetLogEvents", [
    "logGroupName" => "/test/group",
    "logStreamName" => "test/stream",
    "startFromHead" => true
]);

foreach ($logPaginator as $logPage) { // Infinite loop here
    foreach ($logPage["events"] ?? [] as $logEvent) {
        echo $logEvent["message"];
    }
}
```

**Expected behavior**
Loop ends when no more events are present.

**Screenshots**
N/A

**Additional context**
N/A

stobrien89:
Hi @pagelypete,

Apologies for the delay here. We've merged a fix for this— it will be available in tomorrow's release.  Let us know if you're still having issues!

PR #2449, opened 2022-05-10
fix: GetLogEvents auto-pagination

stobrien89:
*Issue #, if available:* https://github.com/aws/aws-sdk-php/issues/2237

*Description of changes:*
Fixes a bug caused by atypical pagination behavior. Per the [Cloudwatch Logs documentation](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html): 

> If you have reached the end of the stream, it returns the same token you passed in.

This causes an infinite loop when iterating over pages because the ResultPaginator is unable to detect the end of the stream.  This is the only known API with this behavior. It is also one of two (the other being a Lightsail API) AWS APIs that use a `nextForwardToken` as an `output_token`, so it may be worth checking with the Lightsail team to see if their API's behavior is similar.


src/ResultPaginator.php
```diff
@@ -29,6 +29,10 @@ class ResultPaginator implements \Iterator
     /** @var int Number of operations/requests performed. */
     private $requestCount = 0;
 
+    /** @var array Operations whose last page of results
+     *             returns the previous page's NextToken. */
+    private $lastPageHasPreviousToken = ['GetLogEvents'];
```
SamRemis, line 34: This variable is named like a bool, should be something like $lastPageHasPreviousTokenOperations, but more succinct if you can come up with something

src/ResultPaginator.php
```diff
@@ -127,10 +127,24 @@ public function valid()
         }
 
         if ($this->nextToken || !$this->requestCount) {
+            $lastPageHasPreviousToken = $this->config['output_token'] === 'nextForwardToken';
```
SamRemis, line 130: Change this variable name please

SamRemis (approved):

PR #2454, opened 2022-05-30
nitpicking, remove useless else{}

divinity76:
it's not like we're using some evil __get() magic to actually do something when pkHandle is fetched?
this else looks useless.

*Issue #, if available:*

*Description of changes:*

nitpicking


stobrien89:
Hi @divinity76,

Thanks for the contribution! Nice catch.  LGTM.

Issue #2393, opened 2022-03-02
Realpath cache not cleared correctly for Web Identity token file


MayerMan:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/aws-sdk-php/v3/api/index.html)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-php-sdk) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-php/issues) and didn't find any solution
  
**Describe the bug**
In an AWS EKS cluster we are getting the following error from time to time when building ElasticSearch clients in our code like described in the docs (https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/service_es-data-plane.html). 

`Warning: file_get_contents(/var/run/secrets/eks.amazonaws.com/serviceaccount/token): failed to open stream: No such file or directory`

**Version of AWS SDK for PHP?**
v3.209.30

**Version of PHP (`php -v`)?**
PHP 7.4.3 (cli) (built: Nov 25 2021 23:16:22) ( NTS )

**To Reproduce (observed behavior)**
Imitate the Kubernetes Service Account token update and run a script that periodically tries to get credentials with AssumeRoleWithWebIdentityCredentialProvider

1. initial setup
* save a valid token on disk (token1)
* create a symbolic link which targets that file (token -> token1)
* use the path of the symbolic link as the value for the env variable (AWS_WEB_IDENTITY_TOKEN_FILE=token)
* run a script with a loop that tries to get the credentials and make sure the credentials are returned correctly

2. provoking the error
* make sure the script is running and that credentials are provided correctly
* move token1 to token2
* recreate the symbolic link (token -> token 2)

=> credential providing fails while realpath cache for token still targets token1

**Expected behavior**
If token cannot be read realpath cache should be cleared and another try should be made to read the token.

MayerMan:
To me it looks like #2043 fixed the issue, but #2115 reintroduced it with another change.

yenfryherrerafeliz:
Hi @MayerMan, I will close this issue since the following [PR](https://github.com/aws/aws-sdk-php/pull/2460) is now merged and may fix this issue.

Thanks!

PR #2460, opened 2022-06-07
fix: Warning when symbolic link is pointed to a different path

yenfryherrerafeliz:
*Issue #2393*

*Description of changes:*
AssumeRoleWithWebIdentityCredentialProvider try to load the token from a symbolic link, and when this symbolic link is updated and pointed to a path that is different from the previous one, a warning is generated due that is_readable is used to validate if the file exists, and in the first try it has in cache the old token file location, and therefore file_get_contents generate the warning since the token file in cache does not exist. To fix this behavior I removed the is_readable from the logic and I used file_get_contents supressing the warning, which when the file does not exists it will return false, and when the returned value is false the implementation clear the real path cache and another try is made to read the token.


stobrien89 (approved):

stobrien89:
LGTM :shipit:

PR #2462, opened 2022-06-07
fix: Copy file fails when bucket is an access point arn

yenfryherrerafeliz:
*Issue  2299*

*Description of changes:*
S3 client fails to copy when the source bucket is specified through an access point arn, and this is caused because the path of the object specified should contain "/object/" in between the generated bucket url and the object to be copied, and the current implementation does not have this special handling. To fix this behavior I added a condition in the function that generates the source path, where I validate if is an access point then I add the "/object/" part to the path to be returned.

### DIFF
--- a/tests/S3/ObjectCopierTest.php
+++ b/tests/S3/ObjectCopierTest.php
@@ -78,6 +78,7 @@ class ObjectCopierTest extends TestCase
             function (CommandInterface $cmd, RequestInterface $req) {
 
                 switch($cmd->getName()) {
+                    case 'UploadPartCopy':
                     case 'CopyObject':
                         $this->assertEquals(
                             'mydest-123456789012.s3-accesspoint.us-west-2.amazonaws.com',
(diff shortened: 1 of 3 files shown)


tests/S3/ObjectCopierTest.php
```diff
@@ -111,26 +112,6 @@ function (CommandInterface $cmd, RequestInterface $req) {
                         );
                         break;
 
-                    case 'UploadPartCopy':
```
stobrien89, line 114: Why are we removing this case?

yenfryherrerafeliz: Technically I am not removing this case, instead I am just combining it with the "CopyObject" case. The reason for that is because they both do the same.

stobrien89: Ah, I skimmed over lines 84-88 because they were collapsed.  That makes sense to me

stobrien89 (approved):

stobrien89:
Look good to me.  Just needs a rebase to be run against the current CI.

SamRemis (approved):

stobrien89 (approved):

Issue #2445, opened 2022-05-08
S3 client believes every object exists.


manarth:
### Describe the bug

When invoking 'doesObjectExist()` with the S3 client, the API returns TRUE for non-existing objects.

### Expected Behavior

The API should return TRUE if the specified bucket and key exists.
The API should return FALSE if the specified bucket does not exist, or if the specified bucket and key do not exist.

### Current Behavior

The API always returns true.

### Reproduction Steps

```php
$params = [
  'Bucket' => 'my-bucket-which-exists',
  // Non-existent key name.
  'Key' => 'sdsgfdsgregerdfdgdd',
];
$result = $s3Client->doesObjectExist($params['Bucket'], $params['Key']);
```

### Possible Solution

The method `checkExistenceWithCommand()` in the `S3ClientTrait` executes a defined command.
If the command does not throw an exception, TRUE is returned.

```php
    /**
     * Determines whether or not a resource exists using a command
     *
     * @param CommandInterface $command Command used to poll for the resource
     *
     * @return bool
     * @throws S3Exception|\Exception if there is an unhandled exception
     */
    private function checkExistenceWithCommand(CommandInterface $command)
    {
        try {
            $result = $this->execute($command);
            return true;
        } catch (S3Exception $e) {
            if ($e->getAwsErrorCode() == 'AccessDenied') {
                return true;
            }
            if ($e->getStatusCode() >= 500) {
                throw $e;
            }
            return false;
        }
    }
```

The `doesObjectExist()` method uses a `HeadObject` command to check if the bucket exists.

```php
    /**
     * @see S3ClientInterface::doesObjectExist()
     */
    public function doesObjectExist($bucket, $key, array $options = [])
    {
        return $this->checkExistenceWithCommand(
            $this->getCommand('HeadObject', [
                    'Bucket' => $bucket,
                    'Key'    => $key
                ] + $options)
        );
    }
```

When executing a `HeadObject` command for a non-existing key, the command does not throw an Exception.
Instead, it returns the `404` response as an `Aws\Result` object.

```
class Aws\Result#2316 (2) {
  private $data =>
  array(1) {
    '@metadata' =>
    array(4) {
      'statusCode' =>
      int(404)
      'effectiveUri' =>
      string(55) "https://<redacted>.s3.amazonaws.com/sdsgfdsgregerdfdgdd"
      'headers' =>
      array(5) {
        ...
      }
      'transferStats' =>
      array(1) {
        ...
      }
    }
  }
[...]

PR #2463, opened 2022-06-08
feat: add streamwrapper support for new object and bucket existence methods

stobrien89:
*Issue #, if available:*
#2445

*Description of changes:*
Adds support for `doesObjectExistV2` and `doesBucketExistV2` within the S3 streamwrapper. 

### DIFF
--- a/src/S3/S3ClientTrait.php
+++ b/src/S3/S3ClientTrait.php
@@ -4,6 +4,7 @@ namespace Aws\S3;
 use Aws\Api\Parser\PayloadParserTrait;
 use Aws\CommandInterface;
 use Aws\Exception\AwsException;
+use Aws\CacheInterface;
 use Aws\HandlerList;
 use Aws\ResultInterface;
 use Aws\S3\Exception\PermanentRedirectException;
@@ -94,9 +95,14 @@ trait S3ClientTrait
     /**
      * @see S3ClientInterface::registerStreamWrapper()
      */
-    public function registerStreamWrapper()
+    public function registerStreamWrapper($useV2Existence = null)
     {
-        StreamWrapper::register($this);
+        StreamWrapper::register(
+            $this,
+            's3',
+            null,
+            $useV2Existence
+        );
     }
 
     /**
--- a/src/S3/StreamWrapper.php
+++ b/src/S3/StreamWrapper.php
@@ -8,6 +8,7 @@ use Aws\S3\Exception\S3Exception;
 use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\Stream;
 use GuzzleHttp\Psr7\CachingStream;
+use phpDocumentor\Reflection\Types\Boolean;
 use Psr\Http\Message\StreamInterface;
 
 /**
@@ -97,6 +98,9 @@ class StreamWrapper
     /** @var bool Keeps track of whether stream has been flushed since opening */
     private $isFlushed = false;
 
+    /** @var bool Whether or not to use V2 bucket and object existence methods */
+    private static $useV2Existence = false;
+
     /**
      * Register the 's3://' stream wrapper
      *
@@ -107,8 +111,10 @@ class StreamWrapper
     public static function register(
         S3ClientInterface $client,
         $protocol = 's3',
-        CacheInterface $cache = null
+        CacheInterface $cache = null,
+        $V2Existence = false
     ) {
+        self::$useV2Existence = $V2Existence;
         if (in_array($protocol, stream_get_wrappers())) {
             stream_wrapper_unregister($protocol);
         }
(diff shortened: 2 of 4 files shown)


src/S3/S3ClientTrait.php
```diff
@@ -94,9 +94,14 @@ public function copyAsync(
     /**
      * @see S3ClientInterface::registerStreamWrapper()
      */
-    public function registerStreamWrapper()
+    public function registerStreamWrapper($useV2Existence = false)
     {
-        StreamWrapper::register($this);
+        StreamWrapper::register(
```
SamRemis, line 99: Should we not add the variable on s3 clients so we don't have to change this call?

src/S3/StreamWrapper.php
```diff
@@ -323,10 +328,18 @@ private function createStat($path, $flags)
     private function statDirectory($parts, $path, $flags)
     {
         // Stat "directories": buckets, or "s3://"
-        if (!$parts['Bucket'] ||
-            $this->getClient()->doesBucketExist($parts['Bucket'])
-        ) {
-            return $this->formatUrlStat($path);
+        if (self::$useV2Existence) {
```
SamRemis, line 331: you can restructure this if you want.  
```
if (!parts['Bucket']) {
     return $this->formatUrlStat($path); 
}   else if (self::$useV2Existence
      && $this->getClient()->doesBucketExist($parts['Bucket']
      ) {
     return $this->formatUrlStat($path); 
}  else if (!self::$useV2Existence
      && $this->getClient()->doesBucketExist($parts['Bucket']
     )  {
     return $this->formatUrlStat($path); 
      } 
}

```
Not saying this is better or worse, just offering another option where we don't duplicate the bucket check 
You could also extract calling the legacy or V2 method into another method:

```
private function checkBucketExistance($bucket) {
if (self::useV2) {
      return $this->getClient()->doesBucketExistV2($bucket)
      }
return $this->getClient()->doesBucketExist($bucket)
}
```

src/S3/StreamWrapper.php
```diff
@@ -592,16 +605,24 @@ private function validate($path, $mode)
...
-        if ($mode == 'x' &&
-            $this->getClient()->doesObjectExist(
-                $this->getOption('Bucket'),
-                $this->getOption('Key'),
-                $this->getOptions(true)
-            )
-        ) {
-            $errors[] = "{$path} already exists on Amazon S3";
+        if ($mode === 'x') {
+            if (self::$useV2Existence) {
```
SamRemis, line 609: Same as above, there's a way to reformat this.  You can also have a string as the variable method, and have it be 
```
$method = self::$useV2Existence ? "doesObjectExistV2" : "doesObjectExist";
$this->getClient()->$method();
```

Look at example 2 [here](https://www.php.net/manual/en/functions.variable-functions.php).

There is a precedent for this being used in our code, and if you choose this way, you can make the checkBucketExistance method I mentioned above into just checkExistance, which takes in $type (bucket or object) and $name (for instance, bucket name).

src/S3/S3ClientTrait.php
stobrien89: What do you think about adding a registerStreamWrapperV2) method instead?  I thought about adding it to the S3 client, but this is the only customization that uses object and bucket existence checks.

src/S3/StreamWrapper.php
```diff
@@ -107,8 +111,10 @@ class StreamWrapper
     public static function register(
         S3ClientInterface $client,
         $protocol = 's3',
-        CacheInterface $cache = null
+        CacheInterface $cache = null,
+        $V2Existence = false
```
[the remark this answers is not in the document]
stobrien89: yep, for some reason `v2` didn't come naturally here. thanks

src/S3/StreamWrapper.php
stobrien89: I like this idea better.  Much cleaner than I imagined it

src/S3/StreamWrapper.php
```diff
@@ -830,10 +839,12 @@ private function createSubfolder($path, array $params)
         $params['Body'] = '';
 
         // Fail if this pseudo directory key already exists
-        if ($this->getClient()->doesObjectExist(
+        $method = self::$useV2Existence ? 'doesObjectExistV2' : 'doesObjectExist';
```
SamRemis, line 842: Since each of these are duplicated once, you could have a method below that is checkObjectExists() and that makes the two calls below; not a huge deal

```
$this->checkObjectExist($params['Bucket'], $params['Key']);

private function checkObjectExists($bucket, $key) {
      $method = self::$useV2Existence ? 'doesObjectExistV2' : 'doesObjectExist';
      $this->getClient()->method($bucket, $key);
}
```

Something like that.  Just so modifying it in the future is easier.

SamRemis (approved):

PR #2468, opened 2022-06-10
chore: bump guzzle 

stobrien89:
*Description of changes:*
Bumping Guzzle version per https://github.com/guzzle/guzzle/security/advisories/GHSA-f2wf-25xc-69c9

### DIFF
--- a/.changes/nextrelease/bump-guzzle.json
+++ b/.changes/nextrelease/bump-guzzle.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "",
+        "description": "Bump Guzzle 6 and 7 due to security advisory"
+    }
+]
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require": {
         "php": ">=5.5",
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
This change doesn't make sense. You still allow Guzzle 5, which contains the vulnerability.

GrahamCampbell:
Similarly, the version contains of `guzzlehttp/psr7` also allow resolution of vulnerable versions of that package. See https://github.com/guzzle/psr7/security/advisories/GHSA-q7rv-6hp3-vh96. I'd actually recommend one version later than the one that was strictly lowest, as the security fix introduced a bug that was later fixed. `^1.8.5 || ^2.3` is what I would recommend, if you want to bump these dependencies. Just remember you have a trade off that composer may just resolve an older version of the aws sdk before you bumped the version, to enable installation of vulnerable versions of guzzle, which could be worse.

stobrien89:
Hey @GrahamCampbell,

Thanks for reaching out!  We were advised by our security team to make this change and we really had no choice other than to allow Guzzle 5 for the time being to avoid potentially breaking some customers.  We'll likely change this when we deprecate older versions of PHP.  

Thank you for the tip on `guzzlehttp/psr7`.  We'll take this into consideration.

Issue #2481, opened 2022-06-27
Deprecated use of strlen() with PHP 8.1.7


benshawuk:
### Describe the bug

When creating a new S3Client object using PHP 8.1.7, I get this error:
strlen(): Passing null to parameter #1 ($string) of type string is deprecated in file vendor/aws/aws-sdk-php/src/S3/S3Client.php on line 432

### Expected Behavior

Error does not occur

### Current Behavior

strlen(): Passing null to parameter #1 ($string) of type string is deprecated in file vendor/aws/aws-sdk-php/src/S3/S3Client.php on line 432

### Reproduction Steps

$s3 = new \Aws\S3\S3Client([
            'version' => 'latest',
            'region'  => REGION,
            'debug'   => false,
            'credentials' => [
                'key'    => ACCESS_KEY,
                'secret' => SECRET
            ],
        ]);

        print_r($s3->listBuckets());

### Possible Solution

//$bucketLen = strlen($bucket);
$bucketLen = 0;

This temporarily resolves the issue, but I don't understand what the function is actually doing.
I'm guessing it needs to not get "null" passed to strlen function.

### SDK version used

3.228

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.1.7, MacOS12.4

yenfryherrerafeliz:
Hi @benshawuk, thanks for opening this issue. 

We do not support PHP 8.1.7 yet, nonetheless I expect to post a PR to prevent this warning.

Thanks!

PR #2403, opened 2022-03-18
PHP 8.0, 8.1, PHPUnit 9 support + automated check for php compatibility

Zombaya:
I've expanded on the excellent work done by @27pchrisl  in PR #2332.  I rebased his branch onto the latest version of master, upgraded the new tests to be able to use phpunit 9 and added a new CI-job that lints all code in `src` to make sure it is compatible with all supported versions of php (5.5-8.1) using [phpcompatibility](https://github.com/PHPCompatibility/PHPCompatibility).

This should add support for php 8.0 and php 8.1 to the project and make sure no code will trigger bugs because of changes in php.

This should probably fix #2378, #2367

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2483, opened 2022-06-28
fix: Warning when we pass null to strlen from isBucketDnsCompatible

yenfryherrerafeliz:
*Issue #2481*

*Description of changes:*
 When using PHP 8.x and we pass null to S3Client::isBucketDnsCompatible, we get a warning stating that is deprecated to pass a parameter as null when a type is defined for it, so that to prevent this warning I add a logic that validates if the variable is null, then, an empty string is given to strlen().


tests/S3/S3ClientTest.php
```diff
@@ -79,6 +79,19 @@ public function testValidatesDnsBucketNames($bucket, $valid)
         $this->assertEquals($valid, S3Client::isBucketDnsCompatible($bucket));
     }
 
+    public function testValidateDnsBucketNamesPassingNull()
```
SamRemis, line 82: This test passes with or without the above code update

yenfryherrerafeliz: Which PHP version did you test with?

SamRemis: Though personally I'd say this doesn't necessarily need a test; the code is pretty self explanatory- anyone with passing knowledge of PHP 8 will understand why we put the `?:` there and know not to remove it.  The functionality hasn't changed, so a unit test isn't necessarily doing anything here. That said, this change is included in this PR https://github.com/aws/aws-sdk-php/pull/2403/files, so do we want to merge it?

SamRemis:
Fixed by https://github.com/aws/aws-sdk-php/commit/af584e7188dba00d76b7ec2306ac69efa2ad8483

Issue #2492, opened 2022-07-08
Remove Unused AWS services: Cannot redeclare `Aws\constantly() previously declared in aws-sdk-php/src/functions.php `


kayw-geek:
### Describe the bug

When I used the new feature “Remove unused AWS services”, I got a PHP fatal error that “PHP Fatal error:  Cannot redeclare `Aws\constantly()` (previously declared in `vendor/aws/aws-sdk-php/src/functions.php:20`)” and it printed twice.
I wonder if we need to regenerate autoload files in the static method `removeUnusedServices()`

### Expected Behavior

No error

### Current Behavior

Cannot redeclare `Aws\constantly()`

### Reproduction Steps

Put the “Aws\\Script\\Composer\\Composer::removeUnusedServices” to Composer event `pre-autoload-dump` 

### SDK version used

3.231.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

8.1.7

stobrien89:
Hi @kayw-geek,

Thanks for the report and sorry about the issues.  Could you provide some reproduction steps?

kayw-geek:
> 
I tried in locally  and CI, I got same results.
 I used Laravel framework, so my composer script have executed the`Illuminate\\Foundation\\ComposerScripts::postAutoloadDump` static method on post-autoload-dump event, it will require autoload file.

kayw-geek:
https://github.com/aws/aws-sdk-php/pull/2495 The PR will solve this problem

stobrien89:
Thanks again for reporting and the comments on #2495 @kayw-geek. The fix will be available in today's release!

PR #2495, opened 2022-07-09
bugfix: fix function redeclaration in removeunusedservices

stobrien89:
*Issue #, if available:*
#2492

*Description of changes:*
Addresses a conflict with autoload files/imports when script is invoked.

### DIFF
--- a/src/Script/Composer/Composer.php
+++ b/src/Script/Composer/Composer.php
@@ -1,9 +1,6 @@
 <?php
 namespace Aws\Script\Composer;
 
-require_once __DIR__ . '/../../functions.php';
-
-use Aws;
 use Composer\Script\Event;
 use Symfony\Component\Filesystem\Filesystem;
 
@@ -42,7 +39,8 @@ class Composer
     public static function buildServiceMapping()
     {
         $serviceMapping = [];
-        $source = Aws\manifest();
+        $manifest = __DIR__ . '/../../data/manifest.json.php';
+        $source = include($manifest);
 
         foreach ($source as $key => $value) {
             $serviceMapping[$value['namespace']] = $key;
--- a/src/Script/Composer/Composer.php
+++ b/src/Script/Composer/Composer.php
@@ -39,11 +39,10 @@ class Composer
     public static function buildServiceMapping()
     {
         $serviceMapping = [];
-        $manifest = __DIR__ . '/../../data/manifest.json.php';
-        $source = include($manifest);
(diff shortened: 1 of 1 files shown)


kayw-geek:
> _Issue #, if available:_ #2492
> 
> _Description of changes:_ Addresses a conflict with autoload files/imports when script is invoked.
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

I find the reason for the problem a long time
I already tried your solution, It has solved my problem, thank you!

src/Script/Composer/Composer.php
```diff
@@ -42,7 +39,8 @@ public static function removeUnusedServices(
     public static function buildServiceMapping()
     {
         $serviceMapping = [];
-        $source = Aws\manifest();
+        $manifest = __DIR__ . '/../../data/manifest.json.php';
```
kayw-geek, line 42: Do we need the definition of the variable `$mainfest`?

src/Script/Composer/Composer.php
```diff
@@ -42,7 +39,8 @@ public static function removeUnusedServices(
...
+        $source = include($manifest);
```
kayw-geek, line 43: I think when the `manifest.json.php` file is not found, more should be used `require()`

stobrien89: Agree that this should throw an error when the file isn't found.  Thanks!

src/Script/Composer/Composer.php
stobrien89: I threw this PR up pretty quickly and (without checking) thought the combination of the two lines might throw us over the PSR-recommended line length.  Looks like it's well below the max recommended line length (63 char).  Nice catch

src/Script/Composer/Composer.php
```diff
@@ -92,7 +89,7 @@ private static function removeServiceDirs(
             }
         }
         $event->getIO()->write(
-            "Removed $deleteCount AWS services"
+            "Removed $deleteCount AWS service" . ($deleteCount === 1 ? '' : 's')
```
SamRemis, line 92: I think this is simpler, but [ngettext](https://www.php.net/manual/en/function.ngettext.php) is a thing in PHP.  I like this better, but I guess ngettext is more idiomatic?

SamRemis (approved):

src/Script/Composer/Composer.php
kayw-geek: The `$deleteCount` probably will be `0`

Issue #2451, opened 2022-05-19
Custom blacklisted headers on SignatureV4


stephpy:
### Describe the feature

**BACKWARD COMPATIBLE**

Hi,

At this moment, SignatureV4 use a defined list of blacklisted headers. [See Method](https://github.com/aws/aws-sdk-php/blob/master/src/Signature/SignatureV4.php#L40-L74) and it cannot be overriden by client.

I can understand in most of case we have not to sign theses headers but in my case I need to sign them.

Would you accept a change in this way ? I would make the PR.

### Use Case

We develop an API (Symfony) + a FRONTEND (React) and for large files we decided than FRONTEND will directly push files to S3.

Workflow:
   - FRONTEND asks for a presigned url to API with some data like content_type, content_length ...
   - API validate the request and if user has rights to upload this kind of media
   - If it's ok, API generate a presigned url and give it to FRONTEND
   - FRONTEND uses this presigned to upload the media

If we don't sign headers like content-type, content-length, FRONTEND could push any other media ....

### Proposed Solution

```
<?php

class SignatureV4 implements SignatureInterface
{
   // ....

    /**
     * The following headers are not signed because signing these headers
     * would potentially cause a signature mismatch when sending a request
     * through a proxy or if modified at the HTTP client level.
     *
     * @var array
     */
    private array $blacklistedHeaders = [
        'cache-control' => true,
        'content-type' => true,
        'content-length' => true,
        'expect' => true,
        'max-forwards' => true,
        'pragma' => true,
        'range' => true,
        'te' => true,
        'if-match' => true,
        'if-none-match' => true,
        'if-modified-since' => true,
        'if-unmodified-since' => true,
        'if-range' => true,
        'accept' => true,
        'authorization' => true,
        'proxy-authorization' => true,
        'from' => true,
        'referer' => true,
        'user-agent' => true,
        'X-Amz-User-Agent' => true,
        'x-amzn-trace-id' => true,
        'aws-sdk-invocation-id' => true,
        'aws-sdk-retry' => true,
    ];

   // .....

    private function getHeaderBlacklist()
    {
        return $this->blacklistedHeaders;
    }

    public function allowHeader($header)
    {
        unset($this->blacklistedHeaders[$header]);
    }

    public function disallowHeader($header)
    {
        $this->blacklistedHeaders[$header] = true;
    }
```

Then

```
<?php

        $signer->allowHeader('content-type');
        $signer->allowHeader('content-length');

```

[...]

PR #2505, opened 2022-08-08
Update SignatureV4.php

stephpy:
# Issue: https://github.com/aws/aws-sdk-php/issues/2451

BC

We can now override header black list by doing:

```
protected function getAwsClient(): S3Client
    {
        $options = [
           // ....
            'signature_provider' => function ($version, $service, $region) {
                if ($version === 's3v4') {
                   // Custom signature class ...
                    return new S3SignatureV4($service, $region);
                }
                  //...
            }
        ];

         return new S3Client($options);
```

Then

```

<?php

namespace ACME;

use Aws\Signature\S3SignatureV4 as BaseS3SignatureV4;

class S3SignatureV4 extends BaseS3SignatureV4
{
    protected function getHeaderBlacklist()
    {
        $list = parent::getHeaderBlacklist();

        unset(
            $list['content-type'],
            $list['content-length']
        );

        return $list;
    }
}
```


jaxwilko:
Hi, is there any update on this? The functionality it provides is really useful to our team :)

SamRemis:
Hi there, 
Sorry for the delay on this.  Can you add a changelog entry following the instructions in the [contributing.md](https://github.com/aws/aws-sdk-php/blob/master/CONTRIBUTING.md) file?  I'm happy to merge this once that's done.

I definitely suggest being careful with this since it may lead to signature mismatch errors if used incorrectly.  Do you think we should add a code comment above the method saying that as a warning?  I'm leaning toward no, but would like another opinion.

LukeTowers:
@SamRemis if you have a moment could you explain how it could lead to signature mismatch errors? That would probably be more helpful than a generic warning message. If we can add a comment to the docblock for the method describing what to watch out for / avoid then that could be helpful otherwise a generic message probably isn't needed here.

SamRemis:
Absolutely.  The reason that list is there to begin with is that S3 checks a hash of the signed headers on the headers that we send.  It is unlikely that it would cause an issue if you add to the header blacklist, but if you remove something from it that is altered down the line, then that would cause the hash to be off.

So for instance if we took the above code and altered it to be `user-agent` that's removed:
```
    protected function getHeaderBlacklist()
    {
        $list = parent::getHeaderBlacklist();
        unset(
            $list['user-agent'],
        );
        return $list;
    }
```

It's not unheard of for customers to modify the user-agent header themselves.  In the case they do, we have `user-agent` signed with one value, then S3 receives a different value as the header.  When they calculate the hash, it'll be different and S3 will throw an exception.

I'm not sure if that's too specific of a use case to be worth throwing a comment in the code.  [The comment that is already there](https://github.com/aws/aws-sdk-php/blob/0293e7b4684827688dba3f3b58d99778a99156ae/src/Signature/SignatureV4.php#L40) should probably suffice.

LukeTowers:
@SamRemis ah, makes sense. To be honest, that is actually exactly the behaviour that I want to happen with my overridden use of that method.

The reason I want to remove Content-Length (& potentially Content-Type) from the blacklist is so that my validation logic when generating the signed URL will actually be guaranteed to work. Right now, the client sends a request to my service to generate the signed URL, and my service will validate the Content-Type & Content-Length the client provides before generating that signed URL. However, with the current logic, my service is unable to include those validated headers in the signed URL and effectively issues the client a blank check for when it actually comes time to send an upload to S3.

For example, consider the following situation:
- MyService wishes to enforce a 10MB upload limit on all uploads sent to S3
- To do this, it asks the client to send the size of the file being uploaded to the service when requesting the signed URL
- It then validates that the size it receives is within the limit and generates a signed URL
- The client then uploads their file to this signed URL. Crucially however, there is nothing stopping the client from lying to the service and sending a length in the initial request that passes the validation but then sending the actual file to S3 with a real size that far exceeds the service determined limit.

With the current implementation I am unable to sign the original validated headers provided by the client that passed my validation and so AWS happily accepts whatever the client provides it to the signed URL. With this change I will be able to override the signer so that I can sign the original headers that passed my validation and if the client attempts to send data that conflicts with those original headers to the actual URL I can rely on S3 preventing that action from occurring by throwing an exception.

stephpy:
Just added a changelog file. I hope my 🥖 english level is understandable. :) 

🍻

.changes/3.234.6
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "Signature",
+        "description": "Allow Override of SignatureV4::getHeaderBlacklist by changing private visibility to protected. Overriding this method is only for specific purposes, you can see usecases on issue #2451 on github. By default, headers are not signed because signing these headers would potentially cause a signature mismatch when sending a request through a proxy or if modified at the HTTP client level.have to be careful with Documentation updates for Database Migration Service (DMS)."
```
LukeTowers, line 5:
```suggestion
        "description": "The visibility of the SignatureV4::getHeaderBlacklist() method has been changed from private to protected to allow overriding the method in extending classes. By default this method prevents headers from being signed to avoid issues with signature mismatches when sending requests through proxies; however you may want to be able to force some headers to be signed for request integrity reasons."
```

stephpy:
Thanks @LukeTowers

SamRemis:
Sorry, you added directly to one of the changelog files; instead, you should make a folder called 'nextrelease' under the .changes folder, then put a file in it named something like `enhancement-header-blacklist-visibility.json` and have it have contents like 

```
    {
        "type": "enhancement", 
        "category": "Signature",
        "description": "The visibility of the SignatureV4::getHeaderBlacklist() method has been changed from private to protected to allow overriding the method in extending classes."
    }
```

SamRemis:
Do you mind changing it so that it gets ingested and added to the changelog properly?

SamRemis:
@LukeTowers, since you mentioned `content-length` specifically, I'd like to mention why it's on the blacklist as a heads up.  There are some load balancers that can strip the `0` from the header which will cause a signature mismatch.  For example, if a request without a body has a signed header of `content-length: 0`, that header will be modified by the load balancer to `content-length:`.

So if you wind up running into that issue, that's why.

LukeTowers:
@SamRemis thanks, I appreciate the heads up! Shouldn't be an issue for our use case though since we won't be submitting uploads with a content length of 0.

SamRemis:
@stephpy thanks for your contribution :) I squashed them, and this should be available via composer after the release tomorrow, which is usually done by 3pm EST

LukeTowers:
Woot woot! Thanks @SamRemis :)

stephpy:
Thanks :)

stof:
Too bad that the method was not renamed to move away from the `blacklist` naming _before_ opening it to inheritance...

SamRemis:
@stobrien89, I think that would be alright, do you agree?  If we are worried about it, we could also keep the blacklist method and just make a method called `getHeaderIgnoreList()` that would just call the `getHeaderBlacklist()` method and mark the `getHeaderBlacklist()` as deprecated.

stobrien89:
I think we'd be better off deprecating `getHeaderBlacklist` and introducing `getHeaderIgnoreList` or similar.  We'll accept a PR for this @stof, @LukeTowers.

LukeTowers:
I'm absolutely swamped at the moment, so feel free to submit the PR if you'd like to see that change @stof

PR #2507, opened 2022-08-12
Add awsQueryCompatible trait support to service

ziyanli-amazon:
*Issue #, if available:*

*Description of changes:*
When awsQueryCompatible trait is found, it's made available as an API property. When this property is found,
the error code is returned by looking up the mapping. This is a pre-requisite for migrating services from
AWSQuery wire protocol to AWSJson.


ziyanli-amazon:
The original plan was abandoned and new plan will take effect in another PR

PR #2518, opened 2022-09-08
Fix self callables

driesvints:
These changes are necessary to solve some deprecation notices for PHP 8.2 where `self::` is used in a callable context. This isn't supported anymore. I didn't add any tests because this should technically get a new 8.2 build. However, the 8.0 and 8.1 builds are also still pending so I figured it didn't make sense to add an 8.2 build before that PR was merged: https://github.com/aws/aws-sdk-php/pull/2403

Here's the upgrade notice: https://github.com/php/php-src/blob/PHP-8.2/UPGRADING#L157-L167

Right now, we're trying to get a passing PHP 8.2 build for Laravel. This library is one of the last blockers to get a passing build. See failures in the Laravel build:

Screenshot 2022-09-08 at 14 34 49

It would be great to get a merge and a release for this @SamRemis. Thanks!

### DIFF
--- a/src/ClientSideMonitoring/ConfigurationProvider.php
+++ b/src/ClientSideMonitoring/ConfigurationProvider.php
@@ -87,7 +87,7 @@ class ConfigurationProvider extends AbstractConfigurationProvider
         $configProviders[] = self::fallback();
 
         $memo = self::memoize(
-            call_user_func_array('self::chain', $configProviders)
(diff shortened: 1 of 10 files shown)


stobrien89:
Hi @driesvints, 

Sorry for the delay. And thanks for the contribution.  LGTM 🚢 .  Will go out in today's release.

PR #2516, opened 2022-09-05
fix source_profile handling

tchernomax:
"$profiles" keys are prefix with "profile " , but source_profile attribut isn't so the SDK never found the source_profile.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2519, opened 2022-09-11
fix source_profile handling

tchernomax:
"$profiles" keys are [prefix with "profile "](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html), but source_profile attribute isn't so the SDK never found the source_profile.

# how to reproduce

~/.aws/config
```
[profile source]
…

[profile role]
role_arn = arn:aws:iam::…:role/…
source_profile = source
```

test.php
```
<?php

use Aws\Sts\StsClient;

$client = new StsClient([
        'region' => 'eu-west-1',
        'version' => '2011-06-15'
]);

$result = $client->getCallerIdentity([]);

print($result);
```

```
export AWS_PROFILE=role
php test.php
```

result: fail  
expected result: ok

But if you use this ~/.aws/config (which is ill formatted according to [the doc](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html)), it works:
```
[profile source]
…

[profile role]
role_arn = arn:aws:iam::…:role/…
source_profile = profile source
```

# note

follow https://github.com/aws/aws-sdk-php/pull/2516 , which is already closed

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -705,6 +705,12 @@ class CredentialProvider
         $sourceProfileName = "";
         if (!empty($roleProfile['source_profile'])) {
             $sourceProfileName = $roleProfile['source_profile'];
+            # in ~/.aws/config all the named profile (except 'default') are
+            # prefix with 'profile '
+            if ($filename == (self::getHomeDir() . '/.aws/config') &&
+                $sourceProfileName != 'default') {
+                $sourceProfileName = 'profile ' . $sourceProfileName;
+            }
             if (!isset($profiles[$sourceProfileName])) {
                 return self::reject("source_profile " . $sourceProfileName
                     . " using profile " . $profileName . " does not exist"


stobrien89 (changes requested):
Hi @tchernomax,

Thanks for the follow-up.  Good catch- this change makes sense to me.  We'd be glad to accept this, but will need the failing test to be adjusted to account for the new (correct) behavior.  You're welcome to do this.  Otherwise, we should be able to carve out some time to address this next week.  Thanks!

tchernomax:
@stobrien89 

> We'd be glad to accept this, but will need the failing test to be adjusted to account for the new (correct) behavior.

In fact I realize the same code handle the `~/.aws/credentials` file, which has other rules: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_profiles.html#assume-role-with-profile

I changed my commit adding check to see if the file is `~/.aws/config` or `~/.aws/credentials` and force-pushed it.

I **think** the tests should be ok now (I didn't run them locally), can you relaunch the CI ?

Thanks

src/Credentials/CredentialProvider.php
```diff
@@ -705,6 +705,12 @@ private static function loadRoleProfile(
         $sourceProfileName = "";
         if (!empty($roleProfile['source_profile'])) {
             $sourceProfileName = $roleProfile['source_profile'];
+            # in ~/.aws/config all the named profile (except 'default') are
+            # prefix with 'profile '
+            if ($filename == (self::getHomeDir() . '/.aws/config') &&
+                $sourceProfileName != 'default') {
+                $sourceProfileName = 'profile ' . $sourceProfileName;
```
yenfryherrerafeliz, line 712: Would you please be able to add a check here for whether `profile ` has been already  appended to the profile. By doing this we keep backward compatibility for customers that are already adding the `profile ` to profile names.

src/Credentials/CredentialProvider.php
```diff
@@ -705,6 +705,12 @@ private static function loadRoleProfile(
         $sourceProfileName = "";
         if (!empty($roleProfile['source_profile'])) {
             $sourceProfileName = $roleProfile['source_profile'];
+            # in ~/.aws/config all the named profile (except 'default') are
+            # prefix with 'profile '
+            if ($filename == (self::getHomeDir() . '/.aws/config') &&
```
yenfryherrerafeliz, line 710: Can you please remove the condition `$filename == (self::getHomeDir() . '/.aws/config')` since I do no think is needed. You can leave the other condition `$sourceProfileName != 'default'`.

yenfryherrerafeliz (changes requested):

stobrien89:
Hi @tchernomax,

I reviewed this today and this behavior has been fixed in the `loadProfiles()` helper method.  Closing, but please let us know if you have any questions.

PR #2523, opened 2022-09-19
feat: Add EndpointV2 provider

stobrien89:
*Description of changes:*

Implements a new endpoint provider. The Endpoints 2.0 initiative across the AWS SDK organization is changing the way service endpoints are determined: Each service will provide an endpoint `ruleset` file: a collection of rules that, based on the presence of certain parameter values, will determine the endpoint to be used (or error to be thrown) for a given request.

 The new endpoint provider is responsible for reading a service's endpoint ruleset and determining an endpoint (or error) to return by comparing rules contained within the ruleset against parameters— both values provided by the client and values that are scoped during rule evaluation. 


src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,120 @@
...
+        foreach($this->rules as $rule) {
+            // newly set parameters via "assign" cannot be shared between
+            // adjacent rules
+            $inputParametersCopy = $inputParameters;
+            $evaluation = $rule->evaluate($inputParametersCopy, $this->standardLibrary);
+            if (!is_null($evaluation)) {
+                return $evaluation;
+            }
+        }
+        return null;
```
SamRemis, line 76: Consider changing all of these to false so we can do a strict evaluation check to know for sure if an endpoint rule has been run yet; otherwise it could wind up being null before it's invoked which seems more error prone

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
```diff
@@ -0,0 +1,431 @@
...
+    public function getAttr($from, $path)
+    {
+        $parts = explode('.', $path);
+        foreach ($parts as $part) {
+            $sliceIdx = strpos($part, '[');
+            if ($sliceIdx !== false) {
+                if (substr($part, -1) !== ']') {
+                    return null;
+                }
+                # take the entire slice except for the last character (which is `]`)
```
SamRemis, line 73: Why use # for comments here instead of //?

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
```diff
@@ -0,0 +1,431 @@
...
+    public function parseArn($arnString)
+    {
+        if (is_null($arnString)
+            || substr( $arnString, 0, 3 ) !== "arn"
+        ) {
+            return null;
+        }
+
+            $arn = [];
+        $parts = explode(':', $arnString, 6);
```
SamRemis, line 235: Is this directly from the SEP?  I feel like this would break in something like an IAM arn where there's two colons in a row: `arn:aws:iam::123456789012:user`.

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
stobrien89: idk, that was your comment 😛 .  Forgot to remove it though

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
stobrien89: This isn't explicitly called out in the spec, but from what I'm seeing IAM will accept the standard format as well. The current formats we allow in the arninterface are: 

`arn:partition:service:region:account-id:resource-id`
`arn:partition:service:region:account-id:resource-type/resource-id`
`arn:partition:service:region:account-id:resource-type:resource-id`

I had to hand-roll this rather than use our built-in arn parser because object lambda arns were not making through the initial validation.

src/EndpointV2/Rule/RuleCreator.php
```diff
@@ -0,0 +1,29 @@
...
+            );
+        }
+
+        switch ($type) {
+            case 'endpoint':
+                return new EndpointRule($spec);
+            case 'error':
+                return new ErrorRule($spec);
+            case 'tree':
+                return new TreeRule($spec);
```
SamRemis, line 25: Alternative - I think both are equally fine ways of coding this.  Instead of the check above, we can just add a default to the switch statement which throws the error.

src/EndpointV2/Rule/TreeRule.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Aws\EndpointV2\Rule;
+
+use Aws\EndpointV2\Ruleset\RulesetStandardLibrary;
+
+class TreeRule extends Rule
+{
+    /** @var array */
+    private $rules;
+
+    public function __construct(array $spec)
```
SamRemis, line 12: Spec feels like an awkward name for the variable.  Specification is what I assume you're going for, but what does the specification contain?  Maybe $ruleDefinition.

src/EndpointV2/Rule/Rule.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Aws\EndpointV2\Rule;
+
+use Aws\EndpointV2\Ruleset\RulesetStandardLibrary;
+
+/**
+ *  A rule within a rule set. All rules contain a conditions property,
+ * which can be empty, and documentation about the rule.
+ */
+abstract Class Rule
```
SamRemis, line 11: Maybe rename to AbstractRule

src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Ensures all corresponding client-provided parameters match
+     * the Ruleset parameters' specified type.
+     *
+     * @return void
+     */
+    private function validateInputParameters(array &$inputParameters)
+    {
+        foreach($this->parameters as $paramName => $paramSpec) {
```
SamRemis, line 85: if we wind up changing spec to definition above, doing it here makes sense too

src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,118 @@
...
+            $inputParam = isset($inputParameters[$paramName]) ? $inputParameters[$paramName] : null;
+
+            if (is_null($inputParam) && !is_null($paramSpec->getDefault())) {
```
SamRemis, line 88: This won't catch empty strings.  I'm assuming that this is intentional and was done for specific use cases, but wanted to make a note

src/EndpointV2/Ruleset/RulesetParameter.php
```diff
@@ -0,0 +1,145 @@
...
+    /** @var array */
+    private $required;
+
+    /** @var string */
+    private $documentation;
+
+    /** @var boolean */
+    private $deprecated;
+
+    public function __construct($name, array $spec)
```
SamRemis, line 33: another place to change "spec"

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
SamRemis, line 73: Whoops, can we still remove this comment?

tests/EndpointV2/partitions.json
```diff
@@ -0,0 +1,101 @@
+{
+  "version": "1.1",
```
SamRemis, line 2: This file should be removed and added via trebuchet

src/EndpointV2/Ruleset/Ruleset.php
stobrien89: Thanks for pointing this out.  I omitted `!empty` as an option because we expect falsey values here, but just put in a check for the empty string case

stobrien89:
Closing in favor of #2550

PR #2530, opened 2022-09-22
Using proper PHP comparison operators

githubfromgui:
*Description of changes:*

This change is mostly focused on using the triple `===` comparison operator. Although, I took the chance to validate the parameters passed to the functions `checkArray` and `isAssociative`, now set as `array`.

### DIFF
--- a/.changes/nextrelease/validator-comparison-operators
+++ b/.changes/nextrelease/validator-comparison-operators
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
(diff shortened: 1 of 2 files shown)


stobrien89:
Hi @githubfromgui,

Thanks for the contribution and apologies for the delay.  Unfortunately, we won't be able to accept the change— this could cause unintended side-effects with our input validation.  I believe at least for some of these examples, loose equality comparisons were a deliberate choice.

PR #2533, opened 2022-09-28
Parse Json error code if it is present in HTTP response header

ziyanli-amazon:
*Description of changes:*
### Motivation
As we decided to parse `AwsQuery` Compatible `Error` in the Http response header, the required changes need to take in place on both SDK client and Service. 
This will be a pre-requisite for migrating services from AWSQuery wire protocol to AWSJson.
### What changed
Error code can be fetched from response content (for example __type#fooException). This PR takes another scenario into consideration, of which when `x-amzn-query-error` header with valid value is found from the Http response. 
- The parsed error code remains unchanged if this header is missing, or contains null or invalid value.
### SEP 
[link](https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/query-protocol-migration-compatibility.md)

### DIFF
--- a/src/Api/ErrorParser/JsonParserTrait.php
+++ b/src/Api/ErrorParser/JsonParserTrait.php
@@ -21,9 +21,9 @@ trait JsonParserTrait
         ) {
             $queryError = $response->getHeaderLine('x-amzn-query-error');
             $parts = explode(';', $queryError);
-            if (isset($parts) && count($parts) == 2 && $parts[0] && $parts[1]) {
-                $error_code = $parts[0];
-                $error_type = $parts[1];
+            if (isset($parts) && count($parts) == 2 && trim($parts[0]) && trim($parts[1])) {
+                $error_code = trim($parts[0]);
+                $error_type = trim($parts[1]);
             }
         }
         if (!isset($error_type)) {
--- a/src/Api/ErrorParser/JsonParserTrait.php
+++ b/src/Api/ErrorParser/JsonParserTrait.php
@@ -15,16 +15,26 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
-        if ($queryError = $response->getHeaderLine('x-amzn-query-error')) {
+        if ($this->api &&
+            $this->api->getMetadata('awsQueryCompatible') &&
+            $response->getHeaderLine('x-amzn-query-error')
+        ) {
+            $queryError = $response->getHeaderLine('x-amzn-query-error');
             $parts = explode(';', $queryError);
-            $error = (isset($parts) and !empty(trim($parts[0]))) ? $parts[0] : null;
+            if (isset($parts) && count($parts) == 2 && $parts[0] && $parts[1]) {
+                $error_code = $parts[0];
+                $error_type = $parts[1];
+            }
+        }
+        if (!isset($error_type)) {
+            $error_type = $code[0] == '4' ? 'client' : 'server';
         }
 
         return [
             'request_id'  => (string) $response->getHeaderLine('x-amzn-requestid'),
-            'code'        => isset($error) ? $error : null,
+            'code'        => isset($error_code) ? $error_code : null,
             'message'     => null,
-            'type'        => $code[0] == '4' ? 'client' : 'server',
+            'type'        => $error_type,
             'parsed'      => $this->parseJson($response->getBody(), $response)
         ];
     }
--- a/src/Api/ErrorParser/JsonParserTrait.php
+++ b/src/Api/ErrorParser/JsonParserTrait.php
@@ -15,15 +15,15 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
-        if ($this->api &&
-            $this->api->getMetadata('awsQueryCompatible') &&
-            $response->getHeaderLine('x-amzn-query-error')
+        if ($this->api
+            && $this->api->getMetadata('awsQueryCompatible')
+            && $response->getHeaderLine('x-amzn-query-error')
         ) {
             $queryError = $response->getHeaderLine('x-amzn-query-error');
             $parts = explode(';', $queryError);
-            if (isset($parts) && count($parts) == 2 && trim($parts[0]) && trim($parts[1])) {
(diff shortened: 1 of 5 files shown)


src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -15,10 +15,14 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
+        if ($queryError = $response->getHeaderLine('x-amzn-query-error')) {
+            $parts = explode(';', $queryError);
+            $error = (isset($parts) and !empty(trim($parts[0]))) ? $parts[0] : null;
```
SamRemis, line 20: We need to add something for type here, where we parse the query error and return type from $parts[1].  That should be passed into 'type' on line 23/27 below.

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -15,10 +15,14 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
+        if ($queryError = $response->getHeaderLine('x-amzn-query-error')) {
```
SamRemis, line 18: We can't be positive this header will only exist on the services that are switching their protocols and ideally would check the models for the `awsQueryCompatible` coral trait

SamRemis (changes requested):

src/Api/ErrorParser/JsonParserTrait.php
ziyanli-amazon: will do that.

src/Api/ErrorParser/JsonParserTrait.php
ziyanli-amazon: discussed offline. Will include the change in the next revision.

tests/TestServiceTrait.php
```diff
@@ -45,6 +45,7 @@ private function generateTestService($protocol)
         ];
         if ($protocol === 'json') {
             $metadata['jsonVersion'] = "1.1";
+            $metadata['awsQueryCompatible'] = "String";
```
SamRemis, line 48: I think this is good, I just want to call it out with a note:
This is a widely used file, and now many tests will have this trait added to them.  This makes no difference without adding the header if you look at the logic above, so it's unlikely to effect any other tests as the logic is now.  This sort of opens the door to unintended consequences in test files if the coder isn't aware of what they're doing or just copies and pastes.  This will be fine if we are diligent, especially since the logic around the header being there can't really change.  

Just want to check that @stobrien89 is okay with this as well

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -15,12 +15,26 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
+        if ($this->api &&
+            $this->api->getMetadata('awsQueryCompatible') &&
+            $response->getHeaderLine('x-amzn-query-error')
+        ) {
+            $queryError = $response->getHeaderLine('x-amzn-query-error');
+            $parts = explode(';', $queryError);
+            if (isset($parts) && count($parts) == 2 && $parts[0] && $parts[1]) {
```
SamRemis, line 24: Since this is on explode, if there are two parts, won't the array always be indexed as normal and the $parts[0] and $parts[1] check is not necessary.

ziyanli-amazon: The `parts[0]` and `parts[1]` checks are to make sure they are not empty strings

SamRemis: Ah, that won't catch whitespace only, so maybe add trim to it if you want to catch that as well

ziyanli-amazon: copy. will add trim()! thanks

SamRemis: actually getHeaderLine should trim this, so I think it's good

ziyanli-amazon: Cool. I've already submitted the commit containing the `trim()`. while it doesn't hurt to trim again, let me know if you wanna me to revert that commit or leave as it is.

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -15,12 +15,26 @@ trait JsonParserTrait
     private function genericHandler(ResponseInterface $response)
     {
         $code = (string) $response->getStatusCode();
+        if ($this->api &&
```
stobrien89, line 18: Could we have these operators precede the expression they pertain to? i.e. 

```
(<expression1>
  && <expression2>
  && <expression3>
) {

Comment (ziyanli-amazon) [reply src/Api/ErrorParser/JsonParserTrait.php:18]: Yes for sure

File: tests/TestServiceTrait.php (hunk 3 above)
Comment (stobrien89) [reply tests/TestServiceTrait.php:48]: This is probably fine, but I'd rather see us add `$args = []` as a parameter to `generateTestService` (like generateTestClient) and check for a key that would tell the method to add the property to a json model's metadata.  Then we can just set another service in the provider like 

`$awsQueryCompatibleService =  $this->generateTestService('json', ['queryCompatible' => true])`

Comment (ziyanli-amazon) [reply tests/TestServiceTrait.php]: Sounds good to me. Commit is submitted.

**Review by stobrien89 — CHANGES_REQUESTED** [review]
requesting two small changes

File: tests/TestServiceTrait.php [hunk 6]
```diff
@@ -37,14 +37,17 @@ private function generateTestClient(Service $service, $args = [])
...
-    private function generateTestService($protocol)
+    private function generateTestService($protocol, $args = [])
     {
         $metadata = [
             "protocol" => $protocol,
             "apiVersion" => "2019-05-01"
         ];
         if ($protocol === 'json') {
             $metadata['jsonVersion'] = "1.1";
+            if (array_key_exists('awsQueryCompatible', $args) && $args['awsQueryCompatible']) {
```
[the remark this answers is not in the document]
ziyanli-amazon: sounds good!

stobrien89 (approved):

SamRemis (approved):

PR #2547, opened 2022-10-21
Emit deprecation warning

SamRemis:
[Related Discussion](https://github.com/aws/aws-sdk-php/discussions/2638)


tests/bootstrap.php
```diff
@@ -34,3 +34,7 @@ class_alias('\PHPUnit_Framework_Error_Warning', '\PHPUnit\Framework\Error\Warnin
     $patchGeneratorPath = __DIR__ . '/bootstrap/PHPUnit_Framework_MockObject_Generator_7.4.php';
     file_put_contents($vendorGeneratorPath, file_get_contents($patchGeneratorPath));
 }
+
+if (PHP_VERSION_ID < 70205) {
```
stobrien89, line 38: Just wanted to see what this would do, but it still doesn't take care of the notice/stack trace from being printed into the test results.  We might want to go with the env var here

src/ClientResolver.php
```diff
@@ -264,6 +264,13 @@ class ClientResolver
             'doc'       => 'Set to false to disable checking for shared aws config files usually located in \'~/.aws/config\' and \'~/.aws/credentials\'.  This will be ignored if you set the \'profile\' setting.',
             'default'   => true,
         ],
+        'suppress_php_deprecation_warning' => [
+            'type'      => 'value',
+            'valid'     => ['bool'],
+            'doc'       => 'Set to false to disable the deprecation warning of PHP versions 7.2.4 and below',
+            'default'   => false,
+            'fn'        => [__CLASS__, '_suppress_php_deprecation_warning']
```
stobrien89, line 272: This should be `_apply_suppress_deprecation_warning`

src/ClientResolver.php
```diff
@@ -934,6 +941,33 @@ public static function _apply_idempotency_auto_fill(
         }
     }
 
+    public static function _suppress_php_deprecation_warning($suppressWarning, array &$args) {
+        $phpVersion = PHP_VERSION_ID;
+        if ($suppressWarning !== true) {
```
stobrien89, line 946: We've constrained this to a bool, so no need for strict equality

src/ClientResolver.php
```diff
@@ -934,6 +941,33 @@ public static function _apply_idempotency_auto_fill(
...
+            if (!empty("AWS_SUPPRESS_PHP_DEPRECATION_WARNING")) {
```
stobrien89, line 947: Missing a `getenv` here

stobrien89 (commented):
A few updates needed, but like how the fallbacks are structured now

stobrien89 (approved):

PR #2549, opened 2022-10-26
Composer don't noise after install

dmitryuk:
After setup "Aws\\Script\\Composer\\Composer::removeUnusedServices" every time I run `composer install` with already exists vendor I see the message:
```
Removed 0 AWS services
```
it doesn't contain readable logic, and, for example, the [same library](https://github.com/googleapis/google-api-php-client/blob/main/src/Task/Composer.php#L56) excludes zero-messages.

The PR removes the message if no package deleted

### DIFF
--- a/src/Script/Composer/Composer.php
+++ b/src/Script/Composer/Composer.php
@@ -88,8 +88,10 @@ class Composer
                 }
             }
         }
-        $event->getIO()->write(
-            "Removed $deleteCount AWS service" . ($deleteCount === 1 ? '' : 's')
-        );
+        if ($deleteCount > 0) {
+            $event->getIO()->write(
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @dmitryuk,

Thanks for the contribution.  And sorry for the delay.  I'm hesitant to change the behavior here without more feedback from the community.   My idea here was that this would act as a reminder that the script was still in use, but I could see how that might be a nuisance for some people.  

I'm going to close this PR for now, pending more feedback.   If you could open a new issue [here](https://github.com/aws/aws-sdk-php/issues/new/choose) (as a feature request) and link this PR, we'll watch to see if this is something the community wants.  Thanks again!

PR #2550, opened 2022-11-01
Dynamic Endpoint Resolution

stobrien89:
*Description of changes:*

Implements a new endpoint provider. The Endpoints 2.0 initiative across the AWS SDK organization is changing the way service endpoints are determined: Each service will provide an endpoint ruleset file: a collection of rules that, based on the presence of certain parameter values, will determine the endpoint to be used (or error to be thrown) for a given request.

The new endpoint provider is responsible for reading a service's endpoint ruleset and determining an endpoint (or error) to return by comparing rules contained within the ruleset against parameters— both values provided by the client and values that are scoped during rule evaluation.


src/S3/BucketEndpointArnMiddleware.php
```diff
@@ -174,15 +188,15 @@ private function generateAccessPointHost(
...
+
         $useFips = $this->config['use_fips_endpoint']->isUseFipsEndpoint();
         $fipsString = $useFips ? "-fips" : "";
 
         if ($arn instanceof OutpostsAccessPointArn) {
             $host .= '.' . $arn->getOutpostId() . '.s3-outposts';
         } else if ($arn instanceof ObjectLambdaAccessPointArn) {
             if (!empty($this->config['endpoint'])) {
-               return $host . '.' . $this->config['endpoint'];
+                return $host . '.' . $this->config['endpoint'];
```
SamRemis, line 199: I think this was a mistake?

src/S3Control/EndpointArnMiddleware.php
```diff
@@ -69,26 +72,30 @@ class EndpointArnMiddleware
...
     public static function wrap(
         Service $service,
-        $region,
-        array $config
-
-    ) {
-        return function (callable $handler) use ($service, $region, $config) {
-            return new self($handler, $service, $region, $config);
+                $region,
+        array   $config,
```
SamRemis, line 76: looks a bit odd having only one of them have the type here, and the indentation is off

src/S3Control/EndpointArnMiddleware.php
```diff
@@ -198,12 +206,16 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
                     }
 
                     // Set modified request
-                    $req = $req
-                        ->withUri($req->getUri()->withHost($host)->withPath($path))
-                        ->withHeader('x-amz-account-id', $arn->getAccountId());
                     if (isset($body)) {
                         $req = $req->withBody($body);
                     }
+                    if ($this->isUseEndpointV2) {
```
SamRemis, line 212: I feel like this logic could be reconstructed differently; maybe put all of the endpointsV2 logic into another method and just have one method that you return the result of:

if ($this->isUseEndpointV2) {
      return EndpointV2LogicMethod();
}

function EndpointV2LogicMethod() {
return nextHandler(command,request)
}

src/Sts/RegionalEndpoints/Configuration.php
```diff
@@ -4,10 +4,12 @@
 class Configuration implements ConfigurationInterface
 {
     private $endpointsType;
+    private $isDefault;
```
SamRemis, line 7: isDefault what?  This variable name should be more explanatory, as of right now I'm not sure what it does

src/Api/Serializer/JsonRpcSerializer.php
```diff
@@ -48,21 +52,41 @@ public function __construct(
...
-    public function __invoke(CommandInterface $command)
+    public function __invoke(
+        CommandInterface $command,
+        $endpointProvider = null,
+        array $clientArgs = null
+    )
     {
-        $name = $command->getName();
-        $operation = $this->api->getOperation($name);
+        $operationName = $command->getName();
```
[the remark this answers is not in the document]
stobrien89: Wanted to get your take on this.  We could encapsulate the logic for JsonRPC and Query serializers into one method.  The rest serializer would need its own.  Creating another middleware for this seemed like an unnecessary complexity— because this all needs to happen before request serialization, we'd have to devise a way to pass the resolved endpoint and headers to the next handler (the serializer) before we (historically) instantiate a request. I can write some methods to encapsulate this stuff- just didn't know if we'd want it out in the open for better readability

src/Api/Serializer/RestSerializer.php
```diff
@@ -40,17 +44,47 @@ public function __construct(Service $api, $endpoint)
...
-        $args = $command->toArray();
-        $opts = $this->serialize($operation, $args);
-        $uri = $this->buildEndpoint($operation, $args, $opts);
+        $operationName = $command->getName();
+        $operation = $this->api->getOperation($operationName);
+        $commandArgs = $command->toArray();
+        $opts = $this->serialize($operation, $commandArgs);
+        $headers = isset($opts['headers']) ? $opts['headers'] : [];
+
+        if ($endpointProvider instanceof EndpointProvider) {
```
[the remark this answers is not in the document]
stobrien89: Currently the way I have things configured, any endpoint provider will be passed in here- this is to make for an easier transition when we deprecate the old stuff.  I can either get rid of the default args or do a check in the awsClient `addRequestBuilder` method to make sure the old providers don't get passed

src/Api/Serializer/RestSerializer.php
```diff
@@ -40,17 +44,47 @@ public function __construct(Service $api, $endpoint)
...
+                $operation,
+                $endpointProvider,
+                $commandArgs,
+                $clientArgs,
+                $operationName
+            );
+            $endpoint = $endpointProvider->resolveEndpoint($providerArgs);
+            $this->endpoint = new Uri($endpoint->getUrl());
+
+            if ($service === 's3') {
```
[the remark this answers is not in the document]
stobrien89: Unfortunately S3 can't remove `Bucket` from the path in their models for backward-compatibility concerns.  We either have to pre-process the model to remove it, or ignore it like we're doing here.  For S3control, we have to remove `AccountId` from the URI and also remove the `required` trait from it, but that's done at client instantiation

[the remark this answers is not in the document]
stobrien89: I'm on board with an s3-specific customization instead of doing it here.  I've already done pre-processing with the S3control model, so I could just do that on the S3 client at instantiation if we're using the new provider

src/Api/Service.php
```diff
@@ -28,6 +31,9 @@ class Service extends AbstractModel
     /** @var array */
     private $waiters = null;
 
+    /** @var boolean */
+    private $modifiedModel = false;
```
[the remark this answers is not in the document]
stobrien89: See the above comment re: the S3 specific customizations

src/AwsClient.php
```diff
@@ -419,6 +478,147 @@ private function addRecursionDetection()
...
+    /**
+     * Retrieves and sets default values used for endpoint resolution.
+     */
+    private function setClientBuiltIns($args)
+    {
+        $builtIns = [];
+        $config = $this->getConfig();
+        $service = $args['service'];
+
+        $builtIns['SDK::Endpoint'] = isset($args['endpoint']) ? $args['endpoint'] : null;
```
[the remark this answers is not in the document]
stobrien89: Agreed

src/AwsClient.php
```diff
@@ -419,6 +478,147 @@ private function addRecursionDetection()
...
+        $builtIns['AWS::Region'] = $this->getRegion();
+        $builtIns['AWS::UseFIPS'] = $config['use_fips_endpoint']->isUseFipsEndpoint()
+            && !(isset($builtIns['SDK::Endpoint']));
+        $builtIns['AWS::UseDualStack'] = $config['use_dual_stack_endpoint']->isUseDualstackEndpoint()
+            && !(isset($builtIns['SDK::Endpoint']));
+        if ($service === 'sts') {
```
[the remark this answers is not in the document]
stobrien89: yeah this is some ugly code. I'm up for anything that makes it look better

src/ClientResolver.php
```diff
@@ -402,9 +407,9 @@ private function getArgMessage($name, $args = [], $useRequired = false)
      * @param mixed  $provided The provided value.
      * @throws \InvalidArgumentException
      */
-    private function invalidType($name, $provided)
+    private function invalidType($name, $provided, $expected = null)
     {
-        $expected = implode('|', $this->argDefinitions[$name]['valid']);
+        $expected = $expected ?: implode('|', $this->argDefinitions[$name]['valid']);
```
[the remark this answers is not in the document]
stobrien89: turns out it works without those changes above.  Not sure why I added them in the first place

src/ClientResolver.php
```diff
@@ -1044,4 +1066,45 @@ private static function isValidRegion($region)
...
+
+    private function _apply_client_context_params(array $args)
+    {
+        if (isset($args['api'])) {
+            if (!is_null($clientContextParams = $args['api']->getClientContextParams())) {
+                foreach($clientContextParams as $paramName => $paramSpec) {
+                    $definition = [
+                        'type' => 'value',
+                        'valid' => [$paramSpec['type']],
+                        'doc' => isset($paramSpec['documentation']) ?
```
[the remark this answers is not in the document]
stobrien89: We look at it like we would in the defaultArgs— the doc is provided in the IAE:

```
Fatal error: Uncaught InvalidArgumentException: Invalid configuration value provided for "UseArnRegion". Expected boolean, but got string(4) "true"

UseArnRegion: (boolean)

  Enables this client to use an ARN's region when constructing an endpoint
  instead of the client's configured region. in /Users/xxx/xxx/xxx/aws-sdk-php/src/ClientResolver.php:417

File: src/ClientResolver.php [hunk 13]
```diff
@@ -1044,4 +1066,45 @@ private static function isValidRegion($region)
...
+                        if (!$fn($args[$paramName])) {
+                            $this->invalidType($paramName, $args[$paramName]);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private static function isValidService($service) {
```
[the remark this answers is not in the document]
stobrien89: When customers provide custom API providers- there are a few examples of this in our tests

src/EndpointV2/EndpointDefinitionProvider.php
```diff
@@ -0,0 +1,61 @@
...
+            $apiVersion = self::getLatest($service);
+        }
+
+        $rulesetPath = $serviceDir . '/' . $apiVersion;
+        if (!is_dir($rulesetPath)) {
+            throw new \InvalidArgumentException(
+                'Invalid api version.'
+            );
+        }
+        $fileName = $getTestFile ? '/endpoint-tests-1' : '/endpoint-rule-set-1';
```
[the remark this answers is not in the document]
stobrien89: Agreed.

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Set of helper functions used to set endpoints and endpoint
+ * properties derived from dynamic endpoint resolution.
+ *
+ * @internal
+ */
+trait EndpointV2SerializerTrait
+{
+    private static $accelerateExclusions = [
+        'CreateBucket' => true,
```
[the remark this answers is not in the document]
stobrien89: I think this could be rolled into the model pre-processing I was talking about

src/Sts/RegionalEndpoints/Configuration.php
stobrien89: Basically just the default value or the 'fallback' we use.  I think the latter might make more sense as a variable name

src/S3Control/EndpointArnMiddleware.php
stobrien89: The diff makes this look even stranger, but those types were there before.  `Service` has a type as well

src/Api/AbstractModel.php
```diff
@@ -20,6 +23,9 @@ public function __construct(array $definition, ShapeMap $shapeMap)
     {
         $this->definition = $definition;
         $this->shapeMap = $shapeMap;
+        if (isset($definition['contextParam'])) {
+            $this->contextParam = $definition['contextParam'];
```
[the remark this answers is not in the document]
stobrien89: There's ever only one for a given shape.  There could potentially be multiple contextParams for a given operation, but they are specified at the input members level

src/Api/Serializer/RestSerializer.php
```diff
@@ -36,21 +40,38 @@ public function __construct(Service $api, $endpoint)
     }
 
     /**
-     * @param CommandInterface $command Command to serialized
+     * @param CommandInterface $command Command to serialize
```
SamRemis, line 43: maybe change this to "Command to serialize into a request" to make it clearer what this is doing

src/Api/Serializer/RestSerializer.php
```diff
@@ -250,4 +268,25 @@ private function hasPayloadParam(StructureShape $input, $payload)
...
         return false;
     }
+
+    private function appendQuery($query, $endpoint)
+    {
+        $append = Psr7\Query::build($query);
+        return $endpoint .= strpos($endpoint, '?') ? "&{$append}" : "?$append";
+    }
+
+    private function getVarspecs($command, $args)
```
SamRemis, line 278: what does varspecs stand for?  Variable specifications?  Feels kind of like a confusing name

src/S3Control/S3ControlClient.php
```diff
@@ -211,9 +215,55 @@ public function __construct(array $args)
...
+            }
+        }
+    }
+
+    private function removeRequiredMember(&$definition)
+    {
+        foreach($definition['shapes'] as &$shape) {
+            if (isset($shape['required'])
+            ) {
+                $found = array_search('AccountId', $shape['required']);
```
SamRemis, line 261: Why are we doing this?

src/Api/Serializer/QuerySerializer.php
```diff
@@ -31,38 +35,54 @@ public function __construct(
...
-    public function __invoke(CommandInterface $command)
+    public function __invoke(
+        CommandInterface $command,
+        $endpointProvider = null,
+        array $clientArgs = null
+    )
     {
-        $operation = $this->api->getOperation($command->getName());
-
+        $operationName = $command->getName();
```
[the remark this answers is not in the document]
stobrien89: I separated that out before my most recent changes.  It's no longer needed

src/Api/Serializer/QuerySerializer.php
```diff
@@ -31,38 +35,54 @@ public function __construct(
...
+        ];
+
+        if ($endpointProvider instanceof EndpointProviderV2) {
+            $this->resolveEndpoint(
+                $endpointProvider,
+                $command,
+                $operation,
+                $commandArgs,
+                $clientArgs,
+                $headers
```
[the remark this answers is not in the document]
stobrien89: I could rename it to something broader like `resolveRequest` or something to that effect? another option would be to just move the call to `setHeaders` outside of this method. Auth schemes are also being added here

src/Api/Serializer/RestSerializer.php
stobrien89: Agreed. pretty confusing.  I just kept the variable name from the original implementation, but will change that

src/S3Control/S3ControlClient.php
stobrien89: AccountId is technically no longer required (can be parsed out of an ARN), but the model will not be changed. Ruby and Python have had to make similar changes.

src/EndpointV2/Ruleset/RulesetEndpoint.php
```diff
@@ -0,0 +1,50 @@
...
+ * Represents a fully resolved endpoint that a
+ * rule returns if input parameters meet its requirements.
+ */
+class RulesetEndpoint
+{
+    /** @var string */
+    private $url;
+
+    /** @var array */
+    private $properties;
```
SamRemis, line 15: add a description

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+
+        foreach($contextParams as $name => $spec) {
+            if (isset($commandArgs[$spec['shape']])) {
+                $scopedParams[$name] = $commandArgs[$spec['shape']];
+            }
+        }
+        return $scopedParams;
+    }
+
+    private function scopeStaticContextParams($staticContextParams)
```
SamRemis, line 107: This method doesn't do anything, remove it

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        $boundParams  = array_merge(
+            array_merge($clientArgs, $contextParams),
+            $staticContextParams
+        );
+
+        return array_merge($boundParams, $endpointCommandArgs);
+    }
+
+    private function scopeContextParams($commandArgs, $contextParams)
```
SamRemis, line 95: rename to bind

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+        $commandArgs,
+        $clientArgs
+    )
+    {
+        $rulesetParams = $endpointProvider->getRuleset()->getParameters();
+        $endpointCommandArgs = $this->filterEndpointCommandArgs(
+            $rulesetParams,
+            $commandArgs
+        );
+        $staticContextParams = $this->scopeStaticContextParams(
```
SamRemis, line 64: remove this call and the method below

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private function normalizeEndpointProviderArgs(
+        $endpointCommandArgs,
+        $clientArgs,
+        $contextParams,
+        $staticContextParams
+    )
+    {
+        $boundParams  = array_merge(
```
SamRemis, line 87: change this so all 3 are called the same way, either sequentially or nested

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+        $providerArgs = $this->resolveProviderArgs(
+            $endpointProvider,
+            $operation,
+            $commandArgs,
+            $clientArgs
+        );
+        $endpoint = $endpointProvider->resolveEndpoint($providerArgs);
+
+        if ($this instanceof RestSerializer) {
+            $this->endpoint = new Uri($endpoint->getUrl());
```
SamRemis, line 44: See if you can move this logic out into the RestSerializer

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @internal
+     */
+    private function resolveRequestOptions(
+        $endpointProvider,
+        $command,
+        $operation,
+        $commandArgs,
+        $clientArgs,
+        &$headers
```
SamRemis, line 32: change headers to be a serializer property

src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -0,0 +1,209 @@
...
+     * Merges endpoint resolution arguments passed from the client
+     * and command and attempts to resolve an endpoint. Headers and
+     * auth schemes may be returned in a resolved endpoint object.
+     * A resolved endpoint uri and headers will be applied to the request.
+     * Auth schemes are applied to the command and compared against the default
+     * auth scheme at signing.
+     *
+     * @internal
+     */
+    private function resolveRequestOptions(
```
SamRemis, line 26: change resolve to set

src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /** @var RulesetStandardLibrary */
+    public $standardLibrary;
+
+    public function __construct(array $ruleset, array $partitions)
+    {
+        $this->version = $ruleset['version'];
+        $this->parameters = $this->createParameters($ruleset['parameters']);
+        $this->rules = $this->createRules($ruleset['rules']);
+        $this->standardLibrary = new RulesetStandardLibrary($partitions);
```
SamRemis, line 30: why do we have this instead of a class like our functions class?

src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,118 @@
...
+     * Return the first rule the parameters match against.
+     *
+     * @return mixed
+     */
+    public function evaluate(array $inputParameters)
+    {
+        $this->validateInputParameters($inputParameters);
+
+        foreach($this->rules as $rule) {
+            $inputParametersCopy = $inputParameters;
```
SamRemis, line 68: Try running the tests without this copy; I think it makes a copy anyhow

src/EndpointV2/Ruleset/Ruleset.php
```diff
@@ -0,0 +1,118 @@
...
+     * Ensures all corresponding client-provided parameters match
+     * the Ruleset parameter's specified type.
+     *
+     * @return void
+     */
+    private function validateInputParameters(array &$inputParameters)
+    {
+        foreach($this->parameters as $paramName => $param) {
+            $inputParam = isset($inputParameters[$paramName]) ? $inputParameters[$paramName] : null;
+
```
SamRemis, line 87: feels like this space shouldn't be here

src/EndpointV2/Rule/TreeRule.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * If a tree rule's conditions evaluate successfully, iterate over its
+     * subordinate rules and return a result if there is one. If any of the
+     * subsequent rules are trees, the function will recurse until it reaches
+     * an error or an endpoint rule
+     *
+     * @return mixed
+     */
+    public function evaluate(
+        array &$inputParameters,
```
SamRemis, line 35: why not just pass this by reference and not copy?

taka-oyama:
The syntax for `BucketEndpointArnMiddleware`'s constructor is broken.

```
Optional parameter $config declared before required parameter $isUseEndpointV2 is implicitly treated as a required parameter

/Users/t-oyama/Projects/github/laravel/framework/vendor/aws/aws-sdk-php/src/S3/BucketEndpointArnMiddleware.php:59
```

This needs immediate attention.

driesvints:
@SamRemis @stobrien89 seems like this PR also broke Laravel's test suite. All of our PHP 8.2 builds are failing: https://github.com/laravel/framework/actions/runs/3463979449/jobs/5795101015

driesvints:
I sent in a PR here: https://github.com/aws/aws-sdk-php/pull/2560

driesvints:
Btw, I'd suggest to add PHP 8.2 (which is out in a few days) builds as soon as you can to catch these errors. Although I'm not 100% sure this use case is covered...

PR #2551, opened 2022-11-01
Undo custom logic to gzip endpoints files

SamRemis:
*Issue #, if available:*

*Description of changes:*


.gitattributes
```diff
@@ -19,6 +19,7 @@ src/data/*.json export-ignore
 src/data/*/*/*.json export-ignore
 src/data/*/*/docs*.json.php export-ignore
 src/data/*/*/examples*.json.php export-ignore
+src/data/*/*/endpoint-tests-*.php export-ignore
```
stobrien89, line 22: `endpoint-tests-*.json` should be covered by line 19. We could probably leave this, but makes sense to change to match docs and examples above

stobrien89 (approved):

Issue #2559, opened 2022-11-15
listObjectsV2 returns different structure between v3.242.1 … v3.243.0


leevigraham:
### Describe the bug

Run the following php code

```php
$command = $client->getCommand(
    'ListObjectsV2',
    [
        'Bucket'  => $bucket,
        'Prefix'  => $prefix,
        'MaxKeys' => 1,
    ]
);

$result = $client->executeCommand($command);
```

### Expected Behavior

`v3.242.1` does not include `contents`

![Screen Shot 2022-11-15 at 3 40 49 pm](https://user-images.githubusercontent.com/25124/201831130-f7f7a25e-bb5c-4a9f-a5d7-56dd54ee66fa.png)

`3.243.0` does include `contents` with size: 0

![Screen Shot 2022-11-15 at 3 40 11 pm](https://user-images.githubusercontent.com/25124/201831151-c26ad376-cc65-4ddd-a0c9-83d2978cf8b8.png)

### Current Behavior

See Expected Behavior

### Reproduction Steps

See "Describe the bug"

### Additional Information/Context

This causes issues downstream in thephpleague/flysystem: https://github.com/thephpleague/flysystem/blob/3.x/src/AwsS3V3/AwsS3V3Adapter.php#L132

### SDK version used

3.243.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

php 8.1

i-just:
I believe this is happening because of ill-formatted endpoint URL. You can see it in the `$results` if you follow @leevigraham's steps. 

On v3.242.1 `src/AwsS3Adapter.php->doesDirectoryExist()`, the `$result` > `@metadata` > `effectiveUri` is correctly encoded, on v3.243.0 and v3.243.1 it's not. It causes the v3.243.1 `doesDirectoryExist()` to check if there's any directory in the bucket you're searching through, ignoring the `prefix` param.

v3.242.1: `https://s3.amazonaws.com/<bucket_name>?list-type=2&prefix=300C.png%2F&max-keys=1`
v3.243.0, v3.243.1: `https://s3.amazonaws.com/<bucket_name>?list-type=2%3Dprefix=300C.png%2F&max-keys=1`

lro-tiierisch:
Hey, I just created a new PHP project on my local machine and installed the AWS PHP SDK because I had to look up some files in our companies S3 storage. After getting some unexpected results I also noticed that the Prefix parameter seems to be ignored in v3.243.

[...]

PR #2560, opened 2022-11-15
Fix default parameter in BucketEndpointArnMiddleware constructor

driesvints:
This PR fixes a breaking change in https://github.com/aws/aws-sdk-php/pull/2550 where a new required constructor parameter was added after a parameter with a default value. This isn't allowed in PHP and will lead to an error being thrown. 

I have no idea if the default value should be `false` or `true` but I figured `false` was the goto value.

It seems this use case isn't covered by a test but I was unsure on how to add one. Would appreciate any guidance with that or just a merge and release so consuming libraries are unblocked. Right now, this is failing Laravel's builds: https://github.com/laravel/framework/actions/runs/3463979449/jobs/5795101015

### DIFF
--- a/src/S3/BucketEndpointArnMiddleware.php
+++ b/src/S3/BucketEndpointArnMiddleware.php
@@ -43,6 +43,7 @@ class BucketEndpointArnMiddleware
      * @param Service $service
      * @param $region
      * @param array $config
+     * @param boolean $isUseEndpointV2
      * @return callable
      */
     public static function wrap(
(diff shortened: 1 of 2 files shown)


mirzazeyrek (approved):

stobrien89:
Hi @driesvints,

Thanks a lot for submitting this.  I needed a quick way to combine the fixes for the corresponding issue, as well as #2559 (and add the same change for `endpointArnMiddleware` which I saw you just added), otherwise I would have merged this.  Really appreciate the help.

PR #2561, opened 2022-11-15
append query - fix checking strpos of ?

i-just:
*Issue #, if available:* 
#2559

*Description of changes:*
`src/Api/Serializer/RestSerializer.php` > `appendQuery()` performs a `strpos` check on a passed through `$endpoint` to see if it contains a `?`, but doesn't take into consideration that it could be at the start of the `$endpoint`. This then lead to the incorrect endpoint being queried (ignoring passed through query params).

This change is to make sure that the strict check is performed.


i-just:
Closing as it's now also being addressed by @stobrien89 in #2562

stobrien89:
Thank you @i-just!

Issue #2572, opened 2022-11-19
SNS Subscribe not working - EndpointV2 ?


drekinov:
### Describe the bug

not possible to use SnsClient to execute http/https subscribe

### Expected Behavior

to subscribe http/https endpoints as with 3.242.1 or older

### Current Behavior

I assume that EndpointV2 is confusing `Endpoint` key as AWS API url and sends post data from command to subscriber URL instead AWS SNS rest api.

### Reproduction Steps

```php
$endpoint = 'https://where-all-data-will-be-send.com/';
$protocol = 'https';
$topicArn = 'arn:::::';
$command = $this->snsClient->getCommand(
            'subscribe',
            [
                'Endpoint' => $endpoint,
                'Protocol' => $protocol, // REQUIRED
                'TopicArn' => $topicArn, // REQUIRED
            ]
        );

        $result = $this->snsClient->execute($command);
```

### Additional Information/Context

I want to credit that SO topic because it was the first indicating that my issue might be SDK related: https://stackoverflow.com/questions/74475524/aws-sdk-sns-subscribe-to-https-endpoint

And that is current documentation confirming my syntax should be fine: https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sns-2010-03-31.html#subscribe

### SDK version used

 \>= 3.243.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

8.1.12, ubuntu 20.04 LTS

stobrien89:
Hi @drekinov,

Thanks for raising this! You are correct— the value for `Endpoint` passed into the command is being used for endpoint ruleset evaluation.  Should have a fix out shortly.  In the meantime, you can provide `endpoint_provider => partitionEndpointProvider::defaultProvider()` into your client config or pin your version to `3.242.1` to mitigate this.

stobrien89:
Hi @drekinov,

This should be resolved on `3.246.1`. Could you try upgrading and confirm? Thanks!

Khel:
Hello @drekinov,

Author from the StackOverflow post here :)

Many thanks. We are currently in contact with AWS support and they have not been able to provide any insight. They say it works.

@stobrien89 Will try to see if it works with the updated SDK.

Regards!

Khel:
@stobrien89 @drekinov I can confirm v3.246.1 fixes this :)

Many thanks. It has been hell the last 4 days :)

stobrien89:
Hi @Khel and @drekinov,

[...]

PR #2580, opened 2022-11-22
bugfix: unnecessary `/` appended to s3 path joins

### LINKED ISSUE #2575: S3 ListObjects broken with Aws\EndpointV2
### Describe the bug

When using an `endpoint` + `bucket_endpoint` with `Aws\EndpointV2` the `listObjects` method of an S3 Client doesn't seem to work as expected. It results in a double-slash URL:

```
GET https://bucket-name.endpoint.dev/bucket-name//?prefix=%2F&encoding-type=url
```

Removing the trailing slash from the `endpoint` argument resolves the error, however the returned results are always empty, as the bucket name path is completely stripped from the effective URL:

```
https://bucket-name.endpoint.dev/?prefix=%2F&encoding-type=url
```

### Expected Behavior

The `ListObjects` call should work with a trailing slash on the endpoint and return the items from the S3 bucket.

### Current Behavior

`ListObjects` produces an invalid URL.

### Reproduction Steps

```php
use Aws\Endpoint\PartitionEndpointProvider;
use Aws\S3\S3Client;

$s3Client = new S3Client( [
	'version' => '2006-03-01',
	'endpoint' => 'https://bucket-name.endpoint.dev/bucket-name/',
	'signature' => 'v4',
	'bucket_endpoint' => true,
	'region' => 'us-east-1',
	'credentials' => [
		'key' => 'admin',
		'secret' => 'password',
	],
] );

var_dump( $s3Client->listObjects( [
	'Bucket' => 'bucket-name',
	'Prefix' => '/',
] ) );
```

### Possible Solution

I do not have suggestions on how to fix `ListObjects` with `Aws\EndpointV2`, but current workarounds are:

* Use `ListObjectsV2` instead
* Use `PartitionEndpointProvider::defaultProvider()` for `endpoint_provider`

_No response_

### SDK version used

3.246.1

PHP 8.0.24

stobrien89:
*Issue #, if available:*
closes #2575, revision to #2572

*Description of changes:*
Update to S3 model pre-processing to remove unnecessary trailing slashes when joining paths.  

Updates to #2573 to disallow overrides of any endpoint parameter built-in values (this should be handled by contextparams) except in the case of  the legacy `'@use_dual_stack_endpoint', '@use_accelerate_endpoint'` and `'@use_path_style_endpoint' ` command options


src/EndpointV2/EndpointV2SerializerTrait.php
```diff
@@ -127,28 +127,25 @@ private function filterEndpointCommandArgs(
...
-                    && strpos($builtIn, 'SDK') === 0
-                ) {
+                if (!empty($value->getBuiltIn())) {
                     continue;
                 }
                 $filteredArgs[$name] = $commandArgs[$name];
             }
         }
+
+        if ($this->api->getServiceName() === 's3') {
```
SamRemis, line 141: Is there no way to take this out of the shared files?

src/S3/S3Client.php
```diff
@@ -678,9 +678,11 @@ private function processEndpointV2Model()
         foreach($definition['operations'] as &$operation) {
             if (isset($operation['http']['requestUri'])) {
                 $requestUri = $operation['http']['requestUri'];
-                $requestUri = str_replace('/{Bucket}', '/', $requestUri);
-                $requestUri = str_replace('//', '/', $requestUri);
-                $requestUri = str_replace('/?', '?', $requestUri);
+                if ($requestUri === "/{Bucket}") {
+                    $requestUri = str_replace('/{Bucket}', '/', $requestUri);
+                } else {
+                    $requestUri = str_replace('/{Bucket}', '', $requestUri);
```
SamRemis, line 684: just trying to undrstand this- why do we only have a trailing slash when the uri is /{Bucket}?

src/Api/Serializer/RestSerializer.php
```diff
@@ -231,20 +231,24 @@ function (array $matches) use ($varDefinitions) {
...
         $path = $this->endpoint->getPath();
-        //accounts for removal of bucket from requestUri path
+
+        //Accounts for trailing '/' in path when custom endpoint
+        //is provided to endpointProviderV2
         if ($this->api->isModifiedModel()
             && $this->api->getServiceName() === 's3'
         ) {
-            $relative = $path . $relative;
+            substr($path, -1) === '/' && substr($relative, 0, 1) === '/'
```
SamRemis, line 241: this just removes it from the relative if the path has a trailing slash and the relative has a trailing?  I feel like there's different logic that could accomplish the same thing; maybe an if statement would be clearer here?  Like in pseudo-code:
```
if (substr($path, -1) === '/' && substr($relative, 0, 1) === '/') {
$path = rtrim($path, '/');
}
$relative = $path . $relative;
```
Of course this won't work if the $path can't be changed if it's used later, but it looks like it's not.  

Also, maybe change `substr($relative, 0, 1)` to `$relative[0]`

src/S3/S3Client.php
stobrien89: We/S3 expect there to be a trailing slash on operations (CreateBucket, DeleteBucket, etc.) where there is no additional path or query.  For the others, if there's more path string beyond `{Bucket}`, a forward slash will already be present, or the query will just be appended.

SamRemis (approved):

PR #2582, opened 2022-11-24
Bearer token test updates and bug fixes

SamRemis:
*Issue #, if available:*

*Description of changes:*


src/ClientResolver.php
```diff
@@ -683,12 +683,15 @@ public static function _apply_endpoint_provider($value, array &$args)
 
             $args['endpoint'] = $result['endpoint'];
 
-            if (
-                empty($args['config']['signature_version'])
-                && isset($result['signatureVersion'])
-            ) {
-                $args['config']['signature_version']
-                    = $result['signatureVersion'];
+            if (empty($args['config']['signature_version'])) {
```
stobrien89, line 686: Shouldn't this already be handled by the defaultProvider?  Unless `bearer` is modeled differently than other auth types I don't think we'll need this check

tests/Token/TokenProviderTest.php
```diff
@@ -121,48 +124,6 @@ public function testCreatesFromCache()
         $this->assertEquals($token->getExpiration(), $found->getExpiration());
     }
 
-    public function testRefreshesFromCache()
```
stobrien89, line 124: Why are we removing this?

stobrien89 (changes requested):

tests/Token/TokenProviderTest.php
SamRemis: It was an accidentally added test.  It's actually tested below in testTokenProviderCachedSuccess.  This test wasn't even using a cache, and doing nothing

stobrien89 (approved):

PR #2626, opened 2023-01-20
bugfix: add ability to calculate ContentMD5 for PutObject and UploadPart

### LINKED ISSUE #2256: PutObject with object lock
Hi, 

I'm trying to upload a file to a AWS S3 which have object lock enabled. Using the CLI it works as expected, however, with the PHP SDK I get the following error:
```
Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters
```

Now I saw that the ApplyChecksumMiddleware could handle this (https://github.com/aws/aws-sdk-php/blob/ff4872f72fac84ce6e384c4f40d30ec7b0d1eeda/src/S3/ApplyChecksumMiddleware.php#L58), but I was curious why checking on 'httpChecksumRequired' is required and this is not always done (f.e. always on the PutObject)

stobrien89:
*Issue #, if available:*
Closes #2256 

*Description of changes:*
Adds `AddContentMD5` parameter, which allows ApplyChecksumMiddleware to compute MD5 checksums for `PutObject` and `UploadPart`.  Additionally, the ObjectLoader, MultiPartUploader and the Transfer utility will now accept an option of `add_content_md5`, which when set to `true` will have the MD5 checksum calculated by ApplyChecksumMiddleware.

### DIFF
--- a/tests/S3/S3ClientTest.php
+++ b/tests/S3/S3ClientTest.php
@@ -2228,4 +2228,47 @@ EOXML;
         );
         $s3->execute($command);
     }
+
+    public function testAutomaticallyComputesMD5ForPutObject()
+    {
+        $s3 = $this->getTestClient('s3');
+        $this->addMockResults($s3, [[]]);
+        $command = $s3->getCommand('PutObject',
+            ['Bucket' => 'foo', 'Key' => 'foo', 'Body' => 'test', 'AddContentMD5' => true]
+        );
+        $command->getHandlerList()->appendSign(
+            Middleware::tap(function ($cmd, $req) {
+                $this->assertSame(
+                    'CY9rzUYh03PK3k6DJie09g==',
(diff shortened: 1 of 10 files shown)


tests/S3/S3ClientTest.php
```diff
@@ -2228,4 +2228,47 @@ public function testAddsForwardSlashIfEmptyPathAndQuery()
...
+                $this->assertSame(
+                    'CY9rzUYh03PK3k6DJie09g==',
+                    $req->getHeader('Content-MD5')[0]
+                );
+            })
+        );
+        $s3->execute($command);
+    }
+
+    public function testAutomaticallyComputesMD5ForUploadPart()
```
SamRemis, line 2250: Please add a test for this that is the inverse that checks that it's not added when AddContentMD5 is empty

SamRemis (approved):

PR #2630, opened 2023-01-26
bugfix: fix integ tests

stobrien89:
*Description of changes:*

Updates crypto test resources, removes deprecated block cipher uploads, switches `S3EncryptionClient` to `S3EncryptionClientV2`.  Same with the KMS materials provider.  Switches `PostObjectV4` test to bucket owner read permissions, removes test for deprecated `PostObjectV2`.

### DIFF
--- a/features/s3/postObject.feature
+++ b/features/s3/postObject.feature
@@ -1,28 +1,14 @@
 @s3 @integ
 Feature: POST object uploads
 
-    Scenario: Upload an object via POST object with Signature V2
-        Given I have an s3 client and I have a file
-        And I have an array of form inputs as following:
-        |key              |value              |
-        |acl              |public-read        |
-        And I provide a json string policy as following:
-        """
-        {"expiration":"2026-04-25T22:53:16Z",
-        "conditions":[{"acl":"public-read"}]}
-        """
-        When I create a POST object SigV2 with inputs and policy
-        And I make a HTTP POST request
-        Then the file called "file.ext" is uploaded
-
     Scenario: Upload an object via POST object with Signature V4
         Given I have an s3 client and I have a file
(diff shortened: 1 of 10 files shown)


features/s3/postObject.feature
```diff
@@ -1,28 +1,14 @@
 @s3 @integ
 Feature: POST object uploads
 
-    Scenario: Upload an object via POST object with Signature V2
```
SamRemis, line 4: I'm pretty sure I'm the one who deleted this since you can't make new S3 buckets that accept V2- for legacy customers, we should add a backlog item to ticket S3 to allowlist an account for this and run it there so we still have this tested for customers using legacy buckets.

tests/Integ/S3Context.php
```diff
@@ -237,20 +237,6 @@ public function iProvideAJsonStringPolicyAsFollowing(PyStringNode $string)
         $this->jsonPolicy = json_encode($policy);
     }
 
-    /**
-     * @When I create a POST object SigV2 with inputs and policy
```
SamRemis, line 241: add this to the backlog item too

tests/Integ/S3EncryptionContext.php
```diff
@@ -93,7 +94,7 @@ public function iEncryptEachFixtureWith($wrapAlgorithm, $alias, $region, $cipher
         ]);
         $keyArn = $this->getKmsArnFromAlias($kmsClient, $alias);
 
-        $materialsProvider = new KmsMaterialsProvider(
```
SamRemis, line 96: Should we be duplicating this and doing a test for v1 and a test for v2?

Or is V1 permanently deprecated

PR #2632, opened 2023-02-01
enhancement: update partition to accept boolean values for variants

stobrien89:
*Description of changes:*

Allows boolean values to be passed for `use_fips_endpoint` and `use_dual_stack_endpoint` Partition config.  This code path can be reached when instantiating a PartitionEndpointProvider.  Use case here is to bypass dynamic endpoint resolution for testing purposes while also being able to use endpoint variants. 

### DIFF
--- a/src/Endpoint/Partition.php
+++ b/src/Endpoint/Partition.php
@@ -287,12 +287,18 @@ final class Partition implements ArrayAccess, PartitionInterface
     {
         $variantTags = [];
         if (isset($options['use_fips_endpoint'])) {
-            if ($options['use_fips_endpoint']->isUseFipsEndpoint()) {
+            $useFips = $options['use_fips_endpoint'];
+            if (is_bool($useFips)) {
+                $useFips && $variantTags[] = 'fips';
+            } elseif ($useFips->isUseFipsEndpoint()) {
                 $variantTags[] = 'fips';
             }
         }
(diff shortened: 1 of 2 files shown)


src/Endpoint/Partition.php
```diff
@@ -287,12 +287,18 @@ private function getVariant(array $options, array $data)
     {
         $variantTags = [];
         if (isset($options['use_fips_endpoint'])) {
-            if ($options['use_fips_endpoint']->isUseFipsEndpoint()) {
+            $useFips = $options['use_fips_endpoint'];
+            if (is_bool($useFips)) {
+                $useFips && $variantTags[] = 'fips';
+            } elseif ($useFips->isUseFipsEndpoint()) {
```
SamRemis, line 293: This is typechecked when we initialize the client, right?  So it'll only be a bool or a config option that contains a method isUseFipsEndpoint?

stobrien89: Yeah, it's typechecked for the top-level config option, but if someone instantiates a `PartitionEndpointProvider`, they have to also instantiate a fips and dualstack config to pass in as options.  This allows `'use_fips_endpoint` => true` (same for dualstack) to be passed in when instantiating the provider.

SamRemis (approved):

PR #2639, opened 2023-02-16
Update functions signature to include return type

PablNico:
*Description of changes:*
These changes just update the function signatures to include their return type. 
*Motivation:*
We have a custom drupal module to create a user for our developers to AWS IAM, every time we give new access to developers we got a huge error message about deprecated functions. I tried to update the package with composer and add `#[\ReturnTypeWillChange]` but it hasn't solved the issue, after further investigation, I managed to solve the warnings by adding the return type into function declaration.

### DIFF
--- a/src/Api/AbstractModel.php
+++ b/src/Api/AbstractModel.php
@@ -36,8 +36,7 @@ abstract class AbstractModel implements \ArrayAccess
     /**
      * @return mixed|null
      */
-    #[\ReturnTypeWillChange]
-    public function offsetGet($offset)
+    public function offsetGet($offset): mixed
     {
         return isset($this->definition[$offset])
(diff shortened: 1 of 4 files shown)


SamRemis:
Hi @PablNico,
Thanks for the PR- unfortunately, we can't accept it until August 15th at the earliest.  We still support back to PHP 5.5, and just [announced our deprecation](https://github.com/aws/aws-sdk-php/discussions/2638) yesterday.  Adding a return type will break customers on versions before 7.0, or 7.1 for `void`.  
I'm looking forward to being able to add return types to our methods then

stobrien89:
Hi @PablNico,

Thanks again for the PR.  Going to close this for now.  Feel free to check back in with us at the beginning of August and we'd be happy to reconsider.

PR #2640, opened 2023-02-17
Update deprecation warning

SamRemis:
*Issue #, if available:*

*Description of changes:*


stobrien89 (commented):
Looks good, but we might want an assertion that the message is not showing up in the error handler in the tests where we disable the message

stobrien89 (approved):

PR #2643, opened 2023-02-23
str_replace fix

marispro:
*Description of changes:*
fix for case sensitive const name replace

### DIFF
--- a/build/Burgomaster.php
+++ b/build/Burgomaster.php
@@ -326,7 +326,7 @@ EOT
         $this->debug("Creating phar stub at $dest");
 
         $alias = $alias ?: basename($dest);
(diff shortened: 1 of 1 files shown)


peterader (approved):

enkhjargal (approved):

yenfryherrerafeliz (approved):

yenfryherrerafeliz:
This change is needed because the replacement logic was wrong due to that the string given to the `str_replace` function was being upper-cased while the text being searched is lower case. 

Please see:
```
// supposing $alias = 'test.phar' then '.phar' was never going to be found on 'TEST.PHAR'. 
constName = str_replace('.phar', '', strtoupper($alias)) . '_PHAR';
```

PR #2651, opened 2023-03-01
make aws-crt-php optional

TingDaoK:
*Issue #, if available:*

- aws/aws-crt-php is about to change the requirement to require installing the native extension `ext-awscrt` first. So that it will break the current customers

*Description of changes:*
- Details of why we change the requirement can be found https://github.com/awslabs/aws-crt-php/pull/89
- Make `aws/aws-crt-php` a suggestion instead of require


stobrien89:
Closing this until we reach a consensus that we will make the package optional.

Issue #2658, opened 2023-03-09
Use region set in environment variables when contacting sts with AssumeRoleWebIdentity


pkruesi:
### Describe the bug

I run some php code inside a container on eks. When I try to get credentials using `$provider = CredentialProvider::defaultProvider();` the default endpoint `sts.amazonaws.com` gets contacted instead of the regional endpoint. 

### Expected Behavior

I expect, that the function `defaultProvider()` when run in a container on eks reads the environment variables `AWS_REGION` and `AWS_DEFAULT_REGION` and uses this variables to determine which region should get used to contact sts to get credentials. For example if `AWS_DEFAULT_REGION=eu-central-1`  is set, the endpoint `sts.eu-central-1.amazonaws.com` should get conacted.

### Current Behavior

The `defaultProvider()` function uses always `us-east-1` when used with `AssumeRoleWithWebIdentity`. Except when a config array with explicit defined region is given to the `defaultProvider($config)` function.

### Reproduction Steps

Run in a container on esk with the following environment varialbes set:
```
AWS_ROLE_ARN=arn:aws:iam:xxxxxxxxxxxxxxxx:role/nexiot-api
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
AWS_STS_REGIONAL_ENDPOINTS=regional
AWS_DEFAULT_REGION=eu-central-1
AWS_REGION=eu-central-1
```
```
<?php

require_once __DIR__ . '/libraries/autoload.php';

use Aws\Credentials\CredentialProvider;
use Aws\Credentials\Credentials;
use Aws\S3\S3Client; 

$provider = CredentialProvider::defaultProvider();

$client = new S3Client([
	'region' => 'eu-central-1',
	'version' => '2006-03-01',
	'credentials' => $provider
]);

$buckets = $client->listBuckets();
```

checked with cloudtrail

```
SELECT eventTime, userIdentity.principalid, eventName, userAgent, awsRegion, eventSource  FROM xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx 
WHERE userAgent like 'aws-sdk-php%'
	and userAgent not like 'aws-sdk-php%ubuntu%'
    and eventTime > '2023-03-09 08:00:00'
```

cloudtrail result (see awsRegion and eventSource)

```
[...]

PR #2659, opened 2023-03-09
Use region set in environment variables when contacting sts

pkruesi:
When I try to get credentials using `$provider = CredentialProvider::defaultProvider();` the default endpoint `sts.amazonaws.com` gets contacted instead of the regional endpoint.

- Evaluate the region environment variables to contact the regional endpoint for sts.
- Relates to issue #2658

*Issue #2658

### DIFF
--- a/.changes/nextrelease/read-environment-vars.json
+++ b/.changes/nextrelease/read-environment-vars.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "defaultProvider",
+        "description:": "Use region set in environment variables when contacting sts."
+    }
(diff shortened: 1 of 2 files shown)


stobrien89:
Hi @pkruesi,

Thanks for the contribution.  We can't accept this change since it alone will not regionalize the STS endpoint and while we may consider supporting `AWS_REGION` sometime in the future, we will definitively not be supporting `AWS_DEFAULT_REGION`.  If we do end up supporting `AWS_REGION`, I imagine it would be at the client configuration level and not specifically in this class. We can continue the discussion in the issue you opened!

PR #2671, opened 2023-04-10
Allow S3EncryptionClient::putObject to accept any type that implement StreamInterface

whizsid:
`putObject` is throwing an exception if user passed a stream implemented to `StreamInterface`. I checked the usage of `$plaintext` and found that `$plaintext` is using only for constructors of `AesGcmEncryptingStream` and `AesEncryptingStream`. Those constructors requiring `StreamInterface` as first arguments. So we do not have to stick into `Stream` in `EncryptionTrait`s.

Now we are doing like this.
```php
$tfh = tmpfile();
fwrite($tfh, (string)$stream);

$result = $encryptedS3Client->putObject([
      '@MaterialsProvider' => $this->kmsProvider,
      '@CipherOptions' => $cipherOptions,
      'Bucket' => $bucket,
      'Key' => $path,
      'Body' => $tfh,
]);
```

With this PR we can save some cost like this:-
```php
$result = $encryptedS3Client->putObject([
      '@MaterialsProvider' => $this->kmsProvider,
      '@CipherOptions' => $cipherOptions,
      'Bucket' => $bucket,
      'Key' => $path,
      'Body' => $stream,
]);
```

### DIFF
--- a/src/Crypto/EncryptionTrait.php
+++ b/src/Crypto/EncryptionTrait.php
@@ -3,7 +3,7 @@ namespace Aws\Crypto;
 
 use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\AppendStream;
-use GuzzleHttp\Psr7\Stream;
+use Psr\Http\Message\StreamInterface;
 
 trait EncryptionTrait
 {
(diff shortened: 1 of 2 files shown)


RanVaknin:
Hi @whizsid ,

We have tried refreshing the CI tests in order to get them to pass but we are running into some unexpected issues. Is there any chance you can create a new PR with those same changes? 

Thanks,
Ran~

PR #2673, opened 2023-04-12
Enhancement update user agent

SamRemis:
*Issue #, if available:*

*Description of changes:*


stobrien89 (commented):
Are endpoint discovery and retry modes the only features we're tracking for this?

src/ClientResolver.php
stobrien89, line 1: Are there any new cases we can add to existing tests for this? If not, we should probably add some new unit tests

stobrien89 (changes requested):

PR #2677, opened 2023-04-20
Bump minimum psr7 versions in light of CVE-2023-29197

GrahamCampbell:


composer.json
```diff
@@ -39,12 +39,12 @@
         "nette/neon": "^2.3",
         "andrewsville/php-token-reflection": "^1.4",
         "psr/cache": "^1.0",
+        "psr/http-message": "^1.0",
         "psr/simple-cache": "^1.0",
         "paragonie/random_compat": ">= 2",
         "sebastian/comparator": "^1.2.3 || ^4.0",
         "yoast/phpunit-polyfills": "^1.0",
-        "dms/phpunit-arraysubset-asserts": "^0.4.0",
-        "psr/http-message": "<1.1"
```
GrahamCampbell, line 47: This change was not necessary. It is v2.0.0 that is the issue with the return types in your tests, not v1.1.0.

SamRemis (approved):

PR #2693, opened 2023-05-16
feature: add request body compression support

stobrien89:
*Description of changes:*
Adds support for compressing request payloads for service/operations that support it. Compatibility is determined by a modeled trait `requestcompression`.  For now, the only compression algorithm supported is `gzip`.


SamRemis (approved):

src/RequestCompression/RequestCompressionMiddleware.php
```diff
@@ -0,0 +1,168 @@
...
+            } else {
+                $minCompressionSz = $config['request_min_compression_size_bytes'];
+            }
+
+            if ($this->isValidCompressionSize($minCompressionSz)) {
+                return $minCompressionSz;
+            }
+        }
+
+        return self::REQUEST_MIN_COMPRESSION_SIZE_BYTES;
```
SamRemis, line 152: When is this called?  If the user does specify a minimum size, then it will be converted into the config object (if it isn't already that type).  If a user does not, then this will be set to a config option by the default provider.  
What codepath will lead the if statement above to be false

src/RequestCompression/RequestMinCompressionSizeBytes/Configuration.php
```diff
@@ -0,0 +1,45 @@
+<?php
+namespace Aws\RequestCompression\RequestMinCompressionSizeBytes;
+
+use Aws;
+use Aws\RequestCompression\Exception\ConfigurationException;
+
+class Configuration implements ConfigurationInterface
```
SamRemis, line 7: This is a lot of lines of code to get a number and a boolean; could we get the config another way?

src/Configuration/ConfigurationResolver.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        $envValue = self::env($key, $expectedType);
+        if (!is_null($envValue)) {
+            return $envValue;
+        }
+
+        if (!isset($config['use_aws_shared_config_files'])
+            || $config['use_aws_shared_config_files'] != false
+        ) {
+            $iniValue = self::ini($key, $expectedType);
```
SamRemis, line 45: are the ini keys always going to be the same as the environment variables?

src/Api/Operation.php
```diff
@@ -139,4 +144,29 @@ private function setContextParams()
...
+     *
+     * @return array
+     */
+    public function getTraits()
+    {
+        return $this->traits;
+    }
+
+    /**
+     * Returns an associative array of operation traits.  Operation traits are
```
SamRemis: Accidentally duplicated code comment

src/ClientResolver.php
```diff
@@ -219,6 +220,20 @@ class ClientResolver
...
+            'type'      => 'value',
+            'valid'     => ['bool', 'callable'],
+            'doc'       => 'Set to true to disable request compression for supported operations',
+            'fn'        => [__CLASS__, '_apply_disable_request_compression'],
+            'default'   => [__CLASS__, '_default_disable_request_compression'],
+        ],
+        'request_min_compression_size_bytes' => [
+            'type'      => 'value',
+            'valid'     => ['int', 'callable'],
+            'doc'       => 'Set to a value between between 0 and 10485760 bytes, inclusive',
```
SamRemis: Let's add docs here that clarify the interaction with disable_request_compression

src/RequestCompressionMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return false;
+    }
+
+    private function hasStreamingTraitWithoutRequiresLength($command, $operation)
+    {
+        foreach ($operation->getInput()->getMembers() as $name => $member) {
+            if (!empty($member['streaming'])
+                && empty($member['requiresLength'])
+                && isset($command[$name])) {
```
SamRemis, line 125: Super small nit: I feel like the last of these 3 should be first logically for readability.  Functionally it is the same though, just feels strange to check something before we know it's even relevant to the problem at hand.

src/RequestCompressionMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+        if (isset($command['@request_min_compression_size_bytes'])
+            && is_int($command['@request_min_compression_size_bytes'])
+            && $this->isValidCompressionSize($command['@request_min_compression_size_bytes'])
+        ) {
+            $this->minimumCompressionSize = $command['@request_min_compression_size_bytes'];
+        }
+        $nextHandler = $this->nextHandler;
+        $operation = $this->api->getOperation($command->getName());
+        $requestBodySize = $request->getBody()->getSize();
+        $compressionTrait = isset($operation['requestcompression'])
```
SamRemis, line 55: shouldn't `requestcompression` be `requestCompression`?

src/RequestCompressionMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+        if (!$this->shouldCompressRequestBody(
+            $compressionTrait,
+            $command,
+            $operation,
+            $requestBodySize
+        )) {
+            return $nextHandler($command, $request);
+        }
+
+        $this->encodings = $compressionTrait['encodings'];
```
SamRemis, line 68: This logic works fine since `encodings` is required by $compressionTrait; do we want to throw an error if it's not there instead of the default "could not access array index" error?

src/RequestCompressionMiddleware.php
stobrien89: It will be modeled as lowercase

src/RequestCompressionMiddleware.php
stobrien89: This is a required field that will be caught by modeling tools, so I don't see it as necessary

src/RequestCompressionMiddleware.php
stobrien89: ack

src/Configuration/ConfigurationResolver.php
```diff
@@ -0,0 +1,180 @@
...
+    private static function getDefaultConfigFilename()
+    {
+        if ($filename = getenv(self::ENV_CONFIG_FILE)) {
+            return $filename;
+        }
+        return self::getHomeDir() . '/.aws/config';
+    }
+
+    /**
+     * Normalizes string values pulled out of ini files and
```
SamRemis, line 156: pulled out of -> retrieved from?

src/RequestCompressionMiddleware.php
```diff
@@ -0,0 +1,160 @@
...
+                && empty($member['requiresLength'])
+            ){
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private function determineMinimumCompressionSize($config) {
+        if (isset($config['request_min_compression_size_bytes'])) {
```
SamRemis, line 134: What gets returned when this isn't set?  I think this check is not necessary since it's always set in `_apply_request_min_compression_size_bytes`

SamRemis (approved):

Issue #2681, opened 2023-04-21
HashingStream incompatible with Psr\Http\Message\StreamInterface still failing


martinzaleski:
### Describe the bug

I'd like to re-open this issue: https://github.com/aws/aws-sdk-php/issues/2644

I strongly believe there's actually a bug that has been unresolved.
The signature of the latest versions of `psr/http-message` and `aws-sdk-php` do not match and I'm still getting the same error.
I fixed it pinning to version `1.0.1` of `psr/http-message` but of course that's not the optimal solution.

### Expected Behavior

Not throw the exception.

### Current Behavior

It runs the same exception due to incompatibility between signatures: 
```
Declaration of Aws\HashingStream::seek($offset, $whence = Aws\SEEK_SET) must be compatible with Psr\Http\Message\StreamInterface::seek(int $offset, int $whence = Psr\Http\Message\SEEK_SET): void in /app/vendor/aws/aws-sdk-php/src/HashingStream.php on line 53
```

### Reproduction Steps

It's a compilation error so no steps to reproduce. I use the latest version of `psr/http-message`

### SDK version used

3.263.15

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.1 OS

SamRemis:
Hi @martinzaleski,
Unfortunately, we can't fix this for now since this repository supports PHP versions down to 5.5 and this would be a breaking change.  We have started a [deprecation campaign](https://github.com/aws/aws-sdk-php/discussions/2638) announcing that we will no longer support versions below 7.2.5 after the minor version bump that will happen on August 15th, 2023.  

Once that happens, we will be able to add return types without breaking customers code.

-Sam

martinzaleski:
Hello @SamRemis, thank you for your prompt response.
I totally understand. In that case, wouldn't it be fair from you to pin `psr/http-message` yourselves in the SDK code?

If I'm not missing anything, the multipart upload to S3 right now is completely broken unless you pin to the aforementioned dependency version. It took us quite some time to understand the issue and we had some bugs in production due to this.

Warm regards,
Martin

walkonthemarz:
> Hello @SamRemis, thank you for your prompt response. I totally understand. In that case, wouldn't it be fair from you to pin `psr/http-message` yourselves in the SDK code?
> 
[...]

PR #2697, opened 2023-05-22
Support latest guzzle promises

GrahamCampbell:


SamRemis:
Hi Graham,
[We announced](https://github.com/aws/aws-sdk-php/discussions/2638) dropping support for versions below 7.2.5 in February which will be effective on August 15th, 2023.  Totally agree with you here.

GrahamCampbell:
Excellent news. 🎉

bmckay959:
I'm also running into the issue mention in #2681 

Is there a recommended fix in the meantime before August 15th?

GrahamCampbell:
You need to upgrade to the very latest version of this SDK, which may result in other packages begin downgraded, but if you do that, the SDK will work properly.

PR #2705, opened 2023-06-07
Allow `psr/http-message` v2

ruudk:
The SDK should not prevent installing v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-json": "*",
         "ext-simplexml": "*",
         "aws/aws-crt-php": "^1.0.4",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.0 || ^2.0"
     },
     "require-dev": {
         "composer/composer" : "^1.10.22",


SamRemis:
Hi @ruudk,

Thanks for opening the PR; unfortunately, we can't merge this til August.  We originally restricted the version because of [this issue](https://github.com/aws/aws-sdk-php/issues/2681) and others like it.  `Aws\HashingStream` is one of a few classes in the SDK that implement an interface from `psr7/http-message`; since we support down to PHP 5.5, we can't implement the class with return types until we drop support for versions below 7.2.5.  

We will be [dropping support for those versions](https://github.com/aws/aws-sdk-php/discussions/2638) on August 15th, and once we do, we can properly implement those interfaces in a way that is compatible with version 2 of `http-message`.  

-Sam

SamRemis:
Closing as this will be addressed in https://github.com/aws/aws-sdk-php/pull/2707/

Thank you for bringing this to our attention

PR #2707, opened 2023-06-08
Deprecation below 7.2.5

SamRemis:
This pull request is a work in progress and will be added to over time.  This will be merged before the release on Tuesday, August 15th, 2023 and will be a minor version bump that ends support for php versions below 7.2.5.

PR #2719, opened 2023-06-28
Update guzzle promises to 2.x

paulsheldrake:
*Description of changes:*

Updating to Guzzle Promises 2.x as this is a requirement of Drupal core


SamRemis:
Hi @paulsheldrake,

We will be updating this on August 15th with [this PR](https://github.com/aws/aws-sdk-php/pull/2707/files).  We implement interfaces from guzzle promises, and if we were to merge this now there would be users whose code would break due to us not implementing the interfaces with return types.  

Luckily, we [announced the deprecation of versions below 7.2.5](https://github.com/aws/aws-sdk-php/discussions/2638) back in February with a date of August 15th.  

Thank you for the PR,

-Sam

PR #2722, opened 2023-07-04
Alternative credential retrieval - threadsafe

paullallier:
Similar to PR [2155](https://github.com/aws/aws-sdk-php/pull/2155)

*Description of changes:*
Fallback to $_SERVER variable for AWS credentials to avoid using non-threadsafe putenv()

Would you consider allowing this change so that we can avoid having to use the non-threadsafe putenv() function to set AWS credentials - which we're storing in a .env file?

I've tried to write a test for it - not sure it's going to work yet.


paullallier:
Found a minor problem. This doesn't seem to fall through to null when `$_SERVER[self::ENV_SESSION]` is undefined.  I'm tempted to remove that one since I don't need it, but maybe there's a better way?
`getenv(self::ENV_SESSION) ?: $_SERVER[self::ENV_SESSION] ?: NULL`

arnaubuch:
Could we get this merged eventually? Thanks!

paullallier:
Any chance of a review and comments on how we could make this an acceptable merge, please?

src/Credentials/CredentialProvider.php
```diff
@@ -289,8 +289,8 @@ public static function env()
     {
         return function () {
             // Use credentials from environment variables, if available
-            $key = getenv(self::ENV_KEY);
-            $secret = getenv(self::ENV_SECRET);
+            $key = getenv(self::ENV_KEY) ?: $_SERVER[self::ENV_KEY];
```
stobrien89, line 292: The `$_SERVER` references should check to see if it is set.  Also, we should fall back to false if it isn't. I think you could add a null coalescing operator to what's already here and add `false` on the right side of the operand.

stobrien89 (changes requested):

paullallier:
I'm not sure why I thought @server would set the $_SERVER variable for the test. Fixed this, but is there a cleaner option? 

Sorry - I should have run the tests locally (which I have now down with PHP 8.3 at least).

paullallier:
Sorry. I'd love to get this into a mergeable state, but getting it to pass the tests on all the test suites appears to be beyond my skill level. :-(

stobrien89:
Hi @paullallier,

Apologies for the delay.  We'll get these fixed up and get the PR merged as soon as possible.

stobrien89:
Hey @paullallier,

Sorry for the very delayed response.  I was reviewing this today and I don't think this is something we can merge.  There are too many tests that depend on credential 'emptiness' and there's not a great way for us to set these values and restore them when running unit tests. Something we'll keep in mind if we end up overhauling our credential providers though.  Thanks again for the PR!

PR #2723, opened 2023-07-09
Add "after" callback for the Transfer class

### LINKED ISSUE #2725: Add the possibility to implement a callback after each file transfer is fulfilled when using the Transfer class in S3
### Describe the feature

The SDK offers the possibility to transfer files to and from an S3 Object Storage using the Transfer class.
Currently it is possible to implement a callback before each Command is executed but there is no possibility of implementing a callback after the Command has been executed.

Example of the "before" callback:
```php
$s3Client->uploadDirectory($path, $bucket, null, [
    'before' => function (\Aws\Command $command) {
        // Do something with the $command
    }
]);
```

I would like to request the feature that we could write another type of callback to be executed "after" the command is executed.
Such a feature would be easy to implement with a minimum amount of changes and would open more of the S3 transfer SDK possibilities.

### Use Case

As I had to implement a buffered S3 client (the buffer being represented by saving files on disk before they will be transferred to S3 later on) I could not easily cleanup the files that were already transferred because there is no "after" callback function.

With the proposed solution one could use something like:
```php
$s3Client->uploadDirectory($path, $bucket, null, [
    'after' => function(\Aws\Result $result, int $index, \GuzzleHttp\Promise\Promise $aggregatePromise) {
        // Do something with $result or check $aggregatePromise
    }
]);
```

### Proposed Solution

Implementing an "after" callback in a similar way to how the "before" callback is working now.
Basically sending a callable as part of the options parameter to the existing functions.
For the upload methods the callable would just need to be passed further to the `Promise\Each::ofLimitAll` as the 3rd parameter (onFulfilled) when creating the upload promise.
For the download methods the callable would just need to be sent as part of the `CommandPool` constructor options as `fulfill`.

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.275.6

PHP 7.2.24-0ubuntu0.18.04.17

madalindk:
Fixes #2725 

*Description of changes:*
Added possibility to execute a callback function after a transfer is fulfilled when using the Transfer class. Implemented similarly to the way the 'before' callback works.

While using the transfer class, particularly the upload methods, I could see that there is a "before" callback I can use but there is no "after" callback. I needed such an "after" callback to cleanup the files as they are getting uploaded. Because of this I have implemented an "after" callback similar to the "before" callback which uses the fulfill callback option. It now can be used as follows:

```php
$s3Client->uploadDirectory($path, $bucket, null, [
    'after' => function(\Aws\Result $result, int $index, \GuzzleHttp\Promise\Promise $aggregatePromise) {
        // Do something with $result or check $aggregatePromise
    }
]);
```

I think such a feature might be helpful to others in the future as well. I have created this PR to see if you also agree having an "after" callback would be a good idea and the code is just a fast copy paste from the "before" callback. If there is agreement that this would be a good idea, I can/will write some more tests and proper documentation about this new callback.


stobrien89:
Hi @Mastertrap21,

I think we'd be open to accepting something like this.  Seems like a perfectly reasonable way to do this and I'd imagine others may like the option to perform a task after transfer operation.  

We generally like to have a feature request (in the form of an issue) open before we decide to merge a community requested feature, so if you could do that (feel free to copy over some of the description from this PR and link it as well), it'd be greatly appreciated.  Some people pay more attention to issues on our repository and we like to give others the chance upvote or chime in about these sorts of things.

Mastertrap21:
Hi @stobrien89, 

Thank you for your quick response, openness and guidance regarding this feature, it's always nice to contribute to a well maintained project.

I have now opened a feature request in the form of an issue here: https://github.com/aws/aws-sdk-php/issues/2725
I will write some specific tests for the new callback tomorrow when I get some extra time.
I'm not sure about the existing checks, I can see that it is failing because of the added nextrelease changelog JSON. Is that something I should look into?

stobrien89 (changes requested):
@madalindk,  

LGTM, just need to move that changelog entry file to `.changes/nextrelease/`. from the SDK base directory.  Looks like it ended up in some old build artifacts.

stobrien89 (approved):

PR #2726, opened 2023-07-11
feature: resolve region from profile and env

stobrien89:
Description of changes:

Adds the ability to configure region via env var and shared config file. This would mean the region is no longer required in the client configuration. This branches from https://github.com/aws/aws-sdk-php/pull/2693, so the changes in question here are _apply_region, _resolve_region some updates to _missing_region. This is an initial draft/proof-of-concept.


src/ClientResolver.php
```diff
@@ -533,10 +534,10 @@ public static function _apply_disable_request_compression($value, array &$args)
             $value = $value();
         }
         if (!is_bool($value)) {
-           throw new IAE(
-              "Invalid configuration value provided for 'disable_request_compression'."
-              . " value must be a bool."
-           );
+            throw new IAE(
```
SamRemis, line 537: Hard to see this on GH, but looks like this spacing should be undone

src/ClientResolver.php
```diff
@@ -92,8 +92,9 @@ class ClientResolver
         'region' => [
             'type'     => 'value',
```
SamRemis, line 93: Please add a changelog file

src/ClientResolver.php
```diff
@@ -1115,15 +1116,33 @@ public static function _default_signing_region(array &$args)
...
+    }
+
+    public static function _default_region(&$args)
+    {
+        return ConfigurationResolver::resolve('region', '', 'string');
+    }
+
     public static function _missing_region(array $args)
     {
         $service = isset($args['service']) ? $args['service'] : '';
```
SamRemis, line 1134: When will service ever be blank?  Is there a good use case for resolving a client without a service using this file?

src/ClientResolver.php
stobrien89: This is actually right.  Just verified locally.

SamRemis (approved):

Issue #1239, opened 2017-04-05
Feature: Configure Services' Endpoint URLs via Environment Variables


bradynpoulsen:
I've been looking the documentation about environment variables and see that you are able to configure credentials, regions, and profiles using environment variables, but I do not see a way to configure the endpoint url option to point to our internal proxy.

My suggestion is to configure the endpoint url if either the `AWS_<REGION>_<SERVICE>_ENDPOINT` or `AWS_<SERVICE>_ENDPOINT` (in that order of priority) environment variables are available.

I suggest making this change in one of these locations (that I have found so far) in order of suggestion preference:
 * [`Aws\ClientResolver::_default_endpoint_provider`](https://github.com/aws/aws-sdk-php/blob/master/src/ClientResolver.php#L615-L619)
 * [`Aws\Endpoint\PartitionEndpointProvider::getPartition`](https://github.com/aws/aws-sdk-php/blob/master/src/Endpoint/PartitionEndpointProvider.php#L38-L47)

For reference purposes:
 * [`Aws\Credentials\CredentialsProvider`](https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/CredentialProvider.php#L214-L229) use of environment variables
 * [PHP `getenv` Documentation](http://php.net/manual/en/function.getenv.php)

bradynpoulsen:
As a heads up, I am starting a PR to provide an implementation of this

imshashank:
@bradynpoulsen You can provide the endpoint as part of the client.

Please have a look at the documentation here: https://docs.aws.amazon.com/aws-sdk-php/v3/guide/guide/configuration.html#endpoint

We would like to know more about your scenario and why would you prefer setting up endpoints as environment variable.

bradynpoulsen:
We have 2 different use cases right now.

1) Local Environment
We inherit everything from the environment currently and are wanting to stub out the APIs. But it feels weird that we can currently configure everything through the environment but have to have some sort of if condition to add the endpoint configuration only during non-production environments.

2) Feature Deployments via Docker
We are deploying our application in a disposable way for integration testing against feature/bug branches to verify issues are being resolved. Same thing as above that everything is currently configured through the environment (excluding the version), but they have to go through a proxy to pass through the outbound firewall.

Maybe it's just me, but it feel inconsistent to allow key things (credentials/environment) to easily be configured through environment variables, but then prevent something like the endpoint configuration from being allowed.

imshashank:
[...]

PR #2737, opened 2023-07-25
feature: add support for per-service endpoint url configuration

stobrien89:
*Issue #, if available:*
https://github.com/aws/aws-sdk-php/issues/1239

*Description of changes:*
Adds the ability to configure endpoints on a per-service basis via the environment and ini file.  Order of precedence is as follows:

- `endpoint` client configuration
- `AWS_ENDPOINT_URL_<SERVICE_IDENTIFIER>` env var
- `services` endpoint in shared config  example: 
```
[default]
region = us-west-2
services = my-services

[services my-services]
s3 =
  endpoint_url = https://s3-endpoint-override.aws
```
- `AWS_ENDPOINT_URL` env var
- `endpoint_url` in shared config.  example:
```
[default]
endpoint_url = http://localhost:1234
```

For backward-compatibility, we will continue to support the `endpoint` client constructor option, however, `endpoint_url` will take precedence, if resolved from any one of the above-mentioned options.


src/ClientResolver.php
```diff
@@ -287,6 +287,20 @@ class ClientResolver
...
+            'type'      => 'value',
+            'valid'     => ['bool'],
+            'doc'       => 'Set to true to disable configured endpoint urls.',
+            'fn'        => [__CLASS__, '_apply_ignore_configured_endpoint_urls'],
+            'default'   => [__CLASS__, '_default_ignore_configured_endpoint_urls'],
+        ],
+        'endpoint_url' => [
+            'type'      => 'value',
+            'valid'     => ['string'],
+            'doc'       => 'The full URI of the webservice. This is only required when connecting to a custom endpoint (e.g., a local version of S3). This will take precedence over the `endpoint` configuration option.',
```
SamRemis, line 300: This doc could be a bit clearer about how this differs from endpoint

src/AwsClient.php
```diff
@@ -554,7 +554,11 @@ private function setClientBuiltIns($args)
         $config = $this->getConfig();
         $service = $args['service'];
 
-        $builtIns['SDK::Endpoint'] = isset($args['endpoint']) ? $args['endpoint'] : null;
+        $builtIns['SDK::Endpoint'] = isset($config['endpoint_url'])
```
SamRemis, line 557: Should we merge this soon so we can add this to the deprecation with new syntax?  Like `$config['endpoint_url'] ?? $args['endpoint'] ?? null;`

src/ClientResolver.php
```diff
@@ -287,6 +287,20 @@ class ClientResolver
             'default'   => false,
             'fn'        => [__CLASS__, '_apply_suppress_php_deprecation_warning']
         ],
+        'ignore_configured_endpoint_urls' => [
+            'type'      => 'value',
+            'valid'     => ['bool'],
+            'doc'       => 'Set to true to disable configured endpoint urls.',
```
SamRemis, line 293: endpoint urls that have been configured where?

stobrien89: Yeah, could use some clarification here.

src/ClientResolver.php
stobrien89: It doesn't really differ from `endpoint`.  It's essentially the same thing with different a different name and configuration locations.  I can add that though.

src/Configuration/ConfigurationResolver.php
```diff
@@ -30,7 +30,8 @@ public static function resolve(
         $key,
         $defaultValue,
         $expectedType,
-        $config = []
+        $config = [],
```
SamRemis, line 33: Aren't config and options _basically_ synonyms?  Why are we differentiating them here?  Seems like we could flatten this into one array

src/Configuration/ConfigurationResolver.php
```diff
@@ -89,8 +96,13 @@ public static function env($key, $expectedType)
...
      * @return null | mixed
      */
-    public static function ini($key, $expectedType, $profile = null, $filename = null)
-    {
+    public static function ini(
+        $key,
+        $expectedType,
+        $profile = null,
+        $filename = null,
+        $options = []
```
SamRemis, line 104: Let's stick with config if we are going to stick with config

src/functions.php
```diff
@@ -503,6 +503,42 @@ function boolean_value($input)
...
+    foreach ($lines as $line) {
+        $line = trim($line);
+
+        if (empty($line) || strpos($line, ';') === 0 || strpos($line, '#') === 0) {
+            continue; // Ignore empty lines and comments
+        }
+
+        if (strpos($line, '[') === 0) {
+            $current_service = '';
+            $section_start = (trim($line, '[]') === $section_name);
```
SamRemis, line 523: So this takes an input, looking for a section name.  Would it make sense to move the logic around a bit?  Maybe make an explicit iterator, and then look for the right section and do the parsing from there until we hit the end of the section?

src/Configuration/ConfigurationResolver.php
stobrien89: `config` is passed in from the client args/config in the client constructor, but I suppose we could flatten additional options into it.

SamRemis (approved):

src/ClientResolver.php
```diff
@@ -281,6 +281,20 @@ class ClientResolver
...
             'default'   => true,
         ],
+        'ignore_configured_endpoint_urls' => [
+            'type'      => 'value',
+            'valid'     => ['bool'],
+            'doc'       => 'Set to true to disable endpoint urls configured using the `endpoint_url` configuration option.',
+            'fn'        => [__CLASS__, '_apply_ignore_configured_endpoint_urls'],
+            'default'   => [__CLASS__, '_default_ignore_configured_endpoint_urls'],
+        ],
+        'endpoint_url' => [
```
SamRemis, line 291: Since this does the same thing as endpoint, why not just have endpoint look in the config variables and environment variables?

src/ClientResolver.php
```diff
@@ -1116,6 +1130,64 @@ public static function _default_signing_region(array &$args)
...
+            return;
+        }
+
+        $args['config']['endpoint_url'] = $value;
+    }
+
+    public static function _default_endpoint_url(array &$args)
+    {
+        if ($args['config']['ignore_configured_endpoint_urls']
+            || !self::isValidService($args['service'])
```
SamRemis, line 1160: This line is just for tests?

src/functions.php
```diff
@@ -503,6 +503,51 @@ function boolean_value($input)
     return null;
 }
 
+/**
+ * Checks if an input is a valid epoch time
```
SamRemis, line 507: This doesn't seem accurate

src/Configuration/ConfigurationResolver.php
```diff
@@ -100,6 +114,20 @@ public static function ini($key, $expectedType, $profile = null, $filename = nul
         // Use INI_SCANNER_NORMAL instead of INI_SCANNER_TYPED for PHP 5.5 compatibility
         //TODO change after deprecation
         $data = @\Aws\parse_ini_file($filename, true, INI_SCANNER_NORMAL);
+
+        if (isset($options['section'])
+            && isset($options['subsection'])
+            && isset($options['key']))
+        {
+            return self::retrieveValueFromIniSubsection(
```
SamRemis, line 122: As this is expected to grow and add more config options to individual services, does it make more sense to grab the subsection early on in the client resolver instead and reference it here and have that be referenced throughout the resolution of the client?

src/functions.php
```diff
@@ -503,6 +503,51 @@ function boolean_value($input)
...
+ * @param $filename
+ * @return array
+ */
+function parse_ini_section_with_subsections($filename, $section_name) {
+    $config = [];
+    $ini_content = file_get_contents($filename);
+    $lines = explode("\n", $ini_content);
+
+    $section_start = false;
+    $current_subsection = '';
```
SamRemis, line 519: Since we know what section we are looking for, would it make sense to instead search the file straight for the section name (using a regex maybe?), then parse each line until the end of the section

SamRemis: If not a regex, then we know the exact string we are looking for, maybe just checking if the line equals that with a trim(), then once we have found it using another loop to parse each value into the config array until we are done

src/ClientResolver.php
stobrien89: More or less. There are other places in the clientresolver where we don't require a service or set a default value when a service is not valid/present.

src/ClientResolver.php
stobrien89: To ensure consistency in naming across all sources.

SamRemis: Alternate option, we could add an alias field and a little logic in the resolver code so that it gets taken care of there

src/functions.php
```diff
@@ -503,6 +503,63 @@ function boolean_value($input)
...
+    $stream = fopen($filename, 'r');
+
+    if (!$stream) {
+        return $config;
+    }
+
+    $current_subsection = '';
+
+    while (!feof($stream)) {
+        $line = trim(fgets($stream));
```
SamRemis, line 524: should we condense these few lines into another function called read_ini_line?  Just to avoid the duplication of the trim, the fgets, and the empty/comment check

SamRemis (approved):

src/ClientResolver.php
```diff
@@ -993,6 +1001,11 @@ public static function _apply_user_agent($inputUserAgent, array &$args, HandlerL
 
     public static function _apply_endpoint($value, array &$args, HandlerList $list)
     {
+        if (empty($value)) {
```
SamRemis, line 1004: Can we move this down in resolution order instead?

src/CloudSearchDomain/CloudSearchDomainClient.php
```diff
@@ -35,6 +36,7 @@ public static function getArguments()
             // (e.g. http://search-blah.{region}.cloudsearch.amazonaws.com)
             return explode('.', new Uri($args['endpoint']))[1];
         };
+        unset($args['endpoint']['default']);
```
SamRemis, line 39: Why is this necessary?  This seems like an endpoint is always required, so it should never go to the default?

SamRemis (approved):

src/CloudSearchDomain/CloudSearchDomainClient.php
stobrien89: The client resolver won't handle the 'required' property if there's a default

PR #2763, opened 2023-08-25
Updated CredentialProvider.php to get the credentials directly from t…

mimarox:
…he config object

*Issue #, if available:*

*Description of changes:*

Updated CredentialProvider.php to get the credentials directly from the config object.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -87,6 +87,7 @@ class CredentialProvider
         $profileName = getenv(self::ENV_PROFILE) ?: 'default';
 
         $defaultChain = [
+            'config' => self::config($config),
             'env' => self::env(),
             'web_identity' => self::assumeRoleWithWebIdentityCredentialProvider($config),
         ];
@@ -99,6 +100,7 @@ class CredentialProvider
                 self::getHomeDir() . '/.aws/config',
                 $config
             );
+
             $defaultChain['process_credentials'] = self::process();
             $defaultChain['ini'] = self::ini();
             $defaultChain['process_config'] = self::process(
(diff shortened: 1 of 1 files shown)


src/Credentials/CredentialProvider.php
```diff
@@ -872,11 +891,11 @@ private static function getSsoCredentials($profiles, $ssoProfileName, $filename,
             $ssoSession = $profiles['sso-session ' . $ssoProfile['sso_session']];
             $ssoOidcClient = new Aws\SSOOIDC\SSOOIDCClient([
                 'region' => $ssoSession['sso_region'],
-                'version' => '2019-06-10',
+                'version' => 'latest',
```
GrahamCampbell, line 894: Why has this changed?

GrahamCampbell (changes requested):

src/Credentials/CredentialProvider.php
mimarox: Sorry, I didn't mean to commit that change. You can savely revert it. Thx.

SamRemis: This line is no longer significant and can now be safely deleted; the `version` input parameter is a consequence of the fact that this SDK is one of the oldest ones we have and the vision for service versioning was very different at the time that this was written.  Services now rarely come out with new versions, and when they do we require those versions to be backward compatible with previous versions.  

Every other SDK defaults to `latest`; we recently made a change so that customers no longer have to specify the version when initializing a client and it will always default to latest.

SamRemis:
What benefit does this PR have over the usual syntax for hard coding the credentials?  Something like the following:
```
$stsClient = new Aws\Sts\StsClient([
    'region' => 'us-east-2',
    'credentials' => [
        'key' => 'foo',
        'secret' => 'bar'
    ]
]);
```

It seems like this is asking for an alternative way to hard-code the credentials and add that to the default provider chain.  If you're adding it to the config option anyway, why not just make a custom provider to get it from the config or just initialize it directly using the above syntax?

stobrien89:
@mimarox,

Thanks for the PR.  We generally like to have an issue open as a feature request (to explain the need/benefit for/of requested changes) prior to merging community PRs.  Since we haven't been provided with any reason or use case where this functionality would be necessary, going to close this for now.  Let us know if you have any questions, or if you'd like to have this PR re-opened.

PR #2781, opened 2023-09-22
Aws models sync

SamRemis:
Sync endpoint files with the source of truth in aws-models.  Mostly stylistic changes here, with a few bugfixes.


stobrien89 (approved):
:shipit:

PR #2790, opened 2023-10-06
enhancement: Endpointv2 Middleware

stobrien89:
*Description of changes:*
This is a draft for proposed changes to endpoint resolution, specifically middleware.  These changes prepend `EndpointV2Middleware` to the build step, which passes a resolved endpoint object to the serializer rather than resolving the endpoint within the serializer itself.  


src/Api/Serializer/RestSerializer.php
```diff
@@ -49,33 +50,28 @@ public function __construct(Service $api, $endpoint)
...
-                $command,
-                $operation,
-                $commandArgs,
-                $clientArgs,
-                $headers
-            );
-            $this->endpoint = new Uri($this->endpoint);
+        $headers = $opts['headers'] ?? [];
+
+        if ($endpoint instanceof RulesetEndpoint) {
```
SamRemis, line 61: This code appears to be the same for all the serializers; is there a reason it's not in a method on the EndpointV2SerializerTrait?

SamRemis (approved):

yenfryherrerafeliz (approved):

PR #2809, opened 2023-11-01
enhancement: querycompatible input middleware

stobrien89:
*Description of changes:*

Adds middleware which compares input values against their modeled type and casts to the modeled type if necessary.  This applies only to scalar input members for query compatible (migrated from query to json) services.  Edit: Booleans are not included in the processing because the validation middleware runs an `is_bool` check on these members and rejects them if they are not the correct type.


src/QueryCompatibleInputMiddleware.php
```diff
@@ -0,0 +1,234 @@
...
+    {
+        switch ($type) {
+            case 'string':
+                return is_string($value);
+            case 'integer':
+            case 'long':
+                return is_int($value);
+            case 'float':
+                return is_float($value);
+            default:
```
kuhe, line 208: is boolean also one of these?

kuhe (approved):

tests/QueryCompatibleInputMiddlewareTest.php
```diff
@@ -0,0 +1,256 @@
...
+use Aws\Api\Service;
+use Aws\Middleware;
+use Aws\QueryCompatibleInputMiddleware;
+use GuzzleHttp\Psr7\Request;
+use Yoast\PHPUnitPolyfills\TestCases\TestCase;
+
+/**
+ * @covers \Aws\QueryCompatibleInputMiddleware
+ */
+class QueryCompatibleInputMiddlewareTest extends TestCase
```
kstich, line 16: Should have a test assserting that non query compatible clients aren't having the handler added.

tests/QueryCompatibleInputMiddlewareTest.php
```diff
@@ -0,0 +1,256 @@
...
+            . "`. The modeled type is `{$type}`."
+        );
+        $service = $this->generateTestService();
+        $client = $this->generateTestClient($service);
+        $command = $client->getCommand(
+            'FooOperation',
+            [$inputParam => $inputValue]
+        );
+
+        $list = new HandlerList();
```
kstich, line 41: Why manually construct a `HandlerList`? Can't the handler be configured on the test client, then these tests use the mocked client?

tests/QueryCompatibleInputMiddlewareTest.php
```diff
@@ -0,0 +1,256 @@
...
+        $command = $client->getCommand(
+            'FooOperation',
+            $input
+        );
+
+        $list = new HandlerList();
+        $list->setHandler(function ($command) {
+            return "success";
+        });
+        $list->prependBuild(Middleware::tap(function (Command $command) {
```
kstich, line 131: Can this be added to the client's `HandlerList` instead of using a fully rebuilt one?

kstich (changes requested):

tests/QueryCompatibleInputMiddlewareTest.php
stobrien89: Yeah, this could definitely be done in a more straightforward way.  I borrowed some of this from another middleware test class and didn't give it a close enough look.

kstich (approved):

PR #2815, opened 2023-11-06
Fix Log Prefix Formation with Multi-Account Organization Support in C…

laiquesantana:
Description:
This PR introduces a necessary change to the AWS SDK PHP to address an issue encountered during the configuration of AWS CloudTrail with the 'Enable for all accounts in my organization' option. This setting introduces an additional directory layer after AWSLogs to accommodate organization IDs, which was not accounted for in the existing SDK logic. As a result, log retrieval for multi-account setups within an organization was failing. The implemented fix normalizes the log prefix, ensuring that the SDK can successfully access and iterate over log files in S3 buckets across multiple AWS accounts with the organization feature enabled.

Impact:
This update ensures seamless functionality for new multi-account organizational setups while preserving the behavior for existing single-account implementations. It's an optional adjustment that only affects users who opt into organization-level logging.

### DIFF
--- a/src/CloudTrail/LogFileIterator.php
+++ b/src/CloudTrail/LogFileIterator.php
@@ -20,7 +20,7 @@ class LogFileIterator extends \IteratorIterator
 {
     // For internal use
     const DEFAULT_TRAIL_NAME = 'Default';
-    const PREFIX_TEMPLATE = 'prefix/AWSLogs/account/CloudTrail/region/date/';
+    const PREFIX_TEMPLATE = 'prefix/AWSLogs/organization/account/CloudTrail/region/date/';
     const PREFIX_WILDCARD = '*';
 
     // Option names used internally or externally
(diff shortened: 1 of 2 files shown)


src/CloudTrail/LogFileIterator.php
```diff
@@ -20,7 +20,7 @@ class LogFileIterator extends \IteratorIterator
 {
     // For internal use
     const DEFAULT_TRAIL_NAME = 'Default';
-    const PREFIX_TEMPLATE = 'prefix/AWSLogs/account/CloudTrail/region/date/';
+    const PREFIX_TEMPLATE = 'prefix/AWSLogs/organization/account/CloudTrail/region/date/';
```
stobrien89, line 23: This might be a breaking change.  Is there a way to tell if 'Enable for all accounts in my organization' is enabled first?

stobrien89 (changes requested):
Hi @laiquesantana,

Apologies for the delay.  I'm not familiar with the specific cloudtrail setting you mentioned, but it seems like this should be configurable (pass in the prefix) or somehow based on whether or not the setting is enabled.

PR #2835, opened 2023-11-22
bugfix: dot segment handling

stobrien89:
*Description of changes:*
Bypasses Guzzle's UriResolver for request paths that contain `../` parent path references.  Previously, the UriResolver would strip dot segments, leading to a situation where objects other than the intended target could be created or accessed.


src/Api/Serializer/RestSerializer.php
```diff
@@ -248,14 +248,24 @@ function (array $matches) use ($varDefinitions) {
             $relative = substr($relative, 1);
         }
 
-        //Append path to endpoint when leading '//...' present
-        // as uri cannot be properly resolved
-        if ($this->api->isModifiedModel()
-            && strpos($relative, '//') === 0
+        //Append path to endpoint when leading '//...' or
+        // '../' present as uri cannot be properly resolved
+        if (($this->api->isModifiedModel()
```
SamRemis, line 253: these parentheses look like they were added by mistake

src/Api/Serializer/RestSerializer.php
```diff
@@ -248,14 +248,24 @@ function (array $matches) use ($varDefinitions) {
...
         ) {
             return new Uri($this->endpoint . $relative);
         }
 
+        if ($this->api->getServiceName() === 's3'
+            && strpos($relative, '../') !== false
+        ){
+            return new Uri(
+                $this->endpoint
+                . (strpos($relative, '/') !== 0 ? '/' : '')
```
SamRemis, line 266: So this is just making relative start with a slash if it doesn't?  I might refactor this to say 
```
if ($relative && $relative[0] != '/') {
     $relative = '/' . $relative;
}
 return new Uri($this->endpoint . $relative);
```

if you think that looks better.

SamRemis (changes requested):

src/Api/Serializer/RestSerializer.php
SamRemis: Also, could we just move this up to the if statement above and short-circuit this there?

SamRemis (approved):

SamRemis (approved):

PR #2838, opened 2023-11-28
Add support for psr/simple-cache v2 and v3

rmn-eloha-jouny:
*Issue #, if available:*
I don't think i need to create an issue for this.

*Description of changes:*
I need this change when peer-dependencies require different version of the `psr/simple-cache` package.
In my case, the package `mobiledetect/mobiledetectlib` require version 3. Since these versions do not change public interface (only type safety was added) i think it is safe to allow this.

### DIFF
--- a/.changes/nextrelease/simple-cache.json
+++ b/.changes/nextrelease/simple-cache.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "",
+    "description": "Adds support for psr/simple-cache v2 and v3"
+  }
+]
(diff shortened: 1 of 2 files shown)


Timic3:
We are also blocked by this issue. Can this get merged any time soon?

stobrien89 (approved):

stobrien89:
@rmn-eloha-jouny,

Thanks for the contribution.  Can you add a changelog entry by creating a json file (e.g. `simple-cache.json`) in `.changes/nextrelease`?  you can label this as an enhancement.  It should follow the format below (you can feel free to copy/paste this).

```
[
  {
    "type": "enhancement",
    "category": "",
    "description": "Adds support for psr/simple-cache v2 and v3"
  }
]
```

PR #2857, opened 2023-12-14
bugfix: cloudfront sigv4a

stobrien89:
*Description of changes:*
Standardizes CRT signing config for S3 and non-S3 services.  Fixes flaky integ test and adds another for CloudfrontKeyValueStore clients.


src/Signature/SignatureV4.php
```diff
@@ -502,15 +503,20 @@ private function CRTRequestFromGuzzleRequest($request)
...
+    protected function signWithV4a(
+        CredentialsInterface $credentials,
+        RequestInterface $request,
+        $signingService,
+        SigningConfigAWS $signingConfig = null
+    ){
         $this->verifyCRTLoaded();
-        $credentials_provider = $this->createCRTStaticCredentialsProvider($credentials);
-        $signingConfig = new SigningConfigAWS([
+        $signingConfig = $signingConfig ?: new SigningConfigAWS([
```
yenfryherrerafeliz, line 513: nit: Should you use the "??" coalescing operator instead of "?:". Because $signingConfig should be either an instance of SigningConfigAWS or null, but we could have the impression that we can pass other types such as boolean, string, etc., and it will be valid.

stobrien89: I think the type hint takes care of that impression, but the null coalescing operation is technically more correct here. I'll change it

yenfryherrerafeliz (approved):

Issue #2847, opened 2023-11-29
Unable to override IMDSv2 service URL


ibexa-yuna:
### Describe the bug

I'm using aws_signing_helper in order to facilitate IAM Roles Anywhere setup.
I'm overriding AWS_EC2_METADATA_SERVICE_ENDPOINT to http://127.0.0.1:9911/
However, the application doesn't use the new endpoint and still trying to fetch InstanceCredentials from 169.254.169.254.

### Expected Behavior

The application would acquire temporary credentials from aws_signing_helper via IAM Roles Anywhere by talking to a service endpoint defined by the environment variable AWS_EC2_METADATA_SERVICE_ENDPOINT.

### Current Behavior

The application fails to authenticate with an error of

`Error retrieving credentials from the instance profile metadata service. (cURL error 7: Failed to connect to 169.254.169.254 port 80: Network is unreachable (see https://curl.haxx.se/libcurl/c/libcurl-errors.html))`
clearly indicating the old address for instance metadata.

### Reproduction Steps

1. Setup application that is using AWS SDK for PHP
2. Make sure that the application doesn't have any other credentials available
3. Setup environment variable with another URL to instance metadata
4. Try calling any AWS service, allowing SDK to authenticate
5. Observe the error and a call to 169.254.169.254, instead of the URL defined in an environment variable `AWS_EC2_METADATA_SERVICE_ENDPOINT`

### Possible Solution

Re-implement https://github.com/aws/aws-sdk-php/pull/2277 after it has been reverted.

### Additional Information/Context

I think the problem is that URL is hardcoded in https://github.com/SamRemis/aws-sdk-php/blob/master/src/Credentials/InstanceProfileProvider.php#L19 and https://github.com/SamRemis/aws-sdk-php/blob/master/src/Credentials/InstanceProfileProvider.php#L230

### SDK version used

3.290.1

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

php 8.1

yenfryherrerafeliz:
Hi @ibexa-yuna, thanks for reporting this. You are experiencing this behavior because the instance profile provider does not support a custom endpoint to resolve credentials. I do not know why the previous implementation was removed, but I will bring this into discussion with the team to add an item in our backlog to address this in the near future. I will provide updates regarding what we decide or our progress in the thread of this issue.

Thanks!

stobrien89:
Hi @ibexa-yuna,

[...]

PR #2859, opened 2023-12-18
feat: IMDS support for providing custom endpoint

yenfryherrerafeliz:
When using IMDS for fetching credentials, customers should be able to provide their custom endpoint when desired, and that is what this change does. Basically, customer can provide a custom endpoint by doing one of the following options: Please note that a valid URI value needs to be provided, otherwise a credential exception will be thrown.
- Providing a parameter called 'ec2_metadata_service_endpoint' to the constructor of the InstanceProfileProvider.
- By setting an environment variable called AWS_EC2_METADATA_SERVICE_ENDPOINT.
- By defining a key-value config in the config file ~/.aws/config

*Issue #2847*

*Description of changes:*

### DIFF
--- a/src/Credentials/CredentialsUtils.php
+++ b/src/Credentials/CredentialsUtils.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Aws\Credentials;
+
+final class CredentialsUtils
+{
+    /**
+     * Determines whether a given host
+     * is a loopback address.
+     *
+     * @param $host
+     *
+     * @return bool
+     */
+    public static function isLoopBackAddress($host): bool
+    {
+        if (!filter_var($host, FILTER_VALIDATE_IP)) {
+            return false;
+        }
+
+        if (filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+            if ($host === '::1') {
+                return true;
+            }
+
+            return false;
+        }
+
+        $loopbackStart = ip2long('127.0.0.0');
+        $loopbackEnd = ip2long('127.255.255.255');
+        $ipLong = ip2long($host);
+
+        return ($ipLong >= $loopbackStart && $ipLong <= $loopbackEnd);
+    }
+}
--- a/src/Credentials/InstanceProfileProvider.php
+++ b/src/Credentials/InstanceProfileProvider.php
@@ -16,17 +16,22 @@ use Psr\Http\Message\ResponseInterface;
  */
 class InstanceProfileProvider
 {
-    const SERVER_URI = 'http://169.254.169.254/latest/';
     const CRED_PATH = 'meta-data/iam/security-credentials/';
     const TOKEN_PATH = 'api/token';
     const ENV_DISABLE = 'AWS_EC2_METADATA_DISABLED';
     const ENV_TIMEOUT = 'AWS_METADATA_SERVICE_TIMEOUT';
     const ENV_RETRIES = 'AWS_METADATA_SERVICE_NUM_ATTEMPTS';
     const CFG_EC2_METADATA_V1_DISABLED = 'ec2_metadata_v1_disabled';
+    const CFG_EC2_METADATA_SERVICE_ENDPOINT = 'ec2_metadata_service_endpoint';
+    const CFG_EC2_METADATA_SERVICE_ENDPOINT_MODE = 'ec2_metadata_service_endpoint_mode';
     const DEFAULT_TIMEOUT = 1.0;
     const DEFAULT_RETRIES = 3;
     const DEFAULT_TOKEN_TTL_SECONDS = 21600;
     const DEFAULT_AWS_EC2_METADATA_V1_DISABLED = false;
+    const ENDPOINT_MODE_IPv4 = 'IPv4';
+    const ENDPOINT_MODE_IPv6 = 'IPv6';
+    private const DEFAULT_METADATA_SERVICE_IPv4_ENDPOINT = 'http://169.254.169.254';
+    private const DEFAULT_METADATA_SERVICE_IPv6_ENDPOINT = 'http://[fd00:ec2::254]';
 
     /** @var string */
     private $profile;
@@ -49,6 +54,12 @@ class InstanceProfileProvider
     /** @var bool|null */
     private $ec2MetadataV1Disabled;
 
+    /** @var string */
+    private $endpoint;
+
+    /** @var string */
+    private $endpointMode;
+
     /**
      * The constructor accepts the following options:
      *
@@ -56,6 +67,11 @@ class InstanceProfileProvider
      * - profile: Optional EC2 profile name, if known.
      * - retries: Optional number of retries to be attempted.
      * - ec2_metadata_v1_disabled: Optional for disabling the fallback to IMDSv1.
+     * - endpoint: Optional for overriding the default endpoint to be used for fetching credentials.
+     *             The value must contain a valid URI scheme. If the URI scheme is not https, it must
+     *             resolve to a loopback address.
+     * - endpoint_mode: Optional for overriding the default endpoint mode (IPv4|IPv6) to be used for
+     *   resolving the default endpoint.
      *
      * @param array $config Configuration options.
      */
@@ -66,6 +82,12 @@ class InstanceProfileProvider
         $this->retries = (int) getenv(self::ENV_RETRIES) ?: ($config['retries'] ?? self::DEFAULT_RETRIES);
         $this->client = $config['client'] ?? \Aws\default_http_handler();
         $this->ec2MetadataV1Disabled = $config[self::CFG_EC2_METADATA_V1_DISABLED] ?? null;
+        $this->endpoint = $config[self::CFG_EC2_METADATA_SERVICE_ENDPOINT] ?? null;
+        if (!empty($this->endpoint) && !$this->isValidEndpoint($this->endpoint)) {
+            throw new \InvalidArgumentException('The provided URI "' . $this->endpoint . '" is invalid, or contains an unsupported host');
+        }
+
+        $this->endpointMode = $config[self::CFG_EC2_METADATA_SERVICE_ENDPOINT_MODE] ?? null;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


src/Credentials/Utils/Validator.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Aws\Credentials\Utils;
+
+final class Validator
+{
+    private function __construct() {}
+
+    public static function validateURI(
```
stobrien89, line 9: I think this could be moved to the InstanceProfileProvider class.  validation will vary depending on the provider (see the EcsCredentialProvider)

src/Credentials/InstanceProfileProvider.php
```diff
@@ -66,6 +78,17 @@ public function __construct(array $config = [])
         $this->retries = (int) getenv(self::ENV_RETRIES) ?: ($config['retries'] ?? self::DEFAULT_RETRIES);
         $this->client = $config['client'] ?? \Aws\default_http_handler();
         $this->ec2MetadataV1Disabled = $config[self::CFG_EC2_METADATA_V1_DISABLED] ?? null;
+        $this->endpoint = Validator::validateURI(
```
stobrien89, line 81: what do you think about something like:

```
 $this->endpoint = isset($config['endpoint']) && $this->isValidEndpoint($config['endpoint']) 
    ? $config['endpoint'] 
    : null;
```

`endpoint` is the key we should be looking for in config

src/Credentials/Utils/Validator.php
stobrien89, line 1: I don't think we'll need this

tests/Credentials/InstanceProfileProviderTest.php
```diff
@@ -1349,4 +1349,134 @@ private function fetchMockedCredentialsAndAlwaysExpectAToken($config=[]) {
             return false;
         }
     }
+
+    /**
+     * @dataProvider endpointModeCases
+     */
+    public function testEndpointMode($clientConfig, $env, $config, $expected)
```
stobrien89, line 1356: Let's modularize the tests here.  We should have separate tests for both resolving/validating provided endpoints, as well as using the provided endpoint mode

src/Credentials/Utils/Validator.php
stobrien89: We won't need it to accept arguments aside from the Uri itself.

stobrien89 (changes requested):

src/Credentials/InstanceProfileProvider.php
```diff
@@ -329,4 +351,132 @@ private function shouldFallbackToIMDSv1(): bool
...
+    /**
+     * Resolves the default metadata service endpoint.
+     * If endpoint_mode is resolved as IPv4 then:
+     * - endpoint = http://169.254.169.254
+     * If endpoint_mode is resolved as IPv6 then:
+     * - endpoint = http://[fd00:ec2::254]
+     *
+     * @return string
+     */
+    private function defaultEndpoint(): string
```
stobrien89, line 394: change this to `getDefaultEndpoint()`

src/Credentials/InstanceProfileProvider.php
```diff
@@ -329,4 +351,132 @@ private function shouldFallbackToIMDSv1(): bool
...
+    {
+        $endpointMode = $this->resolveEndpointMode();
+        switch ($endpointMode) {
+            case self::ENDPOINT_MODE_IPv4:
```
stobrien89, line 398: I'd make this the default case

src/Credentials/InstanceProfileProvider.php
```diff
@@ -329,4 +351,132 @@ private function shouldFallbackToIMDSv1(): bool
...
+    /**
+     * This method makes sure that the endpoint mode is either IPv4 or IPv6.
+     * @param string $endpointMode the endpoint mode to the validation against to.
+     * @param bool $ignoreNulls when this parameter is false then, an exception will be thrown
+     *  if the value for $endpointMode is null.
+     * @param $exceptionClass
+     *
+     * @return string|null
+     */
+    private function validateEndpointMode(
```
stobrien89, line 463: We won't be needing this

tests/Credentials/InstanceProfileProviderTest.php
stobrien89, line 1: These look good. I would create a separate test for bad config (invalid endpoints)

stobrien89 (changes requested):

src/Credentials/InstanceProfileProvider.php
```diff
@@ -329,4 +346,111 @@ private function shouldFallbackToIMDSv1(): bool
...
+                'string',
+                ['use_aws_shared_config_files' => true]
+            );
+        }
+
+        if (!$this->isValidEndpoint($endpoint)) {
+            throw new CredentialsException('The provided URI "' . $endpoint . '" is not a valid URI scheme');
+        }
+
+        if (str_ends_with('/', $endpoint) === false) {
```
stobrien89, line 375: This function is only available in PHP 8+

src/Credentials/CredentialsUtils.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Aws\Credentials;
+
+final class CredentialsUtils
```
stobrien89, line 5: This is not in the specification for imds endpoint configuration, but checking whether or not we should do this.

src/Credentials/InstanceProfileProvider.php
```diff
@@ -66,6 +77,12 @@ public function __construct(array $config = [])
         $this->retries = (int) getenv(self::ENV_RETRIES) ?: ($config['retries'] ?? self::DEFAULT_RETRIES);
         $this->client = $config['client'] ?? \Aws\default_http_handler();
         $this->ec2MetadataV1Disabled = $config[self::CFG_EC2_METADATA_V1_DISABLED] ?? null;
+        $this->endpoint =$config[self::CFG_EC2_METADATA_SERVICE_ENDPOINT] ?? $config['endpoint'] ?? null;
+        if ($this->endpoint !== null && !$this->isValidEndpoint($this->endpoint)) {
```
stobrien89, line 81: I think an empty check would be better here: `empty($this->endpoint)`

src/Credentials/InstanceProfileProvider.php
```diff
@@ -329,4 +346,111 @@ private function shouldFallbackToIMDSv1(): bool
...
+        }
+
+        // We make sure that if is a no secure host then it must be a loop back address.
+        $parsedUri = parse_url($uri);
+        if ($parsedUri['scheme'] !== 'https') {
+            $host = trim($parsedUri['host'], '[]');
+
+            return CredentialsUtils::isLoopBackAddress($host)
+                || in_array(
+                    $parsedUri['scheme'] . '://' . $parsedUri['host'],
```
stobrien89, line 449: Shouldn't we use `$uri` here?

src/Credentials/CredentialsUtils.php
stobrien89: Checked and I think this is okay.  We can always loosen restrictions later, but probably a good move from a security standpoint.

stobrien89 (changes requested):
Just a few minor things

src/Credentials/InstanceProfileProvider.php
```diff
@@ -66,6 +80,12 @@ public function __construct(array $config = [])
         $this->retries = (int) getenv(self::ENV_RETRIES) ?: ($config['retries'] ?? self::DEFAULT_RETRIES);
         $this->client = $config['client'] ?? \Aws\default_http_handler();
         $this->ec2MetadataV1Disabled = $config[self::CFG_EC2_METADATA_V1_DISABLED] ?? null;
+        $this->endpoint =$config[self::CFG_EC2_METADATA_SERVICE_ENDPOINT] ?? $config['endpoint'] ?? null;
```
stobrien89, line 83: I'd just make this `ec2_metadata_service_endpoint` based on our discussion today.

stobrien89 (changes requested):

src/Credentials/CredentialsUtils.php
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Determines whether a given host
+     * is a loopback address.
+     *
+     * @param $host
+     *
+     * @return bool
+     */
+    public static function isLoopBackAddress($host): bool
```
stobrien89, line 17: We should add a test for this class/method.

stobrien89 (changes requested):
Looks good to me.  We should add a `CredentialsUtilsTest` class and test the `isLoopbackAddress` method directly.  After that, I think we're ready to ship.

src/Credentials/CredentialsUtils.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Aws\Credentials;
+
+final class CredentialsUtils
+{
+    private function __construct() {}
```
stobrien89, line 7: Since we only have one static method, let's take the constructor out for now

src/Credentials/InstanceProfileProvider.php
```diff
@@ -49,13 +54,22 @@ class InstanceProfileProvider
...
+    private $endpointMode;
+
     /**
      * The constructor accepts the following options:
      *
      * - timeout: Connection timeout, in seconds.
      * - profile: Optional EC2 profile name, if known.
      * - retries: Optional number of retries to be attempted.
      * - ec2_metadata_v1_disabled: Optional for disabling the fallback to IMDSv1.
+     * - endpoint: Optional for overriding the default endpoint to be used for fetching credentials.
```
stobrien89, line 70: Let's add to this "the value must be a valid URI scheme, and if it it not https, the host must resolve to a loopback address.

src/Credentials/InstanceProfileProvider.php
```diff
@@ -66,6 +80,12 @@ public function __construct(array $config = [])
         $this->retries = (int) getenv(self::ENV_RETRIES) ?: ($config['retries'] ?? self::DEFAULT_RETRIES);
         $this->client = $config['client'] ?? \Aws\default_http_handler();
         $this->ec2MetadataV1Disabled = $config[self::CFG_EC2_METADATA_V1_DISABLED] ?? null;
+        $this->endpoint = $config[self::CFG_EC2_METADATA_SERVICE_ENDPOINT] ?? null;
+        if (!empty($this->endpoint) && !$this->isValidEndpoint($this->endpoint)) {
+            throw new \InvalidArgumentException('The provided URI "' . $this->endpoint . '" is not a valid URI scheme');
```
stobrien89, line 85: This could be misleading, as there are technically two requirements for this.  I think "The provided URI is invalid, or contains an unsupported host". Works better here.

stobrien89 (changes requested):
Two more minor things

stobrien89 (approved):

Issue #2527, opened 2022-09-21
PHP8 is_resource behavioral change affects AbstractMonitoringMiddleware


mhsdef:
### Describe the bug

`AbstractMonitoringMiddleware::prepareSocket`'s `is_resource` check ([#](https://github.com/aws/aws-sdk-php/blob/a3a5711c00673d18b571199858ffee95296c5f0d/src/ClientSideMonitoring/AbstractMonitoringMiddleware.php#L238)):

```
private function prepareSocket($forceNewConnection = false)
{
    if (!is_resource(self::$socket)
        || $forceNewConnection
        || socket_last_error(self::$socket)
    ) {
        self::$socket = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
        socket_clear_error(self::$socket);
        socket_connect(self::$socket, $this->getHost(), $this->getPort());
    }

    return self::$socket;
} 
```

Looks affected by https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object

### Expected Behavior

The check to cover both PHP7 and PHP8 behaviors.

### Current Behavior

The check only covers <PHP8 behavior.

### Reproduction Steps

See this:
https://onlinephp.io/c/3574d

### Possible Solution

Check for `!self::$socket instanceof Socket` as well to cover PHP8.

### SDK version used

master

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.0.23

yenfryherrerafeliz:
Hi @hewsut, thanks for opening this issue. I agree with the suggested change. I will mark this issue to be reviewed so we can address it further.

Thanks!

MattPearce:
@yenfryherrerafeliz this issue is blocking our PHP v8 upgrade, this patch will fix the issue if you could please merge it:

[0001-Switch-resource-check-to-class-check-for-PHP-v8-comp.patch](https://github.com/aws/aws-sdk-php/files/13983101/0001-Switch-resource-check-to-class-check-for-PHP-v8-comp.patch)

yenfryherrerafeliz:
Hi @MattPearce, I have left a comment on the PR. Would you please be able to take look to it.

Thanks!

yenfryherrerafeliz:
This issue has been fixed by the following [PR](https://github.com/aws/aws-sdk-php/pull/2865). Closing for now.

Thanks!

PR #2865, opened 2024-01-19
Switch resource check to class check for PHP v8 compatibility

MattPearce:
*Issue #, if available:*
https://github.com/aws/aws-sdk-php/issues/2527

*Description of changes:*
In PHP v8 on, Sockets are now classes instead of resources. This PR fixes the PHP v8 compatibility issue in this library.

### DIFF
--- a/src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
+++ b/src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
@@ -235,7 +235,7 @@ abstract class AbstractMonitoringMiddleware
         // Before version 8, sockets are resources
         // After version 8, sockets are instances of Socket
         if (PHP_MAJOR_VERSION >= 8) {
-            return !(self::$socket instanceof Socket);
+            return !(self::$socket instanceof \Socket);
         } else {
             return is_resource(self::$socket);
         }
--- a/src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
+++ b/src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
@@ -224,6 +224,23 @@ abstract class AbstractMonitoringMiddleware
         return $event;
     }
 
+    
+    /**
+     * Checks if the socket is created.
+     *
+     * @return bool Returns true if the socket is created, false otherwise.
+     */
+    private function isSocketCreated()
+    {
+        // Before version 8, sockets are resources
+        // After version 8, sockets are instances of Socket
+        if (PHP_MAJOR_VERSION >= 8) {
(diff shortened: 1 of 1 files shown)


src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
```diff
@@ -224,6 +224,23 @@ protected function populateResultEventData(
...
+     * Checks if the socket is created.
+     *
+     * @return bool Returns true if the socket is created, false otherwise.
+     */
+    private function isSocketCreated()
+    {
+        // Before version 8, sockets are resources
+        // After version 8, sockets are instances of Socket
+        if (PHP_MAJOR_VERSION >= 8) {
+                !(self::$socket instanceof Socket);
```
yenfryherrerafeliz, line 238: Do you mean to do a return statement as following:
```
return self::$socket instanceof \Socket;
```
If so, could you please make that adjustment?

Thanks.

MattPearce: Yes, sorry about that - not sure how I missed that - I've added the return statement and properly referenced the PHP Socket class

src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
```diff
@@ -224,6 +224,23 @@ protected function populateResultEventData(
...
+     * Checks if the socket is created.
+     *
+     * @return bool Returns true if the socket is created, false otherwise.
+     */
+    private function isSocketCreated()
+    {
+        // Before version 8, sockets are resources
+        // After version 8, sockets are instances of Socket
+        if (PHP_MAJOR_VERSION >= 8) {
+            return !(self::$socket instanceof \Socket);
```
yenfryherrerafeliz, line 238: Thanks for the previous change, however, there is something else we would need to address and is that I think the expression should be:
```php
return (self::$socket instanceof \Socket);
```
instead of:
```php
return !(self::$socket instanceof \Socket);
```
because the function will return false even when the socket is already created, since you are prefixing the expression with a negation symbol.

What do you think?

yenfryherrerafeliz (changes requested):

src/ClientSideMonitoring/AbstractMonitoringMiddleware.php
MattPearce: You're right, the logic was inverted - have fixed it now

yenfryherrerafeliz (approved):

yenfryherrerafeliz:
Thanks @MattPearce for your contribution. We will try to get this merged by tomorrow.

PR #2875, opened 2024-01-31
chore: add docs build workflow

yenfryherrerafeliz:
Run a docs build workflow in order to catch errors in this staged before merging. 

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,0 +1,107 @@
+name: Build API documentation
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  build_docs:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php-versions: ['5.6']
+    name: Build API documentation for PHP ${{ matrix.php-versions }}
+    steps:
+      - name: Setup PHP with Xdebug
+        uses: shivammathur/setup-php@v2
+        with:
+          coverage: xdebug
+          php-version: ${{ matrix.php-versions }}
+          ini-values: xdebug.overload_var_dump=0, memory_limit=4G, phar.readonly=false
+
+      - name: Checkout CodeBase
+        uses: actions/checkout@v4
+
+      - name: Replace platform requirements
+        run: |
+          sed -i 's/"php": ">=7.2.5"/"php": ">=5.6.0"/g' composer.json
+
+      - name: Validate composer.json and composer.lock
+        run: composer validate
+
+      - name: Install dependencies
+        run: composer update ${{ matrix.composer-options }} --no-interaction --prefer-source
+
+      - name: Prepare OS environment
+        run: |
+          sudo sed -i s/deb.debian.org/archive.debian.org/g /etc/apt/sources.list
+          sudo sed -i 's|security.debian.org|archive.debian.org/|g' /etc/apt/sources.list
+          sudo sed -i '/stretch-updates/d' /etc/apt/sources.list
+          
+          sudo apt-get -y update && \
+          sudo apt-get -y install git wget zip unzip libzip-dev libssl-dev libtidy-dev python3 cmake python3-distutils-extra python3-apt
+          
+          PHP_CONF_DIR=$(php -i | grep "Scan this dir for additional .ini files" | awk '{print $9}')
+          sudo touch "$PHP_CONF_DIR"/memory.ini \
+          && sudo chmod 666 "$PHP_CONF_DIR"/memory.ini \
+          && sudo echo "memory_limit = 5048M;" >> "$PHP_CONF_DIR"/memory.ini
+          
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
+name: Docs Build Workflow
```
stobrien89, line 1: Let's call this "Build API documentation"

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+
+on:
+  push:
```
stobrien89, line 4: Let's do this for pull requests only.  The automation account will only push if documentation builds are successful, so this is not needed.

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+  pull_request:
+    branches:
+      - master
+jobs:
+  build_docs:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php-versions: ['5.6']
+        composer-options: ['']
```
stobrien89, line 16: not sure this is needed.

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+          sudo sed -i 's|security.debian.org|archive.debian.org/|g' /etc/apt/sources.list
+          sudo sed -i '/stretch-updates/d' /etc/apt/sources.list
+          
+          sudo apt-get -y update && \
+          sudo apt-get -y install git wget zip unzip libzip-dev libssl-dev libtidy-dev python3 cmake python3-distutils-extra python3-apt
+          
+          curl -O https://bootstrap.pypa.io/pip/3.5/get-pip.py && \
+          python3 get-pip.py && \
+          rm get-pip.py && \
+          pip install awscli
```
stobrien89, line 51: won't be needing the cli

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+  
+          # Git Secrets Install
```
stobrien89, line 53: won't be needing this step

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+    branches:
+      - master
+jobs:
+  build_docs:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php-versions: ['5.6']
+        composer-options: ['']
+    name: Build Docs Test for PHP ${{ matrix.php-versions }} ${{ matrix.composer-options }}
```
stobrien89, line 17: Let's also call this "Build API documentation"

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+              
+              except FileNotFoundError:
+                print('php file not found : ', file_path)
+            
+          path = sys.argv[1]
+          worker = FnReturnHints()
+          worker.walkThrough(path)
+          
+          EOF
+      - name: Run Function Return Hits Removal
```
stobrien89, line 118: I would just call this "Remove Return Types"

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,126 @@
...
+          
+          sudo touch "$PHP_CONF_DIR"/phar.ini \
+          && sudo chmod 666 "$PHP_CONF_DIR"/phar.ini \
+          && sudo echo "phar.readonly = Off;" >> "$PHP_CONF_DIR"/phar.ini
+          
+          sudo touch "$PHP_CONF_DIR"/timezone.ini \
+          && sudo chmod 666 "$PHP_CONF_DIR"/timezone.ini \
+          && sudo echo "date.timezone ='America/New_York'" >> "$PHP_CONF_DIR"/timezone.ini
+
+      - name: Prepare Function Return Hints Script
```
stobrien89, line 73: I would call them "Return Types" rather "Return Hints"

stobrien89 (changes requested):
Just semantics and a few unnecessary steps.  Otherwise looks good!

.github/workflows/docs-build.yml
stobrien89: Still seeing this step here.  We don't need to install git secrets.

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,116 @@
...
+          
+          EOF
+      - name: Run remove return types script
+        run: |
+          python3 remove-return-types.py ./src \
+          && rm -f remove-return-types.py
+      - name: Run API docs build
+        run: |
+          make package \
+          && make api-get-apigen \
```
stobrien89, line 115: We can remove this step, it's called by `make api`.

stobrien89 (approved):

PR #2877, opened 2024-02-08
fix(MultipartUpload): changed ObjectURL for CompleteMultipartUpload into effectiveUri

idsulik:
*Description of changes:* I spent several hours to figure out why some files return ObjectURL with "**%2F**", but some files with "**/**", it turned out that **CompleteMultipartUpload** puts "Location" URL to ObjectURL while PutObject and CopyObject put the **effectiveUri**, changed this so CompleteMultipartUpload will return **effecitveUri** if it's available and "Location" if it's not available.

image

### DIFF
--- a/src/S3/PutObjectUrlMiddleware.php
+++ b/src/S3/PutObjectUrlMiddleware.php
@@ -49,7 +49,14 @@ class PutObjectUrlMiddleware
                             : null;
                         break;
                     case 'CompleteMultipartUpload':
-                        $result['ObjectURL'] = $result['Location'];
+                        $objectUrl = null;
+                        $effectiveUri = $result['@metadata']['effectiveUri'] ?? null;
+                        if ($effectiveUri !== null) {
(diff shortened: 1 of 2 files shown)


idsulik:
Another way to fix it is to replace "%2F" with "/" for the location's URI, which one is preferable?

idsulik:
@stobrien89 Hi!
Should I do anything else or I just need to wait?

yenfryherrerafeliz:
Hi @idsulik, thanks for opening this PR. This has been fixed already which you could confirm [here](https://github.com/aws/aws-sdk-php/blob/c2455c54c2f8b04296fd14df7181684c89fa079c/src/S3/PutObjectUrlMiddleware.php#L52). We just urldecoded to the `Location` param. Your suggested changes would be also another way to have this fixed, but by the time I had to fix an reported issue regarding this, I was not aware of this PR.

Closing this PR for now.

Thanks!

PR #2879, opened 2024-02-14
Feature: Multi-Auth

stobrien89:
*Description of changes:*
Adds support for the service and operation-level `auth` trait, which allows a service to specify multiple auth schemes.  This trait will contain a non-empty priority-ordered list of auth schemes to select from.


tests/Auth/AuthSchemeResolverTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $identity = $this->createMock(IdentityInterface::class);
+        $resolver->selectAuthScheme(['non.existent#scheme'], $identity);
+    }
+
+    public function testSelectAuthSchemePrioritizesFirstCompatibleScheme()
+    {
+        $resolver = new AuthSchemeResolver();
+        $identity = $this->createMock(IdentityInterface::class);
+        // Assuming the identity is compatible with 'v4' but not 'v4a'
+        $this->assertEquals('v4', $resolver->selectAuthScheme(['aws.auth#sigv4', 'aws.auth#sigv4a'], $identity));
```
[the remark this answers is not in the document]
stobrien89: You can just ignore that comment above this line.  It's incorrect needs to be removed.  Identity is not evaluated for sigv4 or sigv4a

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -227,10 +227,21 @@ private function bindContextParams(
...
+
+        if (isset($authScheme['name'])) {
+            $command['@context']['signing_service'] = $authScheme['name'];
+        }
+
+        if (isset($authScheme['region'])) {
+            $command['@context']['signing_region'] = $authScheme['region'];
+        } elseif (isset($authScheme['signingRegionSet'])) {
+            $command['@context']['signing_region_set'] = $authScheme['signingRegionSet'];
+        }
```
yenfryherrerafeliz, line 244: Are there any default values for:
- $command['@context']['signing_region']
- $command['@context']['signing_region_set']

Or one of those is guarantee to be always present in the resolved authScheme?

I ask this because, what would happen if none of those conditions are met?

stobrien89: If neither of those conditions are met, we just use the signing region (which is always set) from the client config.  This happens [here](https://github.com/aws/aws-sdk-php/blob/fb67a49453c2b1a51268156b70adeba6c619e743/src/AwsClient.php#L412).  We're not doing anything with the region set right now because its presence indicates a `sigv4a` signature version and we hard-code the `*` value into the CRT signing config. This won't change until we start seeing multiple values in region sets.

yenfryherrerafeliz (approved):

PR #2882, opened 2024-02-21
Fix denylisting x-amz-user-agent in SignatureV4

eduard-sukharev:
When checking headers for signature inclusion, they are lowercased, thus `X-Amz-User-Agent` is never skipped from signature, even though it is blacklisted exactly like that.

*Description of changes:*
X-Amz-User-Agent turned to lowercase in blacklist, as all checks with blacklist are with lowercased headers.

### DIFF
--- a/src/Signature/SignatureV4.php
+++ b/src/Signature/SignatureV4.php
@@ -67,7 +67,7 @@ class SignatureV4 implements SignatureInterface
             'from'                  => true,
             'referer'               => true,
             'user-agent'            => true,
-            'X-Amz-User-Agent'      => true,
+            'x-amz-user-agent'      => true,
             'x-amzn-trace-id'       => true,
             'aws-sdk-invocation-id' => true,
             'aws-sdk-retry'         => true,


eduard-sukharev:
@stobrien89 @yenfryherrerafeliz could you guys have a look at this, please?
and also I don't know why the pipeline fails on some PHP versions, it's pretty much a straightforward fix

yenfryherrerafeliz (approved):

GrahamCampbell (approved):

rmoqpd5t:
I think the problem is here:
```
    private function moveHeadersToQuery(array $parsedRequest)
    {
        //x-amz-user-agent shouldn't be put in a query param
        unset($parsedRequest['headers']['X-Amz-User-Agent']);

        foreach ($parsedRequest['headers'] as $name => $header) {
            $lname = strtolower($name);
            if (substr($lname, 0, 5) == 'x-amz') {
                $parsedRequest['query'][$name] = $header;
            }
            $blacklist = $this->getHeaderBlacklist();
```
The unset assumes the constant X-Amz-User-Agent has that case so it does not match clients that use x-amz-user-agent

stobrien89:
Apologies for the long delay here. This header never belonged in the blacklist anyway- we need to sign it because it's sent with every AWS request.  The reason this has worked for so long is because of the casing (which several of you have pointed out), but changing the casing causes signature mismatches due to a missing signed counterpart to a canonical header. Closing for now, but let us know if you have any questions.

PR #2884, opened 2024-02-22
feat: account id endpoint resolution support

yenfryherrerafeliz:
This change add account_id as part of the identity resolution, from the different credentials provider. It also validates whether an account should have been resolved based on the configure option account_id_endpoint_mode, and when this option is not disabled then, it prepends a middleware that resolves the identity, from the provided credential provider and, it validates the account id based on the account_id_endpoint_mode option. When an identity is resolved by this middleware then, this identity is carry over in a command property bag called @context, which is then reused by the signer middleware, and any other middleware that needs to resolve identity. This is done for avoiding having to resolve identity multiple times in a single request. The property is: $command['@context']['resolved_identity'];


src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +303,11 @@ private function normalizeAuthScheme(array $authScheme) : array
 
         return $normalizedAuthScheme;
     }
+
+    private function appendAccountIdParameter(&$commandArgs)
```
stobrien89, line 307: I think we need to have the account id mode available in the args.  Identity might be resolved in other places and not always because of account id retrieval, so I think this should be done conditionally.

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +303,11 @@ private function normalizeAuthScheme(array $authScheme) : array
...
+    {
+        if (isset($commandArgs['@context']['resolved_identity'])) {
+            $commandArgs['AWS::Auth::AccountId'] = $commandArgs['@context']['resolved_identity']->getAccountId();
```
stobrien89, line 310: The normalized key `AccountId` should be added to the command args

src/AccountIdEndpointMiddleware.php
```diff
@@ -0,0 +1,79 @@
...
+     * @var callable $credentialsProvider
+     */
+    private $credentialsProvider;
+
+    /**
+     * @param callable $nextHandler
+     * @param string $accountIdEndpointMode
+     * @param callable $credentialsProvider
+     */
+    public function __construct($nextHandler, $accountIdEndpointMode, $credentialsProvider)
```
stobrien89, line 35: Might want to consider passing an array of args/config here and pulling the endpoint mode out of that.  Never know what other values we may need access to in the future

src/AwsClient.php
```diff
@@ -534,6 +536,22 @@ private function addEndpointV2Middleware()
...
+        $accountIdEndpointMode = !empty($args['account_id_endpoint_mode'])
+            ? $args['account_id_endpoint_mode']
+            : $config['account_id_endpoint_mode'];
+        if ($accountIdEndpointMode !== 'disabled') {
+            $this->handlerList->prependBuild(
+                AccountIdEndpointMiddleware::wrap(
+                    $accountIdEndpointMode,
+                    $this->credentialProvider
+                ),
+                'account_id_endpoint-middleware'
```
stobrien89, line 550: We should use dashes here

src/AccountIdEndpointMiddleware.php
```diff
@@ -0,0 +1,79 @@
...
+            return new self($handler, $accountIddEndpointMode, $credentialsProvider);
+        };
+    }
+
+    public function __invoke($command)
+    {
+        $nextHandler = $this->nextHandler;
+        $fnCredentialsProvider = $this->credentialsProvider;
+        $resolvedIdentity = $fnCredentialsProvider()->wait();
+        if (empty($resolvedIdentity->getAccountId())) {
```
stobrien89, line 61: This should be simplified.  We should use shorter, more concise and separate messages for the exception and warning cases and leave the finer details to doc annotations.  For raising a warning, we use `trigger_error`

stobrien89 (changes requested):

src/AwsClient.php
yenfryherrerafeliz: Middleware has been removed.

src/AccountIdEndpointMiddleware.php
yenfryherrerafeliz: Middleware has been removed.

src/Credentials/CredentialsInterface.php
```diff
@@ -49,4 +49,6 @@ public function isExpired();
      * @return array
      */
     public function toArray();
+
+    public function getAccountId();
```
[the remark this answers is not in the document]
yenfryherrerafeliz: Done.

src/AwsClient.php
```diff
@@ -558,7 +558,7 @@ private function setClientContextParams($args)
     /**
      * Retrieves and sets default values used for endpoint resolution.
      */
-    private function setClientBuiltIns($args)
+    private function setClientBuiltIns($args, $clientConfig)
     {
         $builtIns = [];
         $config = $this->getConfig();
```
stobrien89, line 564: We could refactor this: it's available in `$clientConfig`

src/AwsClient.php
stobrien89, line 1: We could refactor `getEndpointProviderArgs` or `addEndpointV2Middleware` to pull the account id settings into the args passed into the middleware.

src/ClientResolver.php
stobrien89, line 1: Just the changes we talked about with having both a default and apply function.

stobrien89: And throwing an IAE when the values aren't in the permitted values.

src/AwsClient.php
```diff
@@ -582,6 +582,8 @@ private function setClientBuiltIns($args)
             $builtIns['AWS::S3::ForcePathStyle'] = $config['use_path_style_endpoint'];
             $builtIns['AWS::S3::DisableMultiRegionAccessPoints'] = $config['disable_multiregion_access_points'];
         }
+        $builtIns['AWS::Auth::AccountId'] = new AccountIdLazyResolver($this->credentialProvider, $clientConfig['account_id_endpoint_mode']);
+
```
stobrien89, line 586: Looks like `AWS::Auth::AccountIdEndpointMode` is also considered a builtin per the EP2.0 spec.

stobrien89 (changes requested):

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -112,6 +117,9 @@ public function __invoke(CommandInterface $command)
     private function resolveArgs(array $commandArgs, Operation $operation) : array
     {
         $rulesetParams = $this->endpointProvider->getRuleset()->getParameters();
+
+        $this->resolveAccountIdArgument($commandArgs, $rulesetParams);
```
stobrien89, line 121: Why not just do this conditionally based on whether or not `AccountId` and `AccountIdMode` are set?

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +311,50 @@ private function normalizeAuthScheme(array $authScheme) : array
...
+            return;
+        }
+
+        $accountIdEndpointMode = $this->clientArgs['AccountIdEndpointMode'];
+        if ($accountIdEndpointMode === 'disabled') {
+            return;
+        }
+
+        $credentialsProviderFn = $this->credentialsProvider;
+        $identity = $credentialsProviderFn()->wait();
```
stobrien89, line 339: I'd do a check here to see if the identity is set on the command already.  It might be, dependent on whether or not a service models auth schemes using the `auth` trait.

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +311,50 @@ private function normalizeAuthScheme(array $authScheme) : array
...
+        $commandArgs['@context']['resolved_identity'] = Create::promiseFor($identity);
+        if (empty($identity->getAccountId())) {
+            $messageFn = function ($mode) {
```
stobrien89, line 342: The message contains too much information for a warning or exception.  I don't think this should be a local callable either- we should split this out into another method.  Use heredoc/nowdoc if you really need to, but I think the message could be contained within two < 80 char lines.  I don't think we should have a separate exception for this either.  I think we should be using a `UnresolvedEndpointException` if we hit the required case. I realize there's one place where I use an InvalidArgumentException— that should be changed too.

stobrien89 (changes requested):

src/EndpointV2/EndpointV2Middleware.php
yenfryherrerafeliz: It is done conditionally at the method implementation. If AccountId is not part of the ruleset parameters or AccountIdEndpointMode is disabled then, it will abort the execution. Please see the method implementation.

src/EndpointV2/EndpointV2Middleware.php
yenfryherrerafeliz: We could use UnresolvedEndpointException, but throwing an exception that relate more with the issue can be some times more useful for customers, since it is more clear about what the caused the problem. I also checked other SDKs such as boto3 and they throw a specific exception for it as well. 

I agree that the message should be shorter. I am going to work on that.

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +312,57 @@ private function normalizeAuthScheme(array $authScheme) : array
...
+    {
+        static $ACCOUNT_ID_PARAM = 'AccountId';
+        $accountIdEndpointMode = $this->clientArgs['AccountIdEndpointMode'];
+        if (!isset($rulesetParams[$ACCOUNT_ID_PARAM]) || $accountIdEndpointMode === 'disabled') {
+            return;
+        }
+
+        $identity = null;
+        if (isset($commandArgs['@context']['resolved_identity'])) {
+            $identity = $commandArgs['@context']['resolved_identity']->wait();
```
stobrien89, line 337: resolved_identity should be an instance of `IdentityInterface`

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -67,13 +72,15 @@ public function __construct(
         callable $nextHandler,
         EndpointProviderV2 $endpointProvider,
         Service $api,
-        array $args
+        array $args,
+        callable $credentialsProvider
```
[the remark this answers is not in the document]
yenfryherrerafeliz: I think it should for now, based on the current implementation. But I am open to any suggestion.

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -303,4 +312,57 @@ private function normalizeAuthScheme(array $authScheme) : array
...
+            return;
+        }
+
+        $identity = null;
+        if (isset($commandArgs['@context']['resolved_identity'])) {
+            $identity = $commandArgs['@context']['resolved_identity']->wait();
+        } else {
+            $credentialsProviderFn = $this->credentialsProvider;
+            $identity = $credentialsProviderFn()->wait();
+            $commandArgs['@context']['resolved_identity'] = Create::promiseFor($identity);
```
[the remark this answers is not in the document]
yenfryherrerafeliz: I think is safer to wrap the identity into a promise. I know we discussed that we are not logging $command in any of the middlewares, however, it could be done unintentionally in the future, or it could be appended by a custom customer implementation since the handler list is open for appending more middlewares. Wrapping the identity into a promise will make sure that if command is logged it will not log any credential values.

What do you think?

src/Middleware.php
```diff
@@ -144,6 +144,8 @@ function (TokenInterface $token)
 
                 if ($signer instanceof S3ExpressSignature) {
                     $credentialPromise = $config['s3_express_identity_provider']($command);
+                } elseif (isset($command['@context']['resolved_identity'])) {
+                    $credentialPromise = $command['@context']['resolved_identity'];
```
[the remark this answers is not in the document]
yenfryherrerafeliz: See my previous comment.

src/AwsClient.php
stobrien89: I'd still consider refactoring this

src/ClientResolver.php
```diff
@@ -290,6 +290,13 @@ class ClientResolver
             'doc'       => 'Set to false to disable checking for shared aws config files usually located in \'~/.aws/config\' and \'~/.aws/credentials\'.  This will be ignored if you set the \'profile\' setting.',
             'default'   => true,
         ],
+        'account_id_endpoint_mode' => [
+            'type'      => 'value',
+            'valid'     => ['string'],
+            'doc'       => 'To decide whether account_id must a be a required resolved credentials parameter. If this configuration is set to disabled, then account_id is not required. If set to preferred a warning will be logged when account_id is not resolved, and when set to required an exception will be thrown if account_id is not resolved.',
```
stobrien89, line 296: "decides whether account_id must a be a required resolved endpoint resolution parameter"

src/Credentials/CredentialProvider.php
```diff
@@ -291,9 +292,20 @@ public static function env()
...
+
             if ($key && $secret) {
                 return Promise\Create::promiseFor(
-                    new Credentials($key, $secret, getenv(self::ENV_SESSION) ?: NULL)
+                    new Credentials(
+                        $key,
+                        $secret,
+                        getenv(self::ENV_SESSION) ?: NULL,
+                        null,
+                        $accountId
```
stobrien89, line 307: Could probably use the elvis operator here too

src/Credentials/CredentialProvider.php
```diff
@@ -537,11 +549,18 @@ public static function ini($profile = null, $filename = null, array $config = []
...
+            }
+
             return Promise\Create::promiseFor(
                 new Credentials(
                     $data[$profile]['aws_access_key_id'],
                     $data[$profile]['aws_secret_access_key'],
-                    $data[$profile]['aws_session_token']
+                    $data[$profile]['aws_session_token'],
+                    null,
+                    $accountId
```
stobrien89, line 563: Can probably use null coalescing here

src/Credentials/InstanceProfileProvider.php
```diff
@@ -216,11 +216,17 @@ public function __invoke($previousCredentials = null)
...
+                    $accountId = $result['AccountId'];
+                }
+
                 $credentials = new Credentials(
                     $result['AccessKeyId'],
                     $result['SecretAccessKey'],
                     $result['Token'],
-                    strtotime($result['Expiration'])
+                    strtotime($result['Expiration']),
+                    $accountId
```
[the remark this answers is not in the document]
stobrien89: This still applies

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -17,6 +19,8 @@
  */
 class EndpointV2Middleware
 {
+    const ACCOUNT_ID_PARAM = 'AccountId';
+    const ACCOUNT_ID_ENDPOINT_MODE_PARAM = 'AccountIdEndpointMode';
```
stobrien89, line 23: newline needed here

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -36,24 +40,28 @@ class EndpointV2Middleware
 
     /** @var array */
     private $clientArgs;
```
stobrien89, line 42: newline needed here

src/EndpointV2/EndpointV2Middleware.php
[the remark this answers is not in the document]
stobrien89: Tested both and it's logged regardless of whether or not it's a promise.  The multi-auth PR contains some updates to the trace middleware which scrubs this from the logs.  I'd still like to keep `resolved_identity` "unwrapped" i.e. not contained within a promise

src/EndpointV2/EndpointV2Middleware.php
stobrien89, line 42: newline needed here

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -112,6 +121,11 @@ public function __invoke(CommandInterface $command)
     private function resolveArgs(array $commandArgs, Operation $operation) : array
     {
         $rulesetParams = $this->endpointProvider->getRuleset()->getParameters();
+
+        if (isset($rulesetParams[self::ACCOUNT_ID_PARAM]) && isset($rulesetParams[self::ACCOUNT_ID_ENDPOINT_MODE_PARAM])) {
```
stobrien89, line 125: Need to make this multi-line

stobrien89 (changes requested):

tests/AccountIdEndpointSupport/SourceAccountIdEndpointTest.php
stobrien89, line 1: We shouldn't be creating test classes that don't correspond to a class in our codebase.  It looks like this should go in the `StsClient` tests.  I'd split this out into 4 separate tests for readability

src/Credentials/InstanceProfileProvider.php
stobrien89, line 1: Same as the ecs provider- I think we could add an assertion or two on this in the test class like we're doing for some of the others

src/Credentials/EcsCredentialProvider.php
stobrien89, line 1: I think we could add an assertion or two on this in the test class like we're doing for some of the others

stobrien89 (changes requested):

stobrien89 (approved):
🚢 Ready to ship after ruleset testing

stobrien89 (approved):

yenfryherrerafeliz (commented):
Integ and Smoke tests passed:

....................

60 scenarios (60 passed)
230 steps (230 passed)
4m49.99s (41.13Mb)

......................

116 scenarios (116 passed)
232 steps (232 passed)
0m22.00s (86.33Mb)

stobrien89 (approved):

src/Credentials/Credentials.php
```diff
@@ -101,6 +110,7 @@ public function __unserialize($data)
         $this->secret = $data['secret'];
         $this->token = $data['token'];
         $this->expires = $data['expires'];
+        $this->accountId = $data['accountId'];
```
devfrey, line 113: This change broke my application in production. The `Credentials` class is serialized by `CredentialProvider::cache()`. Previously cached credentials don't have this key, resulting in an `Undefined array key "accountId"` error exception.

stobrien89: Hi @devfrey, apologies for the disruption.  Definitely an oversight on our part.  https://github.com/aws/aws-sdk-php/pull/2996 will most likely go out in today's release.

devfrey: Thank you for the swift resolution!

PR #2891, opened 2024-03-04
feat: multi-auth

stobrien89:
**Description of changes**:

Adds support for the service and operation-level auth trait, which allows a service to model multiple signature versions. When modeled, this trait will contain a non-empty priority-ordered list of signature versions to select from.  Selection is dependent on a few different factors: e.g. `v4a` signatures require installation of the crt extension; `bearer` signatures require that you are using bearer token credentials.  Selection will be performed on a per-request basis, with the operation-level modeled `auth` taking precedence over the service level modeled `auth`.

### DIFF
--- a/src/Auth/AuthSchemeResolver.php
+++ b/src/Auth/AuthSchemeResolver.php
@@ -1,0 +1,169 @@
+<?php
+
+namespace Aws\Auth;
+
+use Aws\Auth\Exception\UnresolvedAuthSchemeException;
+use Aws\Identity\AwsCredentialIdentity;
+use Aws\Identity\BearerTokenIdentity;
+use GuzzleHttp\Promise\PromiseInterface;
+
+/**
+ * Houses logic for selecting an auth scheme modeled in a service's `auth` trait.
+ * The `auth` trait can be modeled either in a service's metadata, or at the operation level.
+ */
+class AuthSchemeResolver implements AuthSchemeResolverInterface
+{
+    const UNSIGNED_BODY = '-unsigned-body';
+
+    /**
+     * @var string[] Default mapping of modeled auth trait auth schemes
+     *               to the SDK's supported signature versions.
+     */
+    private static $defaultAuthSchemeMap = [
+        'aws.auth#sigv4' => 'v4',
+        'aws.auth#sigv4a' => 'v4a',
+        'smithy.api#httpBearerAuth' => 'bearer',
+        'smithy.auth#noAuth' => 'anonymous'
+    ];
+
+    /**
+     * @var array Mapping of auth schemes to signature versions used in
+     *            resolving a signature version.
+     */
+    private $authSchemeMap;
+    private $tokenProvider;
+    private $credentialProvider;
+
+
+    public function __construct(
+        callable $credentialProvider,
+        callable $tokenProvider = null,
+        array $authSchemeMap = []
+    ){
+        $this->credentialProvider = $credentialProvider;
+        $this->tokenProvider = $tokenProvider;
+        $this->authSchemeMap = empty($authSchemeMap)
+            ? self::$defaultAuthSchemeMap
+            : $authSchemeMap;
+    }
+
+    /**
+     * Accepts a priority-ordered list of auth schemes and an Identity
+     * and selects the first compatible auth schemes, returning a normalized
+     * signature version.  For example, based on the default auth scheme mapping,
+     * if `aws.auth#sigv4` is selected, `v4` will be returned.
+     *
+     * @param array $authSchemes
+     * @param $identity
+     *
+     * @return string
+     * @throws UnresolvedAuthSchemeException
+     */
+    public function selectAuthScheme(
+        array $authSchemes,
+        array $args = []
+    ): string
+    {
+        $failureReasons = [];
+
+        foreach($authSchemes as $authScheme) {
+            $normalizedAuthScheme = $this->authSchemeMap[$authScheme] ?? $authScheme;
+
+            if ($this->isCompatibleAuthScheme($normalizedAuthScheme)) {
+                if ($normalizedAuthScheme === 'v4' && !empty($args['unsigned_payload'])) {
+                    return $normalizedAuthScheme . self::UNSIGNED_BODY;
+                }
+
+                return $normalizedAuthScheme;
+            } else {
+                $failureReasons[] = $this->getIncompatibilityMessage($normalizedAuthScheme);
+            }
+        }
+
+        throw new UnresolvedAuthSchemeException(
+            'Could not resolve an authentication scheme: '
+            . implode('; ', $failureReasons)
+        );
+    }
+
+    /**
+     * Determines compatibility based on either Identity or the availability
+     * of the CRT extension.
+     *
+     * @param $authScheme
+     *
+     * @return bool
+     */
+    private function isCompatibleAuthScheme($authScheme): bool
+    {
+        switch ($authScheme) {
+            case 'v4':
+            case 'anonymous':
+                return $this->hasAwsCredentialIdentity();
+            case 'v4a':
+                return extension_loaded('awscrt') && $this->hasAwsCredentialIdentity();
+            case 'bearer':
+                return $this->hasBearerTokenIdentity();
+            default:
+                return false;
+        }
+    }
+
+    /**
+     * Provides incompatibility messages in the event an incompatible auth scheme
+     * is encountered.
+     *
+     * @param $authScheme
+     *
+     * @return string
+     */
+    private function getIncompatibilityMessage($authScheme): string
+    {
+        switch ($authScheme) {
+            case 'v4':
+                return 'Signature V4 requires AWS credentials for request signing';
+            case 'anonymous':
+                return 'Anonymous signatures require AWS credentials for request signing';
+            case 'v4a':
+                return 'The aws-crt-php extension and AWS credentials are required to use Signature V4A';
+            case 'bearer':
+                return 'Bearer token credentials must be provided to use Bearer authentication';
+            default:
+                return "The service does not support `{$authScheme}` authentication.";
+        }
+    }
+
+    /**
+     * @return bool
+     */
+    private function hasAwsCredentialIdentity(): bool
+    {
+        $fn = $this->credentialProvider;
+        $result = $fn();
+
+        if ($result instanceof PromiseInterface) {
+            return $result->wait() instanceof AwsCredentialIdentity;
+        }
+
+        return $result instanceof AwsCredentialIdentity;
+    }
+
+    /**
+     * @return bool
+     */
+    private function hasBearerTokenIdentity(): bool
+    {
+        if ($this->tokenProvider) {
+            $fn = $this->tokenProvider;
+            $result = $fn();
+
+            if ($result instanceof PromiseInterface) {
+                return $result->wait() instanceof BearerTokenIdentity;
+            }
+
+            return $result instanceof BearerTokenIdentity;
+        }
+
+        return false;
+    }
+}
--- a/src/Auth/AuthSelectionMiddleware.php
+++ b/src/Auth/AuthSelectionMiddleware.php
@@ -1,0 +1,99 @@
+<?php
+namespace Aws\Auth;
+
+use Aws\Api\Service;
+use Aws\CommandInterface;
+use Closure;
+use GuzzleHttp\Promise\Promise;
+
+/**
+ * Handles auth scheme resolution. If a service models and auth scheme using
+ * the `auth` trait and the operation or metadata levels, this middleware will
+ * attempt to select the first compatible auth scheme it encounters and apply its
+ * signature version to the command's `@context` property bag.
+ *
+ * IMPORTANT: this middleware must be added to the "build" step.
+ *
+ * @internal
+ */
+class AuthSelectionMiddleware
+{
+    /** @var callable */
+    private $nextHandler;
+
+    /** @var AuthSchemeResolverInterface */
+    private $authResolver;
+
+    /** @var Service */
+    private $api;
+
+    /**
+     * Create a middleware wrapper function
+     *
+     * @param AuthSchemeResolverInterface $authResolver
+     * @param Service $api
+     * @return Closure
+     */
+    public static function wrap(
+        AuthSchemeResolverInterface $authResolver,
+        Service $api
+    ): Closure
+    {
+        return function (callable $handler) use ($authResolver, $api) {
+            return new self($handler, $authResolver, $api);
+        };
+    }
+
+    /**
+     * @param callable $nextHandler
+     * @param $authResolver
+     * @param callable $identityProvider
+     * @param Service $api
+     */
+    public function __construct(
+        callable $nextHandler,
+        AuthSchemeResolverInterface $authResolver,
+        Service $api
+    )
+    {
+        $this->nextHandler = $nextHandler;
+        $this->authResolver = $authResolver;
+        $this->api = $api;
+    }
+
+    /**
+     * @param CommandInterface $command
+     *
+     * @return Promise
+     */
+    public function __invoke(CommandInterface $command)
+    {
+        $nextHandler = $this->nextHandler;
+        $serviceAuth = $this->api->getMetadata('auth') ?: [];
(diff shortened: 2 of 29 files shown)


src/S3/S3Client.php
```diff
@@ -502,9 +502,9 @@ public function createPresignedRequest(CommandInterface $command, $expires, arra
         $command = clone $command;
         $command->getHandlerList()->remove('signer');
         $request = \Aws\serialize($command);
-        $signing_name = empty($command->getAuthSchemes())
-            ? $this->getSigningName($request->getUri()->getHost())
-            : $command->getAuthSchemes()['name'];
+        $signing_name = isset($command['@context']['signing_service'])
+            ? $command['@context']['signing_service']
+            :$this->getSigningName($request->getUri()->getHost());
```
yenfryherrerafeliz, line 507: nit: a space is missing here at:
> :$this->getSigningName($request->getUri()->getHost());
```
: $this->getSigningName($request->getUri()->getHost());
```

src/TraceMiddleware.php
```diff
@@ -40,6 +42,13 @@ class TraceMiddleware
         '/\["key.{27,36}Stream.{9}\]=>\s+.{7}\d{2}\) "\X{16,64}"/U' => '["key":[CONTENT KEY]]',
     ];
 
+    private static $sensitiveContext = [
+        'resolved_identity' => [
+            AwsCredentialIdentity::class => true,
+            BearerTokenIdentity::class => true
```
yenfryherrerafeliz, line 48: Should we add the interface here that these subclasses implements instead?, just because it is more explicit in which type of objects should be considered as sensitive.

yenfryherrerafeliz (changes requested):

src/Auth/AuthSelectionMiddleware.php
```diff
@@ -0,0 +1,120 @@
...
+     * @param CommandInterface $command
+     *
+     * @return Promise
+     */
+    public function __invoke(CommandInterface $command)
+    {
+        $nextHandler = $this->nextHandler;
+        $serviceAuth = $this->api->getMetadata('auth') ?: [];
+        $operation = $this->api->getOperation($command->getName());
+        $operationAuth = isset($operation['auth']) ? $operation['auth'] : [];
```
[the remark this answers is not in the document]
stobrien89: Because this middleware is directly in the `test-phar` test path (it instantiates an S3 client), we have to make sure this is backward-compatible.  at least until we update our doc generation.

src/AwsClient.php
```diff
@@ -420,31 +427,20 @@ private function addSignatureMiddleware()
             if (!empty($c['@context']['signing_service'])) {
                 $name = $c['@context']['signing_service'];
             }
+            if (!empty($c['@context']['signature_version'])) {
+                $version = $c['@context']['signature_version'];
```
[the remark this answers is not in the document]
stobrien89: That's fair.  I just left it because it's been that way for ages.

src/Identity/AwsCredentialIdentity.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Aws\Identity;
+
+/**
+ * Denotes the use of standard AWS credentials.
+ *
+ * @internal
+ */
+class AwsCredentialIdentity implements IdentityInterface
+{
+    public function getExpiration() {}
```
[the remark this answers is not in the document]
stobrien89: I don't think it's necessary. I've mentioned before that we're using these for Identification purposes only since Credentials, Token, etc. already implement the desired functionality.  I've marked them as internal, but maybe I need to be a bit more explicit about what they are used for?

src/TraceMiddleware.php
stobrien89: What would that look like? I think it's fine to show which identity has been selected- I'm only concerned about exposing auth strings

src/Auth/AuthSchemeResolver.php
```diff
@@ -0,0 +1,153 @@
...
+    {
+        $failureReasons = [];
+
+        foreach($authSchemes as $authScheme) {
+            $normalizedAuthScheme = isset($this->authSchemeMap[$authScheme])
+                ? $this->authSchemeMap[$authScheme]
+                : $authScheme;
+
+            if ($this->isCompatibleAuthScheme($normalizedAuthScheme)) {
+                if ($normalizedAuthScheme === 'v4' && !empty($args['unsigned_payload'])) {
```
yenfryherrerafeliz, line 74: Is it guaranteed that $args will always have a `unsigned_payload` key set?, if so, why not to evaluate `$args['unsigned_payload']` as a boolean type?

yenfryherrerafeliz (changes requested):

src/Auth/AuthSchemeResolver.php
stobrien89: In the default case it is always set.  But if this is instantiated and invoked independently it's not

src/Identity/AwsCredentialIdentity.php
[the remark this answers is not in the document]
kstich: I'd both be explicit about that in docs and make these abstract classes/methods. These classes could technically be instantiated and used, however invalidly, while satisfying the defined contract. Marking them abstract doesn't harm functionality of the already implemented pieces, as they've implemented it, and you still get accurate type checking via `instanceof`.

src/Identity/IdentityInterface.php
kstich, line 1: Should have more documentation for this interface. What's it used for? What is that method's expected contract?

src/AwsClient.php
```diff
@@ -407,45 +414,34 @@ private function addSignatureMiddleware()
...
+            switch ($authType) {
                 case 'none':
-                    $version = 'anonymous';
+                    $signatureVersion = 'anonymous';
                     break;
                 case 'v4-unsigned-body':
-                    $version = 'v4-unsigned-body';
+                    $signatureVersion = 'v4-unsigned-body';
                     break;
-                case 'bearer':
```
kstich, line 432: I may have missed it, but where did resolving a `bearer` entry land?

src/Identity/AwsCredentialIdentity.php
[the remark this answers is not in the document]
stobrien89: Puzzled as to why I pushed back on this initially, but I think my wires were crossed when thinking about how this would work in the implementation/inheritance hierarchy.  Updated.

src/Command.php
```diff
@@ -60,27 +60,23 @@ public function getHandlerList()
...
+     *             in a Command's context, if applicable.
      */
     public function getAuthSchemes()
     {
+        trigger_error(__METHOD__ . ' is deprecated.  Auth schemes '
+        . 'resolved using the service `auth` trait or via endpoint resolution '
+        . 'can now be found in the command `@context` property.`'
+        , E_USER_DEPRECATED);
+
         return $this->authSchemes ?: [];
```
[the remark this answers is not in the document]
stobrien89: This is an attempt to walk back a poor design choice made during the initial EP 2.0 implementation.  The `authSchemes` property was only set (using the preceding method) _after_ endpoint resolution if the resolved endpoint contained auth schemes. If an auth scheme was resolved from the model, whether it be the operation or at the service metadata level, this would return nothing.

src/Command.php
```diff
@@ -60,27 +60,23 @@ public function getHandlerList()
         return $this->handlerList;
     }
 
-    /**
-     * For overriding auth schemes on a per endpoint basis when using
-     * EndpointV2 provider. Intended for internal use only.
-     *
-     * @param array $authSchemes
-     *
-     * @internal
```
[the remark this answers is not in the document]
stobrien89: Updated to emit a deprecation notice.

src/AwsClient.php
stobrien89: As far as I'm aware, there hasn't been a service that has modeled `bearer` in an operation's `authtype`, but looking at the bearer token spec I see it's supported.  I figured multi-auth would render this useless, but I'll re-add it just in case

src/Identity/IdentityInterface.php
stobrien89: updated

src/QueryCompatibleInputMiddleware.php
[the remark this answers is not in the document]
stobrien89: I realized we hadn't adhered to any particular spacing when declaring return types, but I'll save this for another PR. There are other places this needs to be updated

src/AwsClient.php
```diff
@@ -407,45 +414,37 @@ private function addSignatureMiddleware()
...
                 case 'none':
-                    $version = 'anonymous';
+                    $signatureVersion = 'anonymous';
                     break;
                 case 'v4-unsigned-body':
-                    $version = 'v4-unsigned-body';
+                    $signatureVersion = 'v4-unsigned-body';
                     break;
                 case 'bearer':
                     $version = 'bearer';
```
kstich, line 443: Shouldn't this be `signatureVersion` now? Is there a test set for this?

src/Command.php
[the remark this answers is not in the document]
kstich: Endpoint resolution based auth should continue setting this content and just not using it, a method overload could be added to suppress the deprecation notice for that call. The method should continue functioning as it did for its case while deprecated.

src/Command.php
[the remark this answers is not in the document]
kstich: These should note that setting this content does not result in it being consumed.

src/Identity/AwsCredentialIdentity.php
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Denotes the use of standard AWS credentials.
+ *
+ * @internal
+ */
+abstract class AwsCredentialIdentity implements IdentityInterface
+{
+    /**
+     * Returns a UNIX timestamp, if available, representing the expiration
```
kstich, line 13: What does it do if one isn't available? Is this same conditional part of `BearerTokenIdentity` and not documented, or do those always have expirations?

The return type should also be noted.

src/Identity/BearerTokenIdentity.php
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Denotes the use of Bearer Token credentials.
+ *
+ * @internal
+ */
+abstract class BearerTokenIdentity implements IdentityInterface
+{
+    /**
+     * Returns a UNIX timestamp representing the expiration
```
kstich, line 13: The return type should be noted.

kstich (changes requested):
Has the integration test suite been run against these changes?

src/AwsClient.php
stobrien89: There wasn't a test on this when bearer was implemented. I've added a case in `AwsClientTest` (`testSignOperationsWithAnAuthType`) which covers this.

src/Identity/AwsCredentialIdentity.php
stobrien89: Updated. The annotations for both identity types should be identical

stobrien89:
Integration and smoke test suites have been run and are passing.

src/Command.php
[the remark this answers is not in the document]
stobrien89: Added an error suppression operator to the call, but definitely open to overriding `__call` if you think we should go that route

src/Auth/AuthSchemeResolver.php
```diff
@@ -0,0 +1,153 @@
...
+     * @return string
+     */
+    private function getIncompatibilityMessage($authScheme): string
+    {
+        switch ($authScheme) {
+            case 'v4a':
+                return 'The aws-crt-php extension must be installed to use Signature V4A';
+            case 'bearer':
+                return 'Bearer token credentials must be provided to use Bearer authentication';
+            default:
```
kstich, line 128: minor: Should there be a case for `v4` but having no AWS credentials?

src/Auth/AuthSchemeResolver.php
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @param $authScheme
+     *
+     * @return bool
+     */
+    private function isCompatibleAuthScheme($authScheme): bool
+    {
+        switch ($authScheme) {
+            case 'v4':
+            case 'anonymous':
```
kstich, line 102: Does `anonymous` auth require AWS credentials? Does it require any credentials at all?

stobrien89: It technically doesn't, but [`AnonymousSignature`](https://github.com/aws/aws-sdk-php/blob/master/src/Signature/AnonymousSignature.php#L15C5-L20C6) requires an instance of `CredentialsInterface` for signing even though it isn't used.  When customers set `credentials` to `false` in a client constructor, a dummy credentials object is created.

src/Auth/AuthSchemeResolver.php
stobrien89: I'll add one.  There's a chance someone could provide a callable that doesn't return an instance of `CredentialsInterface` for `credentials` in the client constructor.

stobrien89: Updated.  Also changed the way we're handling the providers and added some new tests in `AuthSchemeResolverTest`

src/Auth/AuthSchemeResolver.php
kstich: Should there be a separate instance then, one specific for anonymous?

stobrien89: Ideally.  Currently it doesn't matter if a customer sets credentials to `false` or not— the SDK will accept whatever credentials are available to pass into the `AnonymousSignature` signing functions (e.g. in the case of `authtype` being set to `none`).  To remain backward compatible, we still have to accept `AwsCredentialIdentity` here and an instance of `CredentialsInterface` in `AnonymousSignature`'s signing functions, which makes the implementation of `AnonymousCredentials`, `AnonymousIdentity`, etc. seem like a lot of effort for minimal gain.

kstich: This case statement is saying that `anonymous` only accepts AWS credentials though, and couldn't be provided bearer credentials (even if ignored). Maybe this should just be a `return true` case on its own?

stobrien89: >This case statement is saying that anonymous only accepts AWS credentials though, and couldn't be provided bearer credentials (even if ignored).

As nonsensical as it seems that is actually true.  Bearer takes a different path than other signature types and `anonymous` is in the group of [signatures that require an instance of `CredentialsInterface`](https://github.com/aws/aws-sdk-php/blob/29a79bac02997f3053559f6961a0e83622a14f88/src/Middleware.php#L151,L159).

kstich: To close the loop, this is done [when `_apply_credentials`](https://github.com/aws/aws-sdk-php/blob/master/src/ClientResolver.php#L586) gets a value of `false` and creates an empty `Credentials` object, which this PR is adjusting to implement the `AwsCredentialIdentity` interface to pass this check.

kstich (approved):

src/Auth/AuthSchemeResolver.php
stobrien89: Exactly- that's the explicitly configured case, and there's another where [`anonymous` could be set via `authtype`](https://github.com/aws/aws-sdk-php/blob/master/src/AwsClient.php#L426,L428) and a non-empty `Credentials` object could be used.  

Just for the sake of clarity, (bearer)`Token` and `TokenInterface` are not related to `Credentials` and `CredentialsInterface`, which is why we're checking for `AwsCredentialIdentity`

yenfryherrerafeliz (approved):

yenfryherrerafeliz (approved):

rubas:
🚨 This breaks `Aws\SecretsManager\SecretsManagerClient`.

```
$client = new SecretsManagerClient([
      'credentials' => new Credentials(),
      'region'      => 'eu-central-1',
    ]);
    
$client->getSecretValue(['SecretId' => $name]);

---
 
Aws\Auth\Exception\UnresolvedAuthSchemeException  in AuthSchemeResolver.php on line 83:

Could not resolve an authentication scheme: Signature V4 requires AWS credentials for request signing
```

PR #2892, opened 2024-03-06
enhancement: update auth scheme resolution

stobrien89:
*Description of changes:*
Updates to auth scheme resolution, which allows for multiple auth schemes to be modeled in priority order.  For example, a service might prioritize `sigv4a`, which requires the `awscrt` extension, but list `sigv4` as a fallback.  

### DIFF
--- a/src/EndpointV2/EndpointV2Middleware.php
+++ b/src/EndpointV2/EndpointV2Middleware.php
@@ -3,6 +3,7 @@ namespace Aws\EndpointV2;
 
 use Aws\Api\Operation;
 use Aws\Api\Service;
+use Aws\Auth\Exception\UnresolvedAuthSchemeException;
 use Aws\CommandInterface;
 use Closure;
 use GuzzleHttp\Promise\Promise;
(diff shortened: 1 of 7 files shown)


src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -247,19 +248,32 @@ private function resolveAuthScheme(array $authSchemes) : array
 
         foreach($authSchemes as $authScheme) {
             if (isset(self::$validAuthSchemes[$authScheme['name']])) {
+                switch($authScheme['name']) {
```
yenfryherrerafeliz, line 251: nit: Could we maybe create a separated method for validating if the authScheme is valid?.  That would make the code more clear. You have a similar implementation in multi-auth PR [here](https://github.com/aws/aws-sdk-php/pull/2891/files#diff-27f526732eeef580d2bed18ed21b00547a25f6cebf14cc91e552b043f72526c3R85). Could you reuse that same method and  in this case just pass Identity as null.

yenfryherrerafeliz (changes requested):

src/EndpointV2/EndpointV2Middleware.php
stobrien89: I think a separate method is a good idea, but the logic needs to be different for reasons I'll explain elsewhere

Issue #2887, opened 2024-02-26
InstanceProfile credentials provider causing open_basedir restriction error since 3.297.0


mensler:
### Describe the bug

With version 3.297.0 of the SDK the custom endpoint for IMDS was introduced in the InstanceProfileProvider.
It calls `ConfigurationResolver::resolve()` with the option `use_aws_shared_config_files` set to true, which causes an ErrorException because of the open_basedir setting.

### Expected Behavior

The config option `use_aws_shared_config_files` should not be set to true, but reflect the client setting instead, so no error is thrown because of open_basedir restrictions.

### Current Behavior

```
ErrorException: is_readable(): open_basedir restriction in effect. File(/usr/share/httpd/.aws/config) is not within the allowed path(s): (/var/www/:/tmp/:/usr/share/GeoIP/)
#38 [internal](0): is_readable
#37 /vendor/aws/aws-sdk-php/src/Configuration/ConfigurationResolver.php(111): Aws\Configuration\ConfigurationResolver::ini
#36 /vendor/aws/aws-sdk-php/src/Configuration/ConfigurationResolver.php(47): Aws\Configuration\ConfigurationResolver::resolve
#35 /vendor/aws/aws-sdk-php/src/Credentials/InstanceProfileProvider.php(368): Aws\Credentials\InstanceProfileProvider::resolveEndpoint
#34 /vendor/aws/aws-sdk-php/src/Credentials/InstanceProfileProvider.php(252): Aws\Credentials\InstanceProfileProvider::request
#33 /vendor/aws/aws-sdk-php/src/Credentials/InstanceProfileProvider.php(107): Aws\Credentials\InstanceProfileProvider::Aws\Credentials\{closure}
#32 [internal](0): Generator::current
#31 /vendor/guzzlehttp/promises/src/Coroutine.php(71): GuzzleHttp\Promise\Coroutine::__construct
#30 /vendor/guzzlehttp/promises/src/Coroutine.php(86): GuzzleHttp\Promise\Coroutine::of
#29 /vendor/aws/aws-sdk-php/src/Credentials/InstanceProfileProvider.php(101): Aws\Credentials\InstanceProfileProvider::__invoke
#28 /vendor/aws/aws-sdk-php/src/Credentials/CredentialProvider.php(214): Aws\Credentials\CredentialProvider::Aws\Credentials\{closure}
```

### Reproduction Steps

* Restrict `open_basedir` in php.ini with excluded Apache working directory.
[...]

PR #2894, opened 2024-03-06
fix: set use_aws_shared_config_files opts-in

yenfryherrerafeliz:
This change makes the ConfigurationResolver to load configuration from aws shared config files based on the configuration `use_aws_shared_config_files` either provided as provider arguments or as service client arguments.

*Issue #2887*

*Description of changes:*

### DIFF
--- a/src/Credentials/InstanceProfileProvider.php
+++ b/src/Credentials/InstanceProfileProvider.php
@@ -60,6 +60,9 @@ class InstanceProfileProvider
     /** @var string */
     private $endpointMode;
 
+    /** @var array */
+    private $config;
+
     /**
      * The constructor accepts the following options:
      *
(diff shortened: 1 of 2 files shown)


src/Credentials/InstanceProfileProvider.php
```diff
@@ -60,6 +60,9 @@ class InstanceProfileProvider
     /** @var string */
     private $endpointMode;
 
+    /** @var bool */
```
stobrien89, line 63: This property just feels like it doesn't belong to this class.  Could we set a `$config` property instead and pass that the the resolver?

stobrien89 (changes requested):

stobrien89 (approved):
:shipit:

PR #2895, opened 2024-03-07
Check based on Value 

AKHIL-882:
The modified _apply_region function now checks if the $value parameter is set. If it is set, it assigns the value to $args['region']. If $value is not set, it calls the _default_region function to retrieve the default region value and assigns it to $args['region'].

After setting the region value, the function proceeds to check if the value is empty. If it is empty, it invokes the _missing_region function, indicating that the region is missing from the arguments.

Overall, the modified function ensures that the region parameter is either explicitly set or assigned a default value, and it handles the case where the region value is empty by calling a separate function to handle the missing region scenario.

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -1190,10 +1190,15 @@ class ClientResolver
   
     public static function _apply_region($value, array &$args)
     {
+        if (isset($value)) {
+            $args['region'] = $value;
+        } else {
+            $args['region'] = self::_default_region($args);
+        }
+        
         if (empty($value)) {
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @AKHIL-882,

Thanks for the PR.  The `ClientResolver.php` calls a config key's `'default'` method if a value for that key is not provided in the client constructor, meaning currently `_default_region` is called before `_apply_region`.  `_apply_region` just ensures the value provided in the client constructor _or_ `_default_region` isn't empty before applying it to the client's config.  Unfortunately we can't accept this change.

PR #2906, opened 2024-04-01
feat: add streaming support on JSON RPC parser

yenfryherrerafeliz:
Add event streaming capabilities on the JSON RPC parser, so that operations such as StartLiveTail from CloudwatchLogs are properly handled/parsed. This change also includes an extension of the DecodingEventStreamIterator for handling streamable responses where the stream as response is not seekable, and a handling for when the `event-type` is `initial-response

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php
+++ b/src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace Aws\Api\Parser;
+
+use GuzzleHttp\Psr7;
+use Psr\Http\Message\StreamInterface;
+use Aws\Api\Parser\Exception\ParserException;
+
+/**
+ * @inheritDoc
+ */
+class NonSeekableStreamDecodingEventStreamIterator extends DecodingEventStreamIterator
+{
+    /** @var array $tempBuffer */
+    private $tempBuffer;
+
+    /**
+     * NonSeekableStreamDecodingEventStreamIterator constructor.
+     *
+     * @param StreamInterface $stream
+     */
+    public function __construct(StreamInterface $stream)
+    {
+        $this->stream = $stream;
+        if ($this->stream->isSeekable()) {
+            throw new \InvalidArgumentException('The stream provided must be not seekable.');
+        }
+
+        $this->tempBuffer = [];
+    }
+
+    /**
+     * @inheritDoc
+     *
+     * @return array
+     */
+    protected function parseEvent(): array
+    {
+        $event = [];
+        $this->hashContext = hash_init('crc32b');
+        $prelude = $this->parsePrelude()[0];
+        list(
+            $event[self::HEADERS],
+            $numBytes
+        ) = $this->parseHeaders($prelude[self::LENGTH_HEADERS]);
+        $event[self::PAYLOAD] = Psr7\Utils::streamFor(
+            $this->readAndHashBytes(
+                $prelude[self::LENGTH_TOTAL] - self::BYTES_PRELUDE
+                - $numBytes - self::BYTES_TRAILING
+            )
+        );
+        $calculatedCrc = hash_final($this->hashContext, true);
+        $messageCrc = $this->stream->read(4);
+        if ($calculatedCrc !== $messageCrc) {
+            throw new ParserException('Message checksum mismatch.');
+        }
+
+        return $event;
+    }
+
+    protected function readAndHashBytes($num): string
+    {
+        $bytes = '';
+        while (!empty($this->tempBuffer) && $num > 0) {
+            $byte = array_shift($this->tempBuffer);
+            $bytes .= $byte;
+            $num = $num - 1;
+        }
+
+        $bytes = $bytes . $this->stream->read($num);
+        hash_update($this->hashContext, $bytes);
+
+        return $bytes;
+    }
+
+    // Iterator Functionality
+
+    #[\ReturnTypeWillChange]
+    public function rewind()
+    {
+        $this->currentEvent = $this->parseEvent();
+    }
+
+    public function next()
+    {
+        $this->tempBuffer[] = $this->stream->read(1);
+        if ($this->valid()) {
+            $this->key++;
+            $this->currentEvent = $this->parseEvent();
+        }
+    }
+
+    /**
+     * @return bool
+     */
+    #[\ReturnTypeWillChange]
+    public function valid()
+    {
+        return !$this->stream->eof();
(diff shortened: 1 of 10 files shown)


src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -155,4 +156,46 @@
...
+                if (!empty(self::$streamingCommands[$command->getName()])) {
+                    $command['@http']['stream'] = true;
+                }
+
+                return $handler($command, $request);
+            };
+        };
+    }
+
+    public function startLiveTailCheckingForResults($args): \Generator
```
stobrien89, line 188: Where is this used?

src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -155,4 +156,46 @@
...
+        parent::__construct($args);
+        $this->appendClientCustomMiddlewares();
+    }
+
+    private function appendClientCustomMiddlewares()
+    {
+        $this->getHandlerList()->appendInit($this->getSetStreamingFlagMiddleware(), 'streaming-flag-middleware');
+    }
+
+    private function getSetStreamingFlagMiddleware(): callable
```
stobrien89, line 175: the "getSet" part of the name could be a little confusing.  We could name this `getStreamingFlagMiddleware` or something similar

src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -155,4 +156,46 @@
...
+        'StartLiveTail' => true
+    ];
+
+    public function __construct(array $args)
+    {
+        parent::__construct($args);
+        $this->appendClientCustomMiddlewares();
+    }
+
+    private function appendClientCustomMiddlewares()
```
stobrien89, line 170: I don't have any issues with using a separate function for this, but we might want to call this `addStreamingFlagMiddleware` and make each `add` function for middleware separate.  We won't always be appending middleware- it can be prepending or inserted before or after another middleware

src/CloudWatchLogs/CloudWatchLogsClient.php
yenfryherrerafeliz: This is just a helper method on top of startLiveTail command that yields the results when they are available.

src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -155,4 +156,61 @@
  * @method \Aws\Result updateLogAnomalyDetector(array $args = [])
  * @method \GuzzleHttp\Promise\Promise updateLogAnomalyDetectorAsync(array $args = [])
  */
-class CloudWatchLogsClient extends AwsClient {}
+class CloudWatchLogsClient extends AwsClient {
+    static $streamingCommands = [
```
stobrien89, line 160: Nit: maybe rename this to `$streamingFlagCommands`?

src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -155,4 +156,61 @@
...
+     *
+     * Initiates 'startLiveTail' operation with given arguments, and continuously
+     * checks response stream for session updates or results, yielding each
+     * stream chunk when results are not empty.
+     *
+     * @param array $args Command arguments.
+     *
+     * @return \Generator Yields session update or result stream chunks.
+     */
+    public function startLiveTailCheckingForResults(array $args): \Generator
```
stobrien89, line 203: I think the backslash in Generator is throwing off the type removal script.  If I'm not mistaken, it's looking only for alphabetical characters.  This will need to be updated here and in our release system.

stobrien89 (changes requested):

src/CloudWatchLogs/CloudWatchLogsClient.php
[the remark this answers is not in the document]
yenfryherrerafeliz: This helper method abstract from the customer the need of checking for results. This means that it will just yield a result once logs are available.

src/CloudWatchLogs/CloudWatchLogsClient.php
yenfryherrerafeliz: I do not think so. The intention of this variable is to hold the name of the commands that require streaming, therefore I think $streamingCommands suits better here.

src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @inheritDoc
+ */
+class NonSeekableStreamDecodingEventStreamIterator extends DecodingEventStreamIterator
+{
+    /** @var array $tempBuffer */
+    private $tempBuffer;
+
+    /**
+     * NoSeekStreamDecodingEventStreamIterator constructor.
```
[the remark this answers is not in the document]
yenfryherrerafeliz: Done.

src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php
```diff
@@ -0,0 +1,96 @@
...
+            throw new ParserException('Message checksum mismatch.');
+        }
+
+        return $event;
+    }
+
+    protected function readAndHashBytes($num): string
+    {
+        $bytes = '';
+        if (!empty($this->tempBuffer)) {
```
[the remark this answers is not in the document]
yenfryherrerafeliz: Yes, that is correct. I removed.

src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php
```diff
@@ -0,0 +1,96 @@
...
+     * @inheritDoc
+     *
+     * @return array
+     */
+    protected function parseEvent(): array
+    {
+        $event = [];
+        $this->hashContext = hash_init('crc32b');
+        $prelude = $this->parsePrelude()[0];
+        list(
```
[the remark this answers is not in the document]
yenfryherrerafeliz: I do not think this is needed. I know the way suggested is a more modern way to do it, but we could consider this when further modernizing the SDK.

src/CloudWatchLogs/CloudWatchLogsClient.php
yenfryherrerafeliz: I removed.

stobrien89 (approved):
:shipit: Just make sure to run the smoke and integ tests.  Also, add a changelog entry.

PR #2915, opened 2024-05-06
fix: add s3-outpost as s3 signing service

yenfryherrerafeliz:
Add s3-outpost as a s3 signing service so that when resolving the signer and the service is s3-outpost then, S3SignatureV4 is returned.

*Issue #, if available:*

*Description of changes:*


src/Signature/SignatureProvider.php
stobrien89, line 1: this should be `s3-outposts`

stobrien89 (changes requested):

src/Signature/SignatureProvider.php
yenfryherrerafeliz: Done

stobrien89 (approved):

PR #2916, opened 2024-05-07
feat: new api documentation theme

stobrien89:
*Description of changes:*
Updates API doc generation.  Now using PHPDocumentor for generation with a few overrides to their theming.  


yenfryherrerafeliz (commented):
Would it be possible to have the search result windows pops up after hitting enter? It shows up by typing just a single letter.

stobrien89:
I have no idea what's happening with the encryption traits.  I've updated the type annotations for the method, but it's still coming back with `Unexpected token "[", expected type at offset 0 on line 1`.  Nothing appears to be wrong with either file.

The search modal position has been adjusted so it appears under the search bar.  I think it's fine to show results as the search is being typed.

yenfryherrerafeliz (approved):

PR #2917, opened 2024-05-08
feat: add AppID user agent parameter

yenfryherrerafeliz:
AppID is an user agent parameter that can be sourced from either client config, config file, and environment variable.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -270,6 +270,15 @@ class ClientResolver
             'fn'       => [__CLASS__, '_apply_handler'],
             'default'  => [__CLASS__, '_default_handler']
         ],
+        'app_id' => [
+            'type' => 'value',
+            'valid' => ['string'],
+            'doc' => 'app_id(AppId) is an optional application specific identifier that can be set. 
+             When set it will be appended to the User-Agent header of every request in the form of App/{AppId}. 
+             This value is also sourced from environment variable AWS_SDK_UA_APP_ID or the shared config profile attribute sdk_ua_app_id.',
+            'fn' => [__CLASS__, '_apply_app_id'],
+            'default' => [__CLASS__, '_default_app_id']
+        ],
         'ua_append' => [
             'type'     => 'value',
             'valid'    => ['string', 'array'],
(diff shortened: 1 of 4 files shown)


src/ClientResolver.php
```diff
@@ -917,17 +926,41 @@ public static function _apply_http_handler($value, array &$args, HandlerList $li
         );
     }
 
+    public static function _apply_app_id($inputAppId, array &$args)
+    {
+        // AppId should not be longer than 50 chars
+        static $MAX_APP_ID_LENGTH = 50;
+        if (strlen($inputAppId) > $MAX_APP_ID_LENGTH) {
+            trigger_error("The provided or configured value for AppId, an user agent parameter, exceeds the maximum length of $MAX_APP_ID_LENGTH characters.", E_USER_WARNING);
```
stobrien89, line 934: Let's try to make this multi-line

src/ClientResolver.php
```diff
@@ -917,17 +926,41 @@ public static function _apply_http_handler($value, array &$args, HandlerList $li
         );
     }
 
+    public static function _apply_app_id($inputAppId, array &$args)
```
stobrien89, line 929: let's stick with the convention we use in this class of calling the value `$value`

src/ClientResolver.php
```diff
@@ -917,17 +926,41 @@ public static function _apply_http_handler($value, array &$args, HandlerList $li
...
+        }
+
+        $args['app_id'] = $inputAppId;
+    }
+
+    public static function _default_app_id(array $args)
+    {
+        return ConfigurationResolver::resolve(
+            'sdk_ua_app_id',
+            '',
```
stobrien89, line 944: Not familiar with the spec— the default is an empty string?

tests/ClientResolverTest.php
stobrien89, line 1: We should add a test similar to some of the other user agent tests we have here: something like `testUserAgentAddsAppId`

stobrien89 (changes requested):

src/ClientResolver.php
yenfryherrerafeliz: There is not a default value for app_id. If empty is returned here then, that means customer did not provided or configured an app_id value, and in that case this parameter is just ignored and not appended to the user agent header.

tests/ClientResolverTest.php
yenfryherrerafeliz: I think that is covered already [here](https://github.com/aws/aws-sdk-php/pull/2917/files#diff-e42a2aab192ff729219c1686f0383a82d6f38c2fb355bc7708a917213da71688R1579). The way I do the evaluation for the test is by checking the values from the user agent header from the request being sent.

stobrien89 (changes requested):
Good to 🚢 after a changelog entry

stobrien89 (approved):

PR #2918, opened 2024-05-09
Honor EncodingType in response with ListObjectsV2 as with ListObjects

nanawel:
There seems to be a bug with `ListObjectsV2` when dealing with type encoding in responses. The special middleware that's responsible for url-decoding the values in some nodes can only be triggered with `ListObjects`, so in V2 the client ends up with URL-encoded strings that are not valid in the rest of the application.

https://github.com/aws/aws-sdk-php/blob/0de11adba8f174a76346c9dbb175c2030ff5c8f2/src/S3/S3Client.php#L642-L652

This patch should fix this issue while keeping compatibility with PHP 7.


src/S3/S3Client.php
```diff
@@ -785,7 +785,7 @@ private function getEncodingTypeMiddleware()
         return static function (callable $handler) {
             return function (Command $command, $request = null) use ($handler) {
                 $autoSet = false;
-                if ($command->getName() === 'ListObjects'
+                if (strpos($command->getName(), 'ListObjects') === 0
```
stobrien89, line 788: Can we change this to a `str_starts_with()`?

src/S3/S3Client.php
stobrien89, line 1: Could you either turn [this test](https://github.com/aws/aws-sdk-php/blob/master/tests/S3/S3ClientTest.php#L1048C5-L1089C6) into a parameterized test which call both `listObjects` and `listObjectsV2` _or_ create a separate unit test for `listObjectsV2`?

stobrien89 (changes requested):
Just a minor change and test requested

nanawel:
Hi,
Sorry but in a year (!) I've moved to another project.
Feel free to carry the changes if you think they're useful.

PR #2924, opened 2024-05-17
In a subresource the parameter value will be an empty string.

jeijei4:
*Issue #, if available:*

*Description of changes:*

The documentation explains:
when a request targets a subresource, the corresponding query parameter value will be an empty string ("").

Example:
`"http://s3.amazonaws.com/examplebucket?acl"`

The CanonicalQueryString in this case is as follows:

`UriEncode("acl") + "=" + ""`

https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html#create-canonical-request

### DIFF
--- a/src/Signature/SignatureV4.php
+++ b/src/Signature/SignatureV4.php
@@ -348,6 +348,11 @@ class SignatureV4 implements SignatureInterface
         $qs = '';
         ksort($query);
         foreach ($query as $k => $v) {
+            if(is_numeric($k)){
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @jeijei4,

Thanks for the PR. Apologies for the delayed reply. Thanks for raising this — `Psr7\Query::parse` produces `['acl' => null]` for subresource parameters, and the existing null-coalescing in `getCanonicalizedQuery` already handles this correctly by emitting `acl=` as the spec requires. No change is needed at this time.

PR #2927, opened 2024-05-24
fix: clear cached path for ECS auth token file

yenfryherrerafeliz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -97,13 +97,24 @@ class EcsCredentialProvider
     private function getEcsAuthToken()
     {
         if (!empty($path = getenv(self::ENV_AUTH_TOKEN_FILE))) {
-            if (is_readable($path)) {
-                return file_get_contents($path);
+            $token =  @file_get_contents($path);
+            if (false === $token) {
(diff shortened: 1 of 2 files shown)


src/Credentials/EcsCredentialProvider.php
```diff
@@ -97,13 +97,24 @@ public function __invoke()
     private function getEcsAuthToken()
     {
         if (!empty($path = getenv(self::ENV_AUTH_TOKEN_FILE))) {
-            if (is_readable($path)) {
-                return file_get_contents($path);
+            $token =  @file_get_contents($path);
+            if (false === $token) {
+                clearstatcache(true, dirname($path) . "/" . @readlink($path));
```
stobrien89, line 102: instead of hard-coding a forward slash, could you replace it with `\DIRECTORY_SEPARATOR`?

stobrien89 (changes requested):

src/Credentials/EcsCredentialProvider.php
yenfryherrerafeliz: Sure. I think it is safer in that way. 

Thanks!

stobrien89 (approved):

Issue #2794, opened 2023-10-12
Issues with profile-based credentials


BVMiko:
### Describe the bug

When using a specific profile (configured the `AWS_PROFILE` environment variable) and using `role_arn` for the provided profile, the CredentialProvider doesn't search the `~/.aws/credentials` file for the `source_profile`'s credentials. This is different from how it works for both AWS CLI and Python (boto3).

### Expected Behavior

I expect that using the `AWS_PROFILE=foobar` environment variable will properly use STS by pulling the source credentials from `~/.aws/credentials`.

### Current Behavior

When processing the `role_arn`, it only searches for the credentials in the `~/.aws/config` file, instead of the more appropriate `~/.aws/credentials` file.

### Reproduction Steps

With the `~/.aws/config`:
```
[default]
region = us-east-1

[profile foobar]
source_profile = default
role_arn = arn:aws:iam::123:role/OrganizationAccessRole
region = us-east-1
```

And the `~/.aws/credentials` file:
```
[default]
aws_access_key_id = ASIA...
aws_secret_access_key = ...
aws_session_token = ...
```

Use the `AWS_PROFILE=foobar` environment variable.

### Possible Solution

The code [here](https://github.com/aws/aws-sdk-php/blob/6616677d76e39af28138512740199d38a461859f/src/Credentials/CredentialProvider.php#L688) should first check the `~/.aws/credentials` file with:
```php
CredentialProvider::ini($sourceProfileName, null, $config)
```
and on rejection it should try the `~/.aws/config` file as it is doing currently:
```php
CredentialProvider::ini($sourceProfileName, $filename, $config)
```

### SDK version used

3.283.2

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.2.11; Ubuntu Linux 22.04

yenfryherrerafeliz:
Hi @BVMiko, sorry to hear about your issues. I can confirm the reported behavior so that I will add a needs-review label to it so we can address this further.

Test code:
```php
<?php
require '../vendor/autoload.php';

use Aws\S3\S3Client;
use Aws\Credentials\CredentialProvider;

$client = new S3Client([
    'credentials' => CredentialProvider::ini('profile issue-2794', '/PATH/config'),
    'region' => 'us-east-2'
]);
$response = $client->listBuckets();

print_r($response);

```
/PATH/config
```
[profile issue-2794]
source_profile=default
role_arn=arn:aws:iam::account:role/Admin
[...]

PR #2939, opened 2024-06-07
Support for parsing both config and credentials files

BVMiko:
*Issue #, if available:* #2794

*Description of changes:*
AWS CLI as well as `boto3` implementation handle config & credentials files by reading the contents of both and consolidating them into a single associative array, prior to doing any processing. `aws-sdk-php`, on the other hand, only reads in one file when processing. This causes an issue when using assumed roles while organizing credentials the way recommended in the AWS CLI docs (the "IAM Role" tab [here](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#:r2l:)).

The current `aws-sdk-php` implementation looks like it has set up a workaround to the situation by allowing a specific filename to be passed into several of the functions, so that the developer can choose between the config or credentials file. However this would not work in the aforementioned situation where data is split between both files.

Note that `aws-sdk-php` does actually have a segment of code which combines the files into a single associative array; however there are a few issues:
1. It's only used by `WebIdentity Credentials`; it cannot be used by regular configuration file
2. It doesn't properly merge sections which exist in both files (as is done in `boto3`)
3. It doesn't respect the `AWS_CONFIG_FILE` environment variable which can be used to override the `~/.aws/config` file path. (Note this issue seems to be present in many other locations as well)

This PR has two commits:
1. The first commit fixes the problem in the most minimal way. It replaces the single-ini-file handler `CredentialProvider::loadProfiles()` with a call to the multi-ini-file handler `CredentialProvider::loadDefaultProfiles()`. It also updates the latter to fix the two additional issues mentioned above. There was an additional modification added to handle the legacy filename workaround.
2. The second commit applies the same fix to `TokenProvider` / `SsoTokenProvider`.  Additionally it updates the `CredentialProvider::ini()`, `CredentialProvider::process()` and `CredentialProvider::sso()` functions to support a null value for the filename override.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -102,11 +102,11 @@ class CredentialProvider
             $defaultChain['process_credentials'] = self::process();
             $defaultChain['ini'] = self::ini();
             $defaultChain['process_config'] = self::process(
-                'profile ' . $profileName,
+                $profileName,
                 self::getHomeDir() . '/.aws/config'
             );
             $defaultChain['ini_config'] = self::ini(
-                'profile '. $profileName,
+                $profileName,
                 self::getHomeDir() . '/.aws/config'
             );
         }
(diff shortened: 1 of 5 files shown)


BVMiko:
Hi @yenfryherrerafeliz, Is there any chance of this being reviewed in the near future? It's been a very long time now, and I'm hoping to help fix up this bug if your team is willing.

Should I be pinging someone else, perhaps?

yenfryherrerafeliz:
The proposed fix here can be achieved by setting `AWS_SDK_LOAD_NONDEFAULT_CONFIG` to `true`.

PR #2955, opened 2024-07-17
feat: s3 expires customization

stobrien89:
*Description of changes:*
S3 will soon be changing the `Expires` shape type to `string` due to inconsistencies in handling of timestamp formats.  We will be modifying the S3 model to keep `Expires` type as `timestamp` for backward compatibility.  We will also be adding a new shape to output structures which contain `Expires`: `ExpiresString`.  This will be the unparsed (into timestamp) string format of the `expires` header.

### DIFF
--- a/src/Api/Service.php
+++ b/src/Api/Service.php
@@ -284,7 +284,7 @@ class Service extends AbstractModel
                 $this->definition['operations'][$name],
                 $this->shapeMap
             );
-        } else if ($this->modifiedModel) {
+        } elseif ($this->modifiedModel) {
             $this->operations[$name] = new Operation(
                 $this->definition['operations'][$name],
                 $this->shapeMap
@@ -517,6 +517,7 @@ class Service extends AbstractModel
     public function setDefinition($definition)
     {
         $this->definition = $definition;
+        $this->shapeMap = new ShapeMap($definition['shapes']);
         $this->modifiedModel = true;
     }
 
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -437,6 +437,7 @@ class S3Client extends AwsClient implements S3ClientInterface
             InputValidationMiddleware::wrap($this->getApi(), self::$mandatoryAttributes),
             'input_validation_middleware'
         );
+        $stack->appendSign(ExpiresParsingMiddleware::wrap(), 's3.expires_parsing');
         $stack->appendSign(PutObjectUrlMiddleware::wrap(), 's3.put_object_url');
         $stack->appendSign(PermanentRedirectMiddleware::wrap(), 's3.permanent_redirect');
         $stack->appendInit(Middleware::sourceFile($this->getApi()), 's3.source_file');
@@ -444,8 +445,8 @@ class S3Client extends AwsClient implements S3ClientInterface
         $stack->appendInit($this->getLocationConstraintMiddleware(), 's3.location');
         $stack->appendInit($this->getEncodingTypeMiddleware(), 's3.auto_encode');
         $stack->appendInit($this->getHeadObjectMiddleware(), 's3.head_object');
+        $this->processModel($this->isUseEndpointV2());
         if ($this->isUseEndpointV2()) {
-            $this->processEndpointV2Model();
             $stack->after('builder',
                 's3.check_empty_path_with_query',
                 $this->getEmptyPathWithQuery());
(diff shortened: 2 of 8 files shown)


tests/S3/S3ClientTest.php
```diff
@@ -2497,6 +2498,70 @@ public function testCorrectlyResolvesGlobalEndpointWithoutRegionInConstructor(
...
+                    200,
+                    ['expires' => 'this-is-not-a-timestamp']
+                ));
+            },
+        ]);
+
+        $client->headObject(['Bucket' => 'foo', 'Key' => 'bar']);
+    }
+
+    public function testExpiresRemainsTimestamp() {
```
yenfryherrerafeliz, line 2539: Should this test somehow alter the model definition beforehand and set the type for expire to string, and then test it guarantee backward compatibility by forcing it to be timestamp?.

yenfryherrerafeliz (changes requested):

src/Api/Service.php
```diff
@@ -517,6 +517,7 @@ public function getDefinition()
     public function setDefinition($definition)
     {
         $this->definition = $definition;
+        $this->shapeMap = new ShapeMap($definition['shapes']);
```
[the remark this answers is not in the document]
stobrien89: Correct- the original `shapeMap` is constructed with the original definition, so when the definition is updated, the shapeMap should also be updated if there have been modifications made to any shapes in the definition.  Because we don't currently have a use case for updating the `shapeMap` independent of model definition changes (and customers don't either), I don't think we should add a separate method. This can always be added later if it turns out we need it.

tests/Api/ServiceTest.php
```diff
@@ -269,14 +269,40 @@ public function testModifyModel()
                     'signingName'     => 'qux',
                     'protocol'        => 'yak',
                     'uid'             => 'foo-2016-12-09'
+                ],
```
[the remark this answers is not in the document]
stobrien89: I see what you're saying, but this test should be here, albeit maybe with different test keys and values. This test ensures the shapemap is updated.  I was just lazy and used `Expires` and `timestamp`/`string`, but can change those if need be.

tests/S3/S3ClientTest.php
stobrien89: I'll see if I can re-work this by manually providing an altered api definition

stobrien89: Updated

src/Api/Service.php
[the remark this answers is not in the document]
yenfryherrerafeliz: Understood. Is just that the name of the method does not refer to the only behavior contained within. But, I guess that since is a MUST that shapeMap gets updated when definition it is also updated then, I think it should be ok.

src/S3/S3Client.php
```diff
@@ -766,25 +767,45 @@ public static function _default_s3_express_identity_provider(array $args)
      * Modifies API definition to remove `Bucket` from request URIs.
```
yenfryherrerafeliz, line 767: Could you please update the documentation here to include that it also enforces the field `Expires` to be `timestamp` and that adds the synthetic `ExpiresString`  field.

tests/Api/ServiceTest.php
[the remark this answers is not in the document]
yenfryherrerafeliz: Understood. Not need to change it then.

Thanks!

src/S3/S3Client.php
```diff
@@ -1035,6 +1056,29 @@ public static function applyDocFilters(array $api, array $docs)
             'shapes' => ['PutObjectRequest', 'UploadPartRequest']
         ];
 
+        // Add `ExpiresString` shape to output structures which contain `Expires`
```
yenfryherrerafeliz, line 1059: Something we could consider, maybe later, is to split the implementation of this function in sub-functions for each docs injected here. For example, since you are adding a docs filter for Expires and ExpiresString you can create a specific method for that and just make the call from this general method. By doing this we can keep the code cleaner and more easy to maintain.

yenfryherrerafeliz (changes requested):

src/S3/S3Client.php
stobrien89: done

src/S3/S3Client.php
stobrien89: You make a good point.  If we get the time at some point in the future, I'd like to move this logic off of service clients entirely— it doesn't belong here.  Which is why, for now, I won't add sub-methods for this.  I think one method dedicated to docs customizations is enough of an eyesore for now 😜

yenfryherrerafeliz (approved):

PR #2960, opened 2024-07-20
Revert bad changes to Guzzle version constraints

GrahamCampbell:
The [earlier changes](https://github.com/aws/aws-sdk-php/commit/d550c5e7f704f155fca85e1be52898c1b1407c7f) were silently included without an explanation and have important consequences for everyone else who uses Guzzle for reasons unrelated to this SDK. This may result in people simply not upgrading to the newer version of the SDK, or worse, cause downgrades to Guzzle. Only Guzzle 7.9.x is currently supported for security fixes. None of the versions of Guzzle you currently allow are supported. It's important that people can use Guzzle 7.9.x.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,7 @@ jobs:
     - if:  ${{ matrix.composer-options == '' }}
       name: Package generation
       run: |
-        composer require guzzlehttp/psr7 "^1.9.1" -W
+        composer config platform.php 7.2.5
         composer update
         make package
 
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,8 @@
     },
     "require": {
         "php": ">=7.2.5",
-        "guzzlehttp/guzzle": "^6.5.8 || ^7.4.5 <7.9.0",
-        "guzzlehttp/psr7": "^1.9.1 || ^2.4.5 <2.7.0",
+        "guzzlehttp/guzzle": "^6.5.8 || ^7.4.5",
+        "guzzlehttp/psr7": "^1.9.1 || ^2.4.5",
         "guzzlehttp/promises": "^1.4.0 || ^2.0",
         "mtdowling/jmespath.php": "^2.6",
         "ext-pcre": "*",


stobrien89:
@GrahamCampbell,

Thanks for the attention to this.  The 7.9.0 and 2.6.0 releases broke our CI, which coincidentally happened on the only day we had to release last week (for Amazon internal reasons).  We added the constraints as a precautionary measure to prevent breaking users. We plan to remove the constraints once we're able to pinpoint the cause of the failures.

GrahamCampbell:
OK, but your change does not necessarily have the desired effect. People can just install the version one before you added the constraint. I for sure don;t see anything in the psr7 release that could have broken anything for you.

stobrien89:
I don't think it's necessarily the new Guzzle versions themselves—It has something to do with how other packages are resolving alongside the new versions and we had other priorities get in the way of figuring it out last week.  In any case, this is top of mind and we'll work on getting it resolved this week.

GrahamCampbell:
Pushed another change here. There seems to have been a misunderstanding from whoever created the packaging script. In order to ensure PHP 7.2.5 compatibility, the only correct way is to tell composer to target PHP 7.2.5. I suspect the incorrect script that was in there just so happened to work from a time when it was needed to package for PHP 5, and you were lucky there were no other incompatible peer dependencies loaded.

GrahamCampbell:
I'm not too sure how the actual production phar files get generated, but it will be necessary to do the same composer platform config change when generating the phar, there.

GrahamCampbell:
The test failures seem to be caused by bugs in `aws/aws-crt-php`. 1.0.4 passes the tests, but the 1.2.6 fails. This does not seem to be related to Guzzle.

indy2kro:
Any idea when this could be addressed? It does block upgrades for aws-sdk-php library when guzzle  7.9 is used at the moment.

ldebrouwer:
> Any idea when this could be addressed? It does block upgrades for aws-sdk-php library when guzzle 7.9 is used at the moment.

I can confirm, I'm currently unable to upgrade the SDK in several projects because of this as well.

stobrien89 (approved):

stobrien89:
Sorry for the delay on this. we were shorthanded last week and dealing with some service issues, but looks like we're good to go with these changes.  Thanks @GrahamCampbell for the help and apologies to those who were blocked on SDK updates.

PR #2964, opened 2024-07-22
Remove use of http:// URLs where possible, and update some old links

Synchro:
I was running some static analysis on a project that uses this SDK, and one thing that always gets flagged is `http://` URLs, mostly in documentation. Sometimes these are fine (for example for `example.com` demos, local tests,, or in XML namespaces), but most of the rest don't really have any excuse, and it's just unnecessary noise, so I thought I'd have a go at fixing them!

Anyway, I updated a whole bunch of these. Some changes might not be appropriate, but obvious ones like entries in CHANGELOG, README, composer.json, are worth taking on. I also updated a few documentation links where they had changed (e.g. sensiolabs changing to symfony).


stobrien89:
Hi @Synchro,

Thanks for the contribution!  We would accept this, but without the modifications to `src/data`.  Those files have upstream (service) sources and changes to those files originating from the SDK will be overwritten the next time the models are updated.

stobrien89 (changes requested):
Needing any changes to `src/data` to be reverted.

stobrien89:
Hi @Synchro,

Closing this for now, but let us know if you have a chance to get back around to it!

Synchro:
@stobrien89 I reverted the changes to everything in `src/data` and force-pushed, so this should be clean now.

PR #2968, opened 2024-07-26
fix: s3 200 errors handling

yenfryherrerafeliz:
- This change includes the following: -- Refactor the s3 200 response error handling logic so it can be opened for all s3 operations. -- Migrate some S3 parsers to a result mutator implementation so the code can be a bit cleaner.

*Issue #, if available:*

*Description of changes:*


src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,239 @@
...
+     * Adds a mutator into the list of mutators.
+     *
+     * @param $mutatorName
+     * @param S3ResultMutator $s3ResultMutator
+     * @return void
+     */
+    public function addS3ResultMutator($mutatorName, S3ResultMutator $s3ResultMutator)
+    {
+        if (isset($this->s3ResultMutators[$mutatorName])) {
+            trigger_error("The S3 Result Mutator $mutatorName already exists!", E_USER_WARNING);
```
stobrien89, line 200: Variable should be encapsulated in braces

src/S3/Parser/GetBucketLocationResultMutator.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Aws\S3\Parser;
+
+use Aws\CommandInterface;
+use Aws\ResultInterface;
+use Psr\Http\Message\ResponseInterface;
+
+class GetBucketLocationResultMutator implements S3ResultMutator
+{
+
```
stobrien89, line 11: extra newline here

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,239 @@
...
+use Aws\Api\StructureShape;
+use Aws\CommandInterface;
+use Aws\Exception\AwsException;
+use Aws\ResultInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
+
+class S3Parser extends AbstractParser
+{
+    /**
```
stobrien89, line 18: for properties, I would just do inline docblocks like `/** @var XmlErrorParser */`

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,239 @@
...
+        $this->s3ResultMutators[$mutatorName] = $s3ResultMutator;
+    }
+
+    /**
+     * Removes a mutator from the mutator list.
+     *
+     * @param $mutatorName
+     * @return void
+     */
+    public function removeS3ResultMutator($mutatorName)
```
stobrien89, line 214: missing a void return type

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Adds a mutator into the list of mutators.
+     *
+     * @param $mutatorName
+     * @param S3ResultMutator $s3ResultMutator
+     * @return void
+     */
+    public function addS3ResultMutator($mutatorName, S3ResultMutator $s3ResultMutator)
```
stobrien89, line 197: missing a void return type here too

src/S3/Parser/S3ResultMutator.php
stobrien89, line 1: Most interface classes in the SDK will have the `Interface` suffix.  i.e. this should be named `S3ResultMutatorInterface`

src/S3/Parser/ValidateResponseChecksumResultMutator.php
```diff
@@ -0,0 +1,130 @@
...
+            "checksum" => $checksumToValidate,
+            "checksumHeaderValue" => $checksumHeaderValue,
+        ];
+    }
+
+    /**
+     * @param $checksumPriority
+     * @param ResponseInterface $response
+     */
+    public function chooseChecksumHeaderToValidate(
```
stobrien89, line 117: does this need to be public? same for the method above

src/S3/Parser/ValidateResponseChecksumResultMutator.php
```diff
@@ -0,0 +1,130 @@
...
+use Aws\S3\Exception\S3Exception;
+use Psr\Http\Message\ResponseInterface;
+
+class ValidateResponseChecksumResultMutator implements S3ResultMutator
+{
+    use CalculatesChecksumTrait;
+    /**
+     * @var Service $api
+     */
+    private $api;
```
stobrien89, line 18: newline needed here

src/S3/Parser/ValidateResponseChecksumResultMutator.php
stobrien89, line 1: should we consider making this class (and other mutators) `final`?

stobrien89, line 1: Also needs a separate test class

src/S3/S3Client.php
```diff
@@ -923,19 +926,15 @@ public static function _applyRetryConfig($value, $args, HandlerList $list)
...
-                ),
-                $args['api']
-            )
+        $s3Parser = new S3Parser(
+            $args['parser'],
+            $args['error_parser'],
+            $args['api'],
+            $args['exception_class']
         );
+        $s3Parser->addS3ResultMutator('s3.get-bucket-location', new GetBucketLocationResultMutator());
```
stobrien89, line 935: we can probably drop the `s3.` prefix since this is not being printed to debug logging.  The mutators are already S3-specific.

src/S3/Parser/S3ResultMutator.php
stobrien89, line 1: Should we extend this pattern to other parsers i.e. write a generic `ResultMutatorInterface`? do we see the weird parser pattern with any other service?

src/S3/Parser/ValidateResponseChecksumResultMutator.php
stobrien89: and `@internal`?

tests/S3/Parser/S3ParserTest.php
stobrien89, line 1: I would look at the `RetryableMalformedResponseParser` and `AmbiguousSuccessParser` test classes and try to duplicate some of the tests (if the functionality is still applicaple).  Since we're not using those classes anymore, we should make sure we're not losing coverage

stobrien89 (changes requested):
Looking good so far

src/S3/Parser/ValidateResponseChecksumResultMutator.php
yenfryherrerafeliz: Yes, I think for now it is more appropriate to make them final and internal.

src/S3/Parser/S3ResultMutator.php
yenfryherrerafeliz: I did a quick look around and I did not find any other place where we could need this. However, we can extend this implementation anytime in the future if needed or required.

tests/S3/Parser/S3ParserTest.php
yenfryherrerafeliz: For `RetryableMalformedResponseParser` I did add a new test case in the S3ClientTest that makes sure not parsable xml errors are retried, which is what the test on RetryableMalformedResponseParser should be for.

src/S3/Parser/ValidateResponseChecksumResultMutator.php
yenfryherrerafeliz: It is also public in the `ValidateResponseChecksumParser` class. It also has some testings depending on this method to be public.

src/S3/S3Client.php
```diff
@@ -947,19 +950,15 @@ public static function _applyRetryConfig($value, $args, HandlerList $list)
...
-                $args['api']
-            )
+        $s3Parser = new S3Parser(
+            $args['parser'],
+            $args['error_parser'],
+            $args['api'],
+            $args['exception_class']
         );
+        $s3Parser->addS3ResultMutator('get-bucket-location', new GetBucketLocationResultMutator());
+        $s3Parser->addS3ResultMutator('validate-response-checksum', new ValidateResponseChecksumResultMutator($args['api']));
```
stobrien89, line 960: I think these exceeds the max recommended line length

tests/S3/Parser/GetBucketLocationResultMutatorTest.php
stobrien89, line 1: We should add another test that ensures this isn't applied to operations other than `GetBucketLocation`

stobrien89 (changes requested):

src/S3/S3Client.php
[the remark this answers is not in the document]
yenfryherrerafeliz: Did not see any unused imports.

src/S3/Parser/GetBucketLocationResultMutator.php
```diff
@@ -0,0 +1,41 @@
...
+        CommandInterface $command,
+        ResponseInterface $response
+    ): ResultInterface
+    {
+        if ($command->getName() !== 'GetBucketLocation') {
+            return $result;
+        }
+
+        $location = 'us-east-1';
+        if (preg_match('/>(.+?)<\/LocationConstraint>/', $response->getBody(), $matches)) {
```
stobrien89, line 32: This also exceeds 80 characters.  You could probably get around that by moving the regex to a const

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,256 @@
...
+
+        return true;
+    }
+
+    /**
+     * Checks if the root element of the response body is "Error", which is
+     * when we should try to parse an error from a 200 response from s3.
+     * It is recommended to make sure the stream given is seekable, otherwise
+     * the rewind call will cause a user warning.
+     * @param StreamInterface $responseBody
```
stobrien89, line 167: nit: newline before the param annotation.

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Removes a mutator from the mutator list.
+     *
+     * @param $mutatorName
+     * @return void
+     */
+    public function removeS3ResultMutator($mutatorName): void
+    {
+        if (!isset($this->s3ResultMutators[$mutatorName])) {
+            trigger_error("The S3 Result Mutator {$mutatorName} does not exist!", E_USER_WARNING);
```
stobrien89, line 234: Line length

src/S3/Parser/S3Parser.php
```diff
@@ -0,0 +1,256 @@
...
+     * Returns the list of result mutators available.
+     *
+     * @return array
+     */
+    public function getS3ResultMutators(): array
+    {
+        return $this->s3ResultMutators;
+    }
+
+    public function parseMemberFromStream(StreamInterface $stream, StructureShape $member, $response)
```
stobrien89, line 252: line length

src/S3/Parser/ValidateResponseChecksumResultMutator.php
```diff
@@ -0,0 +1,154 @@
...
+            : $operation['httpChecksum'];;
+        if (null === $checksumInfo) {
+            return $result;
+        }
+
+        // Skip this middleware if the operation doesn't send back a checksum, or the user doesn't opt in
+        $checksumModeEnabledMember = $checksumInfo['requestValidationModeMember'] ?? "";
+        $checksumModeEnabled = $command[$checksumModeEnabledMember] ?? "";
+        $responseAlgorithms = $checksumInfo['responseAlgorithms'] ?? [];
+        if (empty($responseAlgorithms) || strtolower($checksumModeEnabled) !== "enabled") {
```
stobrien89, line 57: line length

src/S3/Parser/ValidateResponseChecksumResultMutator.php
stobrien89: This hasn't been addressed

tests/S3/Parser/S3ParserTest.php
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+
+    public function testAddsS3ResultMutator()
+    {
+        $testField = 'TestField';
+        $testValue = 'TestValue';
+        $s3MutatorName = 's3.test-mutator';
+        $s3Parser = $this->getS3Parser();
+        $s3Parser->addS3ResultMutator($s3MutatorName, new class($testField, $testValue) implements S3ResultMutator
```
stobrien89, line 179: This could use some line breaks/wraps

tests/S3/Parser/S3ParserTest.php
```diff
@@ -0,0 +1,251 @@
...
+
+        $this->assertTrue(isset($mutators[$s3MutatorName]));
+        $this->assertEquals($testValue, $result[$testField]);
+    }
+
+    public function testRemovesS3ResultMutator()
+    {
+        $s3Parser = $this->getS3Parser();
+        $s3MutatorName = 's3.test-mutator';
+        $s3Parser->addS3ResultMutator($s3MutatorName, new class implements S3ResultMutator
```
stobrien89, line 215: Same line break/wrap issue here

stobrien89 (changes requested):
Mostly coding style (line length) and formatting requests.  My comment from the previous round regarding method visibility on the checksum mutator class

src/S3/Parser/S3Parser.php
stobrien89: it appears this wasn't addressed.  Break to a new line— this is over 80 characters

stobrien89 (approved):
LGTM minus one line length comment (on a method signature) that wasn't addressed.  I would just check this one more time and see if any look to be over 80 and adjust accordingly (wrap, etc.)

PR #2972, opened 2024-07-30
chore: retry CURLE_SSL_CONNECT_ERROR failures

### LINKED ISSUE #1546: Random AWS HTTP error: cURL error 35
Hi. I ma randomly receiving the following error:
`AWS HTTP error: cURL error 35: OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to s3.amazonaws.com:443`

Credentials are ok but since some days this error is happening.It occurs on any S3 operation.

I'm using aws.phar 3.54.3 but it happens too with 3.55.12

annuh:
This PR enables retries of the `CURLE_SSL_CONNECT_ERROR` failure.

Our application randomly encounters these errors:
```
AWS HTTP error: cURL error 35: Send failure: Connection reset by peer (see https://curl.haxx.se/libcurl/c/libcurl-errors.html)
```
After a retry, it works. Since this failure should be safe to retry, I think it's a good idea to make the AWS SDK automatically retry this failure using the existing logic.

This could also fix https://github.com/aws/aws-sdk-php/issues/1546.

### DIFF
--- a/.changes/nextrelease/retry_curl35.json
+++ b/.changes/nextrelease/retry_curl35.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "",
+    "description": "Enables retries of the CURLE_SSL_CONNECT_ERROR (35) failure."
+  }
+]
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -80,6 +80,7 @@ class RetryMiddleware
     ) {
         $retryCurlErrors = [];
         if (extension_loaded('curl')) {
+            $retryCurlErrors[CURLE_SSL_CONNECT_ERROR] = true;
             $retryCurlErrors[CURLE_RECV_ERROR] = true;
         }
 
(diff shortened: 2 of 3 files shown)


annuh:
@stobrien89 this PR is open for quite some time. I understand this isn't an urgent issue, but our application still encounters this problem, from time to time. 
Could you maybe take a look at this PR someday? 🙂 Thanks in advance!

stobrien89:
Hi @annuh,

Really sorry for the extreme delay here.  We've done some work on retries lately and I would be open to having this as an opt-in only: SSL connect should not be retried by default because it typically indicates a permanent configuration or handshake failure rather than a temporary network glitch. 

We are switching to `Retry\V3\RetryMiddleware` as our new default soon- this middleware accepts `$options['curl_errors']` for customizing the retry behavior.  This middleware supports `standard` and `adaptive` modes only, so if you're looking to use this with the legacy `RetryMiddleware`, we would need to add a similar mechanism there.  Closing for now since we have a workaround, but happy to reopen if you want to rework this or have any questions.

Issue #2954, opened 2024-07-16
Missing functions in Aws\S3\StreamWrapper


p8nut:
### Describe the feature

Add the missing functions:
- stream_set_option
- stream_metadata
to match fully implement the streamWrapper wlass

### Use Case

As I Wrap a Stream in one of my project the interpreter complain about those two function missing.

### Proposed Solution

In the php documentation, the two functions have the return value described as  `Returns true on success or false on failure. If option is not implemented, false should be returned.` so we could simply implements the functions like that: 
```php
<?php

    public function stream_set_option($option, $arg1, $arg2)
    {
        return false;
    }

    public function stream_metadata($path, $option, $value)
    {
        return false;
    }
```
see https://www.php.net/manual/en/streamwrapper.stream-set-option.php
and https://www.php.net/manual/en/streamwrapper.stream-metadata.php

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.316.3

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.1.27

p8nut:
dont know if it was planned but it seems possible to implement file lock functions that are missing using the [s3 object lock](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html)

RanVaknin:
Hi @p8nut ,

Thanks for the Feature request and PR. I think this is a valid request. We will have someone review.

All the best,
Ran~

stobrien89:
Thanks again for the contribution @p8nut! Closing this now.

PR #2979, opened 2024-08-06
add missing stream wrapper functions

p8nut:
*Issue #, if available:*
#2954 
*Description of changes:*
this implement the missing functions from the stream wrapper interface.
all of the missing function return false as the doc describe.
if needed it also throw an E_WARNING


stobrien89:
Hi @p8nut,

Thanks for the contribution.  Would you be able to to add some unit tests covering this behavior in `Test\S3\StreamWrapperTest` ?

p8nut:
as asked i try to add some basic test to the code but for some reason i wasn't able to try them on my computer.
i leave them here for now and will try to validate all of them as soon as possible.

stobrien89:
@p8nut,

Thanks for adding the tests.  Just two things: 

1. for the test methods, can you append `ReturnsFalse` to their names? e.g. `testSetStreamOptionReturnsFalse`.
2. The tests are failing because the mock request is not returning a mock result.  You can fix this by adding 

``` php
$this->addMockResults(
  $this->client,
  [
      new Result(),
  ]
);
```
prior to `$stream = ...` in your tests.

stobrien89 (approved):

stobrien89:
Hi @p8nut,

Hope you don't mind, but I updated the methods- only [`stream_lock()` ](https://www.php.net/manual/en/streamwrapper.stream-lock.php) specifies that an error should be emitted.  Test cases have been updated as well.  Thanks for the contribution!

yenfryherrerafeliz (approved):

PR #2986, opened 2024-08-20
fix: error matcher on booleans

yenfryherrerafeliz:
When the expected value from acceptor is a boolean then, the matcher should:
- if expected is true then an error should be matched.
- if expected is false then no error should has been occurred.

*Issue #, if available:*

*Description of changes:*


stobrien89 (changes requested):
Just one minor change requested.  Thanks @yenfryherrerafeliz!

stobrien89:
We should also publish a changelog entry for this one

stobrien89 (approved):

Issue #2952, opened 2024-07-16
Announcing the upcoming deprecation of PHP runtimes 8.0.x and below


stobrien89:
On ~~January 13, 2025~~ January 21, 2025, in order to continue supporting our customers with tools that are secure and maintainable, AWS will publish a minor version bump of the SDK, which will formally end support for PHP runtimes 8.0.x and below. 

The PHP Group has deprecated PHP 8.0 as of November 30th, 2023, which means no new bug reports, fixes, or changes will be made to PHP 8.0. Many widely-used PHP projects and packages have discontinued support for PHP versions below 8.0, and we will be joining these deprecations to maintain a secure and up-to-date SDK.

**I use a PHP language level below 8.0. What should I do?**
For users below 8.0.x, we recommend updating your PHP language version to a language version supported by the SDK, if possible. Versions of the SDK published early-mid August 2024 through January 13th, 2025 will emit a warning on client construction that your PHP version will be deprecated soon. This can be turned off by setting `suppress_php_deprecation_warning` to `true` in your client constructor, or globally by setting the `AWS_SUPPRESS_PHP_DEPRECATION_WARNING` environment variable to `true`.

**What if I can’t upgrade to PHP 8.1 or higher?**
In the event that you’re unable to migrate to PHP 8.1 or higher, previous versions of the AWS SDK for PHP will continue to operate after the deprecation date. Depending on your installation method, you will either stop receiving updates after the deprecation date, or will need to download a release that is compatible with your PHP version.

**I use a PHP language level of 8.1 or higher. What should I do?**
If you already use a version with continued support, no action is required. You are not impacted by this announcement.

**Conclusion**
If you’re currently using PHP versions 8.0.x or below with the SDK, we encourage you to upgrade to PHP 8.1 or above. Please share your questions, comments, or concerns with us in a direct reply to this issue.

### Links

https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-php-runtimes-8-0-x-and-below-in-the-aws-sdk-for-php/

ishiguro-junya:
@stobrien89 
Does this call to trigger_error() throw an exception?
I didn't expect an exception to occur because E_USER_WARNING was specified, but the error message output I committed last week now throws an exception, so I wanted to share it just in case.
https://github.com/aws/aws-sdk-php/pull/2951/files#diff-4e9ab0a7012bd3e3935430f29adc9a5829b975072e311bd9ca1e8662e98f10b3R1258
[...]

PR #2989, opened 2024-08-23
fix: php deprecated log error level

ishiguro-junya:
*Issue #, if available:* #2952

*Description of changes:*
Fix the log level.


stobrien89:
Hi @ishiguro-junya,

Thanks for the contribution!  The decision to use `E_USER_WARNING` over `E_USER_DEPRECATED` was actually due to `E_USER_WARNING` having a smaller blast radius for those using custom error handlers and also the fact that customers using versions in the affected deprecation ranges will still be able to continue using their current SDK versions; they just won't receive updates after we change the required PHP language level.  We're going to keep the warning level at `E_USER_WARNING`, since this is not a deprecation in the sense of "code stops working".

stobrien89:
Disregard my first comment.  It was posted before I read the discussion in #2952.  I'm fine with changing this.

stobrien89 (approved):

yenfryherrerafeliz (approved):

stobrien89:
Thanks again @ishiguro-junya!

PR #2990, opened 2024-08-26
fix: Update AwsException constructor to accept Throwable instead of Exception

zishang520:
### Overview
This PR updates the `AwsException` constructor to accept a `Throwable` instead of an `Exception` as the `$previous` parameter. This change ensures that all types of throwable errors, including `Error` and `Exception`, are properly handled and passed to the `AwsException`.

### Changes
- Updated the `AwsException` constructor to accept `\Throwable` for the `$previous` parameter.
- Ensured compatibility with all existing error handling mechanisms in the AWS SDK.

### Why this change?
Previously, the `AwsException` constructor only accepted `Exception` as the `$previous` parameter, which caused issues when other types of throwables, like `Error`, were passed. By accepting `Throwable`, we ensure a more robust and flexible error handling process.

### Testing
- Tested with various throwable types to confirm that the error handling behaves as expected.
- No breaking changes observed in the current error-handling workflow.

### Impact
This change enhances the SDK's error-handling capabilities without introducing breaking changes, ensuring better compatibility with a wider range of PHP error types.


stobrien89:
Thanks for the contribution, @zishang520!

PR #2999, opened 2024-09-17
enhancement: flexible checksums

stobrien89:
*Description of changes:*
Adds default checksum calculation, deprecates the `AddContentMD5` config option as S3 plans to deprecate MD5 checksums soon, adds two S3 client config options: `request_checksum_calculation` and `response_checksum_validation` with acceptable values for both being either `when_supported` or `when_required`.  

`request_checksum_calculation` will default to `when_supported`, which means if an operation supports checksum calculation (indicated via modeling), the SDK will calculate the checksum and add the appropriate header.  If it is set to `when_required`, the SDK will only compute and send a checksum when the operation requires an `httpRequestChecksum`.

`response_checksum_validation` also defaults to `when_supported`, which means if an operation models response algorithms, an attempt to validate a response checksum will be made.  `when_required` will only validate response checksums when response checksum are modeled AND `ChecksumMode` is set to `enabled` on the command.


src/S3/S3Client.php
```diff
@@ -239,80 +239,105 @@ class S3Client extends AwsClient implements S3ClientInterface
...
+                'response_checksum_validation' => [
+                    'type' => 'config',
+                    'valid' => ['string'],
+                    'doc' => 'Set to true to disable the usage of'
+                        . ' s3 express session authentication. This is enabled by default.',
+                    'fn' => [__CLASS__, '_apply_response_checksum_validation'],
+                    'default' => [__CLASS__, '_default_response_checksum_validation'],
+                ]
+            ]
+            + $args + [
```
yenfryherrerafeliz, line 273: Question: Why not concatenate the `$args` at the end or at the beginning? We could miss this later since it is sort of hidden.

src/S3/S3Client.php
```diff
@@ -499,6 +527,66 @@ public static function _apply_use_arn_region($value, array &$args, HandlerList $
...
+            $args
+        );
+    }
+
+    public static function _apply_request_checksum_calculation(
+        string $value,
+        array &$args
+    ): void
+    {
+        $value = strtolower($value);
```
yenfryherrerafeliz, line 545: What happens if a customer provide an erroneous value for the parameter  `request_checksum_calculation` being resolved here?. Will it just ignored?

src/S3/S3Client.php
```diff
@@ -499,6 +527,66 @@ public static function _apply_use_arn_region($value, array &$args, HandlerList $
...
+        );
+    }
+
+    public static function _apply_response_checksum_validation(
+        $value,
+        array &$args
+    ): void
+    {
+        $value = strtolower($value);
+        if (array_key_exists($value, self::$checksumValidationOptionEnum)) {
```
yenfryherrerafeliz, line 567: Same here regarding what happens if the customer provided a value that will not be found in `self::$checksumValidationOptionEnum`?.

src/Signature/S3SignatureV4.php
```diff
@@ -94,6 +94,16 @@ public function presign(
                 $this->getPresignedPayload($request)
             );
         }
+
+        //Payload is unknown, checksum will cause requests to fail.
+        if ($request->getMethod() === 'PUT') {
```
yenfryherrerafeliz, line 99: Is it guaranteed that every put request will contain an unknown payload?

yenfryherrerafeliz (changes requested):

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -18,14 +18,28 @@
...
+
+    /**
+     * @var true[]
+     *
+     * S3 Operations for which pre-calculated SHA256
+     * Checksums can be added to the command
+     */
+    private static $sha256 = [
+        'PutObject' => true,
+        'UploadPart' => true,
```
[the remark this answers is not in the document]
stobrien89: In this case the need is for a [custom/synthetic shape `Sha256`](https://github.com/aws/aws-sdk-php/blob/master/src/S3/S3Client.php#L1027,L1032) to only be applied to certain operations if provided as command input.  There might be some need in the future for additional synthetic checksum shapes, but for now I'd rather keep this restricted to a key lookup as opposed to a key lookup + array search. Although performance is a non-issue with an array size of 2

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -34,17 +48,22 @@ class ApplyChecksumMiddleware
...
+        return function (callable $handler) use ($api, $config) {
+            return new self($handler, $api, $config);
         };
     }
 
-    public function __construct(callable $nextHandler, Service $api)
+    public function __construct(
+        callable $nextHandler,
+        Service $api, array
+        $config = []
```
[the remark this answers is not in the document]
stobrien89: My bad. not intentional

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -54,59 +73,38 @@ public function __invoke(
         $next = $this->nextHandler;
         $name = $command->getName();
         $body = $request->getBody();
+        $op = $this->api->getOperation($name);
```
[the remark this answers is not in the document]
stobrien89: Agreed. Just kept the old variable name

.changes/nextrelease/checksumsv2.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "S3",
+    "description": "Adds additional config to "
```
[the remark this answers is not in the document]
stobrien89: Another unintentional omission

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -116,10 +114,82 @@ public function __invoke(
...
+    }
+
+    /**
+     * @param Shape|null $checksumMember
+     * @param string|null $requestedAlgorithm
+     * @param string|null $checksumMemberName
+     *
+     * @return string
+     */
+    private function determineChecksumAlgorithm(
```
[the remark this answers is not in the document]
stobrien89: `$requestedAlgorithm` is a user-provided algorithm. Unless I'm not understanding the question

src/S3/CalculatesChecksumTrait.php
```diff
@@ -16,35 +23,36 @@ trait CalculatesChecksumTrait
...
-                case 'sha1':
-                    return base64_encode(Psr7\Utils::hash($value, $requestedAlgorithm, true));
-                default:
-                    break;
-                throw new InvalidArgumentException(
-                    "Invalid checksum requested: {$requestedAlgorithm}."
-                    . "  Valid algorithms are CRC32C, CRC32, SHA256, and SHA1."
-                );
+
+        if (array_key_exists($requestedAlgorithm, self::$supportedAlgorithms)) {
```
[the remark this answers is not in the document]
stobrien89: Agreed that `array_key_exists` isn't appropriate here.  Changing to `isset`

src/S3/S3Client.php
```diff
@@ -239,80 +239,105 @@ class S3Client extends AwsClient implements S3ClientInterface
...
-                    'bool',
-                    Configuration::class,
-                    CacheInterface::class,
-                    'callable'
+        return
+            [
+                'request_checksum_calculation' => [
+                    'type' => 'config',
+                    'valid' => ['string'],
+                    'doc' => 'Set to true to disable the usage of'
```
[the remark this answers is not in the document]
stobrien89: not intentional.  I'll update it

src/S3/S3Client.php
```diff
@@ -239,80 +239,105 @@ class S3Client extends AwsClient implements S3ClientInterface
...
-            ],
-            's3_express_identity_provider' => [
-                'type'    => 'config',
-                'valid'   => [
-                    'bool',
-                    'callable'
+                'response_checksum_validation' => [
+                    'type' => 'config',
+                    'valid' => ['string'],
+                    'doc' => 'Set to true to disable the usage of'
```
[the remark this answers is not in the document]
stobrien89: same as above. will update

src/S3/S3Client.php
stobrien89: One of these two options (`response_checksum_validation`) needs to be prepended to the list because its value must be resolved before [`apply_api_provider` is called](https://github.com/aws/aws-sdk-php/blob/master/src/ClientResolver.php#L122,#128), due to the fact that `ValidateResponseCheckSumMutator` is applied to the api provider.  I suppose we could just do this with `response_checksum_validation` and move `request_checksum_calcuation` into the larger list of s3 specific config. What do you think?

src/S3/ApplyChecksumMiddleware.php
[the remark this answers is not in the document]
yenfryherrerafeliz: For what I understand from the SEP, customers may provide their own checksum which will be include in the request headers as `x-amz-checksum-CHECKSUM_ALGO` and in such case we should also ignore applying our default one. Unless I misunderstood the SEP?

src/Signature/S3SignatureV4.php
stobrien89: For presigned URLs, yes.  We cannot determine the payload because we are only generating a URL with some headers moved to query params.  The payload is not factored into this, so if a body is provided (or empty) during request serialization, you end up with checksum headers converted into query params in the URL.

stobrien89: S3 `GET` requests are not affected by this given the nature of a `GET` request

src/S3/S3Client.php
stobrien89: I should have thrown an exception there and for the other example you pointed out. Will update.

src/S3/ApplyChecksumMiddleware.php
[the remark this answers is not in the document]
stobrien89: So the workflow in this method is to set the default, then override it IF a user provides an algorithm AND that algorithm is valid.  `requestedAlgorithm` is empty if a user does not provide an algorithm

src/S3/CalculatesChecksumTrait.php
```diff
@@ -16,35 +23,36 @@ trait CalculatesChecksumTrait
...
-                case 'sha1':
-                    return base64_encode(Psr7\Utils::hash($value, $requestedAlgorithm, true));
-                default:
-                    break;
-                throw new InvalidArgumentException(
-                    "Invalid checksum requested: {$requestedAlgorithm}."
-                    . "  Valid algorithms are CRC32C, CRC32, SHA256, and SHA1."
-                );
+
+        if (isset($requestedAlgorithm, self::$supportedAlgorithms)) {
```
yenfryherrerafeliz, line 27: Do you want to check if `$requestedAlgorithm` is set in `self::$supportedAlgorithms`, or you want to check if those two variables are set?.

yenfryherrerafeliz: If is the first case then maybe you made an unintentional mistake. It should be `isset(self::$supportedAlgorithms[$requestedAlgorithm]);`

src/S3/CalculatesChecksumTrait.php
```diff
@@ -16,35 +23,36 @@ trait CalculatesChecksumTrait
...
+            if ($requestedAlgorithm === "crc32") {
                 $requestedAlgorithm = "crc32b";
             }
             return base64_encode(Psr7\Utils::hash($value, $requestedAlgorithm, true));
         }
-    }
 
+        throw new InvalidArgumentException(
+            "Invalid checksum requested: {$requestedAlgorithm}."
+            . "  Valid algorithms are CRC32C, CRC32, SHA256, and SHA1."
```
yenfryherrerafeliz, line 55: Could you please take the valid algorithms from `self::$supportedAlgorithms` instead?. By doing this we do not have to maintain multiple places for supported algorithms when a new one is added.

src/S3/MultipartUploadingTrait.php
```diff
@@ -54,10 +54,17 @@ protected function handleResult(CommandInterface $command, ResultInterface $resu
...
         $partData['PartNumber'] = $command['PartNumber'];
         $partData['ETag'] = $this->extractETag($result);
+
+        $checksumResult = $this instanceof MultipartCopy
+            ? $result['CopyPartResult']
+            : $result;
         if (isset($command['ChecksumAlgorithm'])) {
             $checksumMemberName = 'Checksum' . strtoupper($command['ChecksumAlgorithm']);
-            $partData[$checksumMemberName] = $result[$checksumMemberName];
+            $partData[$checksumMemberName] = $checksumResult[$checksumMemberName];
```
yenfryherrerafeliz, line 63: Is it guaranteed that `$checksumResult` will contain a member `$checksumMemberName`?

src/S3/S3Client.php
```diff
@@ -239,80 +239,107 @@ class S3Client extends AwsClient implements S3ClientInterface
...
-                    'bool',
-                    Configuration::class,
-                    CacheInterface::class,
-                    'callable'
+        return
+            [
+                'request_checksum_calculation' => [
+                    'type' => 'config',
+                    'valid' => ['string'],
+                    'doc' => 'Set to true to disable the usage of'
```
yenfryherrerafeliz, line 259: This parameter's doc still needs to be updated.

yenfryherrerafeliz (changes requested):

src/S3/CalculatesChecksumTrait.php
stobrien89: Whoops.  Rushed through those corrections

src/S3/MultipartUploadingTrait.php
stobrien89: Yes- if a command checksum is provided, we can assume it was valid if we're handling the result.  S3 will send the corresponding encoded value in the `$checksumMemberName` result key.

src/S3/S3Client.php
stobrien89: another silly mistake

src/S3/ApplyChecksumMiddleware.php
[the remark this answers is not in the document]
stobrien89: Updated to verify user-provided checksum headers

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -116,32 +121,114 @@ public function __invoke(
...
-    private function isS3Express(CommandInterface $command): bool
+    private function hasAlgorithmHeader(
+        RequestInterface $request,
+        array $supportedAlgorithms
+    ): bool
     {
-        return isset($command['@context']['signing_service'])
-            && $command['@context']['signing_service'] === 's3express';
+        foreach ($supportedAlgorithms as $supportedAlgorithm) {
+            if ($request->hasHeader('x-amz-checksum-' . $supportedAlgorithm)) {
```
yenfryherrerafeliz, line 227: It should not be validated against our supported algorithms. I know it is odd, but according to the SEP:
> If the user set a checksum header field (i.e. any header prefixed with x-amz-checksum-), then no calculation is necessary. Go to END.

Maybe we should investigate to see if this is the intended behavior here.

tests/S3/Parser/ValidateResponseChecksumResultMutatorTest.php
```diff
@@ -25,25 +25,40 @@ class ValidateResponseChecksumResultMutatorTest extends TestCase
...
     ) {
+        if (!empty($responseAlgorithms)
+            && $responseAlgorithms[0] === 'CRC32C'
+            && !extension_loaded('awscrt')
+        ) {
+            $this->markTestSkipped();
+        }
+
         $s3Client = $this->getTestS3ClientWithResponseAlgorithms(
             'GetObject',
```
yenfryherrerafeliz, line 50: I think you also need to pass the operation here. Instead of `GetObject` it should be `$operation`.

yenfryherrerafeliz (changes requested):
Overall looks good. Just some minor changes. I also did not see a test for when generating a presigned url with PUT operations and the checksum here should be omitted?

src/S3/ApplyChecksumMiddleware.php
stobrien89: I figured this would be the safest way to interpret it since something completely incorrect like `x-amz-checksum-algorithm-foo` would also meet that criteria.  I'll get some confirmation on that

stobrien89: Looks like we shouldn't validate against anything.  Updated.

tests/S3/Parser/ValidateResponseChecksumResultMutatorTest.php
stobrien89: I think I intended to remove the `$operation` param at some point because we were only using `GetObject`, but I'll leave that there in case we want to test others in the future.

yenfryherrerafeliz (approved):

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -54,59 +73,43 @@ public function __invoke(
...
+            $mode,
+            $checksumRequired,
+            $checksumMember,
+            $requestedAlgorithm
+        );
+        if ($shouldAddChecksum) {
+            if (!$this->hasAlgorithmHeader($request)) {
+                $supportedAlgorithms = isset($checksumMember['enum'])
+                    ? array_map('strtolower', $checksumMember['enum'])
+                    : [];
```
yenfryherrerafeliz, line 101: nit: I think this could be simplified to:
```php
$supportedAlgorithms =  array_map('strtolower', $checksumMember['enum'] ?? []);
```

src/S3/CalculatesChecksumTrait.php
```diff
@@ -16,35 +23,37 @@ trait CalculatesChecksumTrait
...
             }
             return base64_encode(Psr7\Utils::hash($value, $requestedAlgorithm, true));
         }
-    }
 
+        $validAlgorithms = implode(', ', array_keys(self::$supportedAlgorithms));
+        throw new InvalidArgumentException(
+            "Invalid checksum requested: {$requestedAlgorithm}."
+            . "  Valid algorithms are {$validAlgorithms}."
+        );
```
yenfryherrerafeliz, line 57: There is another place that validates if the checksum is valid or not. For example [here](https://github.com/aws/aws-sdk-php/pull/2999/files#diff-d65167df11c3be45fbc4360650d5c53535d3408e34ae1e2a33fe12089c5498c1R178-R179). I know one validates if is valid based on what the operation supports and the other one validates if is supported by the SDK in specific. Would not be worth to make that more clear here?, like state why it is not supported.
For example:
```php
  throw new InvalidArgumentException(
      "Invalid checksum requested: {$requestedAlgorithm}."
      . "  Valid algorithms supported by this runtime are {$validAlgorithms}."
  );
```

yenfryherrerafeliz (commented):
It looks great. Just two minor suggestion that are not required to address.

PR #3001, opened 2024-09-20
enhancement: user agent 2.1

yenfryherrerafeliz:
This change provides:
- A builder class for appending metrics
- Default initialization of a metrics builder within command instantiation
- Wraps user agent logic into a single middleware class
- Adds middlewares into the different features from where metrics can be gather, to acomplish this purpose.

*Issue #, if available:*

*Description of changes:*


src/AwsClient.php
```diff
@@ -490,6 +497,11 @@ private function addSignatureMiddleware(array $args)
                 $region = $signingRegionSet
                     ?? $commandSigningRegionSet
                     ?? $region;
+
+                MetricsBuilder::appendMetricsCaptureMiddleware(
```
stobrien89, line 501: question: This function is invoked during the `sign` step.  How is the signature version being passed to the metrics builder by appending middleware to the build step, which has already completed?

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -394,4 +402,23 @@ private function resolveAccountId(): ?string
 
         return $identity->getAccountId();
     }
+
+    private function hookAccountIdMetric($accountId, &$command)
```
stobrien89, line 406: question: can we make a generic method on the MetricsBuilder class that will handle this and other things like it i.e. accountidendpointmetric?

src/MetricsBuilder.php
```diff
@@ -0,0 +1,166 @@
...
+    const ACCOUNT_ID_MODE_DISABLED = "Q";
+    const ACCOUNT_ID_MODE_REQUIRED = "R";
+    const SIGV4A_SIGNING = "S";
+    const RESOLVED_ACCOUNT_ID = "T";
+    const FLEXIBLE_CHECKSUMS_REQ_CRC32 = "U";
+    const FLEXIBLE_CHECKSUMS_REQ_CRC32C = "V";
+    const FLEXIBLE_CHECKSUMS_REQ_CRC64 = "W";
+    const FLEXIBLE_CHECKSUMS_REQ_SHA1 = "X";
+    const FLEXIBLE_CHECKSUMS_REQ_SHA256 = "Y";
+    /** @var int */
```
stobrien89, line 32: nit: newline needed here

src/MetricsBuilder.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @return string
+     */
+    private function encode(): string
+    {
+        return implode(self::$METRIC_SEPARATOR, array_keys($this->metrics));
+    }
+
+    /**
+     * Appends a metric into the internal metrics holder.
```
stobrien89, line 76: nit: can this and other longer docblocks be shortened?

src/MetricsBuilder.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param string $newMetric
+     *
+     * @return bool
+     */
+    private function canMetricBeAppended(string $newMetric): bool
+    {
+        if ($this->metricsSize
+            + (strlen($newMetric) + strlen(self::$METRIC_SEPARATOR))
+            > self::$MAX_METRICS_SIZE) {
```
stobrien89, line 116: nit: closing parentheses and brace should go to a new line

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,326 @@
...
+namespace Aws;
+
+use Aws\EndpointDiscovery\Configuration;
+use Closure;
+use Psr\Http\Message\RequestInterface;
+
+class UserAgentMiddleware
+{
+    const AGENT_VERSION = 2.1;
+    /** @var callable  */
```
stobrien89, line 12: nit: newline needed here

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    /**
+     * Builds the different user agent values.
+     *
+     * @return array
+     */
+    private function buildUserAgentValue(): array
+    {
+        static $fnList = [
```
stobrien89, line 102: suggestion: consider moving to a static property if it could possibly be used by anything else

src/AwsClient.php
yenfryherrerafeliz: Good catch. Metrics here will not be collected. I will do a refactor for this type of cases.

src/EndpointV2/EndpointV2Middleware.php
yenfryherrerafeliz: Yes, I can take a look and see if it fits better there.

src/UserAgentMiddleware.php
stobrien89: Comment still applies

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,333 @@
...
+        } elseif ($accountIdMode === 'required') {
+            $this->metricsBuilder->append(MetricsBuilder::ACCOUNT_ID_MODE_REQUIRED);
+        }
+    }
+
+    /**
+     * Appends the retry mode metric into the metrics builder,
+     * based on the resolved retry config mode.
+     */
+    private function retryConfigMetric(): void
```
stobrien89, line 303: Another case where I think we could use a static mapping

stobrien89 (changes requested):
Looking good, just had a few more comments

src/MetricsBuilder.php
```diff
@@ -0,0 +1,197 @@
...
+     *
+     * @return bool
+     */
+    private function canMetricBeAppended(string $newMetric): bool
+    {
+        if ($this->metricsSize
+            + (strlen($newMetric) + strlen(self::$METRIC_SEPARATOR))
+            > self::$MAX_METRICS_SIZE
+        ) {
+            @trigger_error(
```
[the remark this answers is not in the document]
yenfryherrerafeliz: I have included some test cases for when not more metrics can be appended and if I do not suppress the alert it make my test to fail.

src/S3/ApplyChecksumMiddleware.php
```diff
@@ -144,4 +150,34 @@ private function isS3Express(CommandInterface $command): bool
         return isset($command['@context']['signing_service'])
             && $command['@context']['signing_service'] === 's3express';
     }
+
+    private function hookChecksumAlgorithmMetric($algorithm, $command)
```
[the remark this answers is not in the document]
yenfryherrerafeliz: This method is not needed. I have used another approach here.

src/CommandInterface.php
```diff
@@ -39,4 +39,11 @@ public function hasParam($name);
      * @return HandlerList
      */
     public function getHandlerList();
+
+    /**
+     * Returns the metrics builder instance tied up to this command.
+     *
+     * @return MetricsBuilder
+     */
+    public function getMetricsBuilder();
```
yenfryherrerafeliz, line 48: Would this be a no backward compatible change?

src/Credentials/CredentialSources.php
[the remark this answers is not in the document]
stobrien89: and make it final

src/EndpointV2/EndpointV2Middleware.php
```diff
@@ -98,8 +99,14 @@ public function __invoke(CommandInterface $command)
...
         $commandArgs = $command->toArray();
         $providerArgs = $this->resolveArgs($commandArgs, $operation);
+        if (!empty($providerArgs[self::ACCOUNT_ID_PARAM])) {
+            $command->getMetricsBuilder()->append(MetricsBuilder::RESOLVED_ACCOUNT_ID);
+        }
         $endpoint = $this->endpointProvider->resolveEndpoint($providerArgs);
-
+        $command->getMetricsBuilder()->identifyMetricByValueAndAppend(
+            'account_id_endpoint',
+            $endpoint->getUrl()
```
stobrien89, line 108: question: does the spec really call for the url?

src/UserAgentMiddleware.php
stobrien89: maybe I'm missing something, but I think something like 

```php
$this->metricsBuilder->append(
    self::$staticMap[$retrymode]
);
```
could be done in this method without the need for additional functions

tests/AwsClientTest.php
```diff
@@ -982,4 +985,19 @@ public function testClientParameterOverridesDefaultAccountIdEndpointModeBuiltIns
 
         self::assertEquals($expectedAccountIdEndpointMode, $builtIns['AWS::Auth::AccountIdEndpointMode']);
     }
+
+    public function testAppendsUserAgentMiddleware()
+    {
+        $client = new S3Client([
+            'region' => 'us-east-2',
+            'http_handler' => function (RequestInterface $request) {
+                $userAgentValue = $request->getHeaderLine('User-Agent');
```
stobrien89, line 994: nit: I would add an extra assertion that confirms the header exists before asserting its value

tests/MetricsBuilderTest.php
stobrien89, line 1: we should have some sort of end-to-end test with a client and set of features and make assertions against the output string

src/MetricsBuilder.php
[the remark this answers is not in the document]
yenfryherrerafeliz: I think it will be a good safeguard to know if we are tracking the same feature in two places, which may be an indication of an issue.

src/ClientResolver.php
```diff
@@ -980,66 +980,8 @@ public static function _default_app_id(array $args)
 
     public static function _apply_user_agent($inputUserAgent, array &$args, HandlerList $list)
```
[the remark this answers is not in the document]
yenfryherrerafeliz: This behavior is not deprecated as far as I know.

src/Credentials/CredentialSources.php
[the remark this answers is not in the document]
yenfryherrerafeliz: I agree with making it final. Not sure about moving it into a sub-directory. I think the level it is is fine, because this is intended to be like a enum with the different credential sources.

src/MetricsBuilder.php
```diff
@@ -0,0 +1,335 @@
...
+
+    /**
+     * Appends the credentials metric based on the type of credentials
+     * resolved.
+     *
+     * @param CredentialsInterface $credentials
+     *
+     * @return void
+     */
+    private function appendCredentialsMetric(
```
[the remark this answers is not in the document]
yenfryherrerafeliz: Yes, this suggestion is great. My approach is a little verbose.

src/MetricsBuilder.php
```diff
@@ -0,0 +1,335 @@
...
+            @trigger_error(
+                "The metric `{$newMetric}` "
+                . "can not be added due to size constraints",
+                E_USER_WARNING
+            );
+
+            return false;
+        }
+
+        if (isset($this->metrics[$newMetric])) {
```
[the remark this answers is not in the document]
yenfryherrerafeliz: I will remove the errors for now. I just wanted to have a way to catch if we miss somehow something around gathering or duplicating a metric.

src/EndpointV2/EndpointV2Middleware.php
yenfryherrerafeliz: Actually this feature was removed from the spec. I will remove it.

tests/AwsClientTest.php
yenfryherrerafeliz: This is already handled by the method `getHeaderLine` itself. If the header is not present then it will return an empty string.

tests/MetricsBuilderTest.php
yenfryherrerafeliz: I think that that already exists in the `UserAgentMiddlewareTest` implementation. Please let me know if the tests there is not what you refer.

src/AwsClient.php
```diff
@@ -280,6 +280,11 @@ public function __construct(array $args)
         if (isset($args['with_resolved'])) {
             $args['with_resolved']($config);
         }
+        MetricsBuilder::appendMetricsCaptureMiddleware(
```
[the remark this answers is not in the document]
yenfryherrerafeliz: Not sure if I understand your question. Can you please explain?

src/MetricsBuilder.php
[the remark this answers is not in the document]
yenfryherrerafeliz: I removed the error emitting for now.

src/CommandInterface.php
yenfryherrerafeliz: Removed.

tests/MetricsBuilderTest.php
stobrien89: Missed that, but see it now.

src/AwsClient.php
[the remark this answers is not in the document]
stobrien89: disregard, it's no longer relevant

src/MetricsBuilder.php
stobrien89: Basically I was saying each var/docblock should be separated by a newline

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * Builds and injects the user agent header values.
+ * This middleware must be appended into step where all the
+ * metrics to be gathered are already resolved. As of now it should be
+ * after the signing step.
+ */
+class UserAgentMiddleware
+{
+    const AGENT_VERSION = 2.1;
+    static $userAgentFnList = [
```
stobrien89, line 18: newline needed here

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,339 @@
...
+        'getUserAgentVersion',
+        'getHhvmVersion',
+        'getOsName',
+        'getLangVersion',
+        'getExecEnv',
+        'getEndpointDiscovery',
+        'getAppId',
+        'getMetrics'
+    ];
+    static $metricsFnList = [
```
stobrien89, line 29: newline needed here

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,339 @@
...
+        'appendEndpointMetric',
+        'appendAccountIdModeMetric',
+        'appendRetryConfigMetric',
+    ];
+
+    /** @var callable  */
```
stobrien89, line 35: newline between each var/docblock

src/UserAgentMiddleware.php
```diff
@@ -0,0 +1,339 @@
...
+     */
+    public function __construct(callable $nextHandler, array $args=[])
+    {
+        $this->nextHandler = $nextHandler;
+        $this->args = $args;
+    }
+
+    /**
+     * When invoked, its injects the user agent header into the
+     * request headers.
```
stobrien89, line 70: newline here

stobrien89 (approved):
Approved, pending some nits about newlines

PR #3002, opened 2024-09-24
bugfix: LocationConstraint middleware

stobrien89:
**Description of changes:**

Updates middleware to skip directory buckets (`LocationRestraint` is not supported) and fixes a long-running bug where the middleware would override previously configured `CreateBucketConfiguration`. See https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html#AmazonS3-CreateBucket-request-LocationConstraint

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -234,6 +234,8 @@ use Psr\Http\Message\RequestInterface;
  */
 class S3Client extends AwsClient implements S3ClientInterface
 {
+    private const DIRECTORY_BUCKET_REGEX = '/^[a-zA-Z0-9_-]+--[a-z0-9]+-az\d+--x-s3'
+                                            .'(?!.*(?:-s3alias|--ol-s3|\.mrap))$/';
     use S3ClientTrait;
 
     /** @var array */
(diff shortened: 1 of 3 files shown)


tests/S3/S3ClientTest.php
```diff
@@ -524,6 +531,10 @@ public function testAddsLocationConstraintAutomatically($region, $target, $comma
         } else {
             $this->assertStringNotContainsString($text, $body);
         }
+        //ensures original configuration is not overwritten
+        if ($command === 'CreateBucket') {
```
yenfryherrerafeliz, line 535: At this point $command is not a string anymore, instead it is an instance of Command. The evaluation here could be :
```php 
if ($command->getName() === 'CreateBucket')
```
Or use another naming strategy to keep the name of the command available till this statement.

yenfryherrerafeliz: Also, the assertion here could be more strict and search for the specific node/element. Example:
```php
$this->assertStringContainsString('<Bucket><Type>foo</Type></Bucket>', $body);
```

yenfryherrerafeliz (changes requested):

tests/S3/S3ClientTest.php
stobrien89: First one is a typo, but sure, I'll add the full xml string

src/S3/S3Client.php
```diff
@@ -234,6 +234,8 @@
  */
 class S3Client extends AwsClient implements S3ClientInterface
 {
+    public const DIRECTORY_BUCKET_REGEX = '/^[a-zA-Z0-9_-]+--[a-z0-9]+-az\d+--x-s3'
```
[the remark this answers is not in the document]
stobrien89: I didn't really think about it.  I'll change it to private since the method using it is public

yenfryherrerafeliz (approved):

PR #3006, opened 2024-09-29
PHP 8.4 support (Implicitly marking parameter $request as nullable is deprecated, the explicit nullable type must be used instead)

puzzledmonkey:
This PR adds support for PHP8.4, which has changed how we need to define nullable function parameters - they must be marked explicitly now (i.e. `Service|null $service = null` instead of just `Service $service = null`)

I've tried to minimise the code changes and stick with the code style (i.e. `Service|null` instead of `?Service`) of the original files but please let me know if there is anything you would like to to fix/revisit/explain more.

### DIFF
--- a/build/Burgomaster.php
+++ b/build/Burgomaster.php
@@ -164,7 +164,7 @@ class Burgomaster
         $sourceDir,
         $destDir,
         $extensions = array('php', 'php.gz'),
-        Iterator|null $files = null
+        ?Iterator $files = null
     ) {
         if (!realpath($sourceDir)) {
             throw new \InvalidArgumentException("$sourceDir not found");
--- a/build/Burgomaster.php
+++ b/build/Burgomaster.php
@@ -150,21 +150,21 @@ class Burgomaster
      *
      * Any LICENSE file is automatically copied.
      *
-     * @param string    $sourceDir  Source directory to copy from
-     * @param string    $destDir    Directory to copy the files to that is relative
-     *                              to the the stage base directory.
-     * @param array     $extensions File extensions to copy from the $sourceDir.
-     *                              Defaults to "php" files only (e.g., ['php']).
-     * @param Iterator|null  $files Files to copy from the source directory, each
-     *                              yielded out as an SplFileInfo object.
-     *                              Defaults to a recursive iterator of $sourceDir
+     * @param string        $sourceDir  Source directory to copy from
+     * @param string        $destDir    Directory to copy the files to that is relative
+     *                                  to the the stage base directory.
+     * @param array         $extensions File extensions to copy from the $sourceDir.
+     *                                  Defaults to "php" files only (e.g., ['php']).
+     * @param Iterator|null $files      Files to copy from the source directory, each
+     *                                  yielded out as an SplFileInfo object.
+     *                                  Defaults to a recursive iterator of $sourceDir
      * @throws \InvalidArgumentException if the source directory is invalid.
      */
     function recursiveCopy(
         $sourceDir,
         $destDir,
         $extensions = array('php', 'php.gz'),
-        Iterator $files = null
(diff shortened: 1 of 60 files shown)


Jubeki:
@puzzledmonkey Support for Union Types was only added in PHP 8.0:
See https://www.php.net/manual/en/language.types.declarations.php

This library supports PHP >= 7.2.5.

To support PHP 8.4 and keep backwards compatibility you would need the `?`, which was added in PHP 7.1.

For the maintainers:
The explicit nullable type (`?`) is fully backwards compatible and contain no breaking changes.
See the following article: https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated#change

The explicit declaration is compatible with PHP 7.1 and later, even if you extend the class.

You can also take a look at the following snippet at [PHP Playground at php-play.dev](https://php-play.dev/?c=DwfgDgFmBQ0MYBsCGBnFACACkgTgUwDsAXAYWTXQG9p1awBXAIwQEs50AzeguIlgewLoieFCJwAKEGJwsCAc3QASAG5IE9POgC86AvQQIAlFQC%2B0c9AD0V9ABUILDIlQYn6OPwC2YJH2ZaSBziWAASmOgA7AB0AIwANOiMeHBI9ChaLEToANZ4eGAYRBBa%2BobCAJ5gmT6scFkIFdHw5BgAykxkruh4AB4iBAAmGNj4xF0U1LTTDMxsnNy8AkIiYniSMnKKquqaOnoGxmYWsDb2js6t6O6ePn4sAehBIZjhUXGJyanpmdl5BUUSgdykQqlo%2BmA6g0mi1uh1GBMUAAmKg0OhMOoLHh8QTCUTiKSbBTKNQaPAmSjmSxnBw3K7uAj8bK3Xz%2BBDglSEJJ4Dj8fBhCIADmiABZPik0hlhECWLU2FlgQhKtVrhh8F5%2BJzBs0XBR4YiAMw9fqEYZYXCEUhXKbTdCzTFcbHLPFrDZEWTEnZkilU062WmXboMpkebysh7snqcoTJXn815C0Xi75S4o1SHy7JlJWglXudWavDa2F6zqtEXGgZm0aWxGo212jHzR1LXGrAlE7akva6ADkvV7PpOQA&v=8.3&f=html)

```php
<?php

class ParentClass {
   public function tester(?string $value = null) {}
}

// This class is compatible after PHP 7.1, because it keeps the null type implicitly.
class SubClass extends ParentClass {
     public function tester(string $value = null) {}
}

// This class is compatible after PHP 7.1, because it keeps the null type explicitly.
class SubClass2 {
   public function tester(?string $value) {}
}

// Fatal Error: This class is not compatible with the parent class even before PHP 8.4,
// because the implicit null type is removed from the declaration.
class SubClass3 extends ParentClass {
     public function tester(string $value) {}
}

// Fatal Error: This class is not compatible with the parent class even before PHP 8.4,
// because the implicit null type is removed from the declaration.
class SubClass4 extends ParentClass {
     public function tester(string $value = 'x') {}
}
```

puzzledmonkey:
Sorry @Jubeki I didn't spot your PR for some reason (I did search!) so I'll close this one now.

Jubeki:
@puzzledmonkey you can keep your PR open. My PR was closed, so it would be great if another PR is considered.

puzzledmonkey:
This PR also now uses the backwards-compatible ? syntax rather than |null, and this should therefore mean this PR has no breaking changes at all, and just adds forwards-compatibility for PHP 8.4

yenfryherrerafeliz:
Hi @puzzledmonkey, thanks for your contribution. I will be closing this PR in favor of [this](https://github.com/aws/aws-sdk-php/pull/2995) one which was opened a bit earlier for addressing this. Please track any further updates there.

Thanks!

Issue #2665, opened 2023-03-27
Deprecations messages in symfony 5.4


Spawnrad:
### Describe the bug

Lot of deprecation messages when using this SDK with Symfony 5.4

### Expected Behavior

No deprecation message

### Current Behavior

```
User Deprecated: Method "Iterator::key()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Aws\ResultPaginator" now to avoid errors or add an explicit @return annotation to suppress this message.
User Deprecated: Method "Iterator::next()" might add "void" as a native return type declaration in the future. Do the same in implementation "Aws\ResultPaginator" now to avoid errors or add an explicit @return annotation to suppress this message.
User Deprecated: Method "Iterator::valid()" might add "bool" as a native return type declaration in the future. Do the same in implementation "Aws\ResultPaginator" now to avoid errors or add an explicit @return annotation to suppress this message.
User Deprecated: Method "Iterator::rewind()" might add "void" as a native return type declaration in the future. Do the same in implementation "Aws\ResultPaginator" now to avoid errors or add an explicit @return annotation to suppress this message.

```

### Reproduction Steps

Just run any tests with Symfony 5.4

### SDK version used

3.262.1

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 7.4.33

zerkms:
These warnings come from `Aws\ResultPaginator` and the trivial fix is to add typing to phpdoc for the methods (like it was done here already: https://github.com/aws/aws-sdk-php/pull/2434)

Spawnrad:
Version 3.316.1
Issue already here

RanVaknin:
Hi @Spawnrad @zerkms ,

We have added the `ReturnTypeWillChange` annotation in `Aws\ResultPaginator` which are meant to suppress these warnings.

@Spawnrad  I didnt quite understand your last comment, are you still experiencing this? 

Can you enable the SDK's debug logs and share them with us just in case? 
Can you re-iterate the reproduction instructions to be more clear so I can reproduce this myself?

Thanks,
Ran~

Spawnrad:
@RanVaknin Yes i have upgrade aws sdk and have the same issue 
Here the exact message : 

[...]

PR #3011, opened 2024-10-07
adding return annotations to supress warnings

RanVaknin:
Addresses: https://github.com/aws/aws-sdk-php/issues/2665

Some SDK classes implement only some parts of the `Iterator` interface, causing warnings for customers that are using PHP version 7.x. Adding return type annotations help mitigate those warnings.


stobrien89 (approved):

stobrien89:
Thanks @RanVaknin!

PR #3025, opened 2024-11-12
chore: add query mode header

stobrien89:
*Description of changes:*
Adds opt-in header `x-amzn-query-mode` for query-compatible services (that have migrated to json or another protocol).

### DIFF
--- a/src/AwsClient.php
+++ b/src/AwsClient.php
@@ -12,6 +12,7 @@ use Aws\EndpointV2\EndpointV2Middleware;
 use Aws\Exception\AwsException;
 use Aws\Signature\SignatureProvider;
 use GuzzleHttp\Psr7\Uri;
+use Psr\Http\Message\RequestInterface;
 
 /**
  * Default AWS client implementation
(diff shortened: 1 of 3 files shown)


src/AwsClient.php
```diff
@@ -275,6 +276,7 @@ public function __construct(array $args)
 
         if (!is_null($this->api->getMetadata('awsQueryCompatible'))) {
```
yenfryherrerafeliz, line 277: I think is safer if this condition is scoped inside of the addQueryModeHeader method.
For example:
```php
private function addQueryModeHeader(): void
{
    if (is_null($this->api->getMetadata('awsQueryCompatible'))) {
            return;
    }
    
    ...add the header.
}
```

yenfryherrerafeliz (changes requested):
It looks good. Just a minor change.

src/AwsClient.php
stobrien89: I've thought about it and for readability and to avoid looking for this trait more than once for each client instantiation, I'd like to keep the changes as they are now

yenfryherrerafeliz (approved):

Issue #3026, opened 2024-11-14
Shared config files still checked despite "use_aws_shared_config_files" = false


markinjapan:
### Describe the bug

The `use_aws_shared_config_files` config setting is not respected when resolving the following client configuration options:
* `region`
* `sigv4a_signing_region_set`

This causes the `ConfigurationResolver::resolve()` function to check for `~/.aws/config` files unnecessarily.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

By setting the `use_aws_shared_config_files` config option, when all other config options are resolved, the shared config files should never be checked.

### Current Behavior

Currently shared config files are checked for `region` and `sigv4a_signing_region_set`.

### Reproduction Steps

```php
// Create S3 client
$s3_client = new Aws\S3\S3Client([
    'region' => 'ap-northeast-1',
    'credentials' => new Aws\Credentials\Credentials(
        'MYAWSKEY',
        'xxxxxxxxxxxxxxxx'
    ),
    'version' => '2006-03-01',
    'use_aws_shared_config_files' => false,
]);
```

### Possible Solution

In the methods `ClientResolver::_default_sigv4a_signing_region_set()` and `ClientResolver::_default_region()`, the `$args` argument is not being passed to the `ConfigurationResolver::resolve()` method.

This means the config option `use_aws_shared_config_files` is not respected.

### SDK version used

3.326.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.3.13, MacOS 15.1 (24B83)

yenfryherrerafeliz:
Hi @markinjapan, thanks for reporting this. I have opened a [PR](https://github.com/aws/aws-sdk-php/pull/3027) to address this issue.

Thanks!

yenfryherrerafeliz:
Hi @markinjapan, the reported issue here should be fixed now. 

I will close this issue but please feel free to reach out if you need anything else that we can help with.

Thank you!

the6thBook:
Still getting AWS-3.356.42_PHAR

[Tue Oct 28 19:46:09.990728 2025] [php:warn] [pid 10988:tid 10988] [client X.X.X.X:0] PHP Warning:  file_get_contents(/.aws/credentials): Failed to open stream: Permission denied in phar:///XXX/aws.phar/Aws/functions.php on line 409
[...]

PR #3027, opened 2024-11-14
fix: honor use_aws_shared_config_files

yenfryherrerafeliz:
Honor the `use_aws_shared_config_files` config from `default_region` and `default_sigv4a_signing_region_set` by passing `$args` along to the `ConfigurationResolver::resolve` call.

*Issue #3026*

*Description of changes:*

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -1318,12 +1318,13 @@ class ClientResolver
         }
     }
 
-    public static function _default_sigv4a_signing_region_set(array &$args)
+    public static function _default_sigv4a_signing_region_set(array $args)
     {
         return ConfigurationResolver::resolve(
             'sigv4a_signing_region_set',
             '',
-            'string'
+            'string',
+            $args
         );
     }
 
@@ -1335,9 +1336,14 @@ class ClientResolver
         $args['region'] = $value;
     }
 
-    public static function _default_region(&$args)
+    public static function _default_region($args)
     {
-        return ConfigurationResolver::resolve('region', '', 'string');
+        return ConfigurationResolver::resolve(
+            'region',
+            '',
+            'string',
+            $args
+        );
     }
 
     public static function _missing_region(array $args)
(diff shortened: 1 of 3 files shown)


tests/ClientResolverTest.php
stobrien89, line 1: These tests can be removed— we already have unit tests for this in `ConfigurationResolverTest`

stobrien89 (changes requested):
Just a request to remove the tests— we cover that behavior in `ConfigurationResolverTest` and this test class is arguably bloated already.

tests/ClientResolverTest.php
stobrien89: Per our discussion, I agree that moving `resolve` calls to a method on the `ClientResolver` would help enforce passing `$args`.

src/ClientResolver.php
```diff
@@ -1356,6 +1345,36 @@ public static function _missing_region(array $args)
...
+     *
+     * @param $key
+     * @param $expectedType
+     * @param $args
+     *
+     * @return mixed|string
+     */
+    private static function _resolve_from_env_ini(
+        $key,
+        $expectedType,
```
stobrien89, line 1359: Let's add some type hints here

src/ClientResolver.php
```diff
@@ -327,7 +341,7 @@ class ClientResolver
             'valid' => ['array', 'string'],
```
stobrien89, line 341: The configuration resolver technically doesn't handle arrays (nor should it), so I think we'll need to move 'string' to the front here.

src/ClientResolver.php
```diff
@@ -1356,6 +1345,36 @@ public static function _missing_region(array $args)
...
+        $key,
+        $expectedType,
+        $args
+    ) {
+        static $typeDefaultMap = [
+            'int' => 0,
+            'bool' => false,
+            'boolean' => false,
+            'string' => '',
+            'array' => []
```
stobrien89, line 1367: Let's remove 'array'

stobrien89 (changes requested):

src/ClientResolver.php
yenfryherrerafeliz: Good catch. Thanks!

stobrien89 (approved):

PR #3047, opened 2024-12-13
enhancement: user agent 2.1

yenfryherrerafeliz:
*Issue #, if available:*

Related PR: https://github.com/aws/aws-sdk-php/pull/3001
It was reverted previously due to integ tests failures. The `x-amz-user-agent` should have been excluded from metrics gathering.

*Description of changes:*
This change provides:
- A builder class for appending metrics
- Default initialization of a metrics builder within command instantiation
- Wraps user agent logic into a single middleware class
- Adds middlewares into the different features from where metrics can be gather, to acomplish this purpose.
- Move the user agent middleware after the signing step in order to gather signature metrics.
- Add request compression metric gathering.
- Add specific testing for signatures.
- Add specific testing for request compression.
- Add specific testing for s3 encryption clients.
- Add credentials metric gathering logic.
- Add tests for credentials metrics.
- Make existent credentials tests to work with the new field `source`. For example, for tests around profile credentials the source property for credentials MUST be `profile`. The default value for this field is `static`.
- Use static mapping instead of multiple if/else-ifs.
- Remove the endpoint_id metric tracking since it was removed from the spec.
- Make method's name to be verb based.
- Remove the getMetricsBuilder from the CommandInterface.
- Add tests cases to cover credentials sets the correct source when resolved.
- Fix the sso, sso_legacy, and process metrics. They should be all a profile based metric.


yenfryherrerafeliz:
Smoke and Integ tests passed:

```
....................

60 scenarios (60 passed)
230 steps (230 passed)
15m12.41s (44.50Mb)
```

```
..............

112 scenarios (112 passed)
224 steps (224 passed)


```

Issue #3056, opened 2025-01-07
Account-based endpoint does not work if credential source is ECS


emrekaya-insider:
### Describe the bug

Hello, I'm trying to enable Account-based endpoint in my ECS task. But, I reliazed that http://169.254.170.2/v2/credentials does not return AccountId in response, that's why AccountId is being set as null every time.

https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/EcsCredentialProvider.php#L94

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/security-iam-roles.html

<img width="920" alt="image" src="https://github.com/user-attachments/assets/7ce727d7-aa16-4429-82fb-5a3bbf7d05dd" />

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

If response does not have AccountId field, it should be retrieved from shared config or env variable

### Current Behavior

It checks only response

### Reproduction Steps

Sending curl request to retrieve credentials

### SDK version used

3.336.8

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

8.1

emrekaya-insider:
hi @yenfryherrerafeliz , is there any update here, when do you plan to fix this issue

yenfryherrerafeliz:
Hi @emrekaya-insider, sorry for the delay. I have not gotten a complete answer from the service yet. However, it appears to be safe to take the account id from the role. I have opened this [PR](https://github.com/aws/aws-sdk-php/pull/3058) to remediate this issue. Once is reviewed and approved I will merge it.

Thank you!

yenfryherrerafeliz:
Hi @emrekaya-insider, this should be fixed now. 

Thank you!

emrekaya-insider:
hi @yenfryherrerafeliz , it is working now, thank you for help 🙏

PR #3058, opened 2025-01-13
fix: account_id resolution in ecs

yenfryherrerafeliz:
*Issue #3056*

*Description of changes:*
In order to support account id endpoint resolution on ECS when ECS does not return an AccountId field as part of its response. What is being done in this change is trying to populate the AccountId field from the RoleArn returned in the response.

### DIFF
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -93,10 +93,6 @@ class EcsCredentialProvider
                                 $result['AccountId'] = $parsedArn->getAccountId();
                             } catch (\Exception $e) {
                                 // AccountId will be null
-                                trigger_error(
-                                    "An error occured while parsing the ARN string: " . $e->getMessage(),
-                                    E_USER_WARNING
-                                );
                             }
                         }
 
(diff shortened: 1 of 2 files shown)


src/Credentials/EcsCredentialProvider.php
```diff
@@ -86,6 +87,19 @@ public function __invoke()
                         ]
                     )->then(function (ResponseInterface $response) {
                         $result = $this->decodeResult((string)$response->getBody());
+                        if (!isset($result['AccountId']) && isset($result['RoleArn'])) {
+                            try {
+                                $parsedArn = new Arn($result['RoleArn']);
+                                $result['AccountId'] = $parsedArn->getAccountId();
+                            } catch (\Exception $e) {
+                                // AccountId will be null
+                                trigger_error(
```
stobrien89, line 96: Do users have any control over whether or not an arn is returned and parsed successfully?  Some error handlers may convert this into an exception.

stobrien89 (changes requested):

src/Credentials/EcsCredentialProvider.php
yenfryherrerafeliz: I think users should attach a task role to the task definition in order to have a RoleArn returned. Maybe I should avoid this warning and just leave AccountId null when it is not resolve from any of the sources. What do you think?

yenfryherrerafeliz: Warning have been removed.

stobrien89 (approved):

PR #3065, opened 2025-01-27
feat: track checksum calculation validation mode

yenfryherrerafeliz:
*Issue #, if available:*

*Description of changes:*
- For requests, trackes checksum request calculation, by being either when supported or when required.
- For responses, trackers checksum response validation, by being either when supported or when required.

### DIFF
--- a/src/MetricsBuilder.php
+++ b/src/MetricsBuilder.php
@@ -34,6 +34,10 @@ final class MetricsBuilder
     const FLEXIBLE_CHECKSUMS_REQ_CRC64 = "W";
     const FLEXIBLE_CHECKSUMS_REQ_SHA1 = "X";
     const FLEXIBLE_CHECKSUMS_REQ_SHA256 = "Y";
+    const FLEXIBLE_CHECKSUMS_REQ_WHEN_SUPPORTED = "Z";
+    const FLEXIBLE_CHECKSUMS_REQ_WHEN_REQUIRED = "a";
+    const FLEXIBLE_CHECKSUMS_RES_WHEN_SUPPORTED = "b";
+    const FLEXIBLE_CHECKSUMS_RES_WHEN_REQUIRED = "c";
     const CREDENTIALS_CODE = "e";
     const CREDENTIALS_ENV_VARS = "g";
     const CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN = "h";
@@ -139,6 +143,7 @@ final class MetricsBuilder
             'credentials' => 'appendCredentialsMetric',
             'account_id_endpoint_mode' => 'appendAccountIdEndpointMode',
             'account_id_endpoint' => 'appendAccountIdEndpoint',
+            'request_checksum_calculation' => 'appendRequestChecksumCalculationMetric',
         ];
 
         $fn = $appendMetricFns[$featureGroup];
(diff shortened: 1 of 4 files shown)


src/UserAgentMiddleware.php
```diff
@@ -314,4 +315,20 @@ private function appendRetryConfigMetric(): void
             );
         }
     }
+
+    private function appendResponseChecksumValidationMetric(): void {
```
stobrien89, line 319: I probably should have raised this question in prior reviews, but is it necessary that this method lives here? Can this logic not be moved to the `MetricsBuilder` class? same goes for the other metrics `append` methods above.

src/MetricsBuilder.php
```diff
@@ -246,6 +251,20 @@ private function appendCredentialsMetric(
         }
     }
 
+    private function appendRequestChecksumCalculationMetric(
```
stobrien89, line 254: We should probably have assertions on the corresponding test class for this metrics

stobrien89 (changes requested):

src/UserAgentMiddleware.php
yenfryherrerafeliz: I also though on that before, but I think this should be the right place for those methods to live in, and here is why:
- The metrics builder is just a child utility/helper of the UserAgentMiddleware for gathering values that will be used just for populating the user agent itself.
- While, some similar methods are contained within the MetricsBuilder already, such as `appendRequestCompressionMetric`, `appendRequestChecksumMetric`, etc., those methods are there just for simplicity while gathering those metrics from every place those metrics should be tracked; otherwise, we would need to move those conditions to the top level place from where those metrics come from, which would make the code less cleaner in those places.
- Moving those methods to the MetricsBuilder class will require to pass the client arguments also to the metrics builder when it already exists within the UserAgentMiddleware.

What do you think?

src/MetricsBuilder.php
yenfryherrerafeliz: That method is implicitly covered [here](https://github.com/aws/aws-sdk-php/pull/3065/files#diff-8aabda90e2b5d80c8054b3f415b3478853e44ee9d1c48824bc29465837e91f84R1566), however, if you think we need individual tests for those methods I can do it.

src/MetricsBuilder.php
```diff
@@ -287,6 +306,101 @@ private function appendAccountIdEndpoint(string $endpoint): void
...
     }
 
+    /**
+     * Resolves metrics from client arguments.
+     *
+     * @param array $args
+     *
+     * @return void
+     */
+    public function resolveAndAppendFromArgs(array $args = []): void {
```
stobrien89, line 316: Opening brace should go to a newline.  Also needs a test in `MetricsBuilderTest`

src/MetricsBuilder.php
```diff
@@ -287,6 +306,101 @@ private function appendAccountIdEndpoint(string $endpoint): void
...
+    }
+
+    /**
+     * Appends the provided/resolved response checksum validation mode.
+     *
+     * @param array $args
+     *
+     * @return void
+     */
+    private function appendResponseChecksumValidationMetric(array $args): void {
```
stobrien89, line 388: Opening brace needs a newline

stobrien89 (changes requested):
Just a few minor things.  The logic change looks good!

stobrien89 (approved):

PR #3072, opened 2025-02-02
chore: run GHA HandleStaleDiscussions only on aws repo

ziegenberg:
*Issue #, if available:*
N/A

*Description of changes:*
Running this GHA on forked repos will most likely fail and we should prevent this from happening.


stobrien89:
Thanks again for the contribution @ziegenberg!

PR #2923, opened 2024-05-16
feat: remove code for unsupported version v5 of Guzzle

ziegenberg:
*Issue #, if available:*
n/a

*Description of changes:*
Remove code for unsupported version v5 of Guzzle.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3080, opened 2025-02-06
feat: remove code for unsupported version v6 of Guzzle

ziegenberg:
*Issue #, if available:*
n/a

*Description of changes:*
Remove code for unsupported version v6 of Guzzle.

This is based on #2923.

### DIFF
--- a/src/Handler/GuzzleV6/GuzzleHandler.php
+++ b/src/Handler/GuzzleV6/GuzzleHandler.php
@@ -1,88 +1,11 @@
 <?php
+
 namespace Aws\Handler\GuzzleV6;
 
-use Exception;
-use GuzzleHttp\Exception\ConnectException;
-use GuzzleHttp\Exception\RequestException;
-use GuzzleHttp\Promise;
-use GuzzleHttp\Client;
-use GuzzleHttp\ClientInterface;
-use GuzzleHttp\TransferStats;
-use Psr\Http\Message\RequestInterface as Psr7Request;
+trigger_error(sprintf(
+    'Using the "%s" class is deprecated, use "%s" instead.',
+    __NAMESPACE__ . '\GuzzleHandler',
+    \Aws\Handler\Guzzle\GuzzleHandler::class
+), E_USER_DEPRECATED);
 
-/**
- * A request handler that sends PSR-7-compatible requests with Guzzle 6.
- */
-class GuzzleHandler
-{
-    /** @var ClientInterface */
-    private $client;
-
-    /**
-     * @param ClientInterface $client
-     */
-    public function __construct(?ClientInterface $client = null)
-    {
-        $this->client = $client ?: new Client();
-    }
-
-    /**
-     * @param Psr7Request $request
-     * @param array       $options
-     *
-     * @return Promise\Promise
-     */
-    public function __invoke(Psr7Request $request, array $options = [])
-    {
-        $request = $request->withHeader(
-            'User-Agent',
-            $request->getHeaderLine('User-Agent')
-                . ' ' . \GuzzleHttp\default_user_agent()
-        );
-
-        return $this->client->sendAsync($request, $this->parseOptions($options))
-            ->otherwise(
-                static function ($e) {
-                    $error = [
-                        'exception'        => $e,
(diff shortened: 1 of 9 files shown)


ziegenberg:
@stobrien89, I rebased onto the current master.

ziegenberg:
@stobrien89, I rebased onto the current master and fixed the test failure (multiple items in changelog)

src/Handler/GuzzleV6/GuzzleHandler.php
stobrien89, line 1: Could you use a class alias instead? something like this:

```php
<?php

namespace Aws\Handler\GuzzleV6;

@trigger_error(sprintf(
    'Using the "%s" class is deprecated, use "%s" instead.',
    __CLASS__,
    \Aws\Handler\Guzzle\GuzzleHandler::class
), E_USER_DEPRECATED);

class_alias(\Aws\Handler\Guzzle\GuzzleHandler::class, __NAMESPACE__ . '\GuzzleHandler');

File: tests/FunctionsTest.php [hunk 1]
```diff
@@ -211,19 +211,6 @@ public function testDescribeType()
         $this->assertSame('string(4) "test"', Aws\describe_type("test"));
     }
 
-    /**
-     * @covers Aws\default_http_handler()
-     */
-    public function testGuzzleV6HttpHandler()
```
stobrien89, line 217: I'd like to keep this, just with the updated namespace/class

tests/Handler/GuzzleV6/HandlerTest.php
```diff
@@ -1,149 +1,15 @@
...
-use GuzzleHttp\TransferStats;
 use Yoast\PHPUnitPolyfills\TestCases\TestCase;
 
 /**
  * @covers Aws\Handler\GuzzleV6\GuzzleHandler
  */
 class HandlerTest extends TestCase
 {
-    public function set_up()
+    public function testV6ClassExists(): void
```
stobrien89, line 11: Could you add another test with an assertions for the deprecation notice?

src/Handler/GuzzleV6/GuzzleHandler.php
stobrien89: I'd also prefer not to suppress the deprecation notice. What do you think?

stobrien89 (changes requested):
Requested a couple of changes.  Thanks a lot for the work on this.  Happy to make the changes myself if you want me to.

ziegenberg:
> Requested a couple of changes.  Thanks a lot for the work on this.  Happy to make the changes myself if you want me to.

I'm happy to make the changes in a couple of hours tomorrow morning 😊

src/Handler/GuzzleV6/GuzzleHandler.php
ziegenberg: I changed this as requested.

tests/FunctionsTest.php
ziegenberg: Done.

tests/Handler/GuzzleV6/HandlerTest.php
ziegenberg: Done. Expecting deprecations will be deprecated in PHPUnit 10 and upwards: https://github.com/sebastianbergmann/phpunit/issues/5062

_btw: the PHPUnit version in the composer.json can be updated to `^9.6`. We could even go one higher as PHPUnit 10 requires PHP 8.1 or later._

ziegenberg:
Hi @stobrien89, all changes implemented.

tests/Handler/GuzzleV6/HandlerTest.php
stobrien89: Thank you!  We're aware— waiting on a revamp of our tests to the new attributes/decorators which we'll pick up at some point this year.

stobrien89 (approved):

stobrien89:
Thanks again @ziegenberg!

stobrien89:
Turns out `@covers Aws\Handler\GuzzleV6\GuzzleHandler` was causing the unit tests to fail due to the unhandled deprecation notice from the file being loaded.  I removed it.  Should be good to go after this workflow run.

Issue #3078, opened 2025-02-06
RuntimeException: Failed to open stream: Too many open files


nextend:
### Describe the bug

I want to upload a lot of files into the S3 bucket. I use the CommanPool to make concurrent requests.
```
        $commands = [];

            $commands[] = $client->getCommand('putObject', [
                'Bucket'         => $bucketName,
                'Key'              => $path,
                'SourceFile'   => $pathToFile,
                'ACL'             => 'private',
            ]);
//.....
            $pool = new CommandPool($client, $commands, [
                'concurrency' => 10
            ]);

            $promise = $pool->promise();

            $promise->wait();
```

ulimit -n is 1024

Tried to force garbage collection, but it got the same error:
```
            $pool = new CommandPool($client, $commands, [
                'concurrency' => 10,
                'before'      => function (CommandInterface $cmd, $iterKey) {
                    gc_collect_cycles();
                }
            ]);
```

This it what worked, but probably not optimal for memory consumption:
```
            $commands[] = $client->getCommand('putObject', [
                'Bucket' => $bucketNamePrivate,
                'Key'    => $path,
                'Body'   => file_get_contents($pathToFile),
                'ACL'    => 'private',
            ]);
```

Stack trace:
```
array(5) {
  ["type"]=>
  int(1)
  ["message"]=>
  string(8720) "Uncaught RuntimeException: Unable to open "[...].json" using mode "r": fopen([...].json): Failed to open stream: Too many open files in [...]public/composer/vendor/guzzlehttp/psr7/src/Utils.php:375
Stack trace:
#0 [internal function]: GuzzleHttp\Psr7\Utils::GuzzleHttp\Psr7\{closure}()
#1 [...]public/composer/vendor/guzzlehttp/psr7/src/Utils.php(387): fopen()
#2 [...]public/composer/vendor/guzzlehttp/psr7/src/LazyOpenStream.php(47): GuzzleHttp\Psr7\Utils::tryFopen()
#3 [...]public/composer/vendor/guzzlehttp/psr7/src/StreamDecoratorTrait.php(33): GuzzleHttp\Psr7\LazyOpenStream->createStream()
#4 [...]public/composer/vendor/guzzlehttp/psr7/src/StreamDecoratorTrait.php(109): GuzzleHttp\Psr7\LazyOpenStream->__get()
[...]

Issue #2335, opened 2021-10-13
S3 putObject leaves file handles open


desjardinspezmax:
We are putting a file on S3 using putObject.
Once the transfer is completed, we are removing the file and try removing the directory. But when we are trying to remove the directory with rmdir, we get an error that the directory is not empty.

We did some searching and we found this bug report having the same behavior than us https://github.com/aws/aws-sdk-php/issues/81

We did the fix suggested by user mtdowling and it fixed our problem.

On the version 3.197.1 of aws-sdk-hp we had no problem. We updated to version 3.198.0 and our code stopped working.

PHP version: PHP 7.4.23 (cli) (built: Sep  3 2021 18:20:46) ( NTS )

```
        $a_Params = array(
            'Bucket' => CUSTOMER_AWS_BUCKET,
            'Key' => $sFileName,
            'SourceFile' => $sFilePath,
            'StorageClass' => $sStorageClass,
        );

        $objResult = self::$objS3Client->putObject($a_Params);
```

This is how we patched it to get it working

`        
$handle = fopen($sFilePath, 'r');

        $a_Params = array(
            'Bucket' => CUSTOMER_AWS_BUCKET,
            'Key' => $sFileName,
            'Body' => $handle,
            'StorageClass' => $sStorageClass,
        );

        $objResult = self::$objS3Client->putObject($a_Params);

        // Explicitly close the file handle when done
        fclose($handle);`

paulhenri-l:
Hi there, I ran into this issue as well. But the given workaround only works when the upload is not a multipart one.

When the upload is a multipart one, there are multiple file handles created and they appear to never be closed.

I'd happily make a PR but I am not sure where would be the best place to close these handles.

https://github.com/aws/aws-sdk-php/blob/58fa9d8b522b0afa260299179ff950c783ff0ee1/src/Multipart/AbstractUploader.php#L44:L78

My guess is that they should be closed somewhere after the `yield`

SamRemis:
hi there, this is off the top of my head and I would need some more time to look into this to say for sure, but if I recall correctly there's not an ideal place for the file handle to be closed.  The file handle still needs to be open when the call to the service is made, so we may have to include a file to be closed in each call and close them individually after a successful response is returned from S3.  
[...]

PR #3085, opened 2025-02-14
fix: closes lazy stream

yenfryherrerafeliz:
*Issue #3078, #2335*

*Description of changes:*
In a put object request if the body of the request is given with the `SourceFile` parameter, internally there is a middleware that creates a lazy stream based on that file, in order to populate the body of the request, however, this stream is not closed and hence leaving a file open in every request. This change fixes this by making sure the stream is closed once the request is completely sent.


yenfryherrerafeliz:
```

66 scenarios (66 passed)
248 steps (248 passed)
8m35.72s (42.07Mb)
```
```
..............
```
```
112 scenarios (112 passed)
224 steps (224 passed)
0m32.90s (95.28Mb)
```

tests/MiddlewareTest.php
```diff
@@ -449,4 +449,44 @@ public function recursionDetectionProvider()
...
+        }
+
+        $sourceFile = $tempDir . '/test-file.txt';
+        file_put_contents($sourceFile, 'fooData');
+        try {
+            $client = new Aws\S3\S3Client([
+                'region' => 'us-west-2',
+                'http_handler' => function ($command, $request) {
+
+                    return new Psr7\Response(200);
```
stobrien89, line 472: nit: extra newline here

stobrien89 (changes requested):
Looks good to me-  There should be a changelog entry for this in case we see some adverse effects.

.changes/nextrelease/fix-closes-internal-stream.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "feature",
+        "category": "Api",
+        "description": "When doing a putObject request and the body is provided with the `SourceFile` parameter then, internally a lazy open stream is created in order to populate the body of the request. This change makes sure that created stream is closed after request completion."
+      }
+]
```
stobrien89, line 7: newline needed here

src/Middleware.php
```diff
@@ -51,8 +51,29 @@ public static function sourceFile(
...
                 ) {
-                    $command[$bodyParameter] = new LazyOpenStream($source, 'r');
+                    $lazyOpenStream = new LazyOpenStream($source, 'r');
+                    $command[$bodyParameter] = $lazyOpenStream;
                     unset($command[$sourceParameter]);
+
+                    $next = $handler($command, $request);
+                    // To avoid failures in some tests cases
+                    if ($next !== null && method_exists($next, 'then')) {
+                        try {
```
stobrien89, line 61: Should this not be:

```php
if ($next !== null && method_exists($next, 'then')) {
    return $next->then(
        function ($result) use ($lazyOpenStream)
            $lazyOpenStream->close();
            return $result;
        }
    )->otherwise(function (\Throwable $e) use ($lazyOpenStream) {
        $lazyOpenStream->close();
        throw $e; // Re-throw the exception
    });
}
```

stobrien89 (changes requested):

.changes/nextrelease/fix-closes-internal-stream.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "feature",
+        "category": "Api",
```
stobrien89, line 4: I would leave this field blank since the `Middleware` class doesn't belong to that namespace

stobrien89 (changes requested):

stobrien89 (approved):

Issue #3086, opened 2025-02-17
Memory Leak and Segfault (Exit Code 139) in Aws\ResultPaginator Due to Repeated Metrics Middleware Registration


iromanko:
### Describe the bug

Starting from version 3.336.0, the Aws\ResultPaginator class continuously adds a metrics middleware each time it is instantiated. This leads to:
	•	An increasing function call stack over time in long-running processes.
	•	High memory consumption due to repeated middleware registrations.
	•	In some cases, PHP crashes with exit code 139 (segmentation fault).

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

The Aws\ResultPaginator class should not repeatedly add the metrics middleware on each instantiation. Middleware should be registered once per client lifecycle or have a mechanism to prevent redundant additions.

### Current Behavior

```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 2097152 bytes) in /srv/app/vendor/aws/aws-sdk-php/src/functions.php on line 67

Call Stack:
    0.0011     480264   1. {main}() /srv/app/reproducer.php:0
   10.3241  131880872   2. Aws\AwsClient->getIterator($name = 'Scan', $args = ['TableName' => 'test']) /srv/app/reproducer.php:13
   10.3243  131883568   3. Aws\ResultPaginator->search($expression = 'Items') /srv/app/vendor/aws/aws-sdk-php/src/AwsClientTrait.php:35
```

### Reproduction Steps

	1.	Run a long-running script that continuously creates Aws\ResultPaginator instances.
	2.	Monitor memory usage and function call stack growth.
	3.	Eventually, PHP may crash with exit code 139.

```php

$sdk = new Aws\Sdk([
    'region'  => 'us-east-2',
    'version' => 'latest',
]);

$dynamoDb = $sdk->createDynamoDb();

while (true) {
    $paginator = $dynamoDb->getIterator('Query', ['TableName' => 'test']);
}
```

### Possible Solution

A possible fix could involve checking if the middleware is already present before adding it again.

### Additional Information/Context

Disabling the metrics middleware prevents the issue, confirming it as the root cause.

### SDK version used

3.339.14

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.3.8

yenfryherrerafeliz:
[...]

PR #3087, opened 2025-02-17
fix: prevent duplicating metrics middleware

yenfryherrerafeliz:
*Issue #3086*

*Description of changes:*
- Prevent adding the same metrics middleware multiple times.

### DIFF
--- a/src/HandlerList.php
+++ b/src/HandlerList.php
@@ -116,6 +116,19 @@ class HandlerList implements \Countable
     }
 
     /**
+     * Checks if a middleware exists. The middleware
+     * should have been appended with a name.
+     *
+     * @param string $name
+     *
+     * @return bool
+     */
+    public function hasMiddleware(string $name): bool
+    {
+        return isset($this->named[$name]);
+    }
+
+    /**
      * Append a middleware to the init step.
      *
      * @param callable $middleware Middleware function to add.
--- a/src/HandlerList.php
+++ b/src/HandlerList.php
@@ -117,7 +117,8 @@ class HandlerList implements \Countable
 
     /**
      * Checks if a middleware exists. The middleware
-     * should have been appended with a name.
+     * should have been added with a name in order to
+     * use this method.
      *
      * @param string $name
      *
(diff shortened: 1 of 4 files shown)


yenfryherrerafeliz:
```

66 scenarios (66 passed)
248 steps (248 passed)
9m4.85s (42.06Mb)
```
```
..............

112 scenarios (112 passed)
224 steps (224 passed)
0m36.32s (95.28Mb)
```

src/HandlerList.php
```diff
@@ -115,6 +115,19 @@ public function hasHandler()
         return (bool) $this->handler;
     }
 
+    /**
+     * Checks if a middleware exists. The middleware
+     * should have been appended with a name.
```
stobrien89, line 120: Can we change "appended" to "added"? middleware can be appended or prepended.

tests/MetricsBuilderTest.php
```diff
@@ -169,4 +169,31 @@ public function resolveAndAppendFromArgsProvider(): array
...
         ];
     }
+
+    /**
+     * Tests that metric middlewares are appended just once.
+     *
+     * @return void
+     */
+    public function testAppendMetricsCaptureMiddlewareJustOnce(): void {
+        $handlerList = new HandlerList(function (){});
```
stobrien89, line 179: I think there should be an assertion that "appendBuild" is only called once.  Prior to this change, I'm pretty sure the second attempt to append it would just override the first one since they're both in the build step

stobrien89 (changes requested):
Just a doc nit and an extra assertion needed on the test.  We need a changelog entry for this as well

tests/MetricsBuilderTest.php
yenfryherrerafeliz: Actually it will not override it:
```
// The middleware will be added not matter the name
if ($prepend) {
 $this->steps[$step][] = [$middleware, $name];
} else {
    array_unshift($this->steps[$step], [$middleware, $name]);
}

// Here will be overridden, however, the middleware is already added.
if ($name) {
    $this->named[$name] = $step;
}
```

stobrien89 (approved):

PR #3093, opened 2025-02-28
fix: adjust waiter matchers

yenfryherrerafeliz:
*Issue #, if available:*

*Description of changes:*
- Makes matchesPath to return false when the result value does not match the same value and type as the expected value.
- Makes matchesPathAll to return false when an empty array or a non-array value is resolved by the jmespath evaluator.
- Makes matchesPathAny to return false when a non-array value is resolved by the jmespath evaluator.

This change also refactor the test case providers so that they are more idiomatic by naming them.

### DIFF
--- a/.changes/.nextrelease/feature-progress-tracker.json
+++ b/.changes/.nextrelease/feature-progress-tracker.json
@@ -1,7 +1,0 @@
-[
-    {
-        "type": "enhancement",
-        "category": "S3",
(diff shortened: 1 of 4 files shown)


stobrien89 (changes requested):
Looks good to me- just needs a changelog entry

stobrien89 (approved):

PR #3104, opened 2025-03-31
Changing the return type of getLastRequest to include null

AjaniCS:
In this pull request, I have updated the PHPDoc return type to explicitly include `null`, reflecting that the function may return `null` in certain cases. This change helps to avoid raising PHPStan error `method.impossibleType` when asserting that the last request was `null`.

### DIFF
--- a/src/MockHandler.php
+++ b/src/MockHandler.php
@@ -118,7 +118,7 @@ class MockHandler implements \Countable
     /**
      * Get the last received request.
      *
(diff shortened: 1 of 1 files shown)


stobrien89 (approved):

stobrien89:
Thanks for the contribution, @AjaniCS!

AjaniCS:
Thank you too @stobrien89 !

PR #3111, opened 2025-04-29
Update DocModel.php

LDAV:
${} for string interpolation inside double quotes in PHP is deprecated

*Issue #, if available:*

*Description of changes:*


stobrien89:
Hi @LDAV,

Thanks for the contribution.  In this particular case, we're trying to access variables within the doc model that are prefixed with `$`, so we're not actually using it for interpolation (hence the escape char).  I'd prefer to keep the existing code.  Let us know if you have any questions.

PR #3113, opened 2025-05-07
chore: update ChangelogBuilder

stobrien89:
*Description of changes:*
Validates tag input for `ChangelogBuilder`

### DIFF
--- a/build/changelog/ChangelogBuilder.php
+++ b/build/changelog/ChangelogBuilder.php
@@ -104,6 +104,10 @@ class ChangelogBuilder
 
     private function createChangelogJson($changelog, $tag)
     {
+        if (!preg_match('/^\d+\.\d+\.\d+$/', $tag)) {
+            throw new \InvalidArgumentException('Invalid tag value.');
+        }
+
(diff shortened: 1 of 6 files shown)


build/changelog/ChangelogBuilder.php
```diff
@@ -104,6 +104,10 @@ private function createTag($changelogFile)
 
     private function createChangelogJson($changelog, $tag)
     {
+        if (preg_match('/\.\.|\/|\\\\/', $tag)) {
```
yenfryherrerafeliz, line 107: Could this validation be a little more restrictive. For example a tag value, as far as I am aware of, will be 3 integers separated by a dot "." hence a more restrictive regex would be: `^\d+\.\d+\.\d+$`.

What do you think?

yenfryherrerafeliz (changes requested):

build/changelog/ChangelogBuilder.php
stobrien89: I think that's more appropriate.  Updated

yenfryherrerafeliz (approved):

PR #3115, opened 2025-05-07
feat: InstanceProfileProvider updates

stobrien89:
*Description of changes:*
Adds support for fetching account id from IMDS.  Removes support for IMDSv1.


src/Credentials/InstanceProfileProvider.php
```diff
@@ -285,8 +300,8 @@ private function request($url, $method = 'GET', $headers = [])
 
     private function handleRetryableException(
         \Exception $e,
-        $retryOptions,
-        $message
+                   $retryOptions,
+                   $message
```
yenfryherrerafeliz, line 304: nit: Extra spaces here.

tests/Credentials/InstanceProfileProviderTest.php
```diff
@@ -138,6 +145,8 @@ private function getSecureTestClient(
                 switch ($request->getUri()->getPath()) {
                     case '/latest/meta-data/iam/security-credentials':
                     case '/latest/meta-data/iam/security-credentials/':
+                    case '/latest/meta-data/iam/security-credentials-extended':
+                    case '/latest/meta-data/iam/security-credentials-extended/':
```
yenfryherrerafeliz, line 149: nit: Not a big deal, but I think would be easier if you just check whether `$request->getUri()->getPath()` has a forward slash, and if so then you just opt-in for either cases where the path will end with the forward slash or without it. For example if opting for leaving the path at the end:
```php
if (!str_ends_with(($path = $request->getUri()->getPath()), '/')) {
    $path = $path . "/";
}

switch ($path) {
    case "/latest/meta-data/iam/security-credentials/":
        break;
    case "/latest/meta-data/iam/security-credentials-extended/":
        break;
}
```

yenfryherrerafeliz (changes requested):
I like the refactor done in the credentials resolution logic. Thanks for that. Just some nits, but overall looks good.

src/Credentials/InstanceProfileProvider.php
```diff
@@ -241,23 +155,136 @@ public function __invoke($previousCredentials = null)
...
      * @return PromiseInterface Returns a promise that is fulfilled with the
      *                          body of the response as a string.
      */
-    private function request($url, $method = 'GET', $headers = [])
+    private function request($path, $method = 'GET', $headers = [])
     {
-        $disabled = getenv(self::ENV_DISABLE) ?: false;
-        if (strcasecmp($disabled, 'true') === 0) {
+        $disabled = ConfigurationResolver::ini(self::CFG_DISABLE_EC2_METADATA, 'bool')
+            ?? ConfigurationResolver::env(substr(self::ENV_DISABLE, 4), 'bool')
```
yenfryherrerafeliz, line 277: nit: Why not just to use getenv instead? and that will not require to normalize self::ENV_DISABLE and will avoid having to do a method call.

yenfryherrerafeliz (changes requested):

src/Credentials/InstanceProfileProvider.php
stobrien89: `getenv()` requires us to do the string parsing and/or conversion to bool, whereas the call to the configuration provider guarantees we either receive a boolean or null.

yenfryherrerafeliz: We could just do:
```
$disabled = getenv(self::ENV_DISABLE);
if ($disabled == true) {

}
```

And that will avoid doing any null check or boolean conversion. What do you think?

yenfryherrerafeliz (approved):

stobrien89:
Closing as plans have changed

Issue #3119, opened 2025-05-12
"X-Amz-User-Agent" causes signature issue with HTTP proxy


cweiske:
### Describe the bug

We've got a MinIO instance running behind a (to me unkown) HTTP proxy, which removes the empty `X-Amz-User-Agent` header sent by the AWS PHP SDK. This causes a signature error:
> There were headers present in the request which were not signed.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

No signature error.

### Current Behavior

Signature error:
> There were headers present in the request which were not signed.

This is because `SignedHeaders` contains `x-amz-user-agent`, despite it being on the signature blacklist.

### Reproduction Steps

1. Send request to HTTP to MinIO S3 server behind a proxy that removes empty HTTP headers

### SDK version used

3.343.8

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.2.28

yenfryherrerafeliz:
Hi @cweiske, sorry to hear about your issues. You could just add a middleware that removes the header in question. For example:
```php
$client = new Aws\S3\S3Client([
    'region' => 'us-east-2',
]);
$client->getHandlerList()->prependSign(\Aws\Middleware::mapRequest(function (RequestInterface $request) {
    return $request->withoutHeader('X-Amz-User-Agent');
}));
$client->listBuckets();
```

I hope that helps. 

Thanks!

cweiske:
I already provided a patch for this bug; see #3120.

yenfryherrerafeliz:
Hey @cweiske thanks for providing this PR. However, we already have one open for this [here](https://github.com/aws/aws-sdk-php/pull/2882) which we have not merged yet due to some backward compatibilities issues we get after doing so. For now we can not prioritize investigating why those issues happen; nonetheless, the workaround I provided you with does the trick in the meantime.

Thanks!

cweiske:
Where is the list of "backward compatibilities issues" that you are speaking of? All checks in #2882 are green.

yenfryherrerafeliz:
@cweiske the issues are internal integration tests. I understand your need on having this fixed but for now this can't be prioritized. The workaround I provided you with should be enough to do the trick.

Thank you!

PR #3120, opened 2025-05-12
Fix "X-Amz-User-Agent" in blacklist

cweiske:
The header blacklist contains lowercase header names only, except "X-Amz-User-Agent". This causes this header not to be filtered out, leading to an error with some proxies that remove the empty header:
> There were headers present in the request which were not signed.

This commit lowercases the header, which correctly removes the header from the signature.

Related: https://github.com/aws/aws-sdk-php/pull/2259
Resolves: https://github.com/aws/aws-sdk-php/issues/3119

----


yenfryherrerafeliz:
Closing in favor of [this](https://github.com/aws/aws-sdk-php/pull/2882) one.

PR #3144, opened 2025-06-28
bugfix: aws protocol serialization updates

stobrien89:
*Description of changes:*
Updates to serialization for all AWS protocols.


src/Api/Serializer/JsonBody.php
```diff
@@ -45,22 +45,30 @@ public static function getContentType(Service $service)
      * Builds the JSON body based on an array of arguments.
      *
      * @param Shape $shape Operation being constructed
-     * @param array $args  Associative array of arguments
+     * @param array|string $args  Associative array of arguments, or a string.
      *
      * @return string
      */
-    public function build(Shape $shape, array $args)
+    public function build(Shape $shape, array|string $args)
```
stobrien89, line 52: Note for the reviewer: this change accounts for document types (which are always modeled as structures).  It is possible for a document type payload value to be a string.

src/Api/Serializer/JsonRpcSerializer.php
```diff
@@ -62,23 +62,26 @@ public function __invoke(
         $operationName = $command->getName();
         $operation = $this->api->getOperation($operationName);
         $commandArgs = $command->toArray();
+        $body = $this->jsonFormatter->build($operation->getInput(), $commandArgs);
         $headers = [
                 'X-Amz-Target' => $this->api->getMetadata('targetPrefix') . '.' . $operationName,
-                'Content-Type' => $this->contentType
-            ];
+                'Content-Type' => $this->contentType,
+                'Content-Length' => strlen($body)
```
stobrien89, line 69: Note for the reviewer: Sending a `Content-Length` header is required for JSON RPC services: https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#protocol-behaviors https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#protocol-behaviors

The result of `jsonFormatter->build()` will always be a string.

src/Api/Serializer/JsonRpcSerializer.php
```diff
@@ -62,23 +62,26 @@ public function __invoke(
...
+        ];
 
         if ($endpoint instanceof RulesetEndpoint) {
             $this->setEndpointV2RequestOptions($endpoint, $headers);
         }
 
+        $requestUri = $operation['http']['requestUri'] ?? null;
+        $absoluteUri = str_ends_with($this->endpoint, '/')
+            ? $this->endpoint : $this->endpoint . $requestUri;
```
stobrien89, line 78: JSON clients are required to send requests to the root URL (`/`).  Operations will always contain a `requestUri` property https://github.com/aws/aws-sdk-php/blob/master/src/Api/Operation.php#L24C8-L26C10.

https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#protocol-behaviors
https://smithy.io/2.0/aws/protocols/aws-json-1_1-protocol.html#protocol-behaviors

src/Api/Serializer/QuerySerializer.php
```diff
@@ -66,14 +66,17 @@ public function __invoke(
             'Content-Length' => strlen($body),
             'Content-Type'   => 'application/x-www-form-urlencoded'
         ];
+        $requestUri = $operation['http']['requestUri'] ?? null;
 
         if ($endpoint instanceof RulesetEndpoint) {
             $this->setEndpointV2RequestOptions($endpoint, $headers);
         }
+        $absoluteUri = str_ends_with($this->endpoint, '/')
+            ? $this->endpoint : $this->endpoint . $requestUri;
```
stobrien89, line 73: Query clients are also required to send requests to the root URL (`/`). https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#request-serialization

src/Api/Serializer/RestJsonSerializer.php
```diff
@@ -31,12 +31,11 @@ public function __construct(
         $this->jsonFormatter = $jsonFormatter ?: new JsonBody($api);
     }
 
-    protected function payload(StructureShape $member, array $value, array &$opts)
+    protected function payload(StructureShape $member, array|string $value, array &$opts)
     {
-        $body = isset($value) ?
-            ((string) $this->jsonFormatter->build($member, $value))
-            : "{}";
```
stobrien89, line 38: Nonsensical condition here: `$value` is always set.  `jsonFormatter->build()` (found in `JsonBody`) always returns a string, or throws an exception.

src/Api/Serializer/RestSerializer.php
```diff
@@ -114,9 +122,9 @@ private function serialize(Operation $operation, array $args)
         }
 
         if (isset($bodyMembers)) {
-            $this->payload($operation->getInput(), $bodyMembers, $opts);
+            $this->payload($input, $bodyMembers, $opts);
         } else if (!isset($opts['body']) && $this->hasPayloadParam($input, $payload)) {
-            $this->payload($operation->getInput(), [], $opts);
+            $this->payload($input, [], $opts);
```
stobrien89, line 126: `$input` is initialized at the beginning of the function body and is unaltered.

src/Api/Serializer/RestSerializer.php
```diff
@@ -130,12 +138,31 @@ private function applyPayload(StructureShape $input, $name, array $args, array &
...
         $m = $input->getMember($name);
 
+        $type = $m->getType();
         if ($m['streaming'] ||
-           ($m['type'] == 'string' || $m['type'] == 'blob')
+           ($type === 'string' || $type === 'blob')
         ) {
+            //This path skips setting the content-type header usually done in JsonBody
+            //S3 and glacier determine content type in Middleware::ContentType()
+            if (!isset(self::$excludeContentType[$this->api->getServiceName()])) {
```
stobrien89, line 147: Concerned about backward-compatibility here.  This is a path with heavy traffic.  `Content-Type` is determined based on the input for certain S3 and Glacier operations.  Otherwise, if a structure, union, document, etc. member is a payload target, the `Content-Type` header is set in `RestXmlSerializer::payload()` and `RestJsonSerializer::payload()`

For both `rest-xml` and `rest-json`, `Content-Type` for strings and blobs should be the same:

https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html#content-type
https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html#content-type

src/Api/Serializer/RestSerializer.php
```diff
@@ -130,12 +138,31 @@ private function applyPayload(StructureShape $input, $name, array $args, array &
...
+                    case 'blob':
+                        $opts['headers']['Content-Type'] = 'application/octet-stream';
+                        break;
+                }
+            }
+
+            $body = $args[$name];
+            if (!$m['streaming'] && is_string($body)) {
+                $opts['headers']['Content-Length'] = strlen($body);
+            }
```
stobrien89, line 161: While not required by AWS rest-json or rest-xml, we SHOULD be sending the `Content-Length` header if we can reasonably determine it.  From [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length):

> A user agent SHOULD send Content-Length in a request when the method defines a meaning for enclosed content and it is not sending Transfer-Encoding. For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content). A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data.

We can assume that if an operation models a payload, the HTTP semantics (method like `PUT`, `POST`) allow for it.  Because some payloads hitting this path can be non-seekable streams, we can only reasonably determine the content length if it is not marked as `streaming` (potentially sent via transfer encoding) and is a string.

src/Api/Serializer/RestSerializer.php
```diff
@@ -144,13 +171,29 @@ private function applyPayload(StructureShape $input, $name, array $args, array &
...
+            $headerValues = [];
+
+            foreach ($value as $listValue) {
+                $tempOpts = ['headers' => []];
+                $this->applyHeader('temp', $listMember, $listValue, $tempOpts);
+                $convertedValue = $tempOpts['headers']['temp'];
+                $headerValues[] = $convertedValue;
+            }
+
+            $value = $headerValues;
```
stobrien89, line 195: I'm not sure of the depth to which list headers can be modeled as nested (if at all), but our header serialization logic was not properly serializing boolean and timestamp values.  We would just take the array and its values with no alterations (i.e. 1s and 0s for bools and whatever timestamp value happened to be there).

src/Api/Serializer/RestSerializer.php
```diff
@@ -144,13 +171,29 @@ private function applyPayload(StructureShape $input, $name, array $args, array &
...
+        } elseif (!is_null($value)) {
+            switch ($member->getType()) {
+                case 'timestamp':
+                    $timestampFormat = $member['timestampFormat'] ?? 'rfc822';
```
stobrien89, line 199: Per the spec, we should default to `http-date`, which is interoperable with rfc822: https://smithy.io/2.0/spec/http-bindings.html#serialization-rules

src/Api/Serializer/RestSerializer.php
```diff
@@ -183,15 +226,25 @@ private function applyQuery($name, Shape $member, $value, array &$opts)
...
-        } elseif ($value !== null) {
-            $type = $member->getType();
-            if ($type === 'boolean') {
-                $value = $value ? 'true' : 'false';
-            } elseif ($type === 'timestamp') {
-                $timestampFormat = !empty($member['timestampFormat'])
-                    ? $member['timestampFormat']
-                    : 'iso8601';
-                $value = TimestampShape::format($value, $timestampFormat);
+        } elseif ($member instanceof ListShape) {
```
stobrien89, line 238: Same deal as the list headers- we were sending 1s and 0s for bools and whatever timestamp value was provided in the input.

src/Api/Serializer/RestSerializer.php
```diff
@@ -276,19 +329,25 @@ private function hasPayloadParam(StructureShape $input, $payload)
...
         if ($payload) {
             $potentiallyEmptyTypes = ['blob','string'];
-            if ($this->api->getMetadata('protocol') == 'rest-xml') {
+            if ($this->api->getProtocol() == 'rest-xml') {
                 $potentiallyEmptyTypes[] = 'structure';
             }
+
             $payloadMember = $input->getMember($payload);
-            if (in_array($payloadMember['type'], $potentiallyEmptyTypes)) {
+            //unions may also be empty/unset
```
stobrien89, line 346: See test case id (in `input/rest-json.json`) `RestJsonHttpPayloadWithUnsetUnion`.  This behavior seems to only apply to `rest-json` and I was unable to find clear documentation on it, as most of the Smithy docs indicate union types must have one member that is set and not null.

src/Api/Serializer/RestXmlSerializer.php
```diff
@@ -29,7 +29,9 @@ public function __construct(
     protected function payload(StructureShape $member, array $value, array &$opts)
     {
         $opts['headers']['Content-Type'] = 'application/xml';
-        $opts['body'] = $this->getXmlBody($member, $value);
+        $body = $this->getXmlBody($member, $value);
+        $opts['headers']['Content-Length'] = strlen($body);
```
stobrien89, line 33: The result of $this->getXmlBody() will always be a string.

src/RequestCompressionMiddleware.php
```diff
@@ -87,8 +85,12 @@ private function compressRequestBody(
...
         $compressedBody = $fn($body);
 
-        return $request->withBody(Psr7\Utils::streamFor($compressedBody))
-            ->withHeader('content-encoding', $this->encoding);
+        if (!$request->hasHeader('Content-Encoding')) {
+            return $request->withBody(Psr7\Utils::streamFor($compressedBody))
+                ->withHeader('Content-Encoding', $this->encoding);
+        }
+
+        return $request->withAddedHeader('Content-Encoding', $this->encoding);
```
stobrien89: this appends the compression algorithm to an existing `Content-Encoding` header, if present.  See `SDKAppendedGzipAfterProvidedEncoding_restJson1` in `input/rest-json.json`

tests/Api/Serializer/ComplianceTest.php
```diff
@@ -22,14 +22,39 @@ class ComplianceTest extends TestCase
 {
     use UsesServiceTrait;
 
+    public const TEST_CASES_DIR = __DIR__ . '/../test_cases/protocols/input/';
+
+    private static array $excludedCases = [
+        'RestJsonNullAndEmptyHeaders' => true,
```
stobrien89, line 28: This and `NullAndEmptyHeaders` are blocked by Guzzle.  See https://github.com/guzzle/psr7/pull/625

tests/Api/Serializer/ComplianceTest.php
```diff
@@ -22,14 +22,39 @@ class ComplianceTest extends TestCase
...
+        'FlattenedXmlMap' => true,
```
stobrien89: Any of the XML skipped cases you see here are not handled as expected due to the way we construct shapes.  They inherit some of the properties of the structures that target them.  We'll need to implement a way to see a shape's true definition by exposing its `ShapeMap` property.

tests/Api/Serializer/ComplianceTest.php
```diff
@@ -22,14 +22,39 @@ class ComplianceTest extends TestCase
...
+        'FlatNestedXmlMapRequest' => true,
+        'NullAndEmptyHeaders' => true,
+        'RestJsonHttpChecksumRequired' => true,
```
stobrien89, line 30: This one requires an MD5 checksum on a generic service.  We currently handle checksums on a per-service basis.  Need to investigate further.

tests/Api/Serializer/ComplianceTest.php
```diff
@@ -22,14 +22,39 @@ class ComplianceTest extends TestCase
...
+        'MediaTypeHeaderInputBase64' => true,
```
stobrien89, line 31: In this case, because the input is technically already JSON-encoded, we are double encoding the value in https://github.com/aws/aws-sdk-php/blob/6acc8e4f8b7468e604d6f0442284937398c2959a/src/Api/Serializer/RestSerializer.php#L156C9-L164C10.

We can probably add a check to prevent the double encoding, but wanted to open it up for review.

src/Api/Serializer/RestSerializer.php
```diff
@@ -303,13 +363,24 @@ private function getVarDefinitions($command, $args)
...
+                    switch ($member->getType()) {
+                        case 'timestamp':
+                            $timestampFormat = $member['timestampFormat'] ?? 'iso8601';
+                            $value = $this->formatTimestamp($value, $timestampFormat);
+                            break;
+                        case 'boolean':
+                            $value = $this->formatBoolean($value);
+                            break;
+                    }
+                }
```
stobrien89, line 386: We were also handling timestamps and bools incorrectly in URIs.

src/Api/Serializer/JsonBody.php
yenfryherrerafeliz: Understood. I guess that, in such case, the JSON encoded payload will just be that same string.

src/Api/Serializer/JsonRpcSerializer.php
yenfryherrerafeliz: I am curious to know why even when this header was not being set, requests with this serializer still worked. I guess the calculation was done at the service side then.

src/Api/Serializer/JsonRpcSerializer.php
yenfryherrerafeliz: I am not sure if this is the safest way to do this, because adding the needed slash to the endpoint will depend on whether $requestUri has it, and while current implementation appends a slash to $requestUri [here](https://github.com/aws/aws-sdk-php/blob/master/src/Api/Operation.php#L24C8-L26C10), it is not guaranteed this implementation is not changed in the future which could introduce undesired behaviors.

What do you think?

src/Api/Serializer/QuerySerializer.php
yenfryherrerafeliz: Same comment as [here](https://github.com/aws/aws-sdk-php/pull/3144/files#r2175553140).

src/Api/Serializer/JsonRpcSerializer.php
stobrien89: Yeah, I think most of our server implementations don't make assumptions based on the presence or absence of this header, but it certainly saves a bit of extra work on their end if we provide it to them.

src/Api/Serializer/RestSerializer.php
yenfryherrerafeliz: I guess this is the reason why we previously had that issue with the query mode.

src/RequestCompressionMiddleware.php
yenfryherrerafeliz: Why here the request is returned without the compressed body?, isn't needed?, if is not needed then, can we skip compressing the body in line 86?

tests/Api/Serializer/ComplianceTest.php
yenfryherrerafeliz: Mmm, I am curious to see how a value check would be here. In top of my mind I think the most reliable way to check if a value is already a valid json encoded is decoding it and expecting not errors, but that would add the same overhead or even more as if we just don't check nothing here. Unless you have a better suggestion for it.

src/Api/Serializer/JsonRpcSerializer.php
stobrien89: I think it's safe to assume that the implementation in `Operation` won't change— this particular logic has been in place since 2014.  I also think it's safe to assume that both JsonRPC and Query protocols won't model a `requestUri` other than `/` because it's non-compliant with the protocol specification and something we catch via upstream validation.

src/Api/Serializer/RestSerializer.php
stobrien89: similar in nature, but the query mode header is not modeled in a shape anywhere— we just add it to the request if the `queryCompatible` property is present in the model metadata.

src/RequestCompressionMiddleware.php
stobrien89: Good catch— this is wrong.  I'll get it updated

tests/Api/Serializer/ComplianceTest.php
stobrien89: Yeah the extra overhead wouldn't be ideal.  It's not an urgent case, so we can add this to our backlog to investigate.

src/Api/Serializer/XmlBody.php
```diff
@@ -51,7 +56,7 @@ private function startElement(Shape $shape, $name, XMLWriter $xml)
         if ($ns = $shape['xmlNamespace']) {
             $xml->writeAttribute(
                 isset($ns['prefix']) ? "xmlns:{$ns['prefix']}" : 'xmlns',
-                $shape['xmlNamespace']['uri']
+                $ns['uri'] ?? $ns
```
stobrien89: This is a bit of a question mark- it seems upstream changes may be made to change the representation of this value— currently, in all public AWS models, `xmlnameSpace` is modeled as an object with a required field of `uri` and an optional field of `prefix`.  The test models them as scalar values (string), unless it contains a `prefix` field, then it's modeled as an object.

stobrien89: This change is no longer needed.

src/Api/Serializer/RestSerializer.php
```diff
@@ -276,19 +339,25 @@ private function hasPayloadParam(StructureShape $input, $payload)
     {
         if ($payload) {
             $potentiallyEmptyTypes = ['blob','string'];
-            if ($this->api->getMetadata('protocol') == 'rest-xml') {
+            if ($this->api->getProtocol() == 'rest-xml') {
```
yenfryherrerafeliz: nit: we could use `===` instead of `==`.

src/Api/Serializer/XmlBody.php
```diff
@@ -38,7 +38,12 @@ public function build(Shape $shape, array $args)
         $xml = new XMLWriter();
         $xml->openMemory();
         $xml->startDocument('1.0', 'UTF-8');
-        $this->format($shape, $shape['locationName'] ?: $shape['name'], $args, $xml);
+
+        // Check if this is a payload member
+        $payloadMemberName = $shape['payloadMemberName'] ?? null;
```
yenfryherrerafeliz: Why a shape will contain a member `payloadMemberName`?

tests/Api/Serializer/JsonBodyTest.php
```diff
@@ -26,155 +26,135 @@ public function testUsesEmptyHashByDefault()
 
     public function formatProvider()
     {
-        return [
```
yenfryherrerafeliz, line 29: Could you please define everything in an array first, then iterate over and yield it.
Example:
```php
$cases = [
    ['type' => 'string'], ['foo' => 'bar'], '{"foo":"bar"}'],
    [['type' => 'integer'], ['foo' => 1], '{"foo":1}'],
    'Test a blob is base64 encoded' => [
            [
                [
                    'type' => 'structure',
                    'members' => ['foo' => ['type' => 'blob']]
                ],
                ['foo' => 'a'],
                '{"foo":"' .  base64_encode('a') . '"}'
                'type' => 'structure',
                'members' => ['foo' => ['type' => 'blob']]
            ]
]

foreach ($cases as $case) {
    yield $case;
}
```

tests/Api/Serializer/RestJsonSerializerTest.php
```diff
@@ -366,26 +392,120 @@ public function restJsonContentTypeProvider() {
...
+        $this->assertSame($expectedOutput, $request->getBody()->getContents());
+        $this->assertSame(
+            'application/json',
+            $request->getHeaderLine('Content-Type')
+        );
+    }
+
+    public function handlesDocTypeAsPayloadProvider(): \Generator
+    {
+        yield 'string payload' => ['hello', '"hello"'];
```
yenfryherrerafeliz, line 445: Same here, I would define each item in an array, then iterate over and yield it.

yenfryherrerafeliz (changes requested):

src/Api/Serializer/RestSerializer.php
stobrien89: You're right- I just left it as is without thinking about it.  I will update this in a few other places in `RestSerializer` as well

src/Api/Serializer/XmlBody.php
stobrien89: In `RestSerializer`, there are three separate areas where child classes' (`RestXmlSerializer`, `RestJsonSerializer`) `payload()` method is called.  The area in question is [here](https://github.com/aws/aws-sdk-php/pull/3144/files#diff-b91d4a70ff4db4ccd3b8353c2c9fb4b24bc2a323df337b88d75a57ec6b1a839dR173-R177). For rest-xml services, when an input structure models the `payload` trait ([Smithy `httpPayload`](https://smithy.io/2.0/spec/http-bindings.html#httppayload-trait)), we need a way to pass context to `XmlBody` indicating we are dealing with a payload member and preserve its name, so we can properly derive the payload's root element name. 

 I think a safer way to do this would be with a dunder `__` prefix to avoid potential collisions

tests/Api/Serializer/ComplianceTest.php
stobrien89: This is no longer relevant.

tests/Api/Serializer/JsonBodyTest.php
stobrien89: Thanks for the suggestion— I considered this approach, but opted for individual statements for a few reasons: 

- Stack trace points directly to the specific yield statement if there's a failure
- individual yield statements make it easier to modify specific cases without dealing with array syntax (trailing commas)

I think it makes the providers slightly more readable as well- it's clear where each individual case starts

tests/Api/Serializer/RestJsonSerializerTest.php
stobrien89: same comment as above

src/Api/Serializer/RestSerializer.php
stobrien89: updated.

src/Api/Serializer/XmlBody.php
stobrien89: Updated.  Still open to suggestions on how we pass the context

tests/Api/Serializer/JsonBodyTest.php
stobrien89: I'll add that I think it makes sense to yield in a loop when dynamically constructing test cases

yenfryherrerafeliz: Got it. The way it is is fine, not a big deal.

However, you can still get the test case failing flagged using the suggested approach. For example, if you do:
$cases = [
    'my_named_case_1' => ['type' => 'string'], ['foo' => 'bar'], '{"foo":"bar"}'],
    'my_named_case_2' => [['type' => 'integer'], ['foo' => 1], '{"foo":1}'],
]
When one of those tests fail you will get the name you use as the index.

stobrien89: Yeah the key distinction there is that this way, you'll actually get a stack trace printout that points to the exact line the case is declared on— saves you from having to count the indices (if it's not a named key)

yenfryherrerafeliz (approved):

PR #3146, opened 2025-07-09
bugfix: aws protocol deserialization updates

stobrien89:
*Description of changes:*
Updates to deserialization for all AWS protocols.

### DIFF
--- a/src/Api/ErrorParser/JsonParserTrait.php
+++ b/src/Api/ErrorParser/JsonParserTrait.php
@@ -12,41 +12,133 @@ trait JsonParserTrait
 {
     use PayloadParserTrait;
 
-    private function genericHandler(ResponseInterface $response)
+    private function genericHandler(ResponseInterface $response): array
     {
         $code = (string) $response->getStatusCode();
+        $error_code = null;
+        $error_type = null;
+
+        // Parse error code and type for query compatible services
         if ($this->api
             && !is_null($this->api->getMetadata('awsQueryCompatible'))
-            && $response->getHeaderLine('x-amzn-query-error')
+            && $response->hasHeader('x-amzn-query-error')
         ) {
-            $queryError = $response->getHeaderLine('x-amzn-query-error');
-            $parts = explode(';', $queryError);
-            if (isset($parts) && count($parts) == 2 && $parts[0] && $parts[1]) {
-                $error_code = $parts[0];
-                $error_type = $parts[1];
+            $awsQueryError = $this->parseAwsQueryCompatibleHeader($response);
+            if ($awsQueryError) {
+                $error_code = $awsQueryError['code'];
+                $error_type = $awsQueryError['type'];
             }
         }
+
+        // Parse error code from X-Amzn-Errortype header
+        if (!$error_code && $response->hasHeader('X-Amzn-Errortype')) {
+            $error_code = $this->extractErrorCode(
+                $response->getHeaderLine('X-Amzn-Errortype')
+            );
+        }
+
+        $parsedBody = null;
+        $body = $response->getBody();
+        if (!$body->isSeekable() || $body->getSize()) {
+            $parsedBody = $this->parseJson((string) $body, $response);
+        }
+
+        // Parse error code from response body
+        if (!$error_code && $parsedBody) {
+            $error_code = $this->parseErrorFromBody($parsedBody);
+        }
+
         if (!isset($error_type)) {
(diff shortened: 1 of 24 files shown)


src/Api/Parser/XmlParser.php
```diff
@@ -162,12 +165,10 @@ private function parse_timestamp(Shape $shape, $value)
...
     {
-        $namespace = $shape['xmlNamespace']['uri']
-            ? $shape['xmlNamespace']['uri']
-            : '';
-        $prefix = $shape['xmlNamespace']['prefix']
-            ? $shape['xmlNamespace']['prefix']
-            : '';
+        // `xmlNamespace` can be modeled as a string.  When modeled as an object,
+        // it is guaranteed to have a `uri` field.
+        $namespace = $shape['xmlNamespace']['uri'] ?? $shape['xmlNamespace'] ?? '';
```
stobrien89: Also a question mark like indicated in #3144.  This field value is modeled as an object with a required field of `uri` and an optional field of `prefix` in all public AWS models.  It's possible the test suites will be updated to match this.

tests/WrappedHttpHandlerTest.php
stobrien89, line 1: The removals in the expected output are for an empty header value.  They should not be deserialized.

src/Api/Parser/XmlParser.php
stobrien89: This change is no longer needed.

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -12,41 +12,141 @@ trait JsonParserTrait
...
+     *
+     * @param ResponseInterface $response
+     * @return array|null Returns ['code' => string, 'type' => string] or null
+     */
+    private function parseAwsQueryCompatibleHeader(ResponseInterface $response): ?array
+    {
+        $queryError = $response->getHeaderLine('x-amzn-query-error');
+        $parts = explode(';', $queryError);
+
+        if (isset($parts) && count($parts) == 2 && $parts[0] && $parts[1]) {
```
yenfryherrerafeliz: nit: I would prefer to use `is_array($parts)` instead of `isset($parts)`.

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -12,41 +12,141 @@ trait JsonParserTrait
...
+     * Parse error code from X-Amzn-Errortype header
+     *
+     * @param ResponseInterface $response
+     * @return string|null
+     */
+    private function parseErrorTypeHeader(ResponseInterface $response): ?string
+    {
+        $error_code = $response->getHeaderLine('X-Amzn-Errortype');
+        return $this->extractErrorCode($error_code);
+    }
```
yenfryherrerafeliz: I don't think creating a method for this is necessary. Also because it is just used from a single place. It is just adding method calls overhead to the workflow. I would rather extract the header value from line 33 and call `extractErrorCode` from there.
```php
  if (!$error_code && $response->hasHeader('X-Amzn-Errortype')) {
      $error_code = $this->extractErrorCode(
          $response->getHeaderLine('X-Amzn-Errortype')
      );
  }
```

stobrien89: Yeah, got a little carried away with helper methods.  This can be done like you suggested with a descriptive comment

src/Api/ErrorParser/JsonParserTrait.php
stobrien89: We actually don't need either of those checks.  This is a nonsensical check that already existed— I just didn't notice it.  The output of `explode()` is always an array

stobrien89: good catch

src/Api/ErrorParser/JsonParserTrait.php
stobrien89: updated

yenfryherrerafeliz (approved):
Looks good to me. Just make sure integ and smoke tests are ran. 

Thanks for working on this.

stobrien89:
Smoke and Integ tests are passing:

```
............

111 scenarios (111 passed)
222 steps (222 passed)
```

```

66 scenarios (66 passed)
248 steps (248 passed)
```

Just a heads-up: I changed some of the "empty body" conditions around.  We should not skip parsing for non-seekable streams where a call to `getSize()` on the body returns null.  Updated all of the paths that make those calls.

PR #3148, opened 2025-07-13
bugfix: standardize rest protocol URI resolution

stobrien89:
*Description of changes:*

Standardizes endpoint URI resolution logic in RestSerializer::buildEndpoint() to address multiple path resolution issues.

### DIFF
--- a/src/Api/Serializer/RestSerializer.php
+++ b/src/Api/Serializer/RestSerializer.php
@@ -17,7 +17,7 @@ use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Uri;
 use GuzzleHttp\Psr7\UriResolver;
 use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UriInterface;
 
 /**
  * Serializes HTTP locations like header, uri, payload, etc...
@@ -25,7 +25,7 @@ use Psr\Http\Message\StreamInterface;
  */
 abstract class RestSerializer
 {
-    use EndpointV2SerializerTrait;
+    private const TEMPLATE_STRING_REGEX = '/\{([^\}]+)\}/';
 
     private static array $excludeContentType = [
         's3' => true,
@@ -41,9 +41,11 @@ abstract class RestSerializer
     /** @var bool */
     private $isUseEndpointV2;
 
+    use EndpointV2SerializerTrait;
+
     /**
-     * @param Service $api      Service API description
-     * @param string  $endpoint Endpoint to connect to
+     * @param Service $api Service API description
+     * @param string $endpoint Endpoint to connect to
      */
     public function __construct(Service $api, $endpoint)
     {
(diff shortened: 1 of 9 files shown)


src/Api/Serializer/RestSerializer.php
```diff
@@ -198,79 +198,143 @@ private function applyQuery($name, Shape $member, $value, array &$opts)
...
+     */
+    private function expandUriTemplate(Operation $operation, array $args): string
     {
-        $serviceName = $this->api->getServiceName();
-        // Create an associative array of variable definitions used in expansions
         $varDefinitions = $this->getVarDefinitions($operation, $args);
 
-        $relative = preg_replace_callback(
+        return preg_replace_callback(
             '/\{([^\}]+)\}/',
```
yenfryherrerafeliz: Please define the regex in a const.

src/Api/Serializer/RestSerializer.php
```diff
@@ -260,79 +261,143 @@ private function applyQuery($name, Shape $member, $value, array &$opts)
...
+    private function shouldPreservePath(string $key): bool
+    {
+        // Keys with dot segments
+        if (str_contains($key, '.')) {
+            $segments = explode('/', $key);
+            foreach ($segments as $segment) {
+                if ($segment === '.' || $segment === '..') {
+                    return true;
+                }
+            }
```
yenfryherrerafeliz, line 340: It looks like previous implementation carried an issue and is that if you specify a key as follow:
'/mykey/../another-folder/mykey.txt'
We get a signature mismatch error, and we should either get a NoSuchKey error if the object does not exists, or get the object successfully.

yenfryherrerafeliz (changes requested):
I like the implementation. Pretty clean and organized. Just two minor comments.
Thank you!

src/Api/Serializer/RestSerializer.php
stobrien89: I dug into this and this is an issue with signature calculation on S3's end.  We are calculating the signature correctly.  S3 appears to normalize keys/request targets like `../`.  This probably doesn't happen for older buckets where keys like that were allowed to exist.

src/Api/Serializer/RestSerializer.php
stobrien89: done

src/Api/Serializer/RestSerializer.php
yenfryherrerafeliz: Maybe. However, I tested the same request with Java and JavaScript SDKs and we get a NoSuchKey error. Not sure why.

stobrien89: > However, I tested the same request with Java and JavaScript SDKs and we get a NoSuchKey error. Not sure why.

and you've confirmed that both of those SDKs don't remove dot segments from the request target?

In any case, we're not as much concerned with reaching the path-like key as we are _not reaching_ other keys

stobrien89: Had a chance to test JS V3 and with keys that contain dot segments, I receive a `400: UnknownError` instead of a `NoSuchKey` error.  They may have made some recent updates.

yenfryherrerafeliz: Tested and with keys as follows: /mykey/../another-folder/mykey.txt does not work. We will handle this in a separated work item since this is not something introduced by this PR.

yenfryherrerafeliz (approved):
Integ and Smoke tests should be ran before merging.

stobrien89:
Added E2E tests for both `rest-xml` (s3) and `rest-json` (geo services).  Both integ and smoke tests are passing:

```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m19.80s (94.57Mb)
```
```

66 scenarios (66 passed)
248 steps (248 passed)
```

PR #3156, opened 2025-07-31
feat: manual auth schemes configuration

yenfryherrerafeliz:
- Add support to configure manual schemes based on user's preference.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/Auth/AuthSelectionMiddleware.php
+++ b/src/Auth/AuthSelectionMiddleware.php
@@ -134,7 +134,8 @@ class AuthSelectionMiddleware
     private function buildAuthSchemeList(
         array $resolvableAuthSchemeList,
         ?array $commandConfiguredAuthSchemes,
-    ): array {
+    ): array
+    {
         $userConfiguredAuthSchemes = $commandConfiguredAuthSchemes
             ?? $this->configuredAuthSchemes;
 
--- a/src/Auth/AuthSelectionMiddleware.php
+++ b/src/Auth/AuthSelectionMiddleware.php
@@ -29,27 +29,29 @@ class AuthSelectionMiddleware
     private $api;
 
     /** @var array|null */
-    private ?array $userPreferredAuthSchemes;
+    private ?array $configuredAuthSchemes;
 
     /**
      * Create a middleware wrapper function
      *
      * @param AuthSchemeResolverInterface $authResolver
      * @param Service $api
+     * @param array|null $configuredAuthSchemes
+     * 
      * @return Closure
      */
     public static function wrap(
         AuthSchemeResolverInterface $authResolver,
         Service $api,
-        ?array $userPreferredAuthSchemes
+        ?array $configuredAuthSchemes
     ): Closure
     {
         return function (callable $handler) use (
             $authResolver,
             $api,
-            $userPreferredAuthSchemes
+            $configuredAuthSchemes
         ) {
-            return new self($handler, $authResolver, $api, $userPreferredAuthSchemes);
+            return new self($handler, $authResolver, $api, $configuredAuthSchemes);
         };
     }
 
@@ -57,19 +59,19 @@ class AuthSelectionMiddleware
      * @param callable $nextHandler
      * @param AuthSchemeResolverInterface $authResolver
      * @param Service $api
-     * @param array|null $userPreferredAuthSchemes
+     * @param array|null $configuredAuthSchemes
      */
     public function __construct(
         callable $nextHandler,
         AuthSchemeResolverInterface $authResolver,
         Service $api,
-        ?array $userPreferredAuthSchemes=null
+        ?array $configuredAuthSchemes = null
     )
     {
         $this->nextHandler = $nextHandler;
         $this->authResolver = $authResolver;
         $this->api = $api;
-        $this->userPreferredAuthSchemes = $userPreferredAuthSchemes;
+        $this->configuredAuthSchemes = $configuredAuthSchemes;
     }
 
     /**
@@ -125,32 +127,32 @@ class AuthSelectionMiddleware
      * followed by remaining available schemes.
      *
      * @param array $resolvableAuthSchemeList Available auth schemes
-     * @param array|null $commandAuthSchemePreference Command-level preferences (overrides config)
+     * @param array|null $commandConfiguredAuthSchemes Command-level preferences (overrides config)
      *
      * @return array Reordered auth schemes with user preferences first
      */
     private function buildAuthSchemeList(
         array $resolvableAuthSchemeList,
-        ?array $commandAuthSchemePreference,
+        ?array $commandConfiguredAuthSchemes,
     ): array {
-        $userPreferences = $commandAuthSchemePreference
-            ?? $this->userPreferredAuthSchemes;
+        $userConfiguredAuthSchemes = $commandConfiguredAuthSchemes
+            ?? $this->configuredAuthSchemes;
 
-        if (empty($userPreferences)) {
+        if (empty($userConfiguredAuthSchemes)) {
             return $resolvableAuthSchemeList;
         }
 
-        $availableSchemes = array_flip($resolvableAuthSchemeList);
-
-        // Get preferred schemes that are actually available
-        $prioritizedSchemes = array_filter(
-            $userPreferences,
-            fn($scheme) => isset($availableSchemes[$scheme])
+        $prioritizedAuthSchemes = array_intersect(
+            $userConfiguredAuthSchemes,
+            $resolvableAuthSchemeList
         );
 
         // Get remaining schemes not in user preferences
-        $remainingSchemes = array_diff($resolvableAuthSchemeList, $prioritizedSchemes);
+        $remainingAuthSchemes = array_diff(
+            $resolvableAuthSchemeList,
+            $prioritizedAuthSchemes
+        );
 
-        return array_merge($prioritizedSchemes, $remainingSchemes);
+        return array_merge($prioritizedAuthSchemes, $remainingAuthSchemes);
     }
 }
--- a/src/Auth/AuthSelectionMiddleware.php
+++ b/src/Auth/AuthSelectionMiddleware.php
@@ -28,6 +28,9 @@ class AuthSelectionMiddleware
     /** @var Service */
     private $api;
 
+    /** @var array|null */
+    private ?array $userPreferredAuthSchemes;
+
     /**
      * Create a middleware wrapper function
      *
@@ -37,29 +40,36 @@ class AuthSelectionMiddleware
      */
     public static function wrap(
         AuthSchemeResolverInterface $authResolver,
-        Service $api
+        Service $api,
+        ?array $userPreferredAuthSchemes
     ): Closure
     {
-        return function (callable $handler) use ($authResolver, $api) {
-            return new self($handler, $authResolver, $api);
+        return function (callable $handler) use (
+            $authResolver,
+            $api,
+            $userPreferredAuthSchemes
+        ) {
+            return new self($handler, $authResolver, $api, $userPreferredAuthSchemes);
         };
     }
 
     /**
      * @param callable $nextHandler
-     * @param $authResolver
-     * @param callable $identityProvider
+     * @param AuthSchemeResolverInterface $authResolver
      * @param Service $api
+     * @param array|null $userPreferredAuthSchemes
      */
     public function __construct(
         callable $nextHandler,
         AuthSchemeResolverInterface $authResolver,
-        Service $api
+        Service $api,
+        ?array $userPreferredAuthSchemes=null
     )
     {
         $this->nextHandler = $nextHandler;
         $this->authResolver = $authResolver;
         $this->api = $api;
+        $this->userPreferredAuthSchemes = $userPreferredAuthSchemes;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Auth/AuthSelectionMiddleware.php
```diff
@@ -28,6 +28,9 @@ class AuthSelectionMiddleware
     /** @var Service */
     private $api;
 
+    /** @var array|null */
+    private ?array $userPreferredAuthSchemes;
```
stobrien89: nit: can we change this to `$configuredAuthSchemes`?

src/Auth/AuthSelectionMiddleware.php
```diff
@@ -37,29 +40,36 @@ class AuthSelectionMiddleware
      */
     public static function wrap(
         AuthSchemeResolverInterface $authResolver,
-        Service $api
+        Service $api,
+        ?array $userPreferredAuthSchemes
```
stobrien89: same comment as above- and everywhere else it's used

src/Auth/AuthSelectionMiddleware.php
```diff
@@ -37,29 +40,36 @@ class AuthSelectionMiddleware
...
+     * @param AuthSchemeResolverInterface $authResolver
      * @param Service $api
+     * @param array|null $userPreferredAuthSchemes
      */
     public function __construct(
         callable $nextHandler,
         AuthSchemeResolverInterface $authResolver,
-        Service $api
+        Service $api,
+        ?array $userPreferredAuthSchemes=null
```
stobrien89: nit: spacing needed for the default value

src/Auth/AuthSelectionMiddleware.php
```diff
@@ -86,21 +96,61 @@ public function __invoke(CommandInterface $command)
...
+     * Prioritizes auth schemes according to user preference order.
+     * User-preferred schemes that are available will be placed first,
+     * followed by remaining available schemes.
+     *
+     * @param array $resolvableAuthSchemeList Available auth schemes
+     * @param array|null $commandAuthSchemePreference Command-level preferences (overrides config)
+     *
+     * @return array Reordered auth schemes with user preferences first
+     */
+    private function buildAuthSchemeList(
```
stobrien89, line 134: Something like this would be simpler:

```php
private function buildAuthSchemeList(
    array $resolvableAuthSchemeList,
    ?array $commandConfiguredAuthSchemes,
): array {
    $configured = $commandAuthSchemePreference ?? $this->configuredAuthSchemes;

    if (empty($configured)) {
        return $resolvableAuthSchemeList;
    }

    $prioritized = array_intersect($configured, $resolvableAuthSchemeList);
    
    $remaining = array_diff($resolvableAuthSchemeList, $prioritized);

    return array_merge($prioritized, $remaining);
}

File: src/ClientResolver.php [hunk 5]
```diff
@@ -219,6 +219,13 @@ class ClientResolver
             'doc'     => 'An instance of Aws\Auth\AuthSchemeResolverInterface which selects a modeled auth scheme and returns a signature version',
             'default' => [__CLASS__, '_default_auth_scheme_resolver'],
         ],
+        'auth_scheme_preference' => [
+            'type'    => 'value',
+            'valid'   => ['string', 'array'],
+            'doc'     => 'A user\'s preferred auth scheme list',
+            'default' => self::DEFAULT_FROM_ENV_INI,
+            'fn' => [__CLASS__, '_apply_auth_scheme_preference'],
```
stobrien89: I would add a `doc` field with a brief description

src/ClientResolver.php
```diff
@@ -219,6 +219,13 @@ class ClientResolver
             'doc'     => 'An instance of Aws\Auth\AuthSchemeResolverInterface which selects a modeled auth scheme and returns a signature version',
             'default' => [__CLASS__, '_default_auth_scheme_resolver'],
         ],
+        'auth_scheme_preference' => [
+            'type'    => 'value',
+            'valid'   => ['string', 'array'],
+            'doc'     => 'A user\'s preferred auth scheme list',
+            'default' => self::DEFAULT_FROM_ENV_INI,
```
stobrien89: not seeing this defined anywhere

tests/ClientResolverTest.php
```diff
@@ -1519,4 +1520,191 @@ public function testResolveFromEnvIniUseAwsSharedFiles(): void
...
+auth_scheme_preference=smithy.api#noAuth, aws.auth#sigv4 
+EOF;
+
+                        file_put_contents($tempConfigFile, $configContent);
+                        putenv("AWS_CONFIG_FILE=$tempConfigFile");
+                    },
+                    'on_completion' => function () use ($tempDir, $tempConfigFile) {
+                        putenv('AWS_CONFIG_FILE=');
+                        unlink($tempConfigFile);
+                        rmdir($tempDir);
```
stobrien89: This is not needed- it will cause failures.  Unlinking is good enough

tests/ClientResolverTest.php
```diff
@@ -1519,4 +1520,191 @@ public function testResolveFromEnvIniUseAwsSharedFiles(): void
...
+            if ($fns !== null && is_callable($fns['on_completion'])) {
+                call_user_func($fns['on_completion']);
+            }
+        }
+    }
+
+    /**
+     * @return Generator
+     */
+    public function resolvesAuthSchemePreferenceProvider(): Generator
```
stobrien89, line 1629: Since we're doing a lot of the same things repeatedly with the `fns`, can we not just create separate arguments for env and ini values and conditionally add those? Similarly, for cleanup, can we do that conditionally in the body of the test? That would make this test a bit more readable

stobrien89 (changes requested):
Off to a good start- had a couple of suggestions

src/ClientResolver.php
stobrien89: Nevermind, I remember this now

tests/ClientResolverTest.php
yenfryherrerafeliz: So you mean we don't need to remove the directory?
Which or why it will cause failures?

yenfryherrerafeliz: I kept using rmdir since I think is important to clean up any resources we use in a test. I am unliking/deleting the only file I am putting in there, so I don't think it should fail.

src/ClientResolver.php
```diff
@@ -219,6 +219,13 @@ class ClientResolver
             'doc'     => 'An instance of Aws\Auth\AuthSchemeResolverInterface which selects a modeled auth scheme and returns a signature version',
             'default' => [__CLASS__, '_default_auth_scheme_resolver'],
         ],
+        'auth_scheme_preference' => [
+            'type'    => 'value',
+            'valid'   => ['string', 'array'],
+            'doc'     => 'Comma-separated list of authentication scheme preferences in priority order. Configure via environment variable `AWS_AUTH_SCHEME_PREFERENCE`, INI config file `auth_scheme_preference`, or client constructor parameter `auth_scheme_preference` (string or array).\nExample: `AWS_AUTH_SCHEME_PREFERENCE=aws.auth#sigv4a,aws.auth#sigv4,smithy.api#httpBearerAuth`',
+            'default' => self::DEFAULT_FROM_ENV_INI,
```
stobrien89: can we move this up the list between `credentials` and `token`? #3161 will need this value to be resolved prior to the `token` step

tests/Auth/AuthSelectionMiddlewareTest.php
```diff
@@ -313,4 +313,181 @@ private function generateTestService($serviceAuth, $operationAuth, $unsignedPayl
...
+        string $expected,
+    ): void
+    {
+        if (!extension_loaded('awscrt') && $expected === 'v4a') {
+            $this->markTestSkipped(
+                "The awscrt extension is required to run this test."
+            );
+        }
+
+        if (isset($fns['build_source']) && is_callable($fns['build_source'])) {
```
stobrien89: Looks like this might be left over from the last review

tests/ClientResolverTest.php
```diff
@@ -1519,4 +1520,199 @@ public function testResolveFromEnvIniUseAwsSharedFiles(): void
             }
         }
     }
+
+    /**
```
stobrien89, line 1524: A couple of things:

- The restoration of `AWS_CONFIG_FILE` sets it to the tmpdir value
- There should be a case that uses both the ini and env to ensure the right order of precedence
- The callbacks are still somewhat difficult to reason with- to see what actually happens, you have to read from the provider, the helper method and the test method.  I think it would be much simpler to restructure it like:

```php
/**
 * @dataProvider resolvesAuthSchemePreferenceProvider
 */
public function testResolvesAuthSchemePreference(
    array $expectedValue,
    array $args = [],
    ?string $envValue = null,
    ?string $iniValue = null
): void {
    // Store original values
    $originalEnv = getenv('AWS_AUTH_SCHEME_PREFERENCE');
    $originalConfigFile = getenv('AWS_CONFIG_FILE');
    $tempConfigFile = null;
    
    try {
        // Set up environment
        if ($envValue) {
            putenv("AWS_AUTH_SCHEME_PREFERENCE=$envValue");
        }
        
        // Set up INI file
        if ($iniValue) {
            $tempConfigFile = tempnam(sys_get_temp_dir(), 'aws_config_');
            file_put_contents($tempConfigFile, "[default]\nauth_scheme_preference=$iniValue");
            putenv("AWS_CONFIG_FILE=$tempConfigFile");
        }
        
     
        $clientResolver = new ClientResolver(ClientResolver::getDefaultArguments());
        $resolvedConfig = $clientResolver->resolve([
            'service' => 's3',
            'region' => 'us-east-1'
        ] + $args, new HandlerList());
        
        $this->assertArrayHasKey('auth_scheme_preference', $resolvedConfig);
        $this->assertEquals($expectedValue, $resolvedConfig['auth_scheme_preference']);
        
    } finally {
        // Cleanup
        if ($originalEnv !== false) {
            putenv("AWS_AUTH_SCHEME_PREFERENCE=$originalEnv");
        } else {
            putenv("AWS_AUTH_SCHEME_PREFERENCE=");
        }
        
        if ($originalConfigFile !== false) {
            putenv("AWS_CONFIG_FILE=$originalConfigFile");
        } else {
            putenv("AWS_CONFIG_FILE=");
        }
        
        if ($tempConfigFile && file_exists($tempConfigFile)) {
            unlink($tempConfigFile);
        }
    }
}

public function resolvesAuthSchemePreferenceProvider(): Generator
{
    $cases = [
        'provided_at_client_construction' => [
            'expected' => ['smithy.api#httpBearerAuth', 'smithy.api#noAuth'],
            'args' => [
                'auth_scheme_preference' => ['smithy.api#httpBearerAuth', 'smithy.api#noAuth']
            ],
            'env' => null,
            'ini' => null
        ],
        'from_env' => [
            'expected' => ['aws.auth#sigv4a', 'smithy.api#noAuth'],
            'args' => [],
            'env' => 'aws.auth#sigv4a,smithy.api#noAuth',
            'ini' => null
        ],
        'from_env_with_spaces' => [
            'expected' => ['aws.auth#sigv4a', 'smithy.api#noAuth'],
            'args' => [],
            'env' => 'aws.auth#sigv4a,      smithy.api#noAuth',
            'ini' => null
        ],
        'from_ini' => [
            'expected' => ['smithy.api#noAuth', 'aws.auth#sigv4'],
            'args' => [],
            'env' => null,
            'ini' => 'smithy.api#noAuth, aws.auth#sigv4'
        ],
        'from_ini_with_spaces' => [
            'expected' => ['smithy.api#noAuth', 'aws.auth#sigv4', 'aws.auth#sigv4a'],
            'args' => [],
            'env' => null,
            'ini' => 'smithy.api#noAuth,   aws.auth#sigv4 , aws.auth#sigv4a'
        ]
    ];
    
    foreach ($cases as $key => $case) {
        yield $key => $case;
    }
}

Comment (stobrien89) [reply tests/ClientResolverTest.php:1524]: A few things might be missing from that example, but that's the overall spirit of it.  Makes it a little easier to add cases if we need additional coverage over time.  I'm thinking we should start using `set_up` and `tear_down` more for capturing/restoring original values and unsetting env vars/temp directories, but that doesn't need to happen in this PR.

**Review by stobrien89 — CHANGES_REQUESTED** [review]
Looking good- had a few more comments and one request that #3161 depends on

File: src/Auth/AuthSelectionMiddleware.php [hunk 12]
```diff
@@ -86,21 +98,61 @@ public function __invoke(CommandInterface $command)
...
+     *
+     * @param array $resolvableAuthSchemeList Available auth schemes
+     * @param array|null $commandConfiguredAuthSchemes Command-level preferences (overrides config)
+     *
+     * @return array Reordered auth schemes with user preferences first
+     */
+    private function buildAuthSchemeList(
+        array $resolvableAuthSchemeList,
+        ?array $commandConfiguredAuthSchemes,
+    ): array {
```
stobrien89: nit: opening brace goes to a newline. I realize my example didn't do that

stobrien89 (changes requested):
Looks ready to ship.  Just had one small nit.  Also, this needs a changelog entry.

tests/ClientResolverTest.php
```diff
@@ -1519,4 +1520,199 @@ public function testResolveFromEnvIniUseAwsSharedFiles(): void
...
+            file_put_contents($tempConfigFile, $configContent);
+            putenv("AWS_CONFIG_FILE=$tempConfigFile");
+
+            $onCompletionCallbacks[] = function () use (
+                $tempConfigFile,
+                $tempDir,
+                $currentEnvAwsConfigFile
+            ) {
+                if ($currentEnvAwsConfigFile !== false) {
+                    putenv("AWS_CONFIG_FILE=$tempConfigFile");
```
stobrien89: This still sets the original config file to the wrong value

stobrien89 (changes requested):

tests/ClientResolverTest.php
yenfryherrerafeliz: Could you please provide more details for why this will set it to the wrong value?
In line 1545 I try to get the current value for it, if not value is set then getenv will return false, which is the why of my condition here, otherwise, if it has a value then I will set it back.

stobrien89: `$tempConfigFile = $tempDir . 'config';`

is what `AWS_CONFIG_FILE` is being set back to.  Should it not be $`currentEnvAwsConfigFile`?

yenfryherrerafeliz: Ooh, that is true. Thank you!

stobrien89 (approved):

yenfryherrerafeliz:
Integ tests:
```

66 scenarios (66 passed)
248 steps (248 passed)
11m19.14s (44.32Mb)
```

Smoke tests:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
4m3.38s (93.63Mb)
```

PR #3161, opened 2025-08-04
feat: bedrock token provider

stobrien89:
*Description of changes:*
Adds a provider to fetch token values from environment.  If found, sets auth scheme preference to prefer `bearer`.  Additional tests will be added once #3156 is approved and merged.

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -34,6 +34,7 @@ use Aws\Retry\ConfigurationProvider as RetryConfigProvider;
 use Aws\Signature\SignatureProvider;
 use Aws\Token\Token;
 use Aws\Token\TokenInterface;
+use Aws\Token\BedrockTokenProvider;
 use Aws\Token\TokenProvider;
 use GuzzleHttp\Promise\PromiseInterface;
 use InvalidArgumentException as IAE;
@@ -711,8 +712,17 @@ class ClientResolver
         }
     }
 
-    public static function _default_token_provider(array $args)
+    public static function _default_token_provider(array &$args)
     {
+        if (($args['config']['signing_name'] ?? '') === 'bedrock') {
+            // Checks for env value, if present, sets auth_scheme_preference
+            // to bearer auth and returns a provider
+            $provider = BedrockTokenProvider::createIfAvailable($args);
+            if (!is_null($provider)) {
+                return $provider;
+            }
+        }
+
         return TokenProvider::defaultProvider($args);
     }
 
@@ -1130,12 +1140,14 @@ class ClientResolver
     }
 
     public static function _apply_auth_scheme_preference(
-        string|array|null $value,
+        string|array|null &$value,
         array &$args
     ): void
     {
         // Not provided user's preference auth scheme list
         if (empty($value)) {
+            $value = null;
+            $args['config']['auth_scheme_preference'] = $value;
             return;
         }
 
--- a/src/Configuration/ConfigurationResolver.php
+++ b/src/Configuration/ConfigurationResolver.php
@@ -68,7 +68,7 @@ class ConfigurationResolver
      *
      * @return null | mixed
      */
-    public static function env($key, $expectedType)
+    public static function env($key, $expectedType = 'string')
     {
         // Use config from environment variables, if available
         $envValue = getenv(self::$envPrefix . strtoupper($key));
@@ -203,6 +203,7 @@ class ConfigurationResolver
         ) {
             $value = intVal($value);
         }
+
         return $value;
     }
 
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -142,8 +142,12 @@ final class Middleware
      *
      * @return callable
      */
-    public static function signer(callable $credProvider, callable $signatureFunction, $tokenProvider = null, $config = [])
-    {
+    public static function signer(
+        callable $credProvider,
+        callable $signatureFunction,
+        $tokenProvider = null,
+        $config = []
+    ) {
         return function (callable $handler) use ($signatureFunction, $credProvider, $tokenProvider, $config) {
             return function (
                 CommandInterface $command,
--- a/src/Token/BedrockTokenProvider.php
+++ b/src/Token/BedrockTokenProvider.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Aws\Token;
+
+use Aws\Configuration\ConfigurationResolver;
+use Aws\Exception\TokenException;
+use GuzzleHttp\Promise;
+
+/**
+ * Token provider for Bedrock that sources bearer tokens from environment variables.
+ */
+class BedrockTokenProvider extends TokenProvider
+{
+    /** @var string used to resolve the AWS_BEARER_TOKEN_BEDROCK env var */
+    public const TOKEN_ENV_KEY = 'bearer_token_bedrock';
+    public const BEARER_AUTH = 'smithy.api#httpBearerAuth';
+
+    /**
+     * Create a default Bedrock token provider that checks for a bearer token
+     * in the AWS_BEARER_TOKEN_BEDROCK environment variable.
+     *
+     * This provider is automatically wrapped in a memoize function that caches
+     * previously provided tokens.
+     *
+     * @param array $config Optional array of token provider options.
+     *
+     * @return callable
+     */
+    public static function defaultProvider(array $config = []): callable
+    {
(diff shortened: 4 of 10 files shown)


src/Configuration/ConfigurationResolver.php
```diff
@@ -68,7 +68,7 @@ public static function resolve(
      *
      * @return null | mixed
      */
-    public static function env($key, $expectedType)
+    public static function env(string $key, string $expectedType = 'string')
```
yenfryherrerafeliz: I think we should avoid adding types to existent methods/functions since this could break customers if for any reason they are extending this functionality.

tests/Token/BedrockTokenProviderTest.php
```diff
@@ -0,0 +1,232 @@
...
+use Aws\Token\Token;
+use Aws\Token\TokenInterface;
+use GuzzleHttp\Promise;
+use Yoast\PHPUnitPolyfills\TestCases\TestCase;
+
+class BedrockTokenProviderTest extends TestCase
+{
+    use UsesServiceTrait;
+
+    private array $originalEnv = [];
```
yenfryherrerafeliz, line 18: `$originalEnv` does not have to be an array. It can be just a string since we are just dealing with a single env variable.

src/Token/BedrockTokenProvider.php
```diff
@@ -0,0 +1,101 @@
...
+use Aws\Configuration\ConfigurationResolver;
+use Aws\Exception\TokenException;
+use GuzzleHttp\Promise;
+
+/**
+ * Token provider for Bedrock that sources bearer tokens from environment variables.
+ */
+class BedrockTokenProvider extends TokenProvider
+{
+    public const TOKEN_ENV_KEY = 'bearer_token_bedrock';
```
yenfryherrerafeliz, line 15: nit: Can we maybe add a comment that reflects the real variable name here would be `AWS_BEARER_TOKEN_BEDROCK`. I know we are using `ConfigurationResolver::env` which already appends the prefix, but using this helper method is not a requirement.

src/Token/BedrockTokenProvider.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Token provider that creates a token from an environment variable.
+     *
+     * @param string $configKey The configuration key that will be transformed
+     *                          to an environment variable name by ConfigurationResolver
+     *
+     * @return callable
+     */
+    public static function env(string $configKey): callable
```
yenfryherrerafeliz, line 49: I think this function should be called differently. Maybe `getTokenValue` or `getTokenValueFromEnv`.

src/Token/TokenProvider.php
```diff
@@ -92,11 +90,11 @@ public static function defaultProvider(array $config = [])
      *
      * @return callable
      */
-    public static function fromToken(TokenInterface $token)
+    public static function fromToken(TokenInterface $token): callable
```
yenfryherrerafeliz: I think we should avoid adding type hints to existent functions. I also made a similar comment above.

src/Token/BedrockTokenProvider.php
```diff
@@ -0,0 +1,101 @@
...
+     * in the AWS_BEARER_TOKEN_BEDROCK environment variable.
+     *
+     * This provider is automatically wrapped in a memoize function that caches
+     * previously provided tokens.
+     *
+     * @param array $config Optional array of token provider options.
+     *
+     * @return callable
+     */
+    public static function defaultProvider(array $config = []): callable
```
yenfryherrerafeliz, line 29: On this implementation, what would be the usage of this function?. I could not see where it is used.

yenfryherrerafeliz (changes requested):
Looks good. Just some minor changes and some questions.

Thanks!

src/Configuration/ConfigurationResolver.php
stobrien89: That's fair.  Will remove

src/Token/BedrockTokenProvider.php
stobrien89: Yep, will do

src/Token/BedrockTokenProvider.php
stobrien89: It's not used anywhere currently, but the Token Provider specification outlines the need for a 'default chain', so like the other token provider, this is the default chain.  Basically this is here for customers that want to use this provider in in-code client configuration.

src/Token/BedrockTokenProvider.php
stobrien89: I see what you mean, but when we implemented the first token provider, we tried to use the same semantics as the `CredentialProvider` class.  I think we should keep that pattern: https://github.com/aws/aws-sdk-php/blob/92c1ad2cb159129bdaaa1eb01fffd52110d4c6f2/src/Credentials/CredentialProvider.php#L289

tests/Token/BedrockTokenProviderTest.php
stobrien89: It doesn't have to be currently, but it will if they add shared config to the scope of this implementation.  Doesn't matter to me though- we can cross that bridge when we get to it.

stobrien89: Turns out I needed to test additional env vars anyway- keeping for now

src/Token/TokenProvider.php
```diff
@@ -92,11 +90,11 @@ public static function defaultProvider(array $config = [])
      *
      * @return callable
      */
-    public static function fromToken(TokenInterface $token)
+    public static function fromToken($token)
```
yenfryherrerafeliz: Don't need to remove type for this parameter on this method.

src/ClientResolver.php
```diff
@@ -1130,12 +1140,14 @@ public static function _default_auth_scheme_resolver(array $args)
...
     public static function _apply_auth_scheme_preference(
-        string|array|null $value,
+        string|array|null &$value,
         array &$args
     ): void
     {
         // Not provided user's preference auth scheme list
         if (empty($value)) {
+            $value = null;
+            $args['config']['auth_scheme_preference'] = $value;
```
yenfryherrerafeliz, line 1150: Why is needed to set this value to null in this config. Also, Is it guaranteed that `auth_scheme_preference` will be a valid property in `$args['config']`?, otherwise this will cause a runtime error.

src/Middleware.php
```diff
@@ -142,8 +142,12 @@ public static function requestBuilder($serializer)
...
      * @return callable
      */
-    public static function signer(callable $credProvider, callable $signatureFunction, $tokenProvider = null, $config = [])
-    {
+    public static function signer(
+        callable $credProvider,
+        callable $signatureFunction,
+        $tokenProvider = null,
+        $config = []
+    ){
```
yenfryherrerafeliz: nit: I think `){` needs a space to be `) {`.

yenfryherrerafeliz (changes requested):
Clean implementation. Just a question and two minor asks.

Thanks!

src/Token/TokenProvider.php
stobrien89: Whoops.  I'll add it back

src/ClientResolver.php
stobrien89: Currently, when no value is found in the ini file or env, it enters this function as an empty string, which leaves the value in `$args` as an empty string.  Passing by reference and changing `$value` to `null` here changes both `$args['auth_scheme_preference']` and (as shown) sets `$args['config']['auth_scheme_preference']` to null.  `$args['config']` is just another arbitrary array, so setting to `null` shouldn't cause any errors.

yenfryherrerafeliz: Got it. Thanks for clarifying However, I am just trying to understand the use case of this change. Is this needed or required somehow for your implementation?

stobrien89: There are two parts:

- Setting to `null` is more appropriate for an empty or unfound value.
- Adding to the `config` property allows us to retrieve it more easily using `getConfig()`.  This is not necessary, but makes testing the result much easier, because a client's `$args` are not accessible after instantiation.

yenfryherrerafeliz: Sounds good.

stobrien89:
Smoke test run:

```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m30.58s (95.24Mb)
```

Integ test run:
```

66 scenarios (66 passed)
248 steps (248 passed)
8m4.51s (48.33Mb)

Issue #3110, opened 2025-04-23
Bedrock Invoke Agent Stream -  Message checksum mismatch.


adrianpaiva1:
### Describe the bug

I randomly get this Exception while streaming responses with the BedrockAgentRuntimeClient. This is doesn't happen consistently, maybe once every 10 times.

 Aws\Api\Parser\Exception\ParserException Message checksum mismatch. at vendor/aws/aws-sdk-php/src/Api/Parser/NonSeekableStreamDecodingEventStreamIterator.php:55

![Image](https://github.com/user-attachments/assets/7cdfe8bb-2af9-49f6-a7a0-739aab4f8def)

Sample code to reproduce the issue
```php
<?php

namespace App\Console\Commands;

use Aws\BedrockAgentRuntime\BedrockAgentRuntimeClient;
use Aws\Result;
use Illuminate\Console\Command;

class Chat extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'chat';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = '';

    public function __construct()
    {
        parent::__construct();
    }

    /**
     * Execute the console command.
     * @throws \Throwable
     */
    public function handle()
    {
        $client = new BedrockAgentRuntimeClient(
            [
                'version' => config('services.bedrock.version'),
                'region' => config('services.bedrock.region'),
                'credentials' => [
                    'key' => config('services.bedrock.key'),
                    'secret' => config('services.bedrock.secret'),
                ],
                'http' => [
                    'stream' => true,
                ],
            ]
        );

        $data = [
            'agentId' => 'IDHERE',
            'agentAliasId' => 'TSTALIASID',
            'sessionId' => uniqid(),
            'inputText' => 'tell me everything about tylenol',
            'memoryId' => uniqid(),
            'streamingConfigurations' => [
                'streamFinalResponse' => true,
            ],
        ];

        $result = $client->invokeAgent($data);

        foreach ($this->processResult($result) as $chunk) {
            echo  $chunk['data']['bytes'];
        }
    }

    public function processResult(Result $result): \Generator
    {
        $base_data = [
            'session_id' => $result['sessionId'],
            'memory_id' => $result['memoryId'],
        ];

        // Process all events in the completion stream
        foreach ($result['completion'] as $event) {
            $data = $base_data;

            if (isset($event['chunk'])) {
                $data['data'] = $event['chunk'];
[...]

PR #3165, opened 2025-08-08
Fix eventstream partial read

wilsones-berkeley:
*Issue #, if available*
#3110 

*Description of changes:*
This pull request fixes a bug in `NonSeekableStreamDecodingEventStreamIterator` where `readAndHashBytes()` assumed that the underlying stream would always return the full number of requested bytes in a single `read(`) call.

In cases where the stream returns fewer bytes (e.g., partial reads from network streams or large payloads), this could result in truncated payloads and CRC mismatches.

The fix updates `readAndHashBytes()` to:

- Continue reading from the stream until the expected number of bytes have been read or the stream reaches EOF.
- Prevent infinite loops on unexpected empty reads.

A new unit test `testReadAndHashBytesHandlesPartialReads` has been added to verify this behavior by simulating a non-seekable stream that returns partial data in each `read()` call.

All tests pass with `make test`.

### DIFF
--- a/tests/Api/Parser/NonSeekableStreamDecodingEventStreamIteratorTest.php
+++ b/tests/Api/Parser/NonSeekableStreamDecodingEventStreamIteratorTest.php
@@ -5,6 +5,7 @@ namespace Aws\Test\Api\Parser;
 use Aws\Api\Parser\NonSeekableStreamDecodingEventStreamIterator;
 use GuzzleHttp\Psr7\NoSeekStream;
 use GuzzleHttp\Psr7\Utils;
+use Psr\Http\Message\StreamInterface;
 use Yoast\PHPUnitPolyfills\TestCases\TestCase;
 
 class NonSeekableStreamDecodingEventStreamIteratorTest extends TestCase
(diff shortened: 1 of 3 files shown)


yenfryherrerafeliz:
Hi @wilsones-berkeley, thank you so much for this contribution. The fix you have proposed is totally valid!
I took the freedom of modifying the test case so we can test this change against events with the same structure we process them.

tests/Api/Parser/NonSeekableStreamDecodingEventStreamIteratorTest.php
```diff
@@ -67,4 +72,118 @@ public function testValidReturnsTrueOnEOF()
...
+            {
+                if ($this->eof()) {
+                    return '';
+                }
+
+                // Read only up to maxBytesPerRead
+                $readLength = min($length, $this->maxBytesPerRead);
+                return $this->stream->read($readLength);
+            }
+        };
```
stobrien89: This should be 

```suggestion
        $partialReadStream = $this->createMock(StreamInterface::class);
        $partialReadStream->method('isSeekable')->willReturn(false);
        $partialReadStream->method('isReadable')->willReturn(true);
        $partialReadStream->method('read')
            ->willReturnCallback(function ($length) use ($eventStream) {
                if ($eventStream->eof()) {
                    return '';
                }

                $readLength = min($length, 20);
                return $eventStream->read($readLength);
            });
```

yenfryherrerafeliz:
Running integ tets:
```

66 scenarios (66 passed)
248 steps (248 passed)
10m51.03s (49.07Mb)
```

Running smoke tests:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m35.39s (95.98Mb)
```

PR #3174, opened 2025-08-21
bugfix: refresh credentials that are about to expire

### LINKED ISSUE #1714: Memoizing and caching CredentialsProviders return credentials that are just about to expire
My PHP application uses aws-sdk-php v. 3.61.5 to publish messages to an sns topic. The application is running on an ec2 instance with an instance profile that allows it to publish messages to the sns topic.

The application is supposed to publish around 600,000 messages to the SNS topic as fast as it can. Everything works well for the first several hundreds of thousands messages. After several hours, however, it fails with this message: 
```
Aws\Sns\Exception\SnsException

Error executing "Publish" on "https://sns.eu-west-1.amazonaws.com"; AWS HTTP error: Client error: `POST https://sns.eu-west-1.amazonaws.com` resulted in a `403 Forbidden` response:
<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <Error>
    <Type>Sender</Type>
    <Code>ExpiredToke (truncated...)
 ExpiredToken (client): The security token included in the request is expired - <ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <Error>
    <Type>Sender</Type>
    <Code>ExpiredToken</Code>
    <Message>The security token included in the request is expired</Message>
  </Error>
  <RequestId>e6c59e4f-b998-5d99-b926-cafe6ac2711a</RequestId>
</ErrorResponse>

```
So clearly the security token expired. 

I am using the following code to create the sns client: 

```
        $provider = CredentialProvider::defaultProvider();
        $provider = CredentialProvider::cache(
            $provider,
            new DoctrineCacheAdapter(new FilesystemCache('/tmp/cache'))
        );

        $this->client = new SnsClient([
            'region'      => 'eu-west-1',
            'version'     => 'latest',
            'credentials' => $provider
        ]);

```
The caching CredentialsProvider will ask the underlying (default) CredentialsProvider for credentials if the currently cached credentials are expired. And the default CredentialsProvider will refresh the credentials if they are expired. 

The problem, as I see it, is that both the caching CredentialsProvider and the default CredentialsProvider continue to return the existing credentials up to the very last millisecond where they are not expired. This means that just before they expire, the sns client will fetch credentials and use them to call the sns service. By the time the request reaches the sns service, the credentials have expired, and the service consequently rejects the request.

According to AWS docs (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials) refreshed credentials are available from the instance metadata at least five minutes before they expire, so perhaps it would be better for the CredentialsProvider implementations to refresh the token when it expires in less than five minutes?

If I am not using the sdk correctly, some guidance would be highly appreciated :-)

stobrien89:
*Issue #, if available:*
Fixes #1714 

*Description of changes:*
Updates `CredentialProvider::memoize()` to attempt refreshes if credentials will expire within 5 minutes.

### DIFF
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -52,6 +52,7 @@ class CredentialProvider
     const ENV_SESSION = 'AWS_SESSION_TOKEN';
     const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
     const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
+    public const REFRESH_WINDOW = 60;
 
     /**
      * Create a default credential provider that
@@ -224,10 +225,14 @@ class CredentialProvider
                         return $creds;
                     }
 
-                    // Refresh expired credentials.
-                    if (!$creds->isExpired()) {
+                    // Check if credentials are expired or will expire in 1 minute
+                    $needsRefresh = $creds->getExpiration() - time() <= self::REFRESH_WINDOW;
+
+                    // Refresh if expired or expiring soon
+                    if (!$needsRefresh && !$creds->isExpired()) {
                         return $creds;
                     }
+
                     // Refresh the result and forward the promise.
                     return $result = $provider($creds);
                 })
--- a/.changes/nextrelease/bugfix-refresh-credentials.json
+++ b/.changes/nextrelease/bugfix-refresh-credentials.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "bugfix",
+    "category": "Credentials",
+    "description": "Updates `CredentialProvider::memoize()` to refresh credentials that within 1 minute of expiration."
+  }
+]
--- a/tests/Credentials/CredentialProviderTest.php
+++ b/tests/Credentials/CredentialProviderTest.php
@@ -1915,6 +1915,97 @@ EOT;
         $this->assertSame(2, $called);
     }
 
+    public function testMemoizeRefreshesWhenExpiringWithinThreshold(): void
+    {
+        $called = 0;
+        // First credentials expire in 30 seconds (under 60 second threshold)
+        $creds1 = new Credentials('foo', 'bar', null, time() + 30);
+        // Refreshed credentials expire in 1 hour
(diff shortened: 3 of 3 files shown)


src/Credentials/CredentialProvider.php
```diff
@@ -224,10 +224,14 @@ public static function memoize(callable $provider)
                         return $creds;
                     }
 
-                    // Refresh expired credentials.
-                    if (!$creds->isExpired()) {
+                    // Check if credentials are expired or will expire within 5 minutes
+                    $expiresWithinThreshold = $creds->getExpiration() - time() < 300;
```
yenfryherrerafeliz: nit: I would prefer to have the time window `300` declared as a `const`. 
Also, I am curious to know if this shouldn't be optional?, or somehow configurable?

yenfryherrerafeliz (changes requested):
Thanks for working on this. Just a minor ask and a question.

src/Credentials/CredentialProvider.php
stobrien89: I agree that it should be a const.  I also had the same thought (at least about configuration) and mentioned that in the issue.  I think there's a good argument for making this optional/opt-in as well.

src/Credentials/CredentialProvider.php
```diff
@@ -224,10 +225,14 @@ public static function memoize(callable $provider)
                         return $creds;
                     }
 
-                    // Refresh expired credentials.
-                    if (!$creds->isExpired()) {
+                    // Check if credentials are expired or will expire in 1 minute
+                    $expiresThreshold = $creds->getExpiration() - time() <= self::REFRESH_WINDOW;
```
yenfryherrerafeliz: nit: Can you please name this something like `$isWithinExpiresThreshold`?

yenfryherrerafeliz (changes requested):

stobrien89:
Smoke and Integ test runs:

```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m29.49s (95.41Mb)

66 scenarios (66 passed)
248 steps (248 passed)
8m57.59s (48.41Mb)

Issue #2941, opened 2024-06-11
Missing dependency for symfony/filesystem


kkmuffme:
### Describe the bug

https://github.com/aws/aws-sdk-php/blob/master/composer.json

https://github.com/aws/aws-sdk-php/tree/master/src/Script/Composer#removing-unused-services
uses symfony filesystem under the hood though - therefore currently this leads to a fatal error:
https://github.com/aws/aws-sdk-php/blob/master/src/Script/Composer/Composer.php#L24 (https://github.com/aws/aws-sdk-php/blob/master/src/Script/Composer/Composer.php#L6)

### Expected Behavior

no error

### Current Behavior

Fatal error

### Reproduction Steps

https://github.com/aws/aws-sdk-php/tree/master/src/Script/Composer#removing-unused-services

### Possible Solution

As a temporary workaround one can install symfony/filesystem as dependency manually

As a permanent solution this package should be added to the SDK dependencies (require, not require-dev)

### SDK version used

3.312.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

8.3.x - error is version independent

yenfryherrerafeliz:
Hi @kkmuffme, this has been fixed in [this](https://github.com/aws/aws-sdk-php/pull/3179) PR.

Please feel free to reach out again for anything that we may be able to help with.

Thank you!

PR #3179, opened 2025-08-28
chore: add 8.5 support

stobrien89:
Issues: #2941, #3217

*Description of changes:*
Fixes PHP 8.5 deprecation warnings and adds 8.5 to test runs, adds Windows unit test workflow, adds JIT to applicable workflows, adds separate Codecov coverage workflow that runs weekly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,15 +25,15 @@
         "ext-json": "*",
         "ext-simplexml": "*",
         "aws/aws-crt-php": "^1.2.3",
-        "psr/http-message": "^1.0 || ^2.0"
+        "psr/http-message": "^1.0 || ^2.0",
+        "symfony/filesystem": "^v6.4.3 || ^v7.1.0"
     },
     "require-dev": {
         "composer/composer" : "^2.7.8",
         "ext-openssl": "*",
         "ext-dom": "*",
-        "ext-pcntl": "*",
         "ext-sockets": "*",
-        "phpunit/phpunit": "^5.6.3 || ^8.5 || ^9.5",
+        "phpunit/phpunit": "^9.6",
         "behat/behat": "~3.0",
         "doctrine/cache": "~1.4",
         "aws/aws-php-sns-message-validator": "~1.0",
@@ -41,7 +41,6 @@
         "psr/cache": "^2.0 || ^3.0",
         "psr/simple-cache": "^2.0 || ^3.0",
         "sebastian/comparator": "^1.2.3 || ^4.0 || ^5.0",
-        "symfony/filesystem": "^v6.4.0 || ^v7.1.0",
         "yoast/phpunit-polyfills": "^2.0",
         "dms/phpunit-arraysubset-asserts": "^0.4.0"
     },
@@ -49,6 +48,7 @@
         "ext-openssl": "Allows working with CloudFront private distributions and verifying received SNS messages",
         "ext-curl": "To send requests using cURL",
         "ext-sockets": "To use client-side monitoring",
+        "ext-pcntl": "To use client-side monitoring",
         "doctrine/cache": "To use the DoctrineCacheAdapter",
         "aws/aws-php-sns-message-validator": "To validate incoming SNS notifications"
     },
--- a/src/AwsClient.php
+++ b/src/AwsClient.php
@@ -568,8 +568,14 @@ class AwsClient implements AwsClientInterface
             $aliases = \Aws\load_compiled_json($file);
             $serviceId = $this->api->getServiceId();
             $version = $this->getApi()->getApiVersion();
-            if (!empty($aliases['operations'][$serviceId][$version])) {
-                $this->aliases = array_flip($aliases['operations'][$serviceId][$version]);
+            $serviceAliases = null;
+
+            if (!is_null($serviceId) && isset($aliases['operations'][$serviceId])) {
+                $serviceAliases = $aliases['operations'][$serviceId];
+            }
+
+            if ($serviceAliases && isset($serviceAliases[$version])) {
+                $this->aliases = array_flip($serviceAliases[$version]);
             }
         }
     }
--- a/src/LruArrayCache.php
+++ b/src/LruArrayCache.php
@@ -29,6 +29,7 @@ class LruArrayCache implements CacheInterface, \Countable
 
     public function get($key)
     {
+        $key = $key ?? '';
         if (!isset($this->items[$key])) {
             return null;
         }
@@ -49,6 +50,7 @@ class LruArrayCache implements CacheInterface, \Countable
 
     public function set($key, $value, $ttl = 0)
     {
+        $key = $key ?? '';
         // Only call time() if the TTL is not 0/false/null
         $ttl = $ttl ? time() + $ttl : 0;
         $this->items[$key] = [$value, $ttl];
@@ -69,6 +71,7 @@ class LruArrayCache implements CacheInterface, \Countable
 
     public function remove($key)
     {
+        $key = $key ?? '';
         unset($this->items[$key]);
     }
 
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -163,11 +163,13 @@ class RetryMiddleware
             return true;
         }
 
-        if (isset($errorCodes[$error->getAwsErrorCode()])) {
+        $awsCode = $error->getAwsErrorCode();
+        if (!is_null($awsCode) && isset($errorCodes[$awsCode])) {
             return true;
         }
 
-        if (isset($statusCodes[$error->getStatusCode()])) {
+        $status = $error->getStatusCode();
+        if (!is_null($status) && isset($statusCodes[$status])) {
             return true;
         }
 
--- a/src/RetryMiddlewareV2.php
+++ b/src/RetryMiddlewareV2.php
@@ -308,11 +308,13 @@ class RetryMiddlewareV2
             return true;
         }
 
-        if (!empty($errorCodes[$result->getAwsErrorCode()])) {
+        $awsCode = $result->getAwsErrorCode();
+        if (!is_null($awsCode) && isset($errorCodes[$awsCode])) {
             return true;
         }
 
-        if (!empty($statusCodes[$result->getStatusCode()])) {
+        $status = $result->getStatusCode();
+        if (!is_null($status) && isset($statusCodes[$status])) {
             return true;
         }
 
--- a/src/UserAgentMiddleware.php
+++ b/src/UserAgentMiddleware.php
@@ -172,7 +172,11 @@ class UserAgentMiddleware
         if (function_exists('php_uname')
             && !in_array('php_uname', $disabledFunctions, true)
         ) {
-            $osName = "OS/" . php_uname('s') . '#' . php_uname('r');
+            // Replace spaces with underscores to prevent breaking the user agent format
+            $os = str_replace(' ', '_', php_uname('s'));
+            $release = php_uname('r');
+            $osName = "OS/{$os}#{$release}";
+
             if (!empty($osName)) {
                 return $osName;
             }
--- a/tests/Credentials/CredentialProviderTest.php
+++ b/tests/Credentials/CredentialProviderTest.php
@@ -451,10 +451,18 @@ EOT;
     public function testCreatesFromProcessCredentialProvider(): void
     {
         $awsDir = $this->createAwsHome();
-        $ini = <<<EOT
+        if (PHP_OS_FAMILY === 'Windows') {
+            $ini = <<<EOT
 [foo]
-credential_process = echo '{"AccessKeyId":"foo","SecretAccessKey":"bar", "Version":1}'
+credential_process = echo {"AccessKeyId":"foo","SecretAccessKey":"bar","Version":1}
 EOT;
+        } else {
+            $ini = <<<EOT
+[foo]
+credential_process = echo '{"AccessKeyId":"foo","SecretAccessKey":"bar","Version":1}'
+EOT;
+        }
+
         file_put_contents($awsDir . '/credentials', $ini);
 
         $creds = call_user_func(CredentialProvider::process('foo'))->wait();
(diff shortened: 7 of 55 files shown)


composer.json
```diff
@@ -41,14 +41,14 @@
...
         "sebastian/comparator": "^1.2.3 || ^4.0 || ^5.0",
-        "symfony/filesystem": "^v6.4.0 || ^v7.1.0",
         "yoast/phpunit-polyfills": "^2.0",
         "dms/phpunit-arraysubset-asserts": "^0.4.0"
     },
     "suggest": {
         "ext-openssl": "Allows working with CloudFront private distributions and verifying received SNS messages",
         "ext-curl": "To send requests using cURL",
         "ext-sockets": "To use client-side monitoring",
+        "ext-pcntl": "*",
```
GrahamCampbell: Incorrect syntax. The value is meant to be a reason to the user why the package is needed.

stobrien89: Yes- thank you for the reminder.  I just copy/pasted into `suggest` while I was working on the Windows workflows.  I had intended to change this, but probably would have forgotten.

src/AwsClient.php
```diff
@@ -568,8 +568,11 @@ private function loadAliases($file = null)
             $aliases = \Aws\load_compiled_json($file);
             $serviceId = $this->api->getServiceId();
             $version = $this->getApi()->getApiVersion();
-            if (!empty($aliases['operations'][$serviceId][$version])) {
-                $this->aliases = array_flip($aliases['operations'][$serviceId][$version]);
+            $serviceAliases = isset($aliases['operations'][$serviceId ?? ''])
```
yenfryherrerafeliz: I don`t think the fallback here is necessary.
> $aliases['operations'][$serviceId ?? ''] could just be $aliases['operations'][$serviceId]

src/AwsClient.php
```diff
@@ -568,8 +568,11 @@ private function loadAliases($file = null)
...
+                ? $aliases['operations'][$serviceId]
+                : null;
+            if ($serviceAliases && isset($serviceAliases[$version])) {
+                $this->aliases = array_flip($serviceAliases[$version]);
             }
```
yenfryherrerafeliz: I think this could be simplified to:
```php
$serviceAliases = $aliases['operations'][$serviceId ?? ''] ?? [];
$this->aliases = array_flip($serviceAliases[$version] ?? []); // It look it is guarantee that $version will be not null
```

I think it is more readable.

src/LruArrayCache.php
```diff
@@ -29,6 +29,7 @@ public function __construct($maxItems = 1000)
 
     public function get($key)
     {
+        $key = $key ?? '';
         if (!isset($this->items[$key])) {
```
yenfryherrerafeliz, line 33: I prefer the idea of evaluating if the key is null and not fallback to make the key an empty string. Why?, because an empty string is also a valid array key and this could cause unexpected behaviors. Also, when setting the key you also fallback to make it empty.

yenfryherrerafeliz: For example:
```php
if ($key == null || !isset($this->items[$key])) {
    return null;
}

```

src/LruArrayCache.php
```diff
@@ -49,6 +50,7 @@ public function get($key)
 
     public function set($key, $value, $ttl = 0)
     {
+        $key = $key ?? '';
```
yenfryherrerafeliz, line 53: The same here.  I don`t like the idea of making the key an empty string. I think we could either fail because the key is null or do an early return.

src/RetryMiddleware.php
```diff
@@ -177,7 +177,8 @@ private static function isRetryable(
         ) {
             if (method_exists($previous, 'getHandlerContext')) {
                 $context = $previous->getHandlerContext();
-                return !empty($context['errno'])
+                return !empty($context)
+                    && !empty($context['errno'])
```
yenfryherrerafeliz: I think the extra condition could be redundant?, unless `$context` may be null?

src/RetryMiddlewareV2.php
```diff
@@ -322,7 +322,8 @@ private static function isRetryable(
         ) {
             if (method_exists($previous, 'getHandlerContext')) {
                 $context = $previous->getHandlerContext();
-                return !empty($context['errno'])
+                return !empty($context)
+                    && !empty($context['errno'])
```
yenfryherrerafeliz: Same comment as [here](https://github.com/aws/aws-sdk-php/pull/3179/files#r2571993749).

src/UserAgentMiddleware.php
```diff
@@ -172,7 +172,11 @@ private function getOsName(): string
         if (function_exists('php_uname')
             && !in_array('php_uname', $disabledFunctions, true)
         ) {
-            $osName = "OS/" . php_uname('s') . '#' . php_uname('r');
+            // Replace spaces with underscores to prevent breaking the user agent format
+            $os = str_replace(' ', '_', php_uname('s'));
```
yenfryherrerafeliz, line 176: Thank you!

tests/Credentials/CredentialProviderTest.php
```diff
@@ -512,10 +544,20 @@ public function testCreatesFromProcessCredentialWithFilenameParameterOverSharedF
...
     {
         $awsDir = $this->createAwsHome();
-        $ini = <<<EOT
+        if (PHP_OS_FAMILY === 'Windows') {
+            $ini = <<<EOT
+[baz]
+credential_process = echo {"AccessKeyId":"foo","SecretAccessKey":"bar","Version":1}
+EOT;
+        file_put_contents($awsDir . '/mycreds', $ini);
+        putenv("AWS_SHARED_CREDENTIALS_FILE={$awsDir}/mycreds");
```
yenfryherrerafeliz: I guess this is redundant. The same thing happens after line 561, unless I am missing something?

tests/S3/TransferTest.php
```diff
@@ -307,6 +334,10 @@ public function testDebugFalse()
 
     public function testDownloadsObjectsWithAccessPointArn()
     {
+        if (PHP_OS_FAMILY === 'Windows') {
+            $this->markTestSkipped('S3 access point ARN downloads have path handling issues on Windows');
```
yenfryherrerafeliz, line 338: What does this mean?, customers can't do downloads using access points ARNs on Windows then?

yenfryherrerafeliz (changes requested):
I think falling back to empty string in every array access could be problematic, but we can leave this open for discussion.

src/AwsClient.php
stobrien89: Agree that it's more readable, but it breaks one of our tests on the behavior.  I'll split into two separate if statements for readability

src/LruArrayCache.php
stobrien89: I don't particularly like falling back to an empty string either, but it's official guidance from the deprecation guide: 

> Using null as an array offset[ ¶](https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.core.using-null-as-an-array-offset)
Using [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) as an array offset or when calling [array_key_exists()](https://www.php.net/manual/en/function.array-key-exists.php) is now deprecated. Instead an empty string should be used.

I'll try to come up with a better way around this

tests/Credentials/CredentialProviderTest.php
stobrien89: slightly different format: for Linux, it's wrapped in an outer string

tests/S3/TransferTest.php
stobrien89: possibly.  Will need to add to the backlog to investigate.  I tried for a while to get this working

src/LruArrayCache.php
stobrien89: dug into this a bit further- avoided in areas where checks could be performed in another way, but in this class and `MultiRegionClient`, we break backward-compatibility if we do anything else.  From the [PHP Manual](https://www.php.net/manual/en/language.types.array.php):

>[Null](https://www.php.net/manual/en/language.types.null.php) will be cast to the empty string, i.e. the key null will actually be stored under "".

In those classes we were setting/getting `''` keys and that's behavior we should retain.

src/RetryMiddleware.php
stobrien89: Removed in the latest update

src/RetryMiddleware.php
```diff
@@ -163,11 +163,13 @@ private static function isRetryable(
             return true;
         }
 
-        if (isset($errorCodes[$error->getAwsErrorCode()])) {
+        $awsCode = $error->getAwsErrorCode();
+        if (!(is_null($awsCode)) && isset($errorCodes[$awsCode])) {
```
yenfryherrerafeliz: nit: I think the parenthesis are not necessary for the is_null evaluation.

src/RetryMiddlewareV2.php
```diff
@@ -308,11 +308,13 @@ private static function isRetryable(
             return true;
         }
 
-        if (!empty($errorCodes[$result->getAwsErrorCode()])) {
+        $awsCode = $result->getAwsErrorCode();
+        if (!(is_null($awsCode)) && isset($errorCodes[$awsCode])) {
```
yenfryherrerafeliz: nit: parenthesis not necessary.

tests/Credentials/CredentialProviderTest.php
yenfryherrerafeliz: I maybe don't understand. The same code as follow:
```
file_put_contents($awsDir . '/mycreds', $ini);
putenv("AWS_SHARED_CREDENTIALS_FILE={$awsDir}/mycreds");
```
Is repeated in line 561:
```
file_put_contents($awsDir . '/mycreds', $ini);
putenv("AWS_SHARED_CREDENTIALS_FILE={$awsDir}/mycreds");
```

Unless I am missing something?

tests/FunctionsTest.php
```diff
@@ -198,7 +198,7 @@ public function testDescribeArray()
      */
     public function testDescribeDoubleToFloat()
     {
-        $double = (double)1.3;
+        $double = (float) 1.3;
```
yenfryherrerafeliz: nit: The casting here is not necessary. Neither float or double.

tests/MiddlewareTest.php
```diff
@@ -350,16 +350,18 @@ public function testCanTimeSuccessfulHandlers()
...
+            usleep(1000);
             return Promise\Create::promiseFor(new Result);
         });
         $list->prependInit(Middleware::timer());
         $handler = $list->resolve();
         $request = new Request('GET', 'http://exmaple.com');
         $result = $handler(new Command('Foo'), $request)->wait();
         $this->assertArrayHasKey('total_time', $result['@metadata']['transferStats']);
+        // Windows with JIT may report timing differently
+        $minTime = PHP_OS_FAMILY === 'Windows' ? 0.0001 : 0.001;
```
yenfryherrerafeliz, line 362: I don`t understand this. Can you please provide more context on it? The reason I ask is because I want to understand if when customers are using the SDK they will get different metrics based on the runtime they use?, and if so, we somehow need to document this or make the metrics to be similar not matter the runtime.

src/RetryMiddleware.php
stobrien89: I think this was a bad IntelliJ auto-suggestion that was applied and then pasted into the other file without noticing. Removed in latest update

tests/Credentials/CredentialProviderTest.php
stobrien89: I misunderstood your original comment. probably a copy/paste error- removed in latest update

tests/FunctionsTest.php
stobrien89: Double is actually needed if we decide to keep this.  Because we support down to 8.1 and versions prior to 8.5 [support the non-canonical `double` cast](https://www.php.net/manual/en/migration85.deprecated.php), I'll need to re-work this to be skipped on 8.5 and up.  The float cast was pretty careless of me to tdo

tests/MiddlewareTest.php
stobrien89: Windows system timers work differently.  `Middleware::timer()` uses `microtime()` and because Windows does not always have high-resolution times enabled, `microtime(true)` (as float) may return smaller deltas than the sleep duration because the sleep duration doesn't always round up the same way.

tests/FunctionsTest.php
yenfryherrerafeliz: I think keeping the cast does not matter much, however, there is not difference in having the assignment there with the cast and without the casting. double and float are just aliases for the same data type.
For example:
```
$value = 1.3;
echo gettype($value); // double

$value = (double)1.3;
echo gettype($value); // double

$value = (float)1.3;
echo gettype($value); // double
```

tests/MiddlewareTest.php
yenfryherrerafeliz: Got it, but based on your answer, there could be times where Windows have high-resolution times enabled then?, but we are assuming it will always be disabled. How do we know when is enabled or disabled?

maciejlew:
Since that change I cannot upgrade:

```
  Problem 1
    - Root composer.json requires aws/aws-sdk-php ^3.368 -> satisfiable by aws/aws-sdk-php[3.368.0, 3.368.1, 3.368.2, 3.369.0].
    - aws/aws-sdk-php[3.368.0, ..., 3.369.0] require symfony/filesystem ^v6.4.3 || ^v7.1.0 || ^v8.0.0 -> found symfony/filesystem[v6.4.3, ..., 6.4.x-dev, v7.1.0-BETA1, ..., 7.4.x-dev, v8.0.0-BETA1, ..., 8.1.x-dev] but it conflicts with your root composer.json require (^5.4).
```

GrahamCampbell:
> Since that change I cannot upgrade

Because you are using an EOL version of Symfony. It's necessary to upgrade to Symfony 6.4.3 or higher to receive the security fixes.

stobrien89:
Hi @maciejlew,

[We decided to allow `filesystem` version 5.4](https://github.com/aws/aws-sdk-php/pull/3237) (latest) since it's in extended LTS.  This is available in the latest SDK version.

PR #3184, opened 2025-09-10
enhancement: add token source metric

stobrien89:
*Description of changes:*
Adds bedrock env metric to user agent and adds test to cover behavior.  Refactors a tangentially related test for brevity/readability.

### DIFF
--- a/tests/Token/BedrockTokenProviderTest.php
+++ b/tests/Token/BedrockTokenProviderTest.php
@@ -3,11 +3,14 @@
 namespace Aws\Tests\Token;
 
 use Aws\Exception\TokenException;
+use Aws\MetricsBuilder;
+use Aws\Result;
 use Aws\Test\UsesServiceTrait;
 use Aws\Token\BedrockTokenProvider;
 use Aws\Token\Token;
 use Aws\Token\TokenInterface;
 use Aws\Token\TokenProvider;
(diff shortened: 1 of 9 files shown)


tests/Token/BedrockTokenProviderTest.php
```diff
@@ -288,6 +291,15 @@ public function testIntegrationWithClient(
...
             $this->assertEquals($expectations['token'], $token->getToken());
+
+            if (isset($envVars['AWS_BEARER_TOKEN_BEDROCK'])
+                && !isset($clientArgs['token'])
+            ) {
+                $this->assertEquals(
+                    TokenSource::BEARER_SERVICE_ENV_VARS->value,
+                    $token->getSource()
+                );
+            }
```
SamRemis, line 304: It would be good to see an else statement here that asserts the opposite.  

This could silently be skipping the assertion if for example there was a typo in a test which sets the $envVars variable and `AWS_BEARE_TOKEN_BEDROCK` is set instead of the correct one.  

Non-blocking because that seems like an unlikely scenario, but the Python team has definitely distilled in me that we should avoid if statements in tests whenever possible, especially without else statements.

stobrien89: Good callout.  I'll get this added ASAP

PR #3189, opened 2025-09-11
Revert psr/http-message version constraint

davidsickmiller:
*Description of changes:*

Hi! I'd like to kindly request that the psr/http-message version requirement be relaxed to allow v1.0 for a while longer.

The psr/http-message package is very widely used, and while the latest version of most packages have been updated to use 2.0, not all of them have.

My understanding is that the only thing added in psr/http-message v2 is typing, so once again allowing the old version shouldn't introduce any bugs.

One package in particular that hasn't been updated -- the latest version of [riverline/multipart-parser](https://packagist.org/packages/riverline/multipart-parser) still requires psr/http-message: ^1.0.  This is used by the latest version of [bref](https://packagist.org/packages/bref/bref).

This isn't the only dependency issue I'm facing for psr/http-message 2.0, and it's preventing me from upgrading to the latest version of the AWS SDK for PHP.

---

### DIFF
--- a/.changes/nextrelease/psr-http-message-1.json
+++ b/.changes/nextrelease/psr-http-message-1.json
@@ -1,0 +1,7 @@
+[
+  {
+    "type": "enhancement",
+    "category": "",
+    "description": "Once again allow psr/http-message 1.0, for increased compatibility with other packages."
+  }
+]


stobrien89:
I have no problem with this change.  I vaguely remember running into some dependency conflicts when running our `--lowest` workflow runs, but they can most likely be worked around.

stobrien89:
LGTM.  I guess I was thinking of something else when looking back on our < 8.1 deprecation.  Only thing this needs is a changelog entry, for which you'll need to create the `nextrelease` dir in the `.changes` directory and create a .json file of any name.  There you'll just format like:

```json
[
    {
        "type": "enhancement",
        "category": "",
        "description": "<description>"
    }
]
```
The category is intentionally left blank as this is a change to dependencies, so it will default to `Aws`.

davidsickmiller:
@stobrien89 Changelog entry added.  Let me know if I haven't got it right.

stobrien89:
Thanks for the contribution @davidsickmiller!

GrahamCampbell:
This is not true. That package only has that in the require-dev dependencies. I am a Guzzle maintainer and Bref user and can confirm this is the case. Guzzle's PSR-7 v1 package only supports PHP 8.1 and not 8.2-8.4. I'd strongly recommend reverting this PR.

davidsickmiller:
> This is not true. That package only has that in the require-dev dependencies. I am a Guzzle maintainer and Bref user and can confirm this is the case. Guzzle's PSR-7 v1 package only supports PHP 8.1 and not 8.2-8.4. I'd strongly recommend reverting this PR.

It looks like you're right about riverline/multipart-parser, and I was wrong.

But the reality remains that I'm working on a large, 10-year old PHP application, and there are many dependencies I'll need to upgrade before I can get access to the latest features and bug fixes from the AWS PHP SDK. Most of these dependencies are easy to upgrade, but a few are not.

Is there any harm that comes from letting folks use the AWS SDK with the old version of php-message?

GrahamCampbell:
Yes, there are subtle bugs that occur on newer PHP versions that may not result in crashes, but may result in incorrect behaviour. It is also no longer in security maintenance.

GrahamCampbell:
Ah, I misread this PR as changing the guzzle psr7 version constraint, hence my comments earlier. That's not what this is.

PR #3190, opened 2025-09-13
Revert "Revert psr/http-message version constraint"

GrahamCampbell:
Reverts aws/aws-sdk-php#3189

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-json": "*",
         "ext-simplexml": "*",
         "aws/aws-crt-php": "^1.2.3",
-        "psr/http-message": "^1.0 || ^2.0"
+        "psr/http-message": "^2.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
It is untrue that `riverline/multipart-parser` requires v1 - that is only a dependency of its test suite. Moreover, Guzzle's PSR-7 v1 package only supports PHP 8.1 and not 8.2-8.4.

I am a Guzzle maintainer and Bref user and can confirm v2 works absolutely fine.

stobrien89:
Thanks @GrahamCampbell, I should have done my due diligence before merging #3189.  Would you mind adding a changelog entry in `.changes/nextrelease`?  Same category and namespace as the previous PR.

GrahamCampbell:
I misread the other change as guzzle psr v1. That's not what this is. This is just the interface package.

Issue #3004, opened 2024-09-26
(short issue description)


gauthierm:
### Describe the issue

The LambdaClient::invokeAsync method has two type definitions, the second of which appears to be incorrect and overrides the first.

As I understand the SDK, `invokeAsync` should return a Promise, but the API docs say it returns a Result. This causes warnings when combined with PHPStan or other type checking tools.

### Links

https://github.com/aws/aws-sdk-php/blob/master/src/Lambda/LambdaClient.php#L80-L81

yenfryherrerafeliz:
Hi @gauthierm, thanks for reporting this. The issue here is that the name of the operation from Lambda conflicts with our method naming convention. For every service operation we declare two methods, one is the synchronous version where we do not alter its name, for example `Invoke`, and the other is the asynchronous one that we suffix it with Async, for example `InvokeAsync`. And this last one is the one that should return a Promise. However, in this particular scenario Lambda has one operation called `Invoke`, for which we declare the two magic methods defined above, and another operation called 'InvokeAsync', for which we do the same. But, the synchronous method declaration for `InvokeAsync` will conflict with the asynchronous method of `Invoke` which will also be `InvokeAsync`.

I will add this to our backlog so I can discuss this further with my team. 

Thanks!

Narkunan:
@yenfryherrerafeliz I'm a PHP developer and would love to work on this issue. We use the AWS SDK in our projects, so I'm familiar with it and excited to contribute.
Could you please assign this issue to me?

stobrien89:
Hi all,

Apologies for the delayed response.  Currently working on a fix, but in the meantime, to work around this, you can use the Lambda client's `getCommand('executeAsync')` and `execute()` or (ironically) `executeAsync()` methods, passing the command into either of those methods.

PR #3195, opened 2025-10-09
bugfix: update operation aliases

### LINKED ISSUE #3194: Several operations return non Aws\Result due to name conflicts with Aws\AwsClient methods (similar to #1936)
### Describe the bug

Some operations on multiple service clients are not returning Aws\Result as expected.
They appear to be shadowed by public methods defined on Aws\AwsClient, same root cause as #1936.

Affected examples I found

- AppSyncClient::getApi() → returns Aws\Api\Service
- AmplifyBackendClient::getToken() → returns GuzzleHttp\Promise\Promise
- IoTDeviceAdvisorClient::getEndpoint() → returns GuzzleHttp\Psr7\Uri
- IotClient::getCommand() → returns Aws\Command
- RedshiftServerlessClient::getCredentials() → returns GuzzleHttp\Promise\FulfilledPromise

All of these seem to dispatch Aws\AwsClient methods instead of invoking the service operation and thus do not yield Aws\Result.

### Regression Issue

### Expected Behavior

Each call above should invoke the service operation and return an Aws\Result (or a promise of it for *Async variants).

### Current Behavior

The calls are resolved to Aws\AwsClient public methods with the same names, returning other types (Aws\Api\Service, GuzzleHttp\Promise\Promise, GuzzleHttp\Psr7\Uri, Aws\Command, GuzzleHttp\Promise\FulfilledPromise).

### Reproduction Steps

```php
<?php

use Aws\AmplifyBackend\AmplifyBackendClient;
use Aws\AppSync\AppSyncClient;
use Aws\Iot\IotClient;
use Aws\IoTDeviceAdvisor\IoTDeviceAdvisorClient;
use Aws\RedshiftServerless\RedshiftServerlessClient;

$result = new AppSyncClient([])->getApi();
var_dump(get_class($result));  // Aws\Api\Service

$result = new AmplifyBackendClient([])->getToken();
var_dump(get_class($result));  // GuzzleHttp\Promise\Promise

$result = new IoTDeviceAdvisorClient([])->getEndpoint();
var_dump(get_class($result));  // GuzzleHttp\Psr7\Uri

$result = new IotClient([])->getCommand(['commandId' => '...']);
// **This does not actually work.**
// PHP Fatal error:  Uncaught TypeError: Cannot access offset of type array in isset or empty in ./vendor/aws/aws-sdk-php/src/AwsClient.php:331
// Stack trace:
// #0 ./bug.php(20): Aws\AwsClient->getCommand(Array)
// #1 {main}
//   thrown in ./vendor/aws/aws-sdk-php/src/AwsClient.php on line 33

$result = new IotClient([])->getCommand('getCardinality');
var_dump(get_class($result));  // Aws\Command

$result = new RedshiftServerlessClient([])->getCredentials();
var_dump(get_class($result));  // GuzzleHttp\Promise\FulfilledPromise
```

### Possible Solution

Aws\AwsClient publicly exposes getApi, getCommand, getCredentials, getEndpoint, getToken, etc. (docs show these and their return types). When a service also has operations named GetApi, GetCommand, GetCredentials, GetEndpoint, or GetToken, the client method is picked instead of the dynamic operation method. This matches the historical ApiGatewayV2Client::getApi incident fixed by introducing an operation alias (GetApiResource) in PR #1944.  ￼

Suggested fixes

Apply the same approach as PR #1944: introduce operation aliases for any service whose operation name collides with Aws\AwsClient public methods (e.g., GetApiResource, GetEndpointOperation, etc.).  ￼

References

- Prior incident: ApiGatewayV2Client->getApi returns Aws\Api\Service (#1936).  ￼
- Fix: ApiGatewayV2 operation alias (PR #1944).  ￼

Thanks for taking a look!

### SDK version used

3.356.33

8.4.13

stobrien89:
*Issue #, if available:*
Closes #3194 

*Description of changes:*
Adds aliases to service client methods that collide with `AwsClient` (base class) methods.


stobrien89:
I believe this would also be the solution for #3004, but the issue is that the `Async` suffix is interpreted as the async variant of `invoke`.  we would need to append a different suffix that makes semantic sense for that operation.

PR #3196, opened 2025-10-14
bugfix: add invokeasync operation alias

### LINKED ISSUE #3004: (short issue description)

stobrien89:
*Issue #, if available:*
Closes #3004 

*Description of changes:*
Adds `InvokeAsynchronous` alias to Lambda's `invokeAsync` operation, which collides with the `invoke` operation's async variant.


yenfryherrerafeliz (changes requested):
Could we maybe rename this to something `invokeFunctionAsync`?. `InvokeAsynchronous` could still be confusing for customers from my point of view.

stobrien89:
> Could we maybe rename this to something invokeFunctionAsync?. InvokeAsynchronous could still be confusing for customers from my point of view.

Unfortunately not- anything with the suffix of `Async` is routed through the client's async path: https://github.com/aws/aws-sdk-php/blob/master/src/AwsClientTrait.php#L69,L72

I think this is close enough to the original method name to make intuitive sense

PR #3198, opened 2025-10-21
bugfix: sts credential provider endpoint mode

### LINKED ISSUE #3197: AssumeRoleWithWebIdentityCredentialProvider defaults to STS regional endpoint in us-east-1
### Describe the bug

After the AWS outage in us-east-1 on Monday October 20, we discovered that we were using the regional STS endpoint in us-east-1 from our application. (I.e. not the global STS service at `sts.amazonaws.com`, but specifically `sts.us-east-1.amazonaws.com`.

We are running a PHP application on Amazon EKS, and are using [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to automatically provide credentials. This integration configures a set of environment variables, including `AWS_STS_REGIONAL_ENDPOINTS=regional` and `AWS_REGION=eu-north-1`:

```
AWS_STS_REGIONAL_ENDPOINTS:   regional
AWS_DEFAULT_REGION:           eu-north-1
AWS_REGION:                   eu-north-1
AWS_ROLE_ARN:                 arn:aws:iam::123456789012:role/SomeRole
AWS_WEB_IDENTITY_TOKEN_FILE:  /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

Our code to initialize the client uses `CredentialProvider::defaultProvider()`:

```php
$cache = new PsrCacheAdapter(new ApcuAdapter());
$provider = CredentialProvider::defaultProvider();
$client = new DynamoDbClient([
    'version' => '2012-08-10',
    'region'  => 'eu-north-1',
    'credentials' => CredentialProvider::cache($provider, $cache),
]);
```

This ends up calling the [STS AssumeRoleWithWebIdentity API](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) through `sts.us-east-1.amazonaws.com` to get credentials.

### Regression Issue

### Expected Behavior

AssumeRoleWithWebIdentityCredentialProvider should either use the regional STS endpoints in the current region, or the global endpoints (which would be automatically mapped to the current region thanks to [recent changes](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_region-endpoints.html#reference_sts_global_endpoint_changes)).

### Current Behavior

AssumeRoleWithWebIdentityCredentialProvider specifically chooses to use `us-east-1` with regional endpoints when no region is specifically configured.

### Reproduction Steps

(This issue is hard to create a good reproducer for, since it is mainly about how the AWS PHP SDK interacts with the environment set up by AWS EKS.)

To reproduce this issue, we need an EKS cluster configured in a different region than `us-east-1`, and a service account in that cluster that is configured with [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

Then run the following PHP script in a pod that uses that service account:

```php
#!/usr/bin/env php
<?php
require_once('vendor/autoload.php');
$client = new \Aws\Sts\StsClient([
    'credentials' => \Aws\Credentials\CredentialProvider::defaultProvider(),
]);
$client->getCallerIdentity([]);
```

We can then use the CloudTrail logs to see the behavior.

In this case, we will get one CloudTrail entry in the us-east-1 region for the `AssumeRoleWithWebIdentity` call. This entry shows that the request was directed at the regional endpoint in us-east-1:

```yaml
{
    # ...
    "eventName": "AssumeRoleWithWebIdentity",
    "awsRegion": "us-east-1",
    # ...
    "additionalEventData": {
        # ...
        "RequestDetails": {
            "awsServingRegion": "us-east-1",
            "endpointType": "regional"
        }
    },
    # ...
    "tlsDetails": {
        # ...
        "clientProvidedHostHeader": "sts.us-east-1.amazonaws.com"
    }
}
```

Meanwhile, we will find the log entry for the `GetCallerIdentity` call in the CloudTrail logs in the region of the EKS cluster:

```yaml
{
    # ...
    "eventName": "GetCallerIdentity",
    "awsRegion": "eu-north-1",
    # ...
    "additionalEventData": {
        # ...
        "RequestDetails": {
            "awsServingRegion": "eu-north-1",
            "endpointType": "regional"
        }
    },
    # ...
    "tlsDetails": {
        # ...
        "clientProvidedHostHeader": "sts.eu-north-1.amazonaws.com"
    }
}
```

### Possible Solution

_No response_

As far as I can tell, the current code that sets the region is the following:

```php
$region = $config['region'] ?? 'us-east-1';
if (isset($config['client'])) {
    $this->client = $config['client'];
} else {
    $this->client = new StsClient([
        'credentials' => false,
        'region' => $region,
        'version' => 'latest'
    ]);
}
```

([AssumeRoleWithWebIdentityCredentialProvider.php#L75-L84](https://github.com/aws/aws-sdk-php/blob/3.356.42/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php#L75-L84))

### SDK version used

3.356.42

PHP 8.2.29 from Docker image php:8.2-fpm-bullseye

stobrien89:
*Issue #, if available:*
Fixes #3197 

*Description of changes:*
Fixes condition where a default client created by certain STS credential providers would call the `us-east-1` regional endpoint rather than the global endpoint.

### DIFF
--- a/tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
+++ b/tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
@@ -6,6 +6,7 @@ use Aws\Command;
 use Aws\Credentials\AssumeRoleWithWebIdentityCredentialProvider;
 use Aws\Credentials\Credentials;
 use Aws\Exception\AwsException;
+use Aws\Middleware;
 use Aws\Result;
 use Aws\Sts\StsClient;
 use Aws\Sts\Exception\StsException;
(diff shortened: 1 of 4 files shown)


yenfryherrerafeliz (approved):

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +397,85 @@ public function testCanDisableInvalidIdentityTokenRetries()
             unlink($tokenPath);
         }
     }
+
+    /**
+     * @dataProvider endpointConfigurationProvider
+     */
+    public function testEndpointConfigurationBasedOnRegion(
+        string $region,
```
yenfryherrerafeliz: We need to make this parameter nullable to fix the test failure, but overall looks good.

stobrien89: ah, went back and added types and just pushed without checking. fixed now

olavmo-sikt:
I tested these changes using the same test script and setup as I used in [#3197](https://github.com/aws/aws-sdk-php/issues/3197), and can confirm that this MR fixes the issue.

From the CloudTrail logs, I can see that the request was sent to the global `sts.amazonaws.com` endpoint, and was correctly routed to the local instance of that service in `eu-north-1`.

```yaml
{
    # ...
    "eventName": "AssumeRoleWithWebIdentity",
    "awsRegion": "us-east-1",
    # ...
    "additionalEventData": {
        # ...
        "RequestDetails": {
            "endpointType": "global",
            "awsServingRegion": "eu-north-1"
        }
    },
    # ...
    "tlsDetails": {
        # ...
        "clientProvidedHostHeader": "sts.amazonaws.com"
    }
}
```

stobrien89:
@olavmo-sikt,

Am I correct in assuming this also fixes the issue that predates our changes to the default STS endpoint configuration?  I have an idea about what may have caused that, but it's not necessarily relevant to this issue with these code changes applied.

PR #3203, opened 2025-10-27
bugfix: sts-credential-providers

stobrien89:
Issue #, if available:
Fixes https://github.com/aws/aws-sdk-php/issues/3197

Description of changes:
Adds support for resolving the region for STS clients created by assume role providers.  By default, the credentials chain will pass client configuration to expose the top-level client's region.  When providers are called outside the context of a client, they will accept `region` in-code configuration, `region` source profile configuration (for roles assumed via profiles), and will attempt to resolve `AWS_REGION` if any of the aforementioned options don't produce a region.

This also adds a notice that prompts to configure a region if a region cannot be found.  These providers will still fall back to `us-east-1`, but this is likely to change in the future.

### DIFF
--- a/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
+++ b/src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
@@ -36,6 +36,7 @@ class AssumeRoleWithWebIdentityCredentialProvider
 
     /** @var integer */
     private $tokenFileReadAttempts;
+
     /** @var string */
     private $source;
 
@@ -72,15 +73,15 @@ class AssumeRoleWithWebIdentityCredentialProvider
         $this->tokenFileReadAttempts = 0;
         $this->session = $config['SessionName']
             ?? 'aws-sdk-php-' . round(microtime(true) * 1000);
-        $region = $config['region'] ?? 'us-east-1';
+
         if (isset($config['client'])) {
             $this->client = $config['client'];
         } else {
-            $this->client = new StsClient([
-                'credentials' => false,
-                'region' => $region,
-                'version' => 'latest'
-            ]);
+            $region = $config['region']
+                ?? getEnv(CredentialProvider::ENV_REGION)
+                ?: null;
+
+            $this->client = $this->createDefaultStsClient($region);
         }
 
         $this->source = $config['source']
@@ -166,5 +167,35 @@ class AssumeRoleWithWebIdentityCredentialProvider
                 $this->source
             );
         });
+    }
+
+    /**
+     * @param string|null $region
+     *
+     * @return StsClient
+     */
+    private function createDefaultStsClient(
+        ?string $region
+    ): StsClient
+    {
+        if (empty($region)) {
+            $region = CredentialProvider::FALLBACK_REGION;
+            trigger_error(
+                'NOTICE: STS client created without explicit `region` configuration.' . PHP_EOL
+                . "Defaulting to {$region}. This fallback behavior may be removed." . PHP_EOL
+                . 'To avoid potential disruptions, configure a region using one of the following methods:' . PHP_EOL
+                . '(1) Pass `region` in the `$config` array when calling the provider,' . PHP_EOL
+                . '(2) Set the `AWS_REGION` environment variable.' . PHP_EOL
+                . 'OR provide an STS client in the `$config` array when creating the provider as `client`.' . PHP_EOL
+                . 'See: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/assume-role-with-web-identity-provider.html'
+                . PHP_EOL,
+                E_USER_NOTICE
+            );
+        }
+
+        return new StsClient([
+            'credentials' => false,
+            'region' => $region
+        ]);
     }
 }
--- a/src/Credentials/CredentialProvider.php
+++ b/src/Credentials/CredentialProvider.php
@@ -52,6 +52,8 @@ class CredentialProvider
     const ENV_SESSION = 'AWS_SESSION_TOKEN';
     const ENV_TOKEN_FILE = 'AWS_WEB_IDENTITY_TOKEN_FILE';
     const ENV_SHARED_CREDENTIALS_FILE = 'AWS_SHARED_CREDENTIALS_FILE';
+    public const ENV_REGION = 'AWS_REGION';
+    public const FALLBACK_REGION = 'us-east-1';
     public const REFRESH_WINDOW = 60;
 
     /**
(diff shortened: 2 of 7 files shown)


src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -72,14 +73,17 @@ public function __construct(array $config = [])
...
-        $region = $config['region'] ?? 'us-east-1';
+        $region = $config['region'] ?? null;
         if (isset($config['client'])) {
             $this->client = $config['client'];
         } else {
             $this->client = new StsClient([
                 'credentials' => false,
-                'region' => $region,
-                'version' => 'latest'
+                'region' => $region ?? 'us-east-1',
```
yenfryherrerafeliz: Could we create a const called DEFAULT_STS_REGION?

src/Credentials/CredentialProvider.php
```diff
@@ -725,8 +725,11 @@ private static function loadRoleProfile(
             }
             $stsClient = new StsClient([
                 'credentials' => $sourceCredentials,
-                'region' => $sourceRegion,
+                'region' => $region ?? 'us-east-1',
```
yenfryherrerafeliz: I think the const DEFAULT_STS_REGION can be created in `CredentialProvider` class.

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +397,85 @@ public function testCanDisableInvalidIdentityTokenRetries()
...
+            "Failed asserting endpoint for: {$description}"
+        );
+
+        unlink($tokenFile);
+    }
+
+    public function endpointConfigurationProvider(): array
+    {
+        return [
+            'explicit us-east-1 uses regional endpoint' => [
```
yenfryherrerafeliz: Could we add an extra test case here with another region that is not us-east-1.

stobrien89:
Going a different direction with this. I'll let you know when it's updated

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -167,4 +168,34 @@ public function __invoke()
...
+     */
+    private function createDefaultStsClient(
+        ?string $region
+    ): StsClient
+    {
+        if (empty($region)) {
+            $region = CredentialProvider::FALLBACK_REGION;
+            trigger_error(
+                'NOTICE: STS client created without explicit `region` configuration.' . PHP_EOL
+                . 'Defaulting to `us-east-1`. This fallback behavior may be removed.' . PHP_EOL
```
yenfryherrerafeliz: Refer to CredentialProvider::FALLBACK_REGION instead of hard-coding the region.

src/Credentials/AssumeRoleWithWebIdentityCredentialProvider.php
```diff
@@ -167,4 +168,34 @@ public function __invoke()
             );
         });
     }
+
+    /**
+     * @param string|null $region
+     *
+     * @return StsClient
+     */
+    private function createDefaultStsClient(
```
yenfryherrerafeliz, line 177: I think this method is redundant. We can use the one defined in `CredentialProvider`. Just change the type of `$credentials` to be `CredentialsInterface|bool` and when called from `AssumeRoleWithWebIdentityCredentialProvider` just pass false as the credentials parameter.

src/Credentials/CredentialProvider.php
```diff
@@ -1026,5 +1027,37 @@ private static function getCredentialsFromSsoService($ssoProfile, $clientRegion,
...
+    private static function createDefaultStsClient(
+        CredentialsInterface $credentials,
+        ?string $region
+    ): StsClient
+    {
+        if (empty($region)) {
+            $region = self::FALLBACK_REGION;
+            trigger_error(
+                'NOTICE: STS client created without explicit `region` configuration.' . PHP_EOL
+                . 'Defaulting to `us-east-1`. This fallback behavior may be removed.' . PHP_EOL
```
yenfryherrerafeliz: Same comment as [here](https://github.com/aws/aws-sdk-php/pull/3203/files#r2467440963)

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +399,168 @@ public function testCanDisableInvalidIdentityTokenRetries()
...
+                'NOTICE: STS client created without explicit `region` configuration.'
+            );
+        }
+
+        $provider = new AssumeRoleWithWebIdentityCredentialProvider($config);
+
+        // Check region
+        $reflection = new \ReflectionClass($provider);
+        $clientProperty = $reflection->getProperty('client');
+        $clientProperty->setAccessible(true);
```
yenfryherrerafeliz: I think this is not needed. The property should be accessible by default.

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +399,168 @@ public function testCanDisableInvalidIdentityTokenRetries()
...
+
+        // Cleanup
+        unlink($tokenFile);
+
+        // Restore original AWS_REGION value
+        if ($originalRegion !== false) {
+            putenv("AWS_REGION={$originalRegion}");
+        } else {
+            putenv("AWS_REGION");
+        }
```
yenfryherrerafeliz: I think this way of restoring `AWS_REGION` is not safe. For example, if a failure occurred then the variable will not be restored. I recommend to use a try {} finally {} block for it.

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +399,168 @@ public function testCanDisableInvalidIdentityTokenRetries()
...
+            'WebIdentityTokenFile' => $tokenFile,
+            'region' => $region
+        ];
+
+        $provider = new AssumeRoleWithWebIdentityCredentialProvider($config);
+
+        // Get the client via reflection
+        $reflection = new \ReflectionClass($provider);
+        $clientProperty = $reflection->getProperty('client');
+        $clientProperty->setAccessible(true);
```
yenfryherrerafeliz: I think the property should be accessible by default.

tests/Credentials/AssumeRoleWithWebIdentityCredentialProviderTest.php
```diff
@@ -396,4 +399,168 @@ public function testCanDisableInvalidIdentityTokenRetries()
...
+
+        // Assert the correct endpoint was called
+        $this->assertEquals(
+            $expectedEndpoint,
+            $capturedEndpoint,
+            "Failed asserting endpoint for region: {$region}"
+        );
+
+        // Cleanup
+        unlink($tokenFile);
```
yenfryherrerafeliz: I think that also here using a try {} finally {} block is safer in order to clean up the file.

yenfryherrerafeliz (changes requested):

Issue #2856, opened 2023-12-14
 fprintf(): supplied resource is not a valid stream resource


VadimSafonov:
### Describe the bug

After fix #2855
new error if Aws\S3\S3Client created with debug with logfn

Uncaught Exception fprintf(): supplied resource is not a valid stream resource
Stack trace: #0 
/project/vendor/guzzlehttp/guzzle/src/Handler/StreamHandler.php(514): fprintf(Resource id #637, '<%s> [%s] ', 'GET https://tes...', 'PROGRESS')
#1 [internal function]: GuzzleHttp\Handler\StreamHandler->GuzzleHttp\Handler\{closure}(7, 0, NULL, 0, 7671, 73602)
#2 
/project/vendor/guzzlehttp/psr7/src/Stream.php(228): fread(Resource id #642, 8192)
#3 
/project/vendor/aws/aws-sdk-php/src/S3/StreamWrapper.php(209): GuzzleHttp\Psr7\Stream->read(8192)
#4 [internal function]: Aws\S3\StreamWrapper->stream_read(8192)
#5 
/project/vendor/symfony/filesystem/Filesystem.php(62): stream_copy_to_stream(Resource id #643, Resource id #645)

### Expected Behavior

no error

### Current Behavior

error

### Reproduction Steps

copy file to s3

### SDK version used

3.293.10

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.2.13 (cli) (built: Nov 24 2023 13:15:01) (NTS) Copyright (c) The PHP Group Zend Engine v4.2.13, Copyright (c) Zend Technologies     with Zend OPcache v8.2.13, Copyright (c), by Zend Technologies

yenfryherrerafeliz:
Hi @VadimSafonov, sorry to hear about your issues. Can you please provide a minimal reproducible code so we can better investigate this?. For example, we would be able to see if happens by you providing a custom logfn, or if happens with the default logfn, which can be seen [here](https://github.com/aws/aws-sdk-php/blob/9ae3bfd036190847dd0a5c3cdc4681dbb0994f5a/src/TraceMiddleware.php#L67). Our default implementation will use echo to display the value.

I look forward to your response.

Thanks!

VadimSafonov:
Hi @yenfryherrerafeliz

I notice the following:
1 with default debug block config I got error, but file has been downloaded
2 without debug block config NO error, file has been downloaded

**There is reproduce code**
```
<?php

[...]

PR #3206, opened 2025-11-01
TraceMiddleware crashes when http option is false

johanrosenson:
```
Undefined array key "debug" at aws-sdk-php/src/TraceMiddleware.php:289
```

*Description of changes:*

When setting the debug.http option to false the above exception will be thrown because `$command['@http']['debug']` doesn't exist, adding a null check fixes the issue.

```
new S3Client([
    'debug' => [
        'logfn' => ...
        'http' => false,
    ],
]);
```

### DIFF
--- a/src/TraceMiddleware.php
+++ b/src/TraceMiddleware.php
@@ -286,7 +286,7 @@ class TraceMiddleware
 
     private function flushHttpDebug(CommandInterface $command)
     {
-        if ($res = $command['@http']['debug']) {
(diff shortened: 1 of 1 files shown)


yenfryherrerafeliz:
Hey @johanrosenson, I have a left a [comment](https://github.com/aws/aws-sdk-php/issues/2856#issuecomment-3637512775) in the issue. This is not the best way to actually fix the issue there. I am currently investigating what we can do to prevent this issue, however, in the mean time feel free to use the workaround in my comment.

Thank you!

Issue #3181, opened 2025-09-05
S3, check if file exists => Error parsing XML: String could not be parsed as XML


mumhftm:
### Describe the bug

Calling `Aws\S3\S3Client->doesObjectExistV2() `

leads to the exception:
`Error parsing response for HeadObject: AWS parsing error: Error parsing XML: String could not be parsed as XML at /aws/aws-sdk-php/src/S3/Parser/S3Parser.php:74`

- Happens **since V351.10.** 
- Last working version is 351.9. 
- I tested almost all versions between 350.0 up to 356.11

Use no special chars (no whitespace or similar) in filenames of files on S3. Paths are simple, e.g. _invoices/2025/506/2025-08-27_506-000030655_invoice.pdf_

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

It should just return a boolean, indicating whether or not the file exists on S3.

### Current Behavior

It leads to the described exception.
Happens with all files.

### Reproduction Steps

1. Place a file to an S3 bucket
2. Assign permissions to access the S3 bucket to an EC2 (with Ubuntu)
3. Install the AWS lib and PHP 
4. Call the doesObjectExistV2()

### Possible Solution

I would - obviously - review the changes:
https://github.com/aws/aws-sdk-php/compare/3.351.9...3.351.10

### SDK version used

351.10

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.3.6, Ubuntu 24.04.1, EC2, S3 Bucket, Access through assigned IAM role

mumhftm:
Could it be related to #3173  ?

stobrien89:
Hi @mumhftm,

Sorry to hear about the issues.  Looking into this now.

stobrien89:
@mumhftm,

I wasn't able to reproduce, but this does seem similar to the issue you linked.  It seems like the server is sending back a body, or some utility is converting the `HEAD` response to a `GET` response.  Would you be able to provide debug logs by adding `'debug' => true` to your client configuration and providing the output for a failed call? Be sure to redact any sensitive information like account numbers.

It doesn't appear that you're using any 3rd party s3 implementations or libraries to access S3, but if you are, could you provide the names of the packages or services you're using?

mumhftm:
Hi @stobrien89 

Thanks.

The problem happens with Laravel 12, using league/flysystem-aws-s3-v3 V3.29 on top of aws-sdk-php. 
[...]

Issue #3124, opened 2025-05-21
When using streaming response mode, exceptions do not include the message text from the HTTP body


glennschmidt:
### Describe the bug

When making a request which uses `['@http' => ['stream' => true]]`, if the server returns a HTTP error status such as `400`, the SDK is unable to read the response body for inclusion in the exception message, and consequently there's no way to see the reason for the error.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The PHP exception should include the error message from the HTTP body, as in non-streaming mode.

```
Fatal error: Uncaught exception 'Aws\BedrockRuntime\Exception\BedrockRuntimeException' with message 'Error executing "ConverseStream" on "https://bedrock-runtime.us-west-2.amazonaws.com/model/x/converse-stream"; AWS HTTP error: Client error: `POST https://bedrock-runtime.us-west-2.amazonaws.com/model/x/converse-stream` resulted in a `400 Bad Request` response:
{"message":"The provided model identifier is invalid."}
```

### Current Behavior

The PHP exception says `Unable to parse error information from response`.

```
Fatal error: Uncaught exception 'Aws\BedrockRuntime\Exception\BedrockRuntimeException' with message 'Error executing "ConverseStream" on "https://bedrock-runtime.us-west-2.amazonaws.com/model/x/converse-stream"; AWS HTTP error: Client error: `POST https://bedrock-runtime.us-west-2.amazonaws.com/model/x/converse-stream` resulted in a `400 Bad Request` response Unable to parse error information from response - Error parsing JSON: Syntax error'
```

### Reproduction Steps

Make a streaming API call, with some invalid parameters so that it would cause a 400 Bad Request response.

```php
$client = new Aws\BedrockRuntime\BedrockRuntimeClient([
    'version' => '2023-09-30',
    'region' => 'us-west-2',
]);

$client->converseStream([
    'modelId' => 'x',
    '@http' => [
        'stream' => true,
    ],
]);
```

### Possible Solution

This could be because the response body stream is non-seekable, and some other part of the SDK has already read it.

### SDK version used

3.343.14

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.1.31, Linux

yenfryherrerafeliz:
[...]

Issue #3018, opened 2024-10-31
Bedrock Runtime converseStream doesn't seem to be streaming


gazzoy:
### Describe the bug

When calling the `converseStream()` method from the AWS PHP SDK, the result  is supposed to be returned as soon as possible, then we can iterate `stream` value in the result until the streaming finished.

However, debug logs show that when calling  `converseStream()` method, we need to wait for the whole streaming finished, then it returns results, which seems weird to me.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

`converseStream()` should return the result as soon as possible.

### Current Behavior

`converseStream()` does NOT return the result as soon as possible rather it waits until the streaming finished, then it returns the result.

### Reproduction Steps

Belows are example code to describe the issue. 

We see debug logs like:
```
2024-10-31 13:10:00 calling converseStream()
2024-10-31 13:10:30 calling converseStream()  <-- it waits 30 seconds until the streaming finished, which is weird.
2024-10-31 13:10:30 looping stream()
2024-10-31 13:10:30 looped stream() <- it never takes time in loop as its streaming already done.
```

I might be wrong, but does it supposed to work as following?
```
2024-10-31 13:10:00 calling converseStream()
2024-10-31 13:10:00 calling converseStream()  <-- it returns the result as soon as possible
2024-10-31 13:10:00 looping stream()
2024-10-31 13:10:30 looped stream() <-- it takes 30 seconds to finishe the streaming.
```

```php

<?php
use Aws\BedrockRuntime\BedrockRuntimeClient;
use Aws\Credentials\CredentialProvider;

$bedrockClient = new BedrockRuntimeClient([
    'version' => 'latest',
    'region' => 'ap-northeast-1',
    'credentials' => CredentialProvider::env(),
]);

$input = 'Describe about aws in 5k texts';

\Log::debug('calling converseStream()');
$result = $bedrockClient->converseStream([
    'modelId' => 'anthropic.claude-3-5-sonnet-20240620-v1:0', // note: Claude 3.5 Sonnet
    'messages' => [
        [
            'content' => [
                [
                    'text' => "\n\nHuman:{$input} \n\nAssistant:",
                ],
            ],
            'role' => 'user',
        ],
    ],
]);
\Log::debug('called converseStream()');

\Log::debug('looping stream()');
foreach ($result['stream'] as $eachStreamChunk)
{
    // do something
}
\Log::debug('looped stream()');
```

### SDK version used

[...]

PR #3214, opened 2025-11-20
fix: add streaming flag and error handling

yenfryherrerafeliz:
*Issues #3124, #3018, #3181:*

*Description of changes:*
- Evaluates if a client's operation requires http streaming flag to be set to true, and if so then it pass the option along.
- Make error parsers to reused a parsed body for scenario where the body is non seekable and can't be consumed again.
- Make sure empty payloads are not parsed.

### DIFF
--- a/src/Api/ErrorParser/JsonRpcErrorParser.php
+++ b/src/Api/ErrorParser/JsonRpcErrorParser.php
@@ -5,8 +5,6 @@ use Aws\Api\Parser\AbstractParser;
 use Aws\Api\Parser\JsonParser;
 use Aws\Api\Service;
 use Aws\CommandInterface;
-use GuzzleHttp\Psr7\CachingStream;
-use GuzzleHttp\Psr7\Response;
 use Psr\Http\Message\ResponseInterface;
 
 /**
--- a/src/Api/Parser/AbstractRestParser.php
+++ b/src/Api/Parser/AbstractRestParser.php
@@ -6,8 +6,6 @@ use Aws\Api\Shape;
 use Aws\Api\StructureShape;
 use Aws\Result;
 use Aws\CommandInterface;
-use GuzzleHttp\Psr7\CachingStream;
-use GuzzleHttp\Psr7\Response;
 use Psr\Http\Message\ResponseInterface;
 
 /**
@@ -93,9 +91,7 @@ abstract class AbstractRestParser extends AbstractParser
             return;
         }
 
-        if (!$body->isSeekable()) {
-            $response = AbstractParser::getResponseWithCachingStream($response);
-        }
+        $response = AbstractParser::getResponseWithCachingStream($response);
 
         if ($member instanceof StructureShape) {
             //Unions must have at least one member set to a non-null value
--- a/.changes/nextrelease/fix-streaming-scenarios-handling.json
+++ b/.changes/nextrelease/fix-streaming-scenarios-handling.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "bugfix",
+        "category": "",
+        "description": "Fixes how response with empty bodies for non-seekable streams are handled and adds streaming flag automatically based on the operation."
+    }
+]


yenfryherrerafeliz:
```
Running integ tets

66 scenarios (66 passed)
248 steps (248 passed)
5m9.29s (45.13Mb)
Running smoke tests
............

111 scenarios (111 passed)
222 steps (222 passed)
0m45.99s (95.47Mb)
```

src/Api/ErrorParser/AbstractErrorParser.php
```diff
@@ -27,23 +28,10 @@ public function __construct(?Service $api = null)
     }
 
     abstract protected function payload(
-        ResponseInterface $response,
+        ResponseInterface|SimpleXMLElement|array $responseOrParsedBody,
```
stobrien89: this would be better as `mixed` so we're not pulling in codec-specific dependencies, but mixing responses and parsed data here doesn't feel right— is it possible to delegate this call back to `extractPayload`?

src/WrappedHttpHandler.php
```diff
@@ -166,19 +166,21 @@ private function parseError(
             throw new \RuntimeException('The HTTP handler was rejected without an "exception" key value pair.');
         }
 
-        $serviceError = "AWS HTTP error: " . $err['exception']->getMessage();
```
stobrien89, line 169: The change in format might affect log parsing.  I'd leave this unless it's accounted for later

src/Api/ErrorParser/XmlErrorParser.php
```diff
@@ -97,15 +109,27 @@ protected function registerNamespacePrefix(\SimpleXMLElement $element)
...
+
+
         $prefix = $this->registerNamespacePrefix($xmlBody);
         $errorBody = $xmlBody->xpath("//{$prefix}Error");
 
         if (is_array($errorBody) && !empty($errorBody[0])) {
             return $this->parser->parse($member, $errorBody[0]);
         }
+
+        throw new ParserException(
```
stobrien89: This could be a breaking change if callers don't expect this

src/CloudWatchLogs/CloudWatchLogsClient.php
stobrien89, line 1: Still has the static `$streamingCommands` property defined (not shown in diff), but it's no longer used. Should be removed

src/Api/ErrorParser/XmlErrorParser.php
```diff
@@ -38,7 +40,17 @@ public function __invoke(
         ];
 
         $body = $response->getBody();
-        if ($body->getSize() > 0) {
+        if (!$body->isSeekable()) {
+            $tempBody = Utils::streamFor();
+            Utils::copyToStream($body, $tempBody);
+            $body = $tempBody;
```
stobrien89: this could be (lazy rather than eager):

```php
if (!$body->isSeekable()) {
    $body = new CachingStream($body);
}

**Review by stobrien89 — CHANGES_REQUESTED** [review]

File: src/Api/ErrorParser/XmlErrorParser.php (hunk 4 above)
Comment (yenfryherrerafeliz) [reply src/Api/ErrorParser/XmlErrorParser.php]: I don't think it is necessary to have a lazy stream there. The whole content is consumed right after this block, at line 52. Unless you are thinking in any other use case?

Comment (yenfryherrerafeliz) [reply src/Api/ErrorParser/XmlErrorParser.php]: I have replaced that logic for $body->getContents() instead, which avoids having to do a copy operation.

File: src/Api/ResponseWrapper.php
Comment (stobrien89) [inline src/Api/ResponseWrapper.php:1]: Converting non-seekable bodies to `CachingStream` or using another Guzzle utility is fine, but wrapping it in a new class adds unnecessary overhead for the common case (seekable streams). The auto-rewind in getBody() also introduces unexpected behavior that differs from standard PSR-7 responses.

I'd suggest reading the body once with getContents(), storing it, and passing the string to methods that need it, or converting to another resource if the body is unseekable and needs to be read multiple times.

**Review by stobrien89 — CHANGES_REQUESTED** [review]

**Comment by yenfryherrerafeliz** [comment]
```
Running integ tets

66 scenarios (66 passed)
248 steps (248 passed)
10m14.82s (43.42Mb)
```
```
Running smoke tests
............

111 scenarios (111 passed)
222 steps (222 passed)
0m35.47s (92.58Mb)
```

File: src/Api/ErrorParser/JsonRpcErrorParser.php [hunk 5]
```diff
@@ -25,6 +27,15 @@ public function __invoke(
...
         ?CommandInterface $command = null
     ) {
+        if (!$response->getBody()->isSeekable()) {
+            $response = new Response(
+                $response->getStatusCode(),
+                $response->getHeaders(),
+                new CachingStream($response->getBody()),
+                $response->getProtocolVersion(),
+                $response->getReasonPhrase()
+            );
```
stobrien89: This could be:


```suggestion
            $seekableStream = new CachingStream($response->getBody());
            $response = $response->withBody($seekableStream);
```

src/Api/ErrorParser/JsonParserTrait.php
```diff
@@ -39,8 +39,13 @@ private function genericHandler(ResponseInterface $response): array
...
-        if (!$body->isSeekable() || $body->getSize()) {
-            $parsedBody = $this->parseJson((string) $body, $response);
+        if ($body->isSeekable()) {
+            $body->rewind();
+        }
+
+        $rawBody = $body->getContents();
+        if (!empty($rawBody)) {
+            $parsedBody = $this->parseJson($rawBody, $response);
         }
```
stobrien89: Could this logic and the logic in `payload()` for JSON and XML parsers be moved into the abstract parsers?

src/Api/Parser/AbstractRestParser.php
```diff
@@ -39,6 +42,30 @@ public function __invoke(
 
         if ($payload = $output['payload']) {
             $this->extractPayload($payload, $output, $response, $result);
+        } else {
+            $body = $response->getBody();
+            if (!$body->isSeekable()) {
+                $response = new Response(
+                    $response->getStatusCode(),
+                    $response->getHeaders(),
+                    new CachingStream($body),
```
stobrien89: same suggestion as above

stobrien89 (changes requested):
A few more comments and an old one (log parsing) that needs to be addressed

src/Api/ErrorParser/JsonParserTrait.php
yenfryherrerafeliz: This logic is like a generic method for parsing JSON payload and is used by `JsonRpcErrorParser` and `RestJsonErrorParser`, and actually it was already there; I just changed a few things such as the empty content evaluation, etc. 

Regarding moving payload into the abstract class, I think we could, but I would like to understand more why we want to do it?

src/WrappedHttpHandler.php
yenfryherrerafeliz: I basically changed this because it was duplicating the error message. What I did in specific was to append the message from the exception when a response is not available, but, if we have a response then we just use the error message from the deserialization process.
```php
if (!isset($err['response'])) {
    $parts = ['response' => null];
    $serviceError .= $err['exception']->getMessage(); // MOVED HERE.
}
```

src/Api/ErrorParser/JsonRpcErrorParser.php
yenfryherrerafeliz: Totally agree. Your approach is cleaner and more elegant!

src/Api/ErrorParser/JsonParserTrait.php
stobrien89: I noticed rewind logic being inserted into the individual parsers (in this case the parser trait) and because at least a rewind seemed to be happening in each of them, thought we could move that logic and possibly some of the logic that follows into `AbstractParser` and/or `AbstractErrorParser`.  Obviously the encoding-specific methods (`parseJson` in this case) need to be in each protocol parser or trait, but was wondering if there was a way some of the similar logic could be shared in the classes mentioned above.

yenfryherrerafeliz: I see what you mean, however, I don`t think that for this small logic or condition we should consider any sort of abstraction. I prefer clarity in this case. What do you think?

stobrien89: I think the fact that the rewind/read block is repeated 7 times and that these classes inherit from a common parent warrants an abstraction.  The logic is not difficult to reason with and I don't think we'd lose clarity from a helper.  Something like this could work in `AbstractParser` and/or `AbstractErrorParser`:


```php

protected function getBodyContents(ResponseInterface $response): string
{
    $body = $response->getBody();
    if ($body->isSeekable()) {
        $body->rewind();
    }

    return $body->getContents();
}

Comment (yenfryherrerafeliz) [reply src/Api/ErrorParser/JsonParserTrait.php]: Agreed. I preferred to make it a static function so it can be called from different places without depending on inheritance.

File: src/Api/ErrorParser/JsonRpcErrorParser.php [hunk 8]
```diff
@@ -25,6 +27,11 @@ public function __invoke(
         ResponseInterface $response,
         ?CommandInterface $command = null
     ) {
+        if (!$response->getBody()->isSeekable()) {
+            $response = $response->withBody(
+                new CachingStream($response->getBody())
```
stobrien89: question: Is this not needed for error handling with other protocols?

src/Api/Parser/AbstractRestParser.php
```diff
@@ -39,6 +41,26 @@ public function __invoke(
...
             $this->extractPayload($payload, $output, $response, $result);
+        } else {
+            $body = $response->getBody();
+            if (!$body->isSeekable()) {
+                $response = $response->withBody(
+                    new CachingStream($body)
+                );
+            }
+
+            $body = $response->getBody();
```
stobrien89: question: why not use the helper method here/

src/Api/Parser/QueryParser.php
```diff
@@ -40,8 +40,10 @@ public function __invoke(
         ResponseInterface $response
     ) {
         $output = $this->api->getOperation($command->getName())->getOutput();
-        $body = $response->getBody();
-        $xml = !$body->isSeekable() || $body->getSize()
+        // Read the full payload, even in non-seekable streams
```
stobrien89, line 43: curious about the rationale here- not saying it's wrong

src/WrappedHttpHandler.php
stobrien89: Ah, I see. So basically a caller will still see the same thing, just printed from somewhere else?

stobrien89 (changes requested):
A few more comments- I just responded to that one question without doing a proper review, but thanks for the quick turnaround

src/Api/Parser/QueryParser.php
yenfryherrerafeliz: Basically, the body of the response is consumed just once in this case and hence it will completely available without the need of a rewind. However, I can also use the helper method here to be extra safe.

src/Api/Parser/AbstractRestParser.php
yenfryherrerafeliz: Missed this one.

src/Api/ErrorParser/JsonRpcErrorParser.php
yenfryherrerafeliz: Yes, it is needed. Thank you for catching it.

src/WrappedHttpHandler.php
yenfryherrerafeliz: Yes, caller will still see the same thing, but with the difference that now error messages will not be duplicated. Previously, if we deserialized a modeled message then we duplicate the error we output.
For example, before we were getting:
```
PHP Fatal error:  Uncaught exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "GetObject" on "https://bucket.s3.us-east-2.amazonaws.com/no-exists"; AWS HTTP error:
Client error: `GET https://bucket.s3.us-east-2.amazonaws.com/no-exists` resulted in a `404 Not Found` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message> (truncated...)
NoSuchKey (client): The specified key does not exist.'

GuzzleHttp\Exception\ClientException: Client error: `GET https://bucket.s3.us-east-2.amazonaws.com/no-exists` resulted in a `404 Not Found` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message> (truncated...)
```

But now, after this fix:
```
PHP Fatal error:  Uncaught exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "GetObject" on "https://bucket.s3.us-east-2.amazonaws.com/no-exists"; AWS HTTP error:
NoSuchKey (client): The specified key does not exist.'

GuzzleHttp\Exception\ClientException: Client error: `GET https://bucket.s3.us-east-2.amazonaws.com/no-exists` resulted in a `404 Not Found` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message> (truncated...)
```

src/Api/Parser/AbstractRestParser.php
```diff
@@ -39,6 +41,16 @@ public function __invoke(
 
         if ($payload = $output['payload']) {
             $this->extractPayload($payload, $output, $response, $result);
+        } else {
+            $response = AbstractParser::getResponseWithCachingStream($response);
+
+            $rawBody = AbstractParser::getBodyContents($response);
```
stobrien89: Could we prevent the double read here by using `getSize()` in the seekable case?  At least for `RestJsonParser`, this method is also called in `payload()`

src/Api/Parser/AbstractRestParser.php
```diff
@@ -75,17 +78,26 @@ private function extractPayload(
...
+
+        if (!$body->isSeekable()) {
+            $response = AbstractParser::getResponseWithCachingStream($response);
+        }
+
+        if ($member instanceof StructureShape) {
             //Unions must have at least one member set to a non-null value
             // If the body is empty, we can assume it is unset
-            if (!empty($member['union']) && ($body->isSeekable() && !$body->getSize())) {
+            $rawBody = AbstractParser::getBodyContents($response);
```
stobrien89: I think this would be a third read if we're using `rest-json`

src/CloudWatchLogs/CloudWatchLogsClient.php
```diff
@@ -224,36 +224,10 @@
  * @method \GuzzleHttp\Promise\Promise updateScheduledQueryAsync(array $args = [])
  */
 class CloudWatchLogsClient extends AwsClient {
-    static $streamingCommands = [
-        'StartLiveTail' => true
-    ];
 
     public function __construct(array $args)
```
stobrien89: We can probably do away with the constructor altogether, but not required

tests/Api/Parser/RestJsonParserTest.php
```diff
@@ -180,4 +182,49 @@ public function testParsesFalsyHeaderValues(): void
...
+            CommandInterface::class
+        )->getMock();
+        $command->method('getName')->willReturn('TestOperation');
+
+        $parser($command, $response);
+
+        // Not error occurred. Test successfully.
+        // Previously, on non-seekable streams it would have failed.
+        // Because, it would have tried to parse an empty string.
+        $this->assertTrue(true);
```
stobrien89: Could we strengthen the test here with some assertions on the parsed result?

src/Api/Parser/AbstractParser.php
```diff
@@ -43,4 +44,27 @@ abstract public function parseMemberFromStream(
...
+    ): ResponseInterface
+    {
+        if (!$response->getBody()->isSeekable()) {
+            return $response->withBody(
+                new CachingStream($response->getBody())
+            );
+        }
+
+        return $response;
+    }
```
stobrien89, line 69: These two methods should have test coverage

tests/Api/Parser/RestJsonParserTest.php
yenfryherrerafeliz: The intention of this test is to make sure we have not exception thrown when getting empty payloads when the body is non-seekable. I added the proper annotation to avoid having asserting to true.

src/Api/Parser/AbstractRestParser.php
```diff
@@ -6,6 +6,8 @@
 use Aws\Api\StructureShape;
 use Aws\Result;
 use Aws\CommandInterface;
+use GuzzleHttp\Psr7\CachingStream;
+use GuzzleHttp\Psr7\Response;
```
stobrien89: Looks like these aren't being used

src/Api/ErrorParser/JsonRpcErrorParser.php
```diff
@@ -1,9 +1,12 @@
 <?php
 namespace Aws\Api\ErrorParser;
 
+use Aws\Api\Parser\AbstractParser;
 use Aws\Api\Parser\JsonParser;
 use Aws\Api\Service;
 use Aws\CommandInterface;
+use GuzzleHttp\Psr7\CachingStream;
+use GuzzleHttp\Psr7\Response;
```
stobrien89: Looks like these aren't being used

src/Api/Parser/AbstractRestParser.php
```diff
@@ -75,17 +83,31 @@ private function extractPayload(
...
                 $body,
                 $member,
                 $this
             );
-        } elseif ($member instanceof StructureShape) {
+
+            return;
+        }
+
+        if (!$body->isSeekable()) {
```
stobrien89: Not blocking, but this check doesn't seem to be needed

stobrien89:
I'd do another run of integ and smoke tests, but approved pending those results!

yenfryherrerafeliz:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m36.50s (93.87Mb)
```

```
...............

127 scenarios (127 passed)
435 steps (435 passed)
22m54.33s (46.52Mb)
```

PR #3215, opened 2025-11-21
chore: update login provider tests using `chmod()`

stobrien89:
*Description of changes:*

One test that changed file permissions was failing in Docker/CodeBuild, and another was giving false coverage. In containerized environments, they run as root, and root can bypass `chmod()` restrictions (can write to read-only files, read no-permission files). Added root user detection to skip these tests when running as root.


yenfryherrerafeliz (approved):
I will approve, however, I would prefer not to skip tests based on environment circumstances. Tests should always be ran. Nonetheless, since those tests run on GitHub CI first I think it can be acceptable for now.

Issue #2998, opened 2024-09-17
Add @method declarations to Interfaces


donatj:
### Describe the feature

Currently `@method` declarations 's only exist on the implementations but not the interface -

Compare

### S3Client

https://github.com/aws/aws-sdk-php/blob/6330c3367f303ef4430fca605df9e411abe34d48/src/S3/S3Client.php#L36-L235

### S3ClientInterface

https://github.com/aws/aws-sdk-php/blob/6330c3367f303ef4430fca605df9e411abe34d48/src/S3/S3ClientInterface.php#L2-L11

### Use Case

It would be very helpful for IDEs and static analysis and other such tools if the `@methods` were also on the interfaces.

Compare current autocompletion for interface vs instance

<img width="695" alt="image" src="https://github.com/user-attachments/assets/99b23d1e-8041-488e-9563-d471d78f90d8">

<img width="804" alt="image" src="https://github.com/user-attachments/assets/399aecd5-9447-4910-b92e-1fe41205fedb">

### Proposed Solution

I _presume these are generated_ but I am unsure - it would be nice if whatever generated these could also append them to the interfaces.

If they are _not_ generated, I'd be more than happy to open a PR copying them over myself.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.321.13

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.3.11 - doesn't matter

yenfryherrerafeliz:
Hi @donatj, this feature has been implemented on [this](https://github.com/aws/aws-sdk-php/pull/3220) PR. Please feel free to reach out for anything else we may help with.

Thank you!

donatj:
@yenfryherrerafeliz Wonderful news, thank you so much!

PR #3220, opened 2025-12-02
feat: add interface method annotation

yenfryherrerafeliz:
*Issue #2998*

*Description of changes:*
- Add method annotation in interfaces so that methods are available and recognized when the variable is declare with its interface type. For example, S3Client and S3ClientInterface.

### DIFF
--- a/build/ClientAnnotator.php
+++ b/build/ClientAnnotator.php
@@ -6,12 +6,19 @@ class ClientAnnotator
 {
     /** @var ReflectionClass */
     private $reflection;
+
+    /** @var ReflectionClass|null */
+    private $reflectionInterface;
+
     /** @var string */
     private $endpoint;
+
     /** @var string[] */
     private $versions;
+
     /** @var array */
     private $methods;
+
     /** @var array */
     private $aliases;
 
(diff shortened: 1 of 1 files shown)


build/ClientAnnotator.php
```diff
@@ -36,6 +50,17 @@ public function updateApiMethodAnnotations()
...
 
+        if ($this->reflectionInterface !== null) {
+            $interfaceUpdater = new ClassAnnotationUpdater(
+                $this->reflectionInterface,
+                $this->getMethodAnnotations(),
+                $this->getDefaultDocComment(),
+                '/^\* @method (\\\\Aws\\\\Result|\\\\GuzzleHttp\\\\Promise\\\\Promise) /'
+            );
+
+            return $interfaceUpdater->update() && $updater->update();
```
stobrien89, line 61: question: can we just call `$interfaceUpdater->update()` here and fall through to the `$updater->update()` call?

yenfryherrerafeliz: I think is safe to evaluate if both executions are successful, because they are both expected to succeed.

stobrien89 (approved):

PR #3223, opened 2025-12-04
feat: cbor protocol support

stobrien89:
*Description of changes:*
Adds support for the Smithy RPC V2 CBOR protocol. https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html

### DIFF
--- a/src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
+++ b/src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
@@ -1,0 +1,159 @@
+<?php
+namespace Aws\Api\ErrorParser;
+
+use Aws\Api\Parser\AbstractParser;
+use Aws\Api\StructureShape;
+use Aws\CommandInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Base implementation for Smithy RPC V2 protocol error parsers.
+ *
+ * @internal
+ */
+abstract class AbstractRpcV2ErrorParser extends AbstractErrorParser
+{
+    private const HEADER_QUERY_ERROR = 'x-amzn-query-error';
+    private const HEADER_ERROR_TYPE = 'x-amzn-errortype';
+    private const HEADER_REQUEST_ID = 'x-amzn-requestid';
+
+    /**
+     * @param ResponseInterface $response
+     * @param CommandInterface|null $command
+     *
+     * @return array
+     */
+    public function __invoke(
+        ResponseInterface $response,
+        ?CommandInterface $command = null
+    ) {
+        $response = AbstractParser::getResponseWithCachingStream($response);
+        $data = $this->parseError($response);
+
+        if (isset($data['parsed']['__type'])) {
+            $data['message'] = $data['parsed']['message'] ?? null;
+        }
+
+        $this->populateShape($data, $response, $command);
+
+        return $data;
+    }
+
+    /**
+     * @param ResponseInterface $response
+     * @param StructureShape $member
+     *
+     * @return array
+     */
+    abstract protected function payload(
+        ResponseInterface $response,
+        StructureShape $member
+    ): array;
+
+    /**
+     * @param StreamInterface $body
+     * @param ResponseInterface $response
+     *
+     * @return mixed
+     */
+    abstract protected function parseBody(
+        StreamInterface $body,
+        ResponseInterface $response
+    ): mixed;
+
+    /**
+     * @param ResponseInterface $response
+     *
+     * @return array
+     */
+    private function parseError(ResponseInterface $response): array
+    {
+        $statusCode = (string) $response->getStatusCode();
+        $errorCode = null;
+        $errorType = null;
+
+        if ($this->api?->getMetadata('awsQueryCompatible') !== null
+            && $response->hasHeader(self::HEADER_QUERY_ERROR)
+            && $awsQueryError = $this->parseQueryCompatibleHeader($response)
+        ) {
+            $errorCode = $awsQueryError['code'];
+            $errorType = $awsQueryError['type'];
+        }
+
+        if (!$errorCode && $response->hasHeader(self::HEADER_ERROR_TYPE)) {
+            $errorCode = $this->extractErrorCode(
+                $response->getHeaderLine(self::HEADER_ERROR_TYPE)
+            );
+        }
+
+        $parsedBody = null;
+        $body = $response->getBody();
+        if ($body->getSize()) {
+            //TODO handle unseekable streams with CachingStream
+            $parsedBody = array_change_key_case($this->parseBody($body, $response));
+        }
+
+        if (!$errorCode && $parsedBody) {
+            $errorCode = $this->extractErrorCode(
+                $parsedBody['code'] ?? $parsedBody['__type'] ?? ''
+            );
+        }
+
+        return [
+            'request_id' => $response->getHeaderLine(self::HEADER_REQUEST_ID),
+            'code'       => $errorCode ?: null,
+            'message'    => null,
+            'type'       => $errorType ?? ($statusCode[0] === '4' ? 'client' : 'server'),
+            'parsed'     => $parsedBody,
+        ];
+    }
+
+    /**
+     * Parse AWS Query Compatible error from header
+     *
+     * @param ResponseInterface $response
+     *
+     * @return array|null Returns ['code' => string, 'type' => string] or null
+     */
+    private function parseQueryCompatibleHeader(ResponseInterface $response): ?array
+    {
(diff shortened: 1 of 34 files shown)


src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
```diff
@@ -0,0 +1,161 @@
...
+        ResponseInterface $response,
+        ?CommandInterface $command = null
+    ) {
+        $data = $this->parseError($response);
+        if ($data['parsed']) {
+            $data['parsed'] = array_change_key_case($data['parsed']);
+        }
+
+        if (isset($data['parsed']['__type'])) {
+            $data['code'] ??= $this->extractErrorCode($data['parsed']['__type']);
```
yenfryherrerafeliz: Why resolving the error code again when it was already resolved with the same logic [here](https://github.com/aws/aws-sdk-php/pull/3223/files#diff-f565ec6e4154ca347cd195cfe9342fa2119b306cb4c6c213a2c80f668ba1485fR99)?

src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
```diff
@@ -0,0 +1,161 @@
...
+        StructureShape $member
+    ): array;
+
+    /**
+     * @param StreamInterface $body
+     * @param ResponseInterface $response
+     *
+     * @return mixed
+     */
+    abstract protected function parseBody(
```
yenfryherrerafeliz, line 60: Why making `parseBody` an abstract implementation? I can see that probably is because `parseBody` depends on the trait `RpcV2ParserTrait`, but this trait usage can be moved into the abstract class, and by doing so you can leave the implementation of parseBody in the abstract class. Also, I see that parseBody is just a layer on top of calling `parseCbor`. Why not calling directly `parseCbor`?

src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
```diff
@@ -0,0 +1,161 @@
...
+            $errorCode = $this->extractErrorCode(
+                $response->getHeaderLine(self::HEADER_ERROR_TYPE)
+            );
+        }
+
+        $parsedBody = null;
+        $body = $response->getBody();
+        if (!$body->isSeekable() || $body->getSize()) {
+            $parsedBody = $this->parseBody($body, $response);
+        }
```
yenfryherrerafeliz: The condition looks for bodies that are either non-seekable or has a size, but, in the case the body is not seekable, we should take into account that the body can be read just once, which means any other logic forward that depends on reading the body again will fails or produce unexpected behaviors. I addressed something similar on [this](https://github.com/aws/aws-sdk-php/pull/3214) PR.

src/Api/Parser/AbstractRpcV2Parser.php
```diff
@@ -0,0 +1,83 @@
...
+ *    self::HEADER_SMITHY_PROTOCOL => static::$smithyProtocol
+ *
+ * @internal
+ */
+abstract class AbstractRpcV2Parser extends AbstractParser
+{
+    private const HEADER_SMITHY_PROTOCOL = 'Smithy-Protocol';
+
+    /** @var string  */
+    protected static string $smithyProtocol;
```
yenfryherrerafeliz, line 24: Can this value different other than `rpc-v2-cbor`?

src/Api/Parser/AbstractRpcV2Parser.php
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @return Result
+     */
+    private function parseResponse(
+        ResponseInterface $response,
+        Operation $operation
+    ): Result
+    {
+        $smithyProtocolHeader = $response->getHeaderLine(self::HEADER_SMITHY_PROTOCOL);
+        if ($smithyProtocolHeader !== static::$smithyProtocol) {
```
yenfryherrerafeliz, line 50: I think you should either define a default value for `$smithyProtocol` or make sure the variable is not null. This could cause undesired behaviors if for any reason the implementation omits overriding this property and assigning a value to it. Another option could be to define an `static abstract method` to force implementators to provide a value, but this is only if the value for this property will be different based on the parser, otherwise I suggest using a const.

src/Api/Parser/RpcV2CborParser.php
```diff
@@ -0,0 +1,54 @@
...
+
+    use RpcV2ParserTrait;
+
+    /**
+     * @param Service $api Service description
+     * @param CborDecoder|null $decoder Used to decode CBOR-encoded response data
+     */
+    public function __construct(
+        Service $api,
+        ?CborDecoder $decoder = null
```
yenfryherrerafeliz: I know allowing dependency injection is good for testing purposes, however, in this scenario I dont see the need of allowing this property to be passed in, unless the type annotated here is either, a class that can be extended/mocked, an interface, or an abstract class. However, in this case `CborDecoder` is a final class and can't be customized at all. From my point of view I think is safe to just initialize internally.

What do yo think?

src/Api/Serializer/RpcV2CborSerializer.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @param Service $api Service API description
+     * @param string $endpoint Endpoint to connect to
+     * @param CborEncoder|null $encoder Used to CBOR-encode PHP values
+     */
+    public function __construct(
+        Service $api,
+        string $endpoint,
+        ?CborEncoder $encoder = null
```
yenfryherrerafeliz: Same as [here](https://github.com/aws/aws-sdk-php/pull/3223/files#r2599525732), I dont see the need to allow this parameter to be passed in.

src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
stobrien89: This was just logic ported over from another query compatible error parser, but moved the case change to the `extractErrorCode()` call and removed this redundant extraction

src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
stobrien89: Because CBOR is specific to rpc-v2-cbor and not other rpc-v2 protocols

src/Api/Parser/AbstractRpcV2Parser.php
stobrien89: yes

src/Api/Parser/AbstractRpcV2Parser.php
stobrien89: this was intentional (and documented in the class docblock).  Implementers must define the protocol- it's self-enforcing

src/Api/Parser/RpcV2CborParser.php
stobrien89: I agree that it seems pointless- my initial thought was that some dependency injection was better than none, but making this extensible would be a footgun, and dependency injection here is unnecessary complexity.  I'll initialize internally

src/Api/ErrorParser/AbstractRpcV2ErrorParser.php
stobrien89: I've removed this for now and added a TODO for when changes get applied to `EventParsingIterator` to decouple it from json and abstract to the protocol error parsers.  For now, we can reasonably only expect seekable bodies from services that support CBOR (no more than 3)

src/Api/Parser/AbstractRpcV2Parser.php
yenfryherrerafeliz: I think an abstract static method would make more sense here, since I understand it is well documented, which is good, but there is not overriding enforcement here. For this looks error prone, since implementators could omit setting a value to it and the issue will be caught at runtime. However, I don`t see this as a blocker since current implementation will not fail.

stobrien89: I considered that, but I see it like this: an implementer will not get very far in testing something as crucial as a serializer or parser before discovering something like this, and in that case, an enforcement function is not necessary.  If this was a lightweight utility that wasn't an internal component, I'd probably go that route

stobrien89:
> Some comments and a general question, why not moving the Cbor folder under the Api folder?, since it is most likely going to be used by implementations under Api.

Forgot to address this top-level comment.  I'll be moving the codec implementation to the `Api` namespace

stobrien89:
Smoke tests:

```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m31.93s (13.32Mb)
```

Integ tests:

```
...............

127 scenarios (127 passed)
435 steps (435 passed)
16m35.16s (15.40Mb)

PR #3233, opened 2025-12-18
chore: update crypto integ test tags

stobrien89:
*Description of changes:*

Updates S3 crypto integ test tags


features/s3EncryptionV2/encryptionV2.feature
```diff
@@ -1,4 +1,4 @@
-@s3EncryptionV2 @integ
+@s3EncryptionV2 @integ @requiresUniqueResources
```
josecorella, line 1: should we add this to v1 and v3?
or are we landing this first to test and then add it to the others?

stobrien89: They're actually already in V1 and V3- V2 was just missing it

Issue #3239, opened 2026-01-06
[S3Client] Method "PutBucketNotificationConfiguration" returns 400 error when including "Filter" options


mad-rb:
### Describe the bug

Hi,

When submitting with the optional "Filter" settings, the route "PutBucketNotificationConfiguration" returns a 400 "MalformedXML" error:

```
'Filter' => [
    'Key' => [
        'FilterRules' => [
            [
                'Name' => 'prefix',
                'Value' => 'images/'
            ]
        ]
    ]
]
```

Without the "Filter" settings it works however.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

The route "PutBucketNotificationConfiguration" should update the notifications configuration of an S3 bucket with specific "Filter" settings.

### Current Behavior

When submitting "Filter" settings according to the structure mentioned in the docs, the route returns a 400 "MalformedXML" error.

Without the "Filter" settings it works however.

### Reproduction Steps

Submit like route docs say, with additional "Filter" settings:

```
'Filter' => [
    'Key' => [
        'FilterRules' => [
            [
                'Name' => 'prefix',
                'Value' => 'images/'
            ]
        ]
    ]
]
```

### Possible Solution

When turning on the debug flag on the S3Client I can see the submitted XML has these tags: `<Filter><Key>...` while the docs say it must be like `<Filter><S3Key>....` So I've tried changing my PHP from `'Key' => [...]` to `'S3Key' => [...]` but that results in a clear 400 with an explicit message that the key is invalid.

The config in the SDK PHP seems to have the right `locationName` in the file `/src/data/s3/2006-03-01/api-2.json.php`:

```
'NotificationConfigurationFilter' => [ 'type' => 'structure', 'members' => [ 'Key' => [ 'shape' => 'S3KeyFilter', 'locationName' => 'S3Key', ], ], ]
```

Version 3.351.0 does handle "Filter" correctly, but later versions don't. Probably due to change in 3.351.11:
https://github.com/aws/aws-sdk-php/commit/c56f4131e066857474c19a10acbddccb171e864f

### Additional Information/Context

Also see duplicate on re:Post:
https://repost.aws/questions/QUpHMmfR93SXOyJNHCgFJn2w/method-putbucketnotificationconfiguration-returns-unexpected-400-malformedxml-error-using-aws-sdk-php-when-including-filter-options

### SDK version used

3.366.2

[...]

PR #3240, opened 2026-01-06
fix: location name on structure members

yenfryherrerafeliz:
*Issue #3239*

*Description of changes:*

When a structure member has the locationName `xmlName trait` then we should use that name instead of the original name.


yenfryherrerafeliz:
We need to find a way to skip the `locationName` just when it is defined at structure level and keep it when defined at member level.

yenfryherrerafeliz:
```
...............

127 scenarios (127 passed)
435 steps (435 passed)
29m20.48s (58.03Mb)
```
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m39.67s (94.67Mb)
```

PR #3245, opened 2026-02-06
feat: support named parameters

### LINKED ISSUE #3243: PHP 8 Named Arguments Not Supported
### Describe the bug

The method `AwsClientTrait::__call` always looks for `$args[0]` but when using named arguments, they will exist as `$args['args']` instead.

For example:
```
$client->createBucket(args: ["Bucket" => "some-bucket-name"]);
```
This will throw an exception with message:
```
Aws\\S3\\S3Client::isDirectoryBucket(): Argument #1 ($bucket) must be of type string, null given, called in \/var\/www\/html\/vendor\/aws\/aws-sdk-php\/src\/S3\/S3Client.php on line 734
```

As a workaround, I've used this snippet in projects that have strict ruling around named arguments:
```php
use Aws\S3\S3Client as OriginalClient;

class S3Client extends OriginalClient {
    public function __call($name, array $args) {
        return parent::__call($name, [$args["args"]]);
    }
}
```

### Regression Issue

### Expected Behavior

Full named arguments support matching the generation documentation and types.

### Current Behavior

Exception thrown when using named arguments:
```
Aws\\S3\\S3Client::isDirectoryBucket(): Argument #1 ($bucket) must be of type string, null given, called in \/var\/www\/html\/vendor\/aws\/aws-sdk-php\/src\/S3\/S3Client.php on line 734
```

### Reproduction Steps

Attempt to run any method that ends up hitting `AwsClientTrait::__call` method

### Possible Solution

Changing:
```php
# in AwsClientTrait::__call
$params = isset($args[0]) ? $args[0] : [];
```

To:
```php
# in AwsClientTrait::__call
$params = $args['args'] ?? $args[0] ?? [];
```

_No response_

### SDK version used

3.369.25

PHP 8.4.15 - Debian GNU/Linux 13 (trixie) - Running in docker from `php:8.4-fpm` image

yenfryherrerafeliz:
Add support to named parameters by evaluating if the named parameter key is present in the arguments given to __call in AwsClientTrait.

Closes #3243 

### DIFF
--- a/tests/AwsClientTest.php
+++ b/tests/AwsClientTest.php
@@ -1018,7 +1018,7 @@ EOT
         $client->listBuckets();
     }
 
-    public function testSupportNamedArgs()
+    public function testSupportsNamedArgs()
     {
         $client = new S3Client([
             'region' => 'us-east-2',
--- a/.changes/nextrelease/feat_named_args_support.json
+++ b/.changes/nextrelease/feat_named_args_support.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "",
+        "description": "Add support to named arguments in the AwsClientTrait __call method."
+    }
+]


tests/AwsClientTest.php
stobrien89, line 1: pedantic nit, but test name should be `testSupportsNamedArgs`

stobrien89 (changes requested):
One minor comment that doesn't need a fix, but a changelog entry should be added under the `enhancement` category.

PR #3251, opened 2026-02-26
Update links to HTTPS

weshooper:
PhpStorm Inspector noticed these links should be updated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "aws/aws-sdk-php",
-    "homepage": "https://aws.amazon.com/sdkforphp",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,14 +1,14 @@
 {
     "name": "aws/aws-sdk-php",
-    "homepage": "http://aws.amazon.com/sdkforphp",
+    "homepage": "https://aws.amazon.com/sdkforphp",
```
stobrien89: Could you change this to `https://aws.amazon.com/sdk-for-php`? Turns out its been changed

PR #3262, opened 2026-03-17
chore: improve build workflows

yenfryherrerafeliz:
*Description of changes:*
Improve build workflows so they are now executed from a centralized orchestration by following a command-runner architecture. 


build/Command/AbstractCommand.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    protected function verbose(string $msg): void
+    {
+        if ($this->verbose) {
+            fwrite(STDOUT, $msg . "\n");
+        }
+    }
+
+    protected function parseOptions(array $shortOptions, array $longOptions = []): array
```
stobrien89: `getOpt()` is notoriously fragile.  We should parse the args manually like in `RemoveServiceCommand`

build/Command/GhReleaseCommand.php
```diff
@@ -0,0 +1,180 @@
...
+                $attempts++;
+                $response = $uploadClient->post("{$uploadUrl}/assets?name={$filename}", [
+                    'headers' => ['Content-Type' => "application/{$filetype}"],
+                    'body'    => Psr7\Utils::tryFopen($buildDir . "/artifacts/{$filename}", 'r')
+                ]);
+                $this->output("{$filename} uploaded to: " . json_decode($response->getBody(), true)['browser_download_url']);
+                $isSuccessful = true;
+            } catch (\GuzzleHttp\Exception\ServerException $e) {
+                $this->output("{$filename} failed to upload:");
+                var_dump($e->getMessage());
```
stobrien89: seems like `$this->error()` or `$this->output()` should be called here

build/Command/GhReleaseCommand.php
```diff
@@ -0,0 +1,180 @@
...
+                            } else {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} was unable to be deleted.");
+                            }
+                        } catch (\GuzzleHttp\Exception\ClientException $e) {
+                            $response = $e->getResponse();
+                            if ($response->getStatusCode() == 204) {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} has successfully been deleted.");
+                            } else {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} was unable to be deleted.");
+                                var_dump($e);
```
stobrien89: seems like $this->error() or $this->output() should be called here too

build/Command/OptionDocsCommand.php
```diff
@@ -0,0 +1,74 @@
...
+            if (isset($value['default']) && !is_callable($value['default'])) {
+                $modifiers[] = 'default=' . \Aws\describe_type($value['default']);
+            }
+
+            if ($modifiers) {
+                $docs = '(' . implode(', ', $modifiers) . ') ' . $docs;
+            }
+
+            $docs = '* - ' . $name . ': ' . $docs;
+            echo wordwrap($docs, 70, "\n*   ") . "\n";
```
stobrien89: should `$this->output()` be called here?

build/Command/GhReleaseCommand.php
```diff
@@ -0,0 +1,180 @@
...
+            return 1;
+        }
+
+        // Add retry middleware
+        $stack = \GuzzleHttp\HandlerStack::create();
+        $stack->push(\GuzzleHttp\Middleware::retry(
+            function ($retries, $request, $response) {
+                $statusCode = $response->getStatusCode();
+                if ($retries < self::MAX_ATTEMPTS && !in_array($statusCode, [200, 201, 202])) {
+                    echo "Attempt failed with status code {$statusCode}: "
```
stobrien89: should `$this->output()` be called here too?

build/Command/RemoveServiceCommand.php
```diff
@@ -0,0 +1,607 @@
...
+            $root . '/.changes/nextrelease/remove-' . $serviceKey . '.json',
+        ];
+
+        $output = [];
+        $exitCode = 0;
+        $escapedPaths = implode(' ', array_map('escapeshellarg', $paths));
+        exec('git add ' . $escapedPaths . ' 2>&1', $output, $exitCode);
+        if ($exitCode !== 0) {
+            $this->error('git add failed: ' . implode("\n", $output));
+            exit($exitCode);
```
stobrien89: Seeing `exit()` called in a handful of places where maybe an exception should be thrown or an error code returned to `doExecute()`?

build/Command/BuildChangelogCommand.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $params = [];
+        $params['verbose'] = $this->verbose;
+        $params['base_dir'] = $this->getProjectRoot() . '/';
+
+        $changelogBuilder = new ChangelogBuilder($params);
+
+        $changelogBuilder->buildChangelog();
+
+        // Omit fixEndpointFile() call - method doesn't exist on ChangelogBuilder
```
stobrien89, line 34: is this comment intentional?

build/Command/PackageCommand.php
```diff
@@ -0,0 +1,86 @@
...
+        $burgomaster->createZip($buildDir . "/artifacts/aws.zip");
+        $burgomaster->createPhar(
+            $buildDir . "/artifacts/aws.phar",
+            null,
+            $autoloaderFilename,
+            'aws-' . \Aws\Sdk::VERSION . '.phar'
+        );
+
+        $burgomaster->startSection('test-phar');
+        $burgomaster->exec('php ' . $buildDir . '/WorkflowCommandRunner.php test-phar');
```
stobrien89, line 79: could the runner path be moved to a constant?

build/Command/PharTestRunnerCommand.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        return 'Runs the PHPUnit + Behat test suites against the phar artifacts.';
+    }
+
+    public function getUsage(): string
+    {
+        return 'php build/WorkflowCommandRunner.php phar-test-runner [-- phpunit/behat options]';
+    }
+
+    protected function doExecute(array $args): int
```
stobrien89, line 22: does this actually execute the tests?

stobrien89 (changes requested):
A couple of changes requested- Not sure how feasible this is for every class, but this is a lot of new logic with no unit tests.  The `RemoveServiceCommand` class should definitely have some unit tests

build/Command/BuildChangelogCommand.php
yenfryherrerafeliz: Yes, I just want to make sure it keeps documented for future revisions since that method is called [here](https://github.com/aws/aws-sdk-php/blob/18035d80bab38c962af8580307a787bf4e15cc47/build/changelog/build-changelog.php#L17), in the old script file,  but is not a member of `ChangelogBuilder`.

yenfryherrerafeliz (commented):
>Not sure how feasible this is for every class, but this is a lot of new logic with no unit tests. The RemoveServiceCommand class should definitely have some unit tests.

Yes, maybe we should have test coverage, however, the reason why I initially decided not to do it was because those scripts did not have either before this improvements, and we are not changing or introducing functionalities, except for the service removal, but just migrating to a different structural and functional architecture. I can definitely add test coverage if needed.

build/Command/PackageCommand.php
yenfryherrerafeliz: I get the build path from a helper static method shared across all commands and moving this into a cost then will force me to not use this helper method.

build/Command/GhReleaseCommand.php
```diff
@@ -0,0 +1,180 @@
...
+                            } else {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} was unable to be deleted.");
+                            }
+                        } catch (\GuzzleHttp\Exception\ClientException $e) {
+                            $response = $e->getResponse();
+                            if ($response->getStatusCode() == 204) {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} has successfully been deleted.");
+                            } else {
+                                $this->output("Failed upload of {$asset['name']} at {$asset['browser_download_url']} was unable to be deleted.");
+                                $this->error($e);
```
stobrien89: Needs `$e->getMessage()` to avoid type mismatch in `error()`

build/Command/RemoveServiceCommand.php
```diff
@@ -0,0 +1,586 @@
...
+     *   2. Match against namespace values
+     *   3. Match against serviceIdentifier values
+     *
+     * @param string $input The user-provided service identifier
+     * @return array|null Associative array with serviceKey, namespace, serviceIdentifier or null
+     */
+    private function resolveService(string $input): ?array
+    {
+        $manifestPath = $this->getProjectRoot() . '/src/data/manifest.json';
+        $manifest = json_decode(file_get_contents($manifestPath), true);
```
stobrien89, line 131: nit: you could use the `.php` version instead of decoding, but not required

build/Command/RemoveServiceCommand.php
```diff
@@ -0,0 +1,586 @@
...
+        file_put_contents(
+            $jsonPath,
+            json_encode($manifest, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES) . "\n"
+        );
+
+        // Regenerate manifest.json.php
+        $phpContent = "<?php\n"
+            . "// This file was auto-generated from sdk-root/src/data/manifest.json\n"
+            . "return " . var_export($manifest, true) . ";\n";
+        file_put_contents($phpPath, $phpContent);
```
stobrien89: Could we use `JsonCompiler` here for consistency?

build/Command/RemoveServiceCommand.php
```diff
@@ -0,0 +1,586 @@
...
+    /**
+     * Removes the namespace from the grandfathered-services.json.php file.
+     *
+     * Loads the PHP file via include, removes the exact-match namespace from the
+     * "grandfathered-services" array, and regenerates the PHP file content.
+     * Logs a warning if the namespace is not present.
+     *
+     * @param string $namespace The PHP namespace segment (e.g., "S3", "ACMPCA")
+     */
+    private function removeFromGrandfatheredPhp(string $namespace): void
```
stobrien89: Could we combine this with the method above (like `removeFromManifest`) and regenerate using `JsonCompiler`?

stobrien89 (changes requested):

build/Command/RemoveServiceCommand.php
yenfryherrerafeliz: That is true, but I think it is safe to do it against the JSON file because it is guaranteed it will always exists. If you think must be done against the PHP file I can change it.

build/Command/RemoveServiceCommand.php
yenfryherrerafeliz: Nice suggestion. It simplifies and reduces the code. Ty!

stobrien89 (approved):
Approved.  Let's hold off on merging until we've both verified artifacts.  Nice job on this- much cleaner.

stobrien89:
Great job on this, but we don't have the capacity to do the internal build changes alongside it right now. Might re-open at some point in the future.  Thank you for everything you've contributed 🫡

Issue #1273, opened 2017-04-30
Memory Leakage


sm2017:
I have a long execute php script (Executes for months)

I find that we have a memory leakage in aws-sdk-php , When I use GetObjectAsync method of S3 client , and after minutes I run the garbage collector , these objects are in memory

	GuzzleHttp\Psr7\Request
	GuzzleHttp\Psr7\Stream
	GuzzleHttp\Psr7\Uri
	Aws\RetryMiddleware
	Aws\S3\PutObjectUrlMiddleware
	Aws\S3\PermanentRedirectMiddleware
	Aws\Command
	Aws\HandlerList

kstich:
Hi @sm2017, could you please provide a little more information, or potentially a small reproduction, of your issue at hand? That will help us to investigate further what's causing your issue and either guide adjustments or work on updates. Thanks!

sm2017:
Run the following code

```php
<?php
require 'vendor/autoload.php';

$sdk = new Aws\Sdk([
    'version' => 'latest',
    'region'  => 'us-west-2'
]);

$s3 = $sdk->createS3();
$promise = $s3->getObjectAsync([
    'Bucket' => 'bucket',
    'Key' => 'memoryLeakage',
]);

$promise->wait();
echo "Before gc_collect_cycles()\n\n\n";
gc_collect_cycles();
echo "\n\n\nAfter gc_collect_cycles()";
```


Please add __destruct method to these classes

	GuzzleHttp\Psr7\Request
	GuzzleHttp\Psr7\Stream
	GuzzleHttp\Psr7\Uri
	Aws\RetryMiddleware
	Aws\S3\PutObjectUrlMiddleware
	Aws\S3\PermanentRedirectMiddleware
	Aws\Command
	Aws\HandlerList


```php
public function  __destruct(){
    echo __METHOD__."\n";
}
```

With no memory leakage you must not see anything between ```Before gc_collect_cycles()``` and ```After gc_collect_cycles()```

As my script is long execute , I see that these objects never free up memory even after days

stof:
Cyclic objects graph are not collected as soon as the variable goes out of scope (as the refcount never reaches 0), but the next time PHP decides to run the GC to delete unused cyclic graphs. `gc_collect_cycles` is a way to force PHP to run it now instead of waiting for PHP to run it (which may be delayed until PHP considers that it needs to free some memory).

An actual memory leakage would *not* release memory during GC. It would just keep increasing.

[...]

PR #3268, opened 2026-04-09
fix: break circular reference cycles

smilkuri:
*Issue #, if available:*
https://github.com/aws/aws-sdk-php/issues/1273

*Description of changes:*
`PresignUrlMiddleware` and `EndpointDiscoveryMiddleware` hold a direct $client reference back to the AwsClient through its HandlerList, creating a circular reference cycle that PHP's refcounting cannot collect. Replace $client with \WeakReference in both classes. The client outlives its own handler chain, so the weak reference resolves for the lifetime of the middleware.

### DIFF
--- a/src/EndpointDiscovery/EndpointDiscoveryMiddleware.php
+++ b/src/EndpointDiscovery/EndpointDiscoveryMiddleware.php
@@ -21,7 +21,7 @@ class EndpointDiscoveryMiddleware
     private static $discoveryCooldown = 60;
 
     private $args;
-    private $client;
+    private \WeakReference $client;
     private $config;
     private $discoveryTimes = [];
     private $nextHandler;
@@ -53,7 +53,7 @@ class EndpointDiscoveryMiddleware
         $config
     ) {
         $this->nextHandler = $handler;
-        $this->client = $client;
+        $this->client = \WeakReference::create($client);
         $this->args = $args;
         $this->service = $client->getApi();
         $this->config = $config;
@@ -91,7 +91,7 @@ class EndpointDiscoveryMiddleware
                 $identifiers = $this->getIdentifiers($op);
 
                 $cacheKey = $this->getCacheKey(
-                    $this->client->getCredentials()->wait(),
+                    $this->client->get()->getCredentials()->wait(),
                     $cmd,
                     $identifiers
                 );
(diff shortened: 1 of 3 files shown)


src/PresignUrlMiddleware.php
stobrien89, line 57: Did some digging and for us, the cycle really starts here. Since the SDK only uses `wrap()` to add this middleware to a client's `HandlerList`, we can make this change and leave the rest of the class as-is:   

```php
public static function wrap(
    AwsClientInterface $client,
    $endpointProvider,
    array $options = []
) {
    $clientRef = \WeakReference::create($client);
    return function (callable $handler) use ($endpointProvider, $clientRef, $options) {
        return new PresignUrlMiddleware($options, $endpointProvider, $clientRef->get(), $handler);
    };
}
```

The 100% foolproof way to do this would be to apply the above change along with your changes, but for all intents and purposes, these are internal components that we don't have to worry about users instantiating independently.

src/EndpointDiscovery/EndpointDiscoveryMiddleware.php
stobrien89, line 48: Same comment as the one I left on `PresignUrlMiddleware`:

```php
    public static function wrap(
        $client,
        $args,
        $config
    ) {
        $clientRef = \WeakReference::create($client);
        return function (callable $handler) use (
            $clientRef,
            $args,
            $config
        ) {
            return new static(
                $handler,
                $clientRef->get(),
                $args,
                $config
            );
        };
    }
```

We can make just this change and leave the rest alone.

stobrien89 (changes requested):
Thanks for the contribution- just two comments.  I led you down the most invasive path

stobrien89 (approved):
Thanks @smilkuri, looks good.  Before merging, we'll need you to push up a changelog entry, which goes in `.changes/nextrelease`- you'll need to create the `nextrelease` dir.  There you'll need to create a json file of any name (e.g. `fix-memory-leak.json`) with the following structure:

```json
[
    {
        "type": "enhancement",
        "category": "",
        "description": "<description_of_changes>"
    }
]
```

The reason we'll use a blank namespace for this is because when empty, the `category` field defaults to the `Aws` namespace, which is the most appropriate here because the changes apply to two unrelated classes.

stobrien89 (approved):

stobrien89:
Thanks for the contribution, @smilkuri! :shipit:

PR #3269, opened 2026-04-13
feat: new retries implementation

yenfryherrerafeliz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -31,6 +31,8 @@ use Aws\Exception\AwsException;
 use Aws\Exception\InvalidRegionException;
 use Aws\Retry\ConfigurationInterface as RetryConfigInterface;
 use Aws\Retry\ConfigurationProvider as RetryConfigProvider;
+use Aws\Retry\Standard\OptIn as NewRetriesOptIn;
+use Aws\Retry\Standard\RetryMiddleware as StandardRetryMiddleware;
 use Aws\Signature\SignatureProvider;
 use Aws\Token\Token;
 use Aws\Token\TokenInterface;
@@ -547,28 +549,42 @@ class ClientResolver
     public static function _apply_retries($value, array &$args, HandlerList $list)
     {
         // A value of 0 for the config option disables retries
-        if ($value) {
-            $config = RetryConfigProvider::unwrap($value);
-
-            if ($config->getMode() === 'legacy') {
-                // # of retries is 1 less than # of attempts
-                $decider = RetryMiddleware::createDefaultDecider(
-                    $config->getMaxAttempts() - 1
-                );
-                $list->appendSign(
-                    Middleware::retry($decider, null, $args['stats']['retries']),
-                    'retry'
-                );
-            } else {
-                $list->appendSign(
-                    RetryMiddlewareV2::wrap(
-                        $config,
-                        ['collect_stats' => $args['stats']['retries']]
-                    ),
-                    'retry'
-                );
-            }
+        if (!$value) {
+            return;
         }
+
+        $config = RetryConfigProvider::unwrap($value);
+
+        if ($config->getMode() === 'legacy') {
+            // # of retries is 1 less than # of attempts
+            $decider = RetryMiddleware::createDefaultDecider(
+                $config->getMaxAttempts() - 1
+            );
+            $list->appendSign(
+                Middleware::retry($decider, null, $args['stats']['retries']),
+                'retry'
+            );
+            return;
+        }
+
+        if (NewRetriesOptIn::isEnabled()) {
+            $list->appendSign(
+                StandardRetryMiddleware::wrap($config, [
+                    'collect_stats' => $args['stats']['retries'],
+                    'service'       => $args['service'],
+                ]),
+                'retry'
+            );
+            return;
+        }
+
+        $list->appendSign(
+            RetryMiddlewareV2::wrap(
+                $config,
+                ['collect_stats' => $args['stats']['retries']]
+            ),
+            'retry'
+        );
     }
 
     public static function _apply_defaults($value, array &$args, HandlerList $list)
--- a/src/ClientResolver.php
+++ b/src/ClientResolver.php
@@ -31,8 +31,8 @@ use Aws\Exception\AwsException;
 use Aws\Exception\InvalidRegionException;
 use Aws\Retry\ConfigurationInterface as RetryConfigInterface;
 use Aws\Retry\ConfigurationProvider as RetryConfigProvider;
-use Aws\Retry\Standard\OptIn as NewRetriesOptIn;
-use Aws\Retry\Standard\RetryMiddleware as StandardRetryMiddleware;
+use Aws\Retry\V3\OptIn as NewRetriesOptIn;
+use Aws\Retry\V3\RetryMiddleware as RetryV3Middleware;
 use Aws\Signature\SignatureProvider;
 use Aws\Token\Token;
 use Aws\Token\TokenInterface;
@@ -569,7 +569,7 @@ class ClientResolver
 
         if (NewRetriesOptIn::isEnabled()) {
             $list->appendSign(
-                StandardRetryMiddleware::wrap($config, [
+                RetryV3Middleware::wrap($config, [
                     'collect_stats' => $args['stats']['retries'],
                     'service'       => $args['service'],
                 ]),
--- a/src/DynamoDb/DynamoDbClient.php
+++ b/src/DynamoDb/DynamoDbClient.php
@@ -7,8 +7,14 @@ use Aws\ClientResolver;
 use Aws\Exception\AwsException;
 use Aws\HandlerList;
 use Aws\Middleware;
+use Aws\Retry\Configuration as RetryConfiguration;
+use Aws\Retry\ConfigurationInterface as RetryConfigurationInterface;
+use Aws\Retry\ConfigurationProvider as RetryConfigurationProvider;
+use Aws\Retry\Standard\OptIn as NewRetriesOptIn;
+use Aws\Retry\Standard\RetryMiddleware as StandardRetryMiddleware;
 use Aws\RetryMiddleware;
 use Aws\RetryMiddlewareV2;
+use GuzzleHttp\Promise\Create;
 
 /**
  * This client is used to interact with the **Amazon DynamoDB** service.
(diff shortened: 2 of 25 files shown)


yenfryherrerafeliz:
I found a few some minor issues that I am addressing. One of them is that for dynamodb is still defaulting the mode to legacy due to how the configuration is created for dynamodb. Also some unused variables that I need to remove.

src/DynamoDb/DynamoDbClient.php
```diff
@@ -130,16 +130,36 @@
...
         $args['retries']['fn'] = [__CLASS__, '_applyRetryConfig'];
         $args['api_provider']['fn'] = [__CLASS__, '_applyApiProvider'];
 
         return $args;
     }
 
+    /** @internal */
+    public static function _defaultRetries(): callable
+    {
+        return \Aws\Retry\ConfigurationProvider::chain(
```
stobrien89: can `defaultProvider()` not be used here?

src/RetryMiddlewareV2.php
```diff
@@ -201,20 +174,96 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
...
             }
 
-            $delayBy = $delayer($attempts++);
+            // Acquire retry quota
+            $isThrottling = $this->isThrottlingError($value);
+            $acquired = $this->quotaManager->acquireRetryQuota($isThrottling);
+            if ($acquired === false) {
+                // Long-polling: sleep and return without retrying
+                if ($this->isLongPollingOperation($cmd)) {
+                    $sleepMs = (int) ($this->baseDelay * 1000);
```
stobrien89: I might be missing how `$baseDelay` factors in, but the pseudocode for long polling operations is:

```
sleep(ExponentialBackoff(attempts - 1))

File: src/RetryMiddlewareV2.php [hunk 3]
```diff
@@ -239,41 +288,81 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
...
+        $maxBackoffSeconds = $this->maxBackoff / 1000.0;
+        $delaySeconds = $jitter * min($base * pow(2, $i), $maxBackoffSeconds);
+        $delayMs = (int) round($delaySeconds * 1000);
+
+        // Check x-amz-retry-after header
+        if ($value instanceof AwsException) {
+            $response = $value->getResponse();
+            if ($response?->hasHeader('x-amz-retry-after')) {
+                $retryAfterStr = $response->getHeaderLine('x-amz-retry-after');
+                if (is_numeric($retryAfterStr)) {
```
stobrien89: nit: `is_numeric` is a little permissive for the specified "is an integer (`[0-9]+`)". I'd consider using `preg_match('/^\d+$/', $retryAfterStr)` here

src/RetryMiddlewareV2.php
```diff
@@ -239,41 +288,81 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
...
+                $jitter = mt_rand(0, $max) / $max;
+            }
+        }
+
+        $maxBackoffSeconds = $this->maxBackoff / 1000.0;
+        $delaySeconds = $jitter * min($base * pow(2, $i), $maxBackoffSeconds);
+        $delayMs = (int) round($delaySeconds * 1000);
+
+        // Check x-amz-retry-after header
+        if ($value instanceof AwsException) {
```
stobrien89: For 200 error responses (S3 and others), this header could be missed

src/RetryMiddlewareV2.php
```diff
@@ -38,155 +42,124 @@ class RetryMiddlewareV2
         'EC2ThrottledException'                     => true,
     ];
 
-    private static $standardTransientErrors = [
+    private static array $standardTransientErrors = [
         'RequestTimeout'            => true,
         'RequestTimeoutException'   => true,
     ];
```
stobrien89: I would double-check that we handle S3's 200 `InternalError` on `S3Client` since it's called out in the spec

src/RetryMiddlewareV2.php
```diff
@@ -38,155 +42,124 @@ class RetryMiddlewareV2
...
             CommandInterface $command,
-            $result
-        ) use ($options, $quotaManager, $retryCurlErrors, $maxAttempts) {
-
-            // Release retry tokens back to quota on a successful result
-            $quotaManager->releaseToQuota($result);
-
-            // Allow command-level option to override this value
-            // # of attempts = # of retries + 1
-            $maxAttempts = (null !== $command['@retries'])
```
stobrien89, line 99: guessing command-level `$maxAttempts` is handled elsewhere- do we need to pass `$command` anymore?

src/Retry/ConfigurationInterface.php
stobrien89, line 1: I'd stay away from changing method signatures on interfaces for backward compatibility

tests/RetryMiddlewareV2Test.php
```diff
@@ -1140,4 +1151,511 @@ public function testReportsHttpStatsForEachRequestEvenIfRetryStatsDisabled()
             ['fizz' => 'buzz'],
         ], $httpStats);
     }
+
+    /**
+     * Tests standard mode behavior with deterministic jitter.
+     */
+    #[DataProvider('standardModeNewTestCasesProvider')]
+    public function testStandardModeRetry(
```
stobrien89: There's another method `testRetriesForStandardMode` that should be looked at for overlap.  Also seeing some missing cases in the provider.  A couple existing cases had their values changed

src/ClientResolver.php
```diff
@@ -560,10 +560,17 @@ public static function _apply_retries($value, array &$args, HandlerList $list)
                     'retry'
                 );
             } else {
+                $retryOptions = [
+                    'collect_stats' => $args['stats']['retries'],
+                    'service' => $args['service'],
+                ];
+                if ($args['service'] === 'sts') {
```
stobrien89: This can be applied on the `StsClient` class

src/RetryMiddlewareV2.php
```diff
@@ -38,155 +42,124 @@ class RetryMiddlewareV2
...
-    private $rateLimiter;
-
-    public static function wrap($config, $options)
+    private static array $longPollingOperations = [
+        'sqs' => ['ReceiveMessage' => true],
+        'states' => ['GetActivityTask' => true],
+        'swf' => ['PollForActivityTask' => true, 'PollForDecisionTask' => true],
+    ];
+
+    private float $baseDelay;
```
stobrien89: Recommend appending `Seconds` to the variable name since we're mixing units

src/RetryMiddlewareV2.php
```diff
@@ -38,155 +42,124 @@ class RetryMiddlewareV2
...
+    private bool $collectStats;
+    private ?\Closure $customDecider;
+    private ?\Closure $delayer;
+    private ?float $exponentialBase;
+    private int $maxAttempts;
+    private int $maxBackoff;
```
stobrien89: Recommend appending `Ms` to variable name

src/RetryMiddlewareV2.php
```diff
@@ -201,20 +174,96 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
...
+                return $this->bindStatsToReturn($value, $requestStats);
+            }
+            $capacityUsed = $acquired;
+
+            // Compute delay
+            $attemptIndex = $attempts - 1; // 0-based for backoff formula
+            if ($this->delayer !== null) {
+                $delayBy = ($this->delayer)($attempts);
+            } else {
+                $delayBy = $this->computeRetryDelay($attemptIndex, $isThrottling, $value);
```
stobrien89: recommend appending `Ms` to variable name

src/RetryMiddlewareV2.php
```diff
@@ -21,7 +21,11 @@ class RetryMiddlewareV2
 {
     use RetryHelperTrait;
 
-    private static $standardThrottlingErrors = [
+    const THROTTLING_BASE_DELAY = 1.0;
+    const DEFAULT_BASE_DELAY = 0.05;
+    const DYNAMODB_BASE_DELAY = 0.025;
```
stobrien89: Recommend adding visibility modifiers and appending `_SECONDS` for clarity

stobrien89 (changes requested):

src/DynamoDb/DynamoDbClient.php
yenfryherrerafeliz: I though on it, but how would you set the max attempts to four?, which is what is being handled here.

src/Retry/ConfigurationInterface.php
yenfryherrerafeliz: Do you think by making getMaxAttempts to support both returns type, int and string would be sufficient to keep backward compatibility?, I think getMaxAttempts returning string was a mistake we probably did.

src/RetryMiddlewareV2.php
yenfryherrerafeliz: Not really. Max attempts is handled at invocation level now.

tests/RetryMiddlewareV2Test.php
yenfryherrerafeliz: It is actually expected to have values changed for existent cases, and the reason is because the retry cost changed from 5 to 14 unless is a throttling error, which would be 5.

src/RetryMiddlewareV2.php
```diff
@@ -239,41 +298,93 @@ public function __invoke(CommandInterface $cmd, RequestInterface $req)
...
+        );
+
+        // Check x-amz-retry-after header
+        $retryAfterHeader = null;
+        if ($value instanceof AwsException) {
+            $response = $value->getResponse();
+            $retryAfterHeader = $response?->getHeaderLine(
+                self::RETRY_AFTER_HEADER
+            ) ?? null;
+        } elseif($value instanceof ResultInterface) {
```
stobrien89: Nit: space needed after `elseif`

src/Retry/ConfigurationInterface.php
stobrien89: I think the docblock was just incorrect- `intval()` returns an int, so it should have been returning an int previously

stobrien89 (approved):
Just one nit, but approved otherwise. Let's hold off on merging until we find out when customers will be notified about this

src/Retry/Standard/OptIn.php
```diff
@@ -0,0 +1,35 @@
...
+final class OptIn
+{
+    public const ENV = 'AWS_NEW_RETRIES_2026';
+
+    private static ?bool $enabled = null;
+
+    public static function isEnabled(): bool
+    {
+        if (self::$enabled === null) {
+            self::$enabled = getenv(self::ENV) === 'true';
```
stobrien89: should be a bit more permissive with the value here.  See `\Aws\boolean_value()`

src/DynamoDb/DynamoDbClient.php
```diff
@@ -159,40 +199,99 @@ public function registerSessionHandler(array $config = [])
...
+
+        self::appendStandardModeRetries($config, $args, $list);
+    }
+
+    private static function appendLegacyModeRetries(
+        $value,
+        RetryConfigurationInterface $config,
+        array &$args,
+        HandlerList $list
+    ): void {
```
stobrien89: Seeing a few places in multi-line method signatures where the opening brace needs to be moved down:


```suggestion
    private static function appendLegacyModeRetries(
        $value,
        RetryConfigurationInterface $config,
        array &$args,
        HandlerList $list
    ): void 
    {
```

src/DynamoDb/DynamoDbClient.php
```diff
@@ -159,40 +199,99 @@ public function registerSessionHandler(array $config = [])
...
+
+    private static function resolveLegacyModeMaxRetries(
+        $value,
+        RetryConfigurationInterface $config
+    ): int {
+        if (
+            NewRetriesOptIn::isEnabled()
+            && is_array($value)
+            && !isset($value['max_attempts'])
+        ) {
```
stobrien89, line 256: I think this is the only one, but I'd check for others:


```suggestion
        if (NewRetriesOptIn::isEnabled()
            && is_array($value)
            && !isset($value['max_attempts'])
        ) {
```

src/ClientResolver.php
```diff
@@ -31,6 +31,8 @@
 use Aws\Exception\InvalidRegionException;
 use Aws\Retry\ConfigurationInterface as RetryConfigInterface;
 use Aws\Retry\ConfigurationProvider as RetryConfigProvider;
+use Aws\Retry\Standard\OptIn as NewRetriesOptIn;
+use Aws\Retry\Standard\RetryMiddleware as StandardRetryMiddleware;
```
stobrien89: recommend a rename here since this middleware also handles `adaptive` mode

stobrien89 (changes requested):
Looks good- clean way to handle this.  A few small things and one structural recommendation: `Retry\Standard` implies there will only be logic handling the `standard` retry mode, but the new middleware also handles `adaptive`.  I would just put all of these in the `Retry` namespace and rename `QuotaManager` to something like `RetryQuota`.

src/Retry/Standard/OptIn.php
yenfryherrerafeliz: According to the spec writer we should strictly consider lowercase `true`.

stobrien89 (approved):
Looks good- thanks @yenfryherrerafeliz!

yenfryherrerafeliz:
With `AWS_NEW_RETRIES_2026` set to `false`:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m38.24s (97.16Mb)
```

```
...................................................

135 scenarios (135 passed)
471 steps (471 passed)
27m57.59s (684.23Mb)
```

With `AWS_NEW_RETRIES_2026` set to `true`:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m39.89s (97.22Mb)
```

```
...................................................

135 scenarios (135 passed)
471 steps (471 passed)
26m10.38s (622.94Mb)
```

PR #3272, opened 2026-04-15
fix: add path validation

yenfryherrerafeliz:
*Description of changes:*
Enhance path validation for safety in FileDownloadHandler.

### DIFF
--- a/.changes/nextrelease/enhance-path-validation.json
+++ b/.changes/nextrelease/enhance-path-validation.json
@@ -1,7 +1,7 @@
 [
     {
-        "type": "api-change",
-        "category": "",
+        "type": "enhancement",
+        "category": "S3/S3Transfer",
         "description": "Enhance path validation in FileDownloadHandler and ResumableDownload implementation to avoid usage of `..` within the destination path."
     }
 ]
--- a/.changes/nextrelease/enhance-path-validation.json
+++ b/.changes/nextrelease/enhance-path-validation.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "api-change",
+        "category": "",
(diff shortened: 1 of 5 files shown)


.changes/nextrelease/enhance-path-validation.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "type": "api-change",
```
stobrien89: I'd categorize this as `enhancement` and put the category as `S3/S3Transfer`

src/S3/S3Transfer/Utils/FileDownloadHandler.php
stobrien89, line 1: Since `fromFile()` knows the resume file's location, I would add a confinement check there that resolves `dirname($data['destination'])` against `dirname($filePath)` using `realpath()` and verify the destination's parent starts with the resume file's parent. The SDK generates the resume file at `$destination.s3tmp.XXXX.resume`, so they should always be in the same directory tree.

Keep the `..` rejection in `fromJson()` as a fast check. Add the `realpath()` confinement in `fromFile()` after deserializing the data and before returning the object. `fromJson()` can't do confinement because it doesn't have a file path to confine against.

stobrien89 (changes requested):
Just one nit and one suggestion

src/S3/S3Transfer/Utils/FileDownloadHandler.php
yenfryherrerafeliz: The API supports specifying a custom location for the generated resume file. Should we consider removing support for this parameter so we can add this validation and keep the flow safer?

PR #3276, opened 2026-04-23
feat: endpoints bdd

yenfryherrerafeliz:
*Description of changes:*
- Add BDD-based endpoint resolution alongside the existing tree ruletset evaluator.
    - Introduce a Bdd package under EndpointV2 namespace containing utilities
      and components used for resolving an endpoint through BDDs.
    - Enhance the EndpointDefinitionProvider to resolve the endpoint rule definitions
      by giving preference to BDDs "endpoint-bdd.json" rules over tree based rules.
    - Enhance EndpointProviderV2 to support both, BDD endpoint resolver and Tree
      endpoint resolver by also giving preference to BDD resolution.
      How is this done:
      - The parameter $ruleset now supports instances of BddRuleset, besides of array to preserve existent
        behavior, and when an instance of BddReulset is present then a BDD evaluator is instantiated which
        later is used to resolve the endpoint.
      - Otherwise, if an array or a Ruleset is passed in then we resolve the endpoint
        with the tree based endpoint resolution, which is the current behavior.
    - Add a new method "getActiveParameters" in EndpointProviderV2 that is used by
      the EndpointV2Middleware to get the active parameters, which internally
      it just evaluates which rulset instance we should be getting the parameters from,
      either from $bddRuleset if not null or the tree ruleset $ruleset.
    - Enhance EndpointV2Middleware to consume the new getActiveParameters.

## Benchmark
```
PHP 8.5.5 | opcache: off | jit: off                                                                                                                                                              
                                                                                                                                                                                                                                         
  ┌────────────────────────┬───────────┬──────────┬──────────┬────────┐                                                                                                                                                                  
  │       Benchmark        │  Metric   │   Tree   │   BDD    │ Delta  │                                                                                                                                                                  
  ├────────────────────────┼───────────┼──────────┼──────────┼────────┤                                                                                                                                                                  
  │ Parse (n=100)          │ median    │ 61.92 µs │ 21.83 µs │ -64.7% │                                                                                                                                                                  
  ├────────────────────────┼───────────┼──────────┼──────────┼────────┤                                                                                                                                                                  
  │ Parse (n=100)          │ p99       │ 85.17 µs │ 34.54 µs │ -59.4% │                                                                                                                                                                  
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Memory                 │ rss delta │ 69.7 KiB  │ 40.8 KiB │ -41.5% │                                                                                                                                                                 
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Cold resolve (389×100) │ median    │ 16.46 µs  │ 12.67 µs │ -23.0% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Cold resolve (389×100) │ p99       │ 33.96 µs  │ 24.29 µs │ -28.5% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Cold resolve (389×100) │ mean      │ 16.98 µs  │ 13.30 µs │ -21.7% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Cold resolve (389×100) │ thrpt/s   │ 58,878    │ 75,173   │ +27.7% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Warm resolve (n=50000) │ median    │ 0.58 µs   │ 0.54 µs  │ -7.0%  │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Warm resolve (n=50000) │ p99       │ 0.79 µs   │ 0.71 µs  │ -10.5% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ Warm resolve (n=50000) │ mean      │ 0.57 µs   │ 0.56 µs  │ -1.3%  │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ On-disk size           │ .json.php │ 148.3 KiB │ 43.2 KiB │ -70.9% │
  ├────────────────────────┼───────────┼───────────┼──────────┼────────┤                                                                                                                                                                 
  │ On-disk size           │ .json     │ 896.3 KiB │ 58.4 KiB │ -93.5% │
  └────────────────────────┴───────────┴───────────┴──────────┴────────┘ 
```


src/EndpointV2/Bdd/BddResultResolver.php
```diff
@@ -0,0 +1,139 @@
...
+            );
+        }
+
+        return $this->buildEndpoint($result['endpoint'], $inputParameters);
+    }
+
+    /**
+     * @throws UnresolvedEndpointException
+     */
+    public function resolveNoMatch(array $inputParameters): RulesetEndpoint
```
stobrien89: Should be using `never` return type for this and `throwError()` since both unconditionally throw

src/EndpointV2/EndpointProviderV2.php
```diff
@@ -66,4 +122,16 @@ private function hashInputParameters($inputParameters)
...
+    /**
+     * @return array|null
+     */
+    public function getActiveParameters(): ?array
+    {
+        if ($this->bddRuleset !== null) {
+            return $this->bddRuleset->getParameters();
+        }
+
+        return $this->ruleset?->getParameters();
```
stobrien89: will this ever be null?

src/EndpointV2/Bdd/BddNodeDecoder.php
```diff
@@ -0,0 +1,68 @@
...
+     * Decodes `$encoded` and verifies that the byte count matches
+     * `$expectedNodeCount`. Returns a flat int array of length
+     * `3 * $expectedNodeCount`.
+     *
+     * @throws UnresolvedEndpointException when the payload is not valid base64
+     *     or its length does not match the declared node count.
+     */
+    public static function decode(string $encoded, int $expectedNodeCount): array
+    {
+        $bytes = base64_decode($encoded, true);
```
stobrien89, line 34: Probably don't have to worry about this, but are we guaranteed not to have any padding/whitespace in the JSON?

src/EndpointV2/EndpointDefinitionProvider.php
```diff
@@ -2,17 +2,72 @@
...
  * and testing.
  */
 class EndpointDefinitionProvider
 {
+    /**
+     * Returns the endpoint definition the client should use to resolve
+     * endpoints. When a service ships a compiled BDD alongside the
+     * legacy ruleset, the BDD takes precedence.
+     */
+    public static function getEndpointDefinition($service, $apiVersion, $baseDir = null)
```
stobrien89: Not seeing this used anywhere internally- is this needed?

tests/EndpointV2/EndpointDefinitionProviderTest.php
```diff
@@ -73,4 +74,105 @@ public function testThrowsExceptionOnMissingFiles($type)
...
+                EndpointDefinitionProvider::getPartitions(),
+                $baseDir
+            );
+
+            $this->assertInstanceOf(
+                \Aws\EndpointV2\Bdd\BddRuleset::class,
+                $parsed
+            );
+        } finally {
+           TestsUtility::cleanUpDir($baseDir);;
```
stobrien89: nit: double semicolon

tests/EndpointV2/EndpointDefinitionProviderTest.php
```diff
@@ -73,4 +74,105 @@ public function testThrowsExceptionOnMissingFiles($type)
...
+            $definition = EndpointDefinitionProvider::getEndpointDefinition(
+                'widget',
+                '2024-01-01',
+                $baseDir
+            );
+
+            $this->assertArrayHasKey('nodes', $definition);
+            $this->assertArrayHasKey('root', $definition);
+        } finally {
+           TestsUtility::cleanUpDir($baseDir);;
```
stobrien89: nit: double semicolon

src/EndpointV2/Bdd/BddRuleset.php
```diff
@@ -0,0 +1,126 @@
...
+                );
+            }
+        }
+
+        $this->parameters = $this->buildParameters($definition['parameters'] ?? []);
+        $this->conditions = $definition['conditions'];
+        $this->results = $definition['results'];
+        $this->root = (int) $definition['root'];
+
+        $nodeCount = isset($definition['nodeCount'])
```
stobrien89: Are there ever cases where a valid BDD ruleset will have 0 nodes?

src/EndpointV2/EndpointProviderV2.php
```diff
@@ -50,13 +102,17 @@ public function resolveEndpoint(array $inputParameters)
...
+        $endpoint = $this->bddEvaluator !== null
+            ? $this->bddEvaluator->evaluate($inputParameters)
+            : $this->ruleset->evaluate($inputParameters);
+
         if ($endpoint === false) {
             throw new UnresolvedEndpointException(
                 'Unable to resolve an endpoint using the provider arguments: '
                 . json_encode($inputParameters)
             );
         }
```
stobrien89, line 116: nit: pretty sure this is dead code if using BDD, might be worth a comment

stobrien89 (changes requested):
Looking good overall, just some comments on small stuff.

src/EndpointV2/EndpointProviderV2.php
yenfryherrerafeliz: Technically, it should not be null when falling back to the tree ruleset. Will remove the nullsafe operator.

src/EndpointV2/Bdd/BddNodeDecoder.php
yenfryherrerafeliz: I would prefer to trust that the encoded value is correct. `strict` will just fail or return false if within the encoded value non-base64 characters are found.

src/EndpointV2/Bdd/BddRuleset.php
yenfryherrerafeliz: Mmm, I don`t think so. According to the smithy specs all those fields should be present. I will remove the unnecessary check.

src/EndpointV2/EndpointDefinitionProvider.php
yenfryherrerafeliz: I was initially going to use this method for resolving the bdd -> tree ruleset, but I decided later to add the `getParsedRuleset` function for it.

src/EndpointV2/EndpointProviderV2.php
```diff
@@ -66,4 +124,16 @@ private function hashInputParameters($inputParameters)
     {
         return md5(serialize($inputParameters));
     }
+
+    /**
+     * @return array|null
+     */
+    public function getActiveParameters(): ?array
```
stobrien89: The return type should be `array` since both ruleset types guarantee an array. In the old ruleset type, the spec guarantees a non-empty `parameters` field, which is why it's assumed in the constructor.

src/EndpointV2/Bdd/BddRuleset.php
```diff
@@ -0,0 +1,123 @@
...
+    /** @var int[] Flat triples: [condIdx, hi, lo, condIdx, hi, lo, ...] */
+    private array $nodes;
+
+    private int $root;
+
+    public readonly RulesetStandardLibrary $standardLibrary;
+
+    public function __construct(array $definition, array $partitions)
+    {
+        foreach (['conditions', 'results', 'nodes', 'root'] as $key) {
```
stobrien89: recommend moving this to a `private const REQUIRED_FIELDS` and adding `nodeCount` since the index below would cause a confusing error with a malformed ruleset.

tests/EndpointV2/EndpointDefinitionProviderTest.php
```diff
@@ -2,10 +2,14 @@
...
 namespace Aws\Test\EndpointV2;
 
+use Aws\EndpointV2\Bdd\BddRuleset;
 use Aws\EndpointV2\EndpointDefinitionProvider;
+use Aws\EndpointV2\Ruleset\Ruleset;
+use Aws\Test\TestsUtility;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
 use Yoast\PHPUnitPolyfills\TestCases\TestCase;
+use function PHPUnit\Framework\assertInstanceOf;
```
stobrien89: nit: unused import

stobrien89 (changes requested):
A few more minor things. And nice improvement with the switch statement in `callFunction()`

yenfryherrerafeliz:
```
............

111 scenarios (111 passed)
222 steps (222 passed)
0m41.02s (97.73Mb)
```

```
...................................................

135 scenarios (135 passed)
471 steps (471 passed)
29m41.81s (649.75Mb)
```

src/EndpointV2/Bdd/BddEvaluator.php
```diff
@@ -0,0 +1,79 @@
...
+            $isComplement = $ref < 0;
+            $base = ($isComplement ? -$ref : $ref) * 3 - 3;
+
+            $condIndex = $nodes[$base];
+            $value = $library->callFunction(
+                $conditions[$condIndex],
+                $inputParameters
+            );
+
+            $condResult = !(is_null($value) || $value === false);
```
stobrien89:
```suggestion
            $condResult = ($value !== null && $value !== false);
```
Micro-optimization here, but `!==` operator is cheaper than negation on the expression result

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
```diff
@@ -313,25 +369,85 @@ public function isVirtualHostableS3Bucket($bucketName, $allowSubdomains)
...
+                $result = $this->booleanEquals(
+                    $this->resolveValue($argv[0], $inputParameters),
+                    $this->resolveValue($argv[1], $inputParameters)
+                );
+                break;
+
+            case 'isSet':
+                $result = $this->is_set(
+                    $this->resolveValue($argv[0], $inputParameters)
+                );
```
stobrien89:
```suggestion
                $arg = $argv[0];
                $result = isset($arg['ref'])
                  ? isset($inputParameters[$arg['ref']])
                  : $this->is_set($this->resolveValue($arg, $inputParameters));
```

`isSet` is the most frequently invoked condition in any ruleset and in practice always takes a simple ref argument (`{"ref": "X"}`). This short-circuits the full `resolveValue()` dispatch (which checks for functions, refs, and templates) and uses PHP's native `isset()` directly. A single hash table lookup instead of multiple method calls per node evaluation

src/EndpointV2/Ruleset/RulesetStandardLibrary.php
stobrien89:
```php
    if (is_array($value)) {
        if (isset($value['fn'])) {
            return $this->callFunction($value, $inputParameters);
        }
        if (isset($value['ref'])) {
            return $inputParameters[$value['ref']] ?? null;
        }
    } elseif (is_string($value) 
              && str_contains($value, '{') 
              && $this->isTemplate($value)
    ) {
        return $this->resolveTemplateString($value, $inputParameters);
    }

    return $value;
```
`resolveValue` is called 1-3× per condition argument for every node traversal. Inlining `isFunc`/`isRef` (which are just `is_array && isset` wrappers) eliminates 2 method calls per invocation. The `str_contains`  for templates (already used in `isTemplate`) is hoisted here to avoid the regex entirely for scalar literals like `'us-east-1'` or `true`

stobrien89 (approved):
Just a few potential optimizations, but not required.  Thanks @yenfryherrerafeliz!

stobrien89 (approved):
Let's do an integ and smoke run before merging, but otherwise looks great. Thanks @yenfryherrerafeliz!

yenfryherrerafeliz:
```                                         
............

111 scenarios (111 passed)
222 steps (222 passed)
0m35.45s (97.75Mb)

```



```
...................................................

135 scenarios (135 passed)
471 steps (471 passed)
42m54.61s (723.68Mb)
```

PR #3279, opened 2026-04-25
Make EcsCredentialProvider retry behavior configurable

kieranbrown:
## Description of changes

Replaces the hardcoded retry condition in `EcsCredentialProvider` with two configurable allowlists, addressing review feedback that retry behavior for HTTP/container credential providers is largely undefined across AWS SDKs and shouldn't be expanded by default.

Two new constructor options:

- `retryable_exceptions` — array of exception class names. Defaults to `[ConnectException::class]`, preserving existing behavior.
- `retryable_error_codes` — array of HTTP status codes to retry when returned in a Guzzle `BadResponseException`. Defaults to `[]`.

The retry decision now lives in a private `isRetryable()` helper that consults both lists.

### Motivation

The EKS Pod Identity Agent's [rate limiter](https://github.com/aws/eks-pod-identity-agent/blob/main/internal/middleware/rate_limiter/rate_limiter.go) returns HTTP 429 when its token bucket is empty. Previously these surfaced immediately as `CredentialsException`. Callers in that environment can now opt in:

```php
new EcsCredentialProvider([
    'retryable_error_codes' => [429],
]);
```

This mirrors the explicit allowlist used by `aws-sdk-go-v2`'s [endpointcreds client](https://github.com/aws/aws-sdk-go-v2/blob/main/credentials/endpointcreds/internal/client/client.go), without changing default behavior for any existing caller.

### Tests

Added coverage for: opt-in 429 retry succeeds, opt-in 429 retry exhausts attempts, custom exception class added via config, custom config replacing defaults, 429 not retried by default. Existing `ConnectException` retry and 401 non-retryable cases unchanged.

---

Generated by AI tools (Claude), and reviewed by Kieran Brown.

### DIFF
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -3,6 +3,7 @@ namespace Aws\Credentials;
 
 use Aws\Arn\Arn;
 use Aws\Exception\CredentialsException;
+use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\GuzzleException;
 use GuzzleHttp\Psr7\Request;
@@ -107,7 +108,11 @@ class EcsCredentialProvider
                     })->otherwise(function ($reason) {
                         $reason = is_array($reason) ? $reason['exception'] : $reason;
 
-                        $isRetryable = $reason instanceof ConnectException;
+                        $isRetryable = $reason instanceof ConnectException
+                            || (
+                                $reason instanceof BadResponseException
+                                && $reason->getResponse()->getStatusCode() === 429
+                            );
                         if ($isRetryable && ($this->attempts < $this->retries)) {
                             sleep((int)pow(1.2, $this->attempts));
                         } else {
--- a/src/Credentials/EcsCredentialProvider.php
+++ b/src/Credentials/EcsCredentialProvider.php
@@ -42,11 +42,22 @@ class EcsCredentialProvider
     /** @var int */
     private $attempts;
 
+    /** @var string[] */
+    private $retryableExceptions;
+
+    /** @var int[] */
+    private $retryableErrorCodes;
+
     /**
      *  The constructor accepts following options:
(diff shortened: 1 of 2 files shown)


stobrien89:
Hi @kieranbrown,

Thanks for the contribution.  We're hesitant to make this default behavior, only because HTTP/Container credential provider retry behavior is largely undefined across AWS SDKs. We'd be open to something like a configurable `'retryable_exceptions'` and/or `'retryable_error_codes'` in `$config` that gets passed at construction and set as properties which are checked where `$isRetryable` is currently set.  I think this would warrant a private `isRetryable()` helper method too.  

If you're open to implementing something like that, we'd accept changes as long as they come with sufficient test coverage.  If not, feel free to open a feature request in our issues section and we could implement this when able.

kieranbrown:
Hey @stobrien89, totally understandable, I've made the changes you've requested

src/Credentials/EcsCredentialProvider.php
```diff
@@ -41,11 +42,22 @@ class EcsCredentialProvider
     /** @var int */
     private $attempts;
 
+    /** @var string[] */
+    private $retryableExceptions;
```
stobrien89, line 46: This should have a default value of `[ConnectException::class]` and any exceptions passed in `retryable_exceptions` should be added to it. Open to alternatives if your use case requires the current pattern

src/Credentials/EcsCredentialProvider.php
```diff
@@ -221,6 +235,31 @@ private function getEcsUri()
...
+     */
+    private function isRetryable($reason): bool
+    {
+        foreach ($this->retryableExceptions as $exceptionClass) {
+            if ($reason instanceof $exceptionClass) {
+                return true;
+            }
+        }
+
+        if ($reason instanceof BadResponseException
```
stobrien89, line 250: Is `RequestException` too broad?

stobrien89 (changes requested):
Looks good- just a suggestion and question

stobrien89:
Hey @kieranbrown, just bumping the above review

PR #3282, opened 2026-05-06
fix: break circular reference cycles from non-static closures

smilkuri:
*Issue #, if available:*
Related to [#1273](https://github.com/aws/aws-sdk-php/issues/1273)

*Description of changes:*
Seven middleware closures in the implicitly capture $this because they aren't declared static, creating circular references between the client and its HandlerList. Fixed by making closures static, extracting needed values into local variables and using WeakReference for the S3MultiRegionClient. Verified all 7 fixes by running the cyclic reference test script and all of them show 0 KB leaked and no cycles detected.

### DIFF
--- a/.changes/nextrelease/fix-memory-leak-2.json
+++ b/.changes/nextrelease/fix-memory-leak-2.json
@@ -1,0 +1,7 @@
+[
+    {
+        "type": "enhancement",
+        "category": "",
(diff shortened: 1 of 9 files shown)


stobrien89 (approved):
Looks good, thanks @smilkuri! I see you spotted an extra one in `Middleware::mapRequest()`- nice catch.

PR #3289, opened 2026-05-23
fix: cast generated header values to strings

GrahamCampbell:
Hi there. The Guzzle maintainer here again. I am working on the next major version of Guzzle with a focus on security hardening and reducing security-foot guns. As part of those changes, the next major version will disallow non-string header values. I will be triggering deprecations soon on the current major version series, which will start to impact the AWS PHP SDK, unless it is first modified to stop passing non-string header values. And thus, that is what this PR does.

### DIFF
--- a/src/Api/Serializer/RestSerializer.php
+++ b/src/Api/Serializer/RestSerializer.php
@@ -159,7 +159,7 @@ abstract class RestSerializer
 
             $body = $args[$name];
             if (!$m['streaming'] && is_string($body)) {
-                $opts['headers']['Content-Length'] = strlen($body);
+                $opts['headers']['Content-Length'] = (string) strlen($body);
             }
 
             // Streaming bodies or payloads that are strings are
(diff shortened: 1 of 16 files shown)


src/Api/Serializer/RestSerializer.php
```diff
@@ -218,10 +218,23 @@ private function applyHeaderMap($name, Shape $member, array $value, array &$opts
     {
         $prefix = $member['locationName'];
         foreach ($value as $k => $v) {
-            $opts['headers'][$prefix . $k] = $v;
+            $opts['headers'][$prefix . $k] = $this->normalizeHeaderValue($v);
         }
     }
 
+    private function normalizeHeaderValue($value)
```
GrahamCampbell: You may notice that this function is not exhaustive. That is because we want to allow non-string-castable values to result in an invalid argument exception thrown by Guzzle. If you'd rather throw your own exception, that's fine too, and could be a reasonable thing to do. Indeed Guzzle Services chooses to do this in the analogous version of this PR https://github.com/guzzle/guzzle-services/pull/212. Let me know your preference, and I can adjust accordingly.

stobrien89 (changes requested):
Thanks @GrahamCampbell, looks good.  Just one thing: could you add a changelog entry by creating `.changes/nextrelease` and adding a json file by any name with contents in the following format:

```
[
    {
        "type": "bugfix",
        "category": "Api",
        "description": "<description>"
    }
]
```

just in the off chance someone hits the new validation paths.

GrahamCampbell:
Done.

stobrien89 (approved):

PR #3293, opened 2026-05-28
bugfix: align multipartcopy with copyobject metadata directive

stobrien89:
*Description of changes:*
Updates `MultipartCopy` to fully align with `CopyObject` metadata directive behavior. When `$config['metadata_directive']` is set to `COPY` (default), source object metadata takes precedence over any matching values provided in `$config['params']`.


SamRemis (approved):
🚢

Issue #3116, opened 2025-05-08
Enable using a different signature algorithm for CloudFront\UrlSigner::getSignedUrl


ianmjones:
### Describe the feature

It would be very useful if `CloudFront\UrlSigner::getSignedUrl()`:

1. Has an optional 4th parameter to specify the [Signature Algo](https://www.php.net/manual/en/openssl.signature-algos.php) that should be used for signing the URL.

*and/or*

2. It uses a more modern and secure algo by default such as [OPENSSL_ALGO_SHA256](https://www.php.net/manual/en/openssl.signature-algos.php#constant.openssl-algo-sha256).

### Use Case

At present, `CloudFront\UrlSigner::getSignedUrl()` does not allow you to change the algorithm used to sign URLs, it always uses SHA1 because under the hood it eventually uses PHP's [openssl_sign()](https://www.php.net/manual/en/function.openssl-sign.php) function without passing it an algorithm, and therefore the default [OPENSSL_ALGO_SHA1](https://www.php.net/manual/en/openssl.signature-algos.php#constant.openssl-algo-sha1) is used.

Increasingly the SHA1 algorithm is not available on a server as OpenSSL v3 disables it by default, this causes an error to be thrown as per PR #2590, to resolve the previously silent failure #2608.

One such server OS that produces this kind of problem is Ubuntu 24.04, which is becoming more and more of a problem as it is rolled out to hosting providers.

### Proposed Solution

The signature of `CloudFront\UrlSigner::getSignedUrl()` could be changed to:

`public function getSignedUrl($url, $expires = null, $policy = null, $algo = null)`

With a corresponding change to `CloudFront\Signer::getSignature()`:

`public function getSignature($resource = null, $expires = null, $policy = null, $algo = null)`

And a corresponding change to `CloudFront\Signer::sign()`:

`private function sign($policy, $algo = null)`

... which is where `$algo` would be passed to `openssl_sign` if set.

---

Alternatively, `OPENSSL_ALGO_SHA256` is passed to `openssl_sign` in `CloudFront\Signer::sign()`.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.343.6 (and earlier)

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 7.2 - PHP 8.4, any OS!

stobrien89:
Hi @ianmjones,

[...]

PR #3315, opened 2026-06-25
feat: add option to allow different algos to sign Cloudfront

guidopili-docebo:
Addresses *Issue #3116*

## Summary

CloudFront's Signer, CookieSigner, UrlSigner, and the CloudFrontClient convenience methods previously hardcoded SHA1 as the signing algorithm (the openssl_sign() default). This change adds an optional $algorithm parameter so callers can pass any OpenSSL-supported algorithm, such as SHA256.

## Changes

- Signer — new `$algorithm` property (fourth constructor parameter, defaults to `OPENSSL_ALGO_SHA1`). Passed through to `openssl_sign()`.
- CookieSigner / UrlSigner — new $algorithm constructor parameter threaded to the internal Signer instance.
- CloudFrontClient — `getSignedUrl()` and `getSignedCookie()` read an optional `'algorithm'` key from the options array and forward it to the respective signer.
- SignerTest — new test asserting that `OPENSSL_ALGO_SHA256` produces a valid, URL-safe signature that differs from the SHA1 output.

Backward compatibility

Fully backward compatible. All new parameters default to `OPENSSL_ALGO_SHA1`, which matches the previous implicit behavior.

## Usage
```
// Direct signer
$signer = new Signer($keyPairId, $privateKey, '', OPENSSL_ALGO_SHA256);

// Via CookieSigner / UrlSigner
$signer = new CookieSigner($keyPairId, $privateKey, OPENSSL_ALGO_SHA256);
$signer = new UrlSigner($keyPairId, $privateKey, OPENSSL_ALGO_SHA256);

// Via CloudFrontClient
$signer->getSignedUrl([
    'url'         => 'https://d111.cloudfront.net/image.jpg',
    'expires'     => time() + 3600,
    'key_pair_id' => $keyPairId,
    'private_key' => $privateKey,
    'algorithm'   => OPENSSL_ALGO_SHA256,
]);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## next release
+## 3.388.6 - 2026-07-14
 
 * `Aws\Crypto` - Fixes AAD concatenation and tag validation
 * `Aws\drs` - Fast recovery of EC2 based drs workloads by skipping the conversion step
--- a/src/Sdk.php
+++ b/src/Sdk.php
@@ -857,7 +857,7 @@ namespace Aws;
  */
 class Sdk
 {
-    const VERSION = '3.388.5';
+    const VERSION = '3.388.6';
 
     /** @var array Arguments for creating clients */
     private $args;


src/CloudFront/CloudFrontClient.php
```diff
@@ -379,7 +382,8 @@ public function getSignedUrl(array $options)
 
         $urlSigner = new UrlSigner(
             $options['key_pair_id'],
-            $options['private_key']
+            $options['private_key'],
+            isset($options['algorithm']) ? $options['algorithm'] : OPENSSL_ALGO_SHA1
```
stobrien89: For cross-sdk parity, we should use algorithm strings (i.e. "SHA-256") and there should be some validation on the algorithm- I would put the supported algorithm strings in an array constant and ensure the passed value (if provided) is one of the supported options

src/CloudFront/CloudFrontClient.php
```diff
@@ -361,6 +361,9 @@ class CloudFrontClient extends AwsClient
      *   URLs for private distributions.
      * - private_key: (string) The filepath to the private key used to sign
      *   CloudFront URLs for private distributions.
+     * - algorithm: (int|string) OpenSSL signature algorithm constant (e.g.
```
stobrien89: docblock should be updated to align with CloudFront's supported algorithms

src/CloudFront/Signer.php
```diff
@@ -96,9 +102,40 @@ public function getSignature($resource = null, $expires = null, $policy = null)
...
+
         return $signatureHash;
     }
 
+    /**
+     * Maps the OpenSSL algorithm to the value CloudFront expects in the
+     * Hash-Algorithm query parameter / cookie attribute. Returns null for
+     * SHA1 (the default; no parameter needed for backward compatibility).
+     */
+    private function getCloudFrontAlgorithmParam()
```
stobrien89: The `$map` and the string-normalization branch can collapse into a single class-level constant if the constructor validates the input up front. That removes the null-return path and lets you replace this method with a lookup.

src/CloudFront/CookieSigner.php
```diff
@@ -12,15 +12,19 @@ class CookieSigner
     ];
 
     /**
-     * @param $keyPairId  string ID of the key pair
-     * @param $privateKey string Path to the private key used for signing
+     * @param $keyPairId  string     ID of the key pair
+     * @param $privateKey string     Path to the private key used for signing
+     * @param $algorithm  int|string OpenSSL signature algorithm constant (e.g.
```
stobrien89: Same rename recommendation as `Signer.php`. Please also update the docblock to list the two supported values (SHA-1, SHA-256) rather than pointing at any OpenSSL constant.

src/CloudFront/CloudFrontClient.php
stobrien89: nit: should use null coalescing here

tests/CloudFront/SignerTest.php
```diff
@@ -78,6 +78,30 @@ public function testReturnsSignatureAndKeyPairId()
...
+            self::PASSPHRASE,
+            OPENSSL_ALGO_SHA256
+        );
+        $signature = $sha256Signer->getSignature('test.mp4', time() + 1000);
+
+        $this->assertArrayHasKey('Signature', $signature);
+        $this->assertArrayHasKey('Key-Pair-Id', $signature);
+        $this->assertArrayHasKey('Hash-Algorithm', $signature);
+        $this->assertSame('SHA256', $signature['Hash-Algorithm']);
+        $this->assertDoesNotMatchRegularExpression('/[\+\=\/]/', $signature['Signature']);
```
stobrien89: The base64 regex assertion checks encoding, not the signature. Please strengthen with one of:

1. RSA is deterministic. Assert a byte-exact expected signature for `fixtures/test.pem` with SHA-256 and a fixed resource/expiry.
2. Or use `openssl_verify` against the public key to confirm the signature actually verifies under SHA-256.

tests/CloudFront/SignerTest.php
stobrien89, line 1: overall comment for when validation is added: add a test that passes an unsupported hash (for example `OPENSSL_ALGO_MD5` or `'sha512'`) and asserts `InvalidArgumentException`. `docs/URL_SIGNER.md` §Error Handling requires this.

stobrien89 (changes requested):
Sorry for the delay @guidopili-docebo and thanks for the contribution.  Left a couple of comments- if you're able to address them, just request another review and I'll take a look ASAP. If not, I can make those changes too.

src/CloudFront/CloudFrontClient.php
guidopili-docebo: I have moved to algorithm strings as a default, while leaving the possibility to pass openssl constants. Also, validation was added. Was your intention to fully drop the constants or does this implementation work?

src/CloudFront/CloudFrontClient.php
guidopili-docebo: Updated docblock and added supported algos

src/CloudFront/Signer.php
guidopili-docebo: This function was dropped as part of the slight refactor that came out from this comments.

src/CloudFront/CookieSigner.php
guidopili-docebo: Implemented the same changes as `Signer.php` (also for `UrlSigner.php`).

tests/CloudFront/SignerTest.php
guidopili-docebo: Added additional check using `openssl_verify` to confirm the signature verifies under SHA-256.

tests/CloudFront/SignerTest.php
guidopili-docebo: Added a test in `SignerTest.php` to verify the expected behaviour of an unsupported algo (SHA512).

src/CloudFront/CloudFrontClient.php
guidopili-docebo: Changed to coalesce.

guidopili-docebo:
@stobrien89 I have implemented the changes suggested in the review. Can I ask for a second review?

stobrien89:
@guidopili-docebo,

Thanks a lot for updating the implementation. I will review and add any further changes as needed so we can get this merged today.

stobrien89:
Hi @guidopili-docebo,

Apologies, had a mishap with rebasing and force-pushing to the branch. I've pushed the commits to a branch on my fork and plan to open another PR unless you're able to re-open this one. Let me know what you'd prefer to do- I'll make sure you're properly credited either way.

guidopili-docebo:
@stobrien89 No worries! I've seen you already opened another PR, that works for me!

PR #3318, opened 2026-07-06
fix: fixes for S3 Encryption Client

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

In S3 Encryption Client: 

- Fix AAD concatenation
- Fix tag validation


josecorella (approved):
lgtm

src/Crypto/DecryptionTraitV2.php
```diff
@@ -401,7 +401,7 @@ protected function getCommitingDecryptingStream(
                     $cipherOptions['TagLength']
                 );
                 $cipherOptions['Aad'] = isset($cipherOptions['Aad'])
-                    ? $cipherOptions['Aad'] + $algorithmSuiteIdAsBytes
+                    ? $cipherOptions['Aad'] . $algorithmSuiteIdAsBytes
```
stobrien89, line 404: are we trying to concatenate to the string provided? Same applies to the other places we do this. An example (in this thread) would be helpful

stobrien89 (commented):
Looks good, just had one question

src/Crypto/DecryptionTraitV2.php
lucasmcdonald3: Yes, concatenate.

For

```php
$aad   = 'document-id:42';
$suite = "\x00\x78"; // algorithmSuiteIdAsBytes
```

Before was

```php
$aad + $suite;  // => 0 // (actually a TypeError at runtime in PHP 8+, so this never worked)
```

Now is

```php
$aad . $suite;  // => "document-id:42\x00\x78" // (desired result)
```

stobrien89 (approved):

PR #3324, opened 2026-07-20
Set DEFAULT_TOKEN_TTL_SECONDS const as string type

thomashockaday:
This deprecation warning appears on the `x-aws-ec2-metadata-token-ttl-seconds` header for PHP 8.2: 

```
Since guzzlehttp/psr7 2.11: Passing int to MessageInterface::withHeader() is deprecated; guzzlehttp/psr7 3.0 requires string|string[]. in /var/app/current/vendor/symfony/deprecation-contracts/function.php on line 25
```

### DIFF
--- a/src/Credentials/InstanceProfileProvider.php
+++ b/src/Credentials/InstanceProfileProvider.php
@@ -7,8 +7,8 @@ use Aws\Exception\InvalidJsonException;
 use Aws\Sdk;
 use GuzzleHttp\Exception\TransferException;
 use GuzzleHttp\Promise;
-use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7\Request;
 use Psr\Http\Message\ResponseInterface;
 
 /**
@@ -26,7 +26,7 @@ class InstanceProfileProvider
     const CFG_EC2_METADATA_SERVICE_ENDPOINT_MODE = 'ec2_metadata_service_endpoint_mode';
     const DEFAULT_TIMEOUT = 1.0;
     const DEFAULT_RETRIES = 3;
-    const DEFAULT_TOKEN_TTL_SECONDS = 21600;
+    const DEFAULT_TOKEN_TTL_SECONDS = '21600';
     const DEFAULT_AWS_EC2_METADATA_V1_DISABLED = false;
     const ENDPOINT_MODE_IPv4 = 'IPv4';
     const ENDPOINT_MODE_IPv6 = 'IPv6';


stobrien89:
Hi @thomashockaday,

curious- did you close because you upgraded to a more recent version where these deprecation warnings have been addressed?

thomashockaday:
Apologies @stobrien89 I did not intentionally close this - a webhook must have triggered somewhere.

I was seeing two types of deprecation warnings on my side:

```
[2026-07-17 12:15:59] staging.WARNING: 
Since guzzlehttp/psr7 2.11: Passing int to MessageInterface::withHeader() is deprecated;
guzzlehttp/psr7 3.0 requires string|string[].
in /var/app/current/vendor/symfony/deprecation-contracts/function.php on line 25

[2026-07-17 12:15:59] staging.WARNING:
Since guzzlehttp/psr7 2.11: Passing int to GuzzleHttp\Psr7\Request::__construct() is deprecated;
guzzlehttp/psr7 3.0 requires string|string[].
in /var/app/current/vendor/symfony/deprecation-contracts/function.php on line 25
```

From what I can tell the first warning applies to the `x-aws-ec2-metadata-token-ttl-seconds` header and the second applies to the `Content-Length` header. I had a pull request elsewhere updating my `aws-sdk-php` version to 3.384.0 which I think should have resolved the second warning but **not the first**. That pull request closing also seems to have closed this one for some reason.

GrahamCampbell:
I don't think this should be merged. AWS SDK 3.384.0 already casts every `InstanceProfileProvider` header value to a string at the PSR-7 boundary via #3297, and generated `Content-Length` and header values were fixed in 3.382.2 via #3289. Changing `DEFAULT_TOKEN_TTL_SECONDS` from an integer to a string is therefore redundant and makes a duration constant less accurately typed. The logs must have come from an older SDK loaded by the affected process, even if 3.384.0 was present in a separate dependency update.

After #3327 is released, upgrading to the latest AWS SDK, Guzzle 7.x, and PSR-7 2.x should be sufficient to stop the remaining compatibility deprecations. It would be worth checking `Aws\Sdk::VERSION` in the affected process and clearing any stale vendor or OPcache state.

thomashockaday:
I've been monitoring our logs since my comment yesterday when the upgrade went out and have no longer seen either of the deprecation warnings I originally reported, which should mean this doesn't need merging thankfully. Whether it's still a valid change or not is probably a matter of opinion, but considering the problem I had was solved by upgrading as you rightly pointed out @GrahamCampbell I'd be happy just to close this. Thanks for the consideration either way!

stobrien89:
Hi @thomashockaday,

I agree with @GrahamCampbell here. Working with him to get #3327 merged today and will close this now.  Thank you for bringing it to our attention.

thomashockaday:
Awesome, thanks both!

Issue #3309, opened 2026-06-18
Document arguments of various clients via proper detailed type definitions


Ocramius:
## Context

Today, I wrote this:

```php
private function instancesToDelete($tags) {
    return $this->rds->describeDBInstances(['Tags' => $tags])
}
```

This almost got me to delete production DBs by accident: `Tags` is completely ignored by `RdsClient#describeDBInstances()`.

/cc @marcostagno @albertomolinari

## The current API "documentation"

This proposal is about documenting clear array shapes for each API endpoint.

Currently, most API endpoints are documented as:

```php
/**
 * @method \Aws\Result describeDBInstances(array $args = [])
 */
```

It suffices to say that **this is absolutely useless rubbish**.

## The proposal

The generated API documentation should change, and become:

```php
/**
 * @method \Aws\Result<list<array{
 *   foo: detailed-type-of-foo,
 *   bar: detailed-type-type-of-bar
 * }>> describeDBInstances(array{
 *   baz?: detailed-type-of-baz,
 *   taz?: detailed-rtype-of-taz
 * } $args = [])
 */
```

This is **actually useful**, self-documenting, and helpful for static analysis tooling (Psalm, PHPStan, Mago, Phan).

### Use Case

This feature removes the need for jumping around web documentation when using the AWS SDK.

### Proposed Solution

The code generator should convert documented types that are used under the hood into PHP types:

* arrays
* lists 
* non-empty-lists
* string
* non-empty-string
* int
* int<1, max>
* etc.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.379.0

### Environment details (Version of PHP (`php -v`)? OS name and version, etc.)

PHP 8.4

Ocramius:
An example of how `asyncaws` generates such API doc:

<img width="1988" height="828" alt="Image" src="https://github.com/user-attachments/assets/ad225dae-a794-4f18-a701-180759a46d87" />

This is really good DX, and prevents the mistake that almost led to production system damage today...

stobrien89:
Thanks for raising this, @Ocramius, and sorry for the close call.

I understand the "jumping around web documentation" and find myself doing this often as a maintainer. The `@method` annotations were written when the SDK supported PHP 5.5, which is why they don't carry the kind of array shape info PHPStan or Psalm can use today.  Luckily I think we're in a place now where we could do something about this.

[...]

Issue #3017, opened 2024-10-29
Exploring Future Directions for the SDK


stobrien89:
### Describe the issue

Hello Everyone,

We’re excited to announce we are currently exploring potential updates and improvements for the AWS SDK for PHP.

### We Need Your Feedback!

Below we’ve listed a few of the most commonly requested features in separate comments (Modularization, Async functionality, etc.). Please react with a 👍 to the features you would like to see in a major update to the SDK, or any other feedback, suggestions, etc. you would like to highlight. If there is something specific you’d like to see, please comment on this issue with your ideas or suggestions.  They can include, but are not limited to:

* Workflows: Are there any pieces of functionality that you rely on and would like to remain consistent?
* API Changes: Are there any APIs or patterns in the current SDK that you find confusing or cumbersome? What would you change?
* Error Handling & Debugging: How can we improve error handling, logging, or debugging tools within the SDK?
* Framework Compatibility: Are there frameworks for which you’d like to see integrated (i.e. native) SDK support? 
* New Language Features: Are there any recent PHP language features you would like us to incorporate in the SDK? 
* Performance Improvements: Are there areas where you feel the SDK could be optimized for better performance?
* Other Suggestions: Any other ideas or pain points that you would like us to consider?

### Next Steps:

Your input will play a key role in shaping the direction we take. We will collect your feedback over the next month and evaluate which updates to prioritize.
<br>

— The AWS SDK for PHP Team

### Links

n/a

stobrien89:
Modularization/Submodules

stobrien89:
Async/IO

stobrien89:
Encapsulation/Type Safety/IDE discoverability

fredericgboutin-yapla:
I would like to stress out that we appreciate the **stability** and the **retro-compatibility**.

For example supporting a wide range of PHP versions and keeping a single major version as much as possible over time, this is very important for us in the long run.

mcharnayprimeo:
Updating this package with composer fail regularly cause of timeout. Is there already a solution for that ?

yenfryherrerafeliz:
Hi @mcharnayprimeo, thanks for commenting here. Could you please open a new issue and provide more details about what exactly you get?, which one is the error, etc. so we can better look into that problem?

Thanks!

asispts:
There are 2 things to consider:
1. Modularization, so we can install only the required client.
2. Making it easier to mock the client for tests.

[...]

PR #3331, opened 2026-07-25
feat: add phpstan method annotations

stobrien89:
*Issue #, if available:* 
#3309 

*Description of changes:*
Adds `phpstan-method` annotations to service client methods for better static analysis support.


tests/Build/PhpStan/phpstan.neon
andrew-demb, line 1: Is it intentionally empty?

andrew-demb (approved):
PHPStan seems to be working properly, but PHPStorm doesn't understand multiline array type declarations in "@phpstan-method" tags and adds a lot of inspection errors to the file. But it will affect only package development, not consumers.

Overall, the original goal was reached successfully.

Image

Image

andrew-demb:
... relevant PhpStorm issues that I've found:
- https://youtrack.jetbrains.com/issue/WI-78397/PHPStorm-creates-an-Expected-variable-for-PHPDoc-method-word-wraps
- https://youtrack.jetbrains.com/issue/WI-75416/Support-multiline-psalm-array-declaration-with-comments

tests/Build/PhpStan/phpstan.neon
stobrien89: It's not. I'll get it updated

build/PhpstanShapeFormatter.php
```diff
@@ -0,0 +1,351 @@
...
+            $memberShape = $member['shape'] ?? null;
+            $rendered = $memberShape !== null
+                ? $this->renderShape($memberShape, $depth + 1)
+                : 'mixed';
+
+            $key = $this->encodeKey($name);
+            $segments[] = "{$key}?: {$rendered}";
+        }
+
+        // Append the unsealed marker so PHPStan/Psalm don't flag
```
Ocramius: Note: this is where I ran into severe issues with AWS, passing completely ignored arguments to the API, and almost destroying production resources

stobrien89: You're right, this is the false-negative we accept with the unsealed marker. Sealing here would catch `Buckt` → `Bucket`, but it would also flag the common pattern of dynamically-built `array<string, mixed>` values passed as `$args` (request context, tracing metadata) where PHPStan/Psalm can't prove key set membership at the call site. What do you think about sealing top-level members, but leaving nested members unsealed?

stobrien89: @Ocramius bumping the above ^

Ocramius: Sorry, I was on vacation

> but it would also flag the common pattern of dynamically-built `array<string, mixed>`

For this, tools like PHPStan/Psalm/Mago/Phan have the concept of a "baseline", which is a sufficient approach to catch this sort of detail.

> What do you think about sealing top-level members, but leaving nested members unsealed?

I think it wouldn't be added value here.

People that use PHPStan/Psalm/etc. can ignore these issues themselves, as an opt-in, either via a pattern or via a baseline.

Reporting these problems is more valuable than taking an upfront decision to suppress them for consumers.

smilkuri (approved):

stobrien89:
@Ocramius, this will be available in today's release. Also, commented on the "future directions" issue with our plans: https://github.com/aws/aws-sdk-php/issues/3017#issuecomment-5562146670

V4 will eliminate most of the problems raised in that issue.

Ocramius:
@stobrien89 nice! 🙌

PR #3334, opened 2026-07-31
bugfix: retry behavior

stobrien89:
*Description of changes:*
Aligns the PHP SDK’s new retry implementation with retry compliance tests by updating the opt-in V3 path to emit compliant retry headers, honor `AWS_MAX_ATTEMPTS` and shared-config max_attempts without requiring `retry_mode`, and preserve retry quota/rate-limiter state correctly across operations on the same client. The change also updates the PHP retry tests to cover the new behavior and long-polling/quota flow.

### DIFF
--- a/src/Retry/RetryHelperTrait.php
+++ b/src/Retry/RetryHelperTrait.php
@@ -6,9 +6,14 @@ use Aws\ResultInterface;
 
 trait RetryHelperTrait
 {
-    private function addRetryHeader($request, $retries, $delayBy)
+    private function addRetryHeader($request, $retries, $maxAttempts = null)
     {
-        return $request->withHeader('aws-sdk-retry', "{$retries}/{$delayBy}");
+        $header = ['attempt=' . ($retries + 1)];
+        if ($maxAttempts !== null) {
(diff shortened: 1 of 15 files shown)


src/Retry/RetryHelperTrait.php
```diff
@@ -6,9 +6,14 @@
 
 trait RetryHelperTrait
 {
-    private function addRetryHeader($request, $retries, $delayBy)
+    private function addRetryHeader($request, $retries, $maxAttempts = null)
```
smilkuri, line 9: We no longer need the delay?

smilkuri (approved):

src/Retry/RetryHelperTrait.php
stobrien89: Nice catch. Need to update this

stobrien89: Actually, reviewed this and for this particular header, this is what we need. Long ago we sent that info in `aws-sdk-retry`, but this adheres to the updated retry header spec (omits ttl, which is optional)

PR #3335, opened 2026-08-05
Add support for transport sharing

GrahamCampbell:
Guzzle 8 added transport sharing, which lets the built-in handlers share DNS, SSL session, and connection cache state, including persistent cURL share handles on PHP 8.5+ that survive across php-fpm requests. Guzzle 7.11 supports the handler-lifetime subset of the feature. This PR exposes the capability through a new `transport_sharing` client option with five string modes matching Guzzle's vocabulary: `none`, `handler_prefer`, `handler_require`, `persistent_prefer`, and `persistent_require`.

The option's semantics are independent of the installed Guzzle version. The SDK feature-detects what the installed Guzzle supports, degrades the prefer modes gracefully (`persistent_prefer` falls back to handler-lifetime sharing on Guzzle 7.11+ and to no sharing on older versions), and throws for the require modes when they cannot be honored, so `persistent_require` is a deployment assertion that connection re-use is actually happening. The option applies when the SDK creates the default HTTP handler, since transport sharing is configured on the Guzzle client constructor rather than per request, and the require modes reject custom `handler` or `http_handler` options. `Aws\Sdk` applies the option to the HTTP handler it shares across the clients it creates. The name `transport_sharing` mirrors Guzzle's client option; renaming it to `http_transport_sharing` to group with the other HTTP options would be a two-token change if preferred. A separate commit adds a CI matrix leg pinning Guzzle to `>=7.11 <8` so the handler-lifetime degradation path is exercised; it can be dropped if unwanted, since the affected tests skip themselves based on feature detection.

### DIFF
--- a/src/Handler/Guzzle/GuzzleHandler.php
+++ b/src/Handler/Guzzle/GuzzleHandler.php
@@ -2,6 +2,7 @@
 namespace Aws\Handler\Guzzle;
 
 use Aws\Handler\HttpHandlerError;
+use Aws\Handler\HttpTransportSharing;
 use GuzzleHttp\Utils;
 use GuzzleHttp\Promise;
 use GuzzleHttp\Client;
@@ -18,11 +19,23 @@ class GuzzleHandler
     private $client;
 
     /**
-     * @param ClientInterface $client
+     * @param ClientInterface|null $client
+     * @param string|null          $transportSharing
      */
-    public function __construct(?ClientInterface $client = null)
-    {
-        $this->client = $client ?: new Client();
+    public function __construct(
+        ?ClientInterface $client = null,
+        ?string $transportSharing = null
+    ) {
+        if ($client !== null && HttpTransportSharing::isRequired($transportSharing)) {
+            throw new \InvalidArgumentException('The provided transport'
+                . ' sharing mode cannot require sharing when a client is'
+                . ' provided. Configure the "transport_sharing" option on'
+                . ' the provided client instead.');
+        }
+
+        $this->client = $client ?: new Client(
+            HttpTransportSharing::toClientConfig($transportSharing)
+        );
     }
 
     /**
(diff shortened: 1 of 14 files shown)


stobrien89:
Thanks for the work on this @GrahamCampbell - Will review as soon as possible.

src/Handler/HttpTransportSharing.php
```diff
@@ -0,0 +1,95 @@
...
+        if (!in_array($mode, self::MODES, true)) {
+            throw new \InvalidArgumentException('The provided transport'
+                . ' sharing mode "' . $mode . '" is invalid. Valid modes are:'
+                . ' "none", "handler_prefer", "handler_require",'
+                . ' "persistent_prefer", "persistent_require".');
+        }
+
+        // Guzzle 8: all modes are understood, and Guzzle enforces the
+        // runtime requirements of the "*_require" modes itself.
+        if (defined(TransportSharing::class . '::PERSISTENT_PREFER')) {
```
stobrien89: `defined()` and `class_exists()` run on every handler construction. Could we cache the detected capability in a static (compute once per process) since it cannot change at runtime?

src/Handler/Guzzle/GuzzleHandler.php
```diff
@@ -18,11 +19,23 @@ class GuzzleHandler
...
+    ) {
+        if ($client !== null && HttpTransportSharing::isRequired($transportSharing)) {
+            throw new \InvalidArgumentException('The provided transport'
+                . ' sharing mode cannot require sharing when a client is'
+                . ' provided. Configure the "transport_sharing" option on'
+                . ' the provided client instead.');
+        }
+
+        $this->client = $client ?: new Client(
+            HttpTransportSharing::toClientConfig($transportSharing)
```
stobrien89, line 37: `ClientResolver::_apply_transport_sharing` resolves the mode and stores it in `$args['transport_sharing']`. That value is then passed to `default_http_handler()` and re-resolved here. Any way to bypass the double resolution?

stobrien89 (changes requested):
Looks good, just had two comments

GrahamCampbell:
Rebased on latest master.

stobrien89 (approved):

Issue #3336, opened 2026-08-06
StreamWrapper:dir_opendir()  $options is documented as accepting string, should be int


cmlara:
### Describe the issue

StreamWrapper:dir_opendir()  is currently documented as the $options value being unused and accepting a string. https://github.com/aws/aws-sdk-php/blob/7dbeb8684ae8c1388f0d5f26e803259299ae59eb/src/S3/StreamWrapper.php#L404

The PHP StreamWrapper docs however document that the signature should be an integer:
` public function streamWrapper::dir_opendir(string $path, int $options): bool`
https://www.php.net/manual/en/streamwrapper.dir-opendir.php

This creates conflicts (static analysis warnings)  when one extends the class and either overrides dir_opendir() or attempts to implement a third party interface that documents the typehint as integer.

Considering the option is unused, the PHP spec is that this should be an integer (PHP will provide an INT), and that it is documentation not an inline typehint this should be non-breaking to update.

### Links

https://docs.aws.amazon.com/aws-sdk-php/v2/api/class-Aws.S3.StreamWrapper.html#_dir_opendir

stobrien89:
Hi @cmlara,

Thanks for raising this. I agree this should be documented as an int and will merge your PR once the tests pass.

stobrien89:
merged #3337.  Thanks for the contribution @cmlara!

PR #3337, opened 2026-08-06
bugfix: Update parameter type for options in dir_opendir docblock

cmlara:
#3336

Update parameter type for $options in dir_opendir() docblock to align with PHP https://www.php.net/manual/en/streamwrapper.dir-opendir.php


stobrien89:
Thanks for the PR @cmlara, 

I agree this fits with the other wrapper options we support.

stobrien89 (approved):

PR #3343, opened 2026-09-02
chore: update empty endpoint properties assertion

stobrien89:
*Description of changes:*
Updates endpoint resolution assertion on empty properties- our implementation provides `null` for an empty properties. Some tests model an empty array, which creates an incorrect expectation.

### DIFF
--- a/tests/EndpointV2/EndpointProviderV2Test.php
+++ b/tests/EndpointV2/EndpointProviderV2Test.php
@@ -151,9 +151,10 @@ class EndpointProviderV2Test extends TestCase
             if (isset($expectedEndpoint['headers'])){
                 $this->assertEquals($expectedEndpoint['headers'], $endpoint->getHeaders());
             }
-            if (isset($expectedEndpoint['properties'])) {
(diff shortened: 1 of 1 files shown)


tests/EndpointV2/EndpointProviderV2Test.php
```diff
@@ -151,7 +151,7 @@ public function testServiceEndpointAndErrorCases(
             if (isset($expectedEndpoint['headers'])){
                 $this->assertEquals($expectedEndpoint['headers'], $endpoint->getHeaders());
             }
-            if (isset($expectedEndpoint['properties'])) {
+            if (!empty($expectedEndpoint['properties'])) {
```
smilkuri: Wondering instead can we normalize the assertion?
```
$this->assertEquals($expectedEndpoint['properties'] ?? [],
  $endpoint->getProperties() ?? [])
```

stobrien89: Yes, I think that would be better

stobrien89: updated

smilkuri (approved):

PR #3347, opened 2026-09-10
PHP CodeSniffer for CI workflow

lukasdchang:
## Description

Adds a `Coding Standards` CI job that runs PHP_CodeSniffer against changed PHP files in `src/`, excluding `src/data`.

The job compares the PR branch against `master`, runs PHPCS with `phpcs.xml.dist`, and reports formatting violations in the workflow logs. When PHPCS finds violations, the job also runs PHPCBF in the CI checkout and prints a fixed-code preview for the affected block. The output includes the local `phpcbf` command developers can run to apply automatic fixes.

Example:

PHPCS workflow output

## Motivation

The SDK already runs static analysis in CI, but it does not currently enforce formatting. This workflow adds formatting checks without adding PHP_CodeSniffer as a package dependency for SDK consumers.

Developers still install PHPCS locally if they want automatic fixes. CI installs PHPCS only inside the GitHub Actions runner.

## Possible Improvements

- Add Reviewdog to publish PHPCS violations as PR comments.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.


.github/workflows/tests.yml
```diff
@@ -11,6 +11,56 @@ permissions:
   contents: read
 
 jobs:
+  coding-standards:
```
stobrien89: can we move this to a separate workflow? like `coding-standards.yml`

stobrien89 (dismissed):
looks good, just had one request and when you push up those changes, let's see how it reacts to a change in one of the middleware files (move some things around in a file such that it falls out of compliance). You can do that in the same revised commit

lukasdchang (commented):
Created separate workflow file for coding-standards CI job

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,182 @@
...
+            echo "::endgroup::"
+
+            echo ""
+            echo "Run the following command locally for PHPCBF to autofix formatting errors:"
+            echo "phpcbf --standard=phpcs.xml.dist \\"
+            sed 's/^/  /; $!s/$/ \\/' changed-source-files.txt
+            echo ""
+
+            exit "$phpcs_exit"
+          fi
```
kuhe: too much code in a string, write this as a shell or php script and then invoke it with arguments from the yml

kuhe (changes requested):

.github/workflows/coding-standards.yml
lukasdchang: "Run PHP_CodeSniffer" job now runs run-phpcs.sh which contains the PHPCS bash output script under a new directory for scripts: .github/scripts/.

lukasdchang:
coding-standards CI workflow job is confirmed safe from security vulnerabilities by running zizmor locally. What run-phpcs.sh does is it:

1. Runs PHPCS on changed files and outputs the failure
2. Outputs a preview of what each file with formatting errors would look like after PHPCBF autofixes them.
3. Outputs a bash command to run locally to make that fix.

I'm looking to iterate on this so that the workflow job makes PR review comments for PHPCS formatting errors similar to the way smithy-lang handles incorrect changelog entries [here](https://github.com/smithy-lang/smithy/blob/main/.github/workflows/changelog-ci.yml). That way developers do not have to sift through the workflow logs to see their formatting error.

kuhe (approved):

smilkuri (approved):

PR #3351, opened 2026-09-14
[DO NOT MERGE] PR to test PHPCS workflow formatter

lukasdchang:
*Description of changes:*
This PR serves as a place to test workflow behavior specifically the way the coding-standards job that uses PHPCS and PHPCBF. 


.github/workflows/docs-build.yml
```diff
@@ -7,7 +7,7 @@ on:
       - master
 jobs:
   build_docs:
-    if: github.repository_owner == 'aws'
+    if: github.repository_owner == 'aws' && github.event.pull_request.head.ref != 'phpcs-workflow-tests'
```
stobrien89, line 10: I think this can be removed from here and the other workflow files unless there's something I'm missing

src/PhpcsWorkflowTest.php
stobrien89, line 1: This will be removed before merging, right?

.github/workflows/coding-standards.yml
stobrien89, line 1: I can see the workflow produces the correct violations/comments (with suggested change)- Let's see if we can get this to publish comments on the PR with a suggestion. Smithy does that with incorrect changelog entries here: https://github.com/smithy-lang/smithy/blob/main/.github/workflows/changelog-ci.yml

stobrien89 (changes requested):
looking good- just a couple of comments

src/PhpcsWorkflowTest.php
lukasdchang: This is a separate branch where I just do testing on it's own PR so I can observe the workflow behavior, it isn't intended to merge. The src/PhpcsWorkflowTest.php file only exists in this testing branch. The actual feature is here https://github.com/aws/aws-sdk-php/pull/3347

.github/workflows/coding-standards.yml
lukasdchang: Sounds good! I'll see if i can implement that similarly. I just wanted to get a sort of proof of concept through just the workflow outputs first before getting PR comments to suggest formatting changes.

src/PhpcsWorkflowTest.php
lukasdchang: This branch also skips the other workflow jobs so I don't have to wait a long time for the coding-standards job to finally run.

